PR #2, opened 2015-11-25
Replaced unicode characters that caused pyang problems

einarnn:
Found what I'd guess were a few stray unicode characters in openconfig-terminal-device.yang that caused pyang to choke.

Cheers,

Einar


aashaikh:
Thanks Einar -- this should be fixed in commit 72e4ee4 .  Caused by locale issues in the editor.

einarnn:
Great, thanks!

einarnn:
Close this off as it's fixed elsewhere.

PR #3, opened 2015-12-29
Fix pattern in network-instance-types.yang.

jta:
This commit adds a missing parenthesis to the route-distinguisher
pattern.


aashaikh:
@robshakir you may want to review and include this fix with the other pending changes planned for the network instance model.

thanks João !

robshakir:
Thanks! I've committed this to the non-public repo. It'll be fixed in the next rev.

Cheers,
r.

PR #4, opened 2015-12-29
Set config to false for state containers.

jta:
This commit fixes two missing config statements in bgp-use-multiple-paths.


aashaikh:
João, thanks for catching this -- we'll make the changes and push a new version.  I also added a check for this to our linter.

PR #7, opened 2016-02-08
Change 'labeled' nomenclature to match IETF and vendor implementaion naming

earies:
Suggest keeping naming uniform per RFC wording and existing vendor implementations unless there is a good reason to diverge.


robshakir:
Thanks Ebben - as per the README, we don't merge pull requests into this repo - but I will make these changes to the model such that they will be released into the next version. 

(As a side note, I had not realised that there was a difference between British and American English for the spelling of labelled.)

PR #12, opened 2016-03-08
Fix invalid regex for type 2 route-distinguisher

earies:
Merely using a pull request to communicate the diff

regex was improperly defined for type 2 route-distinguisher: missing open bracket

../network-instance/openconfig-network-instance-types.yang:91: error: syntax error in pattern: Element '{http://www.w3.org/2001/XMLSchema}pattern': The value '(4[0-2][0-9][0-4][0-9][0-6][0-7][0-2][0-9][0-5]|[0-3][0-9]{9}|[1-9][0-9]{1,8}|[1-9]):65[0-5][0-3][0-5]|[1-5]{2}[0-9]{3}|[1-9]{0,2}[0-9][0-9]|[1-9])' of the facet 'pattern' is not a valid regular expression., line 1


robshakir:
Thanks Ebben - a couple of folks had reported this, and it'll be fixed when we push the next network-instance model. I'll close the pull-req - please re-open it if there's anything else.

Cheers,
r.

PR #16, opened 2016-04-21
pattern statement does not follow W3C guidelines for regular expression. 

mjethanandani:
RFC 6020 is clear that the pattern statement has to follow the regular expression format defined by W3C. These two pattern statements for some reason do not.

### DIFF
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -82,7 +82,7 @@ module openconfig-bgp-policy {
     type union {
       type uint32;
       type string {
-        pattern "^[+-][0-9]+";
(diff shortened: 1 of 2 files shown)


robshakir:
Hi Mahesh,

Based on discussion amongst the group and with a number of implementors, we were not sure that this specification rule was the right one to follow - the issue being that it did not seem to be a common specification across implementors of NMSes or network equipment. 

To this end, OpenConfig models are intended to use [POSIX Extended Regular Expressions](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html) as per the [style guide](https://github.com/openconfig/public/blob/master/doc/openconfig_style_guide.md#regular-expressions).

Kind regards,
r.

PR #73, opened 2017-06-25
Fixes for oc-inet:ip-address.

jta:
This commit fixes a missing operator in the ipv6-address regex, as well
as adding support for zoned addresses in both ipv4-address and
ipv6-address.


aashaikh:
@jta thanks for the catch on the `I` operator ... 

We may need to discuss the addition of zone notation  -- we deliberately defined the base `ipvX-address` types to be the common case, i.e., non-zoned -- with the option to later introduce `ipvX-address-zoned` types for cases where a zoned address is needed.  If I'm not mistaken, we ran into cases where it doesn't make sense to have a type that allows zone indexes.

jta:
I noticed this was broken because I have a case where `neighbor-address` needs to be a zoned IPv6 address, so maybe the correct solution is to not define it as `oc-inet:ip-address` in the first place.

PR #90, opened 2017-08-29
Initial check-in of 0.1.0 WiFi models

mike-albano:
- Left directory structure as it is in wifi-models private repo
- Specifically; yang/wifi/system/oc-sys-wifi-ext.yang now exists, in addition to yang/system/oc-sys.yang
- yang/wifi/types/oc-wifi-types.yang now exists, in addition to yang/types/...yang
- Modified .spec.yml files to match this directory structure.

LMK how that looks.

### DIFF
--- a/release/models/wifi/.spec.yml
+++ b/release/models/wifi/.spec.yml
@@ -1,0 +1,7 @@
+- name: openconfig-wifi-mac
+  docs:
+    - wifi/mac/openconfig-wifi-mac.yang
+    - wifi/types/openconfig-wifi-types.yang
+  build:
+    - wifi/mac/openconfig-wifi-mac.yang
+  run-ci: true
--- a/release/models/wifi/.spec.yml
+++ b/release/models/wifi/.spec.yml
@@ -1,7 +1,0 @@
-- name: openconfig-wifi-mac
-  docs:
-    - wifi/mac/openconfig-wifi-mac.yang
-    - wifi/types/openconfig-wifi-types.yang
-  build:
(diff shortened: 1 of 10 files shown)


release/models/wifi/.spec.yml
```diff
@@ -0,0 +1,7 @@
+- name: openconfig-wifi-mac
```
aashaikh: do we need a spec file here if we have in the `mac` subdir?

release/models/wifi/README.md
```diff
@@ -0,0 +1,34 @@
+# wifi-models
+Development of configuration and operational state models for WiFi
+
+Open Config is a collaboration between network operators to develop and standardize a set of vendor-neutral network configuration models. Models are written using the YANG data modeling language (IETF RFC 6020).
+
+The configuration models in this repository are made available under the Apache 2.0 license (see the LICENSE file). Since the models are intended to be published in the IETF, participants must be willing to adhere to the IETF Note Well statement as well as BCP 78 and BCP 79.
```
aashaikh: do we plan to publish the wifi models in IETF?  if not sure, you can add these statements if/when we do that.

release/models/wifi/.spec.yml
mike-albano: Don't think so. Probably left here inadvertently. Will remove.

release/models/wifi/README.md
mike-albano: I'd say not sure. Removed IETF references from README.

PR #109, opened 2017-11-09
Add SSID to neighbor-table

mike-albano:
*Add additional leaf to neighbors container for SSID name
*Up version to 0.1.1
*Fix spelling mistakes

LMK if preferable to open 'issue' prior to PR when syncing from private-->public.

### DIFF
--- a/release/models/wifi/phy/openconfig-wifi-phy.yang
+++ b/release/models/wifi/phy/openconfig-wifi-phy.yang
@@ -25,7 +25,13 @@ module openconfig-wifi-phy {
   description
     "Model for managing PHY layer configuration of Radio interfaces.";
 
-  oc-ext:openconfig-version "0.1.0";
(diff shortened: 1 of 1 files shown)


aashaikh:
Just suggest we try to use the (roughly) uniform format for the commit message (can see other recent commits to public repo).

mike-albano:
Noted. This would be more like:
"""
Add SSID to neighbor-table for oc-wifi-phy:neighbor-list-state.
This commit adds an additional leaf to neighbors-list-state container for SSID name, as well as fixing spelling mistakes.
"""

aashaikh:
We've generally done something more like:

```
Additions and fixes to wifi phy model.
    
      * (M) release/models/wifi/phy/openconfig-wifi-phy.yang
        - Added ssid leaf to neighbor state
        - Corrected typos
```

mike-albano:
Got it, thanks.

PR #136, opened 2018-04-12
changes in decription of index in openconfig-local-routing.yang next-hop list

ravicodder:
In openconfig-local-routing.yang, the current description of the index in the next-hop list says:

          " An user-specified identifier utilised to uniquely reference
            the next-hop entry in the next-hop list. The value of this
            index has no semantic meaning other than for referencing
            the entry." 

There is no constraint on the index format, which may lead to vendor's specific implementation of the index. This may lead to inconvenience to openconfig-model users as they need to change their index value in payload according to vendor's requirement. This will lead to the bad user experience.

Thus, to have uniform index implementation across vendors. There should be some constrained suggested by the openconfig-model itself.
**[for example: index format as next-hop-interface + next-hop-address]**

Thanks,

### DIFF
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -137,9 +137,10 @@ module openconfig-local-routing {
       type string;
       description
         "An user-specified identifier utilised to uniquely reference
-        the next-hop entry in the next-hop list. The value of this
-        index has no semantic meaning other than for referencing
-        the entry.";
+        the next-hop entry in the next-hop list. User must specify the index 
+	in specific format as below,
+	'next-hop-interface + next-hop-address',
+	For example index can be 'Eth1-1 + 1.1.1.1/32' "
     }
 
     leaf next-hop {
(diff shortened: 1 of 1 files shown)


robshakir:
Hi,

We do not generally accept pull requests on this repo as mentioned in the description. It's better oto open issues to discuss these matters.

The `index` in the `next-hop` list is specified by the configuration - i.e., by the client/NMS. This means that the client has full control as to what to use for that index. Thus, I'm not sure why the target's implementation matters here. Please can you explain?

Thanks,
r.

ravicodder:
Hi,
@robshakir thanks for your comment.
It is my bad, next time I will open an issue for such queries. 

My understanding of the usage of the index field was quite wrong. I thought the index field is used to supply sufficient information to the device so that it can retrieve you back the next-hop from the next-hop list. Such as next-hop address and next-hop interface. 
But actually, it is kind of primary key which will help us retrieve the next-hop and semantically it has nothing to do with actual route.  

I realized it after reading your comment. 

Thanks for your time, I will close the pull request.

PR #218, opened 2019-04-08
Wrong 'parent' definition in openconfig-platform

idefixcert:
The definiton of the parent in the platform points to the reference of it self.
```
type leafref {
        path "../../config/name";
      }
```
But it should point to an other component.
```
type leafref {
        path "../../../component/config/name";
      }
```


idefixcert:
was changed now, so close the request.

Issue #228, opened 2019-05-23
Missing namespace prefixes on identity names in when statements


nkitchen:
openconfig-network-instance.yang contains several `when` statements that compare `config/type` to `openconfig-network-instance-types:DEFAULT_INSTANCE` and do not include the namespace prefix `oc-ni-types`.

For example:
```
container encapsulation {
  when "../config/type != 'DEFAULT_INSTANCE'" {
```
(https://github.com/openconfig/public/blob/5897507e/release/models/network-instance/openconfig-network-instance.yang#L240)

This is not a purely theoretical issue; OpenDaylight Nitrogen actually treats it as an error:

> java.lang.IllegalArgumentException: Parsed QName (http\://openconfig.net/yang/network-instance?revision=2019-04-16)DEFAULT_INSTANCE does not refer to a valid identity

nkitchen:
I can see now that this is more general than DEFAULT_INSTANCE -- it applies to other identity names in the module.

nkitchen:
Already fixed by commit 7d8d7f4e35ff08bc93b23fa34fa353adefccd513

PR #229, opened 2019-05-23
Add missing prefix to DEFAULT_INSTANCE in network instance model.

nkitchen:
    - add prefix to DEFAULT_INSTANCE in when statements

#228


nkitchen:
Sorry, I forgot to add a new revision for this. I will cancel this request and send a corrected one.

PR #242, opened 2019-06-26
Adding Explicit NULL label configuration for SRTE policies.

kavitharp:
Also adding a new ENUM type called explicit-null-label-type.

This new leaf node resides in the SRTE config container under
the policy rules container.

### DIFF
--- a/release/models/policy-forwarding/openconfig-pf-srte.yang
+++ b/release/models/policy-forwarding/openconfig-pf-srte.yang
@@ -159,7 +159,10 @@ module openconfig-pf-srte {
       type explicit-null-label-type;
       description
         "Type of explicit null label that is being pushed on an unlabeled
-        IP packet. If leaf is unspecified, null label is not added."
+        IP packet. If unspecified, the behaviour of the system is undefined,
+        and is subject to local implementation choice.";
+      reference
+        "draft-ietf-idr-segment-routing-te-policy - Section 2.4.4";
     }
   }
 
--- a/release/models/policy-forwarding/openconfig-pf-srte.yang
+++ b/release/models/policy-forwarding/openconfig-pf-srte.yang
@@ -63,23 +63,23 @@ module openconfig-pf-srte {
   typedef explicit-null-label-type {
     type enumeration {
       enum IPV4 {
-        description "Push an IPv4 Explicit NULL label on an unlabeled 
-                    IPv4 packet";
+        description
+          "Push an IPv4 Explicit NULL label on an unlabeled IPv4 packet";
       }
       enum IPV6 {
-        description "Push an IPv6 Explicit NULL label on an unlabeled 
-                    IPv6 packet";
+        description
+          "Push an IPv6 Explicit NULL label on an unlabeled IPv6 packet";
       }
       enum IPV4_IPV6 {
-        description "Push an IPv4 Explicit NULL label on an unlabeled 
-                    IPv4 packet; 
-                    Push an IPv6 Explicit NULL label on an unlabeled 
-                    IPv6 packet";
+        description
+          "Push an IPv4 Explicit NULL label on an unlabeled IPv4 packet;
+          Push an IPv6 Explicit NULL label on an unlabeled IPv6 packet";
       }
       enum NONE {
-        description "No Explicit NULL label will be pushed for both 
-                    IPv4 and IPv6 unlabeled packets";
-      } 
+        description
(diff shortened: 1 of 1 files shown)


kavitharp:
I signed it!

release/models/policy-forwarding/openconfig-pf-srte.yang
```diff
@@ -124,6 +154,13 @@ module openconfig-pf-srte {
         protocol, or interface to the device, or from other static
         entries configured on the system.";
     }
+
+    leaf explicit-null-label {
+      type explicit-null-label-type;
+      description
+        "Indicates type of explicit null label that is being pushed on 
+        an unlabeled IP packet"
```
robshakir: What happens when this leaf is not specified -- is `NONE` the default here? If so, I propose that we simply say that leaving this leaf unspecified says that there should be no null label added.

robshakir: (nit) Please add a full stop at the end of this sentence.

release/models/policy-forwarding/openconfig-pf-srte.yang
```diff
@@ -53,6 +59,30 @@ module openconfig-pf-srte {
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
+
+  typedef explicit-null-label-type {
+    type enumeration {
+      enum IPV4 {
+        description "Push an IPv4 Explicit NULL label on an unlabeled
```
robshakir: The layout for descriptions through OpenConfig models is:

```yang
enum IPV4 {
  description
     "Blah blah blah";
}
```
Please reformat here :-)

kavitharp: Done

release/models/policy-forwarding/openconfig-pf-srte.yang
kavitharp: done

kavitharp: Yes NONE is the default value - I have added to the description. Please let me know if that looks ok.

sulrich:
is NONE is the default case?  absent specification of this leaf, the system's default explicit-null imposition policy should take place.  specifying NONE, should _explicitly_ convey that there is to be _no_ explicit-null label imposed on any traffic. 

this would be in alignment with the text of the draft for this.

[draft](https://tools.ietf.org/html/draft-ietf-idr-segment-routing-te-policy-06) - s2.4.4
> If an Explicit NULL Label Policy Sub-TLV is not present, the decision
>  of whether to push an Explicit NULL label on a given packet is a
>  matter of local policy.

kavitharp:
> is NONE is the default case? absent specification of this leaf, the system's default explicit-null imposition policy should take place. specifying NONE, should _explicitly_ convey that there is to be _no_ explicit-null label imposed on any traffic.
> 
> this would be in alignment with the text of the draft for this.
> 
> [draft](https://tools.ietf.org/html/draft-ietf-idr-segment-routing-te-policy-06) - s2.4.4
> 
> > If an Explicit NULL Label Policy Sub-TLV is not present, the decision
> > of whether to push an Explicit NULL label on a given packet is a
> > matter of local policy.

Oops - Looks like I misunderstood! 
I have changed the description - does that look ok?

release/models/policy-forwarding/openconfig-pf-srte.yang
```diff
@@ -124,6 +154,14 @@ module openconfig-pf-srte {
         protocol, or interface to the device, or from other static
         entries configured on the system.";
     }
+
+    leaf explicit-null-label {
+      type explicit-null-label-type;
+      description
+        "Type of explicit null label that is being pushed on an unlabeled
+        IP packet. If unspecified, the decision of whether to push 
+        an Explicit NULL label on a given packet is a matter of local policy."
```
robshakir: I don't think it's a matter of local policy, right?  This is the local policy that we'd have "If unspecified, the behaviour of the system is undefined, and is subject to local implementation choice.".

robshakir: Can we add a `reference` statement to the draft and the relevant section here?

kavitharp: Changed this.

kavitharp: Done.

PR #258, opened 2019-08-02
Fix when XPath expression to resolve NI 'config/type'

exa-arrcus:


exa-arrcus:
@googlebot I signed it!

PR #259, opened 2019-08-02
Fix when XPath expression to resolve NI 'config/type'

exa-arrcus:


exa-arrcus:
Currently `type` is unresolvable for this when XPath expression
```
openconfig-network-instance.yang:402: error: the node 'type' from module 'openconfig-network-instance' (in node 'network-instance' from 'openconfig-network-instance') is not found
```

PR #262, opened 2019-08-16
Add node for per-syslog-host facility-override

exa-arrcus:


exa-arrcus:
This proposal involves:

* Remove identity `ALL` SYSLOG_FACILITY as this is not an actual facility per RFC5424 but rather an implementation behavior for match criteria only (This poses problems for re-use outside of match)
* Modify the facility filter under selectors to type `union` with an enum value `ALL` (much like is used in other models)
* Add a `facility-override` node for setting a precise facility per remote-server (This is common across major implementations and operator use)

release/models/system/openconfig-system-logging.yang
```diff
@@ -68,12 +68,6 @@ module openconfig-system-logging {
       "IETF RFC 5424 - The Syslog Protocol";
   }
 
-  identity ALL {
```
aashaikh: sorry, I did not quite understand the rationale for removing this -- how does having this identity cause issues for re-use outside of match?  (for something like listing available facilities?)

release/models/system/openconfig-system-logging.yang
```diff
@@ -433,6 +435,14 @@ module openconfig-system-logging {
         "Sets the destination port number for syslog UDP messages to
         the server.  The default for syslog is 514.";
     }
+
+    leaf facility-override {
+      type identityref {
+        base SYSLOG_FACILITY;
+      }
+      description
+        "Override the syslog facility level to the log server";
```
aashaikh: What is the behavior you expect when this leaf is specified?  I'm a little confused by "facility level" here -- do you mean to override the severity? (the identityref implies override of the facility though)

release/models/system/openconfig-system-logging.yang
exa-arrcus: `ALL` is not an actual facility.  Identities should be grouped in such a fashion for multiple uses and if we compare this to the use of unions throughout other models, there is often a need to choose between a precise data-type/identity *or* a catch-all like `ANY` or `ALL` which will have special implementation behavior.  While it might seem logical to group something as such, it would be better to take the union approach and keep these identities specific to the syslog facilities defined per RFC.

This is related to the `facility-override` knob included in this patch.  This provides the ability to override all facility levels towards a `remote-server` which can reuse these identities however `ALL` is not an actual identity to "set"

release/models/system/openconfig-system-logging.yang
exa-arrcus: No - this is to override the facility.  e.g. say all internal processes are logging at facility levels local5/local6/local7 and the operator chooses to override all towards the remote-server to local0.

This feature exists on various major implementations (see below)

```
user@junos# set system syslog host 1.1.1.1 facility-override ?
Possible completions:
  authorization        Authorization system
  daemon               Various system processes
  ftp                  FTP process
  kernel               Kernel
  local0               Local logging option number 0
  local1               Local logging option number 1
  local2               Local logging option number 2
  local3               Local logging option number 3
  local4               Local logging option number 4
  local5               Local logging option number 5
  local6               Local logging option number 6
  local7               Local logging option number 7
  user                 User processes
``` 

```
eos(config)#logging facility ?
  auth    Authorization messages    (facility=4)
  cron    Clock daemon              (facility=15)
  daemon  System daemons            (facility=3)
  kern    Kernel messages           (facility=0)
  local0  Local use 0               (facility=16)
  local1  Local use 1               (facility=17)
  local2  Local use 2               (facility=18)
  local3  Local use 3               (facility=19)
  local4  Local use 4               (facility=20)
  local5  Local use 5               (facility=21)
  local6  Local use 6               (facility=22)
  local7  Local use 7               (facility=23)
  lpr     Line printer subsystem    (facility=6)
  mail    Mail system               (facility=2)
  news    Network news subsystem    (facility=7)
  sys10   System use 10             (facility=10)
  sys11   System use 11             (facility=11)
  sys12   System use 12             (facility=12)
  sys13   System use 13             (facility=13)
  sys14   System use 14             (facility=14)
  sys9    System use 9              (facility=9)
  syslog  Internal syslogd messages (facility=5)
  user    User-level messages       (facility=1)
  uucp    UUCP subsystem            (facility=8)
  <0-23>  Syslog facility value
```

```
RP/0/RP0/CPU0:ios(config)#logging facility ?
  auth    Authorization system
  cron    Cron/at facility
  daemon  System daemons
  kern    Kernel
  local0  Local use
  local1  Local use
  local2  Local use
  local3  Local use
  local4  Local use
  local5  Local use
  local6  Local use
  local7  Local use
  lpr      Line printer system
  mail    Mail system
  news    USENET news
  sys10   System use
  sys11   System use
  sys12   System use
  sys13   System use
  sys14   System use
  sys9    System use
  syslog  Syslog itself
  user    User process
  uucp    Unix-to-Unix copy system
  <cr>    
```

exa-arrcus:
@aashaikh did the previous comments address your questions?

PR #267, opened 2019-08-22
Fix identity prefixes in openconfig-network-instance.yang when stmts

### LINKED ISSUE #42: openconfig-network-instance: various issues with when statements
Hi,

Similar to reported issues in other models I've encountered both compile errors and data validation errors with openconfig-network-instance.

1. While compiling with confdc I hit several errors related to paths in when statements. Below is one example, there were at least 10 others that need to be addressed that I have not included to keep this note brief.

openconfig-network-instance.yang:100: error: the node 'type' from module 'openconfig-network-instance' (in node 'network-instances' from 'openconfig-network-instance') is not found

In this case I modified this line:
           when "../type = 'L2VSI' or ../type = 'L2P2P'" +
                 " or ../type = 'L2L3'" {

to this and the compilation succeeded:
            when "config/type = 'L2VSI' or config/type = 'L2P2P'" +
                 " or config/type = 'L2L3'" {

2. After fixing the compile issues I've run into problems where the when statements did not validate data properly. One failing example from the protocols container is below.

            uses oc-loc-rt:local-static-top {
              when "../config/identifier = 'STATIC'" {
                description
                  "Include static route parameters only when the
                  protocol is set to static";
              }
              description
                "Configuration and state parameters relating to
                static routes";
            }

I added the prefix 'oc-pol-types' to the STATIC value to address this. I suspect other when statements in this model have the same issue but did not specifically check them all.

Thanks,
Paul

LeonGWang:

This Pull Request closes #42

### DIFF
--- a/release/models/network-instance/openconfig-network-instance.yang
+++ b/release/models/network-instance/openconfig-network-instance.yang
@@ -214,8 +214,8 @@ module openconfig-network-instance {
         }
 
         uses l2ni-instance {
-            when "config/type = 'L2VSI' or config/type = 'L2P2P'" +
-                 " or config/type = 'L2L3' or config/type = 'DEFAULT_INSTANCE'" {
+            when "config/type = 'oc-ni-types:L2VSI' or config/type = 'oc-ni-types:L2P2P'" +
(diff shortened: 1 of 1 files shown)


LeonGWang:
Fixed already merged upstream. Closing this PR.

PR #268, opened 2019-08-22
Fix quotation in openconfig-bgp-policy.yang when statements

### LINKED ISSUE #87: incorrect condition check in "when" stmts in openconfig-bgp-policy model
The comparison checks need to be updated to put the enums in quotes. Diff below. Pls note the line numbers do not match with the latest file but the issue is still valid.

``````
785c785
<       when "../config/method=INLINE" {
---
>       when "../config/method='INLINE'" {
844c844
<       when "../config/method=REFERENCE" {
---
>       when "../config/method='REFERENCE'" {
948c948
<       when "../config/method=INLINE" {
---
>       when "../config/method='INLINE'" {
1009c1009
<       when "../config/method=REFERENCE" {
---
>       when "../config/method='REFERENCE'" {
``````

Pls help address in next release.

thanks
alpesh

LeonGWang:

This Pull Request closes #87

### DIFF
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -767,7 +767,7 @@ module openconfig-bgp-policy {
       action";
 
     container inline {
-      when "../config/method=INLINE" {
+      when "../config/method='INLINE'" {
         description
           "Active only when the set-community method is INLINE";
       }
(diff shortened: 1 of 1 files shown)


robshakir:
Leon,

Thanks for this change -- it looks to me like it was merged upstream.

Much appreciated,
r.

PR #269, opened 2019-08-29
Fix route-distinguisher regexes

### LINKED ISSUE #183: route-distinguisher 2 byte number regex
The pattern for <2-byte assigned number> pattern seems incorrect. As an example, it doesn't accept a value of "10000" or "64535". Also, it accepts 2 digits values starting from 0 as in "08".

gmuloc:
* (M)       release/models/network-instance/openconfig-network-instance-types.yang

This Pull Request closes #183


gmuloc:
@googlebot I signed it!

PR #285, opened 2019-09-05
Adjust small description typo in oc-system model

exa-arrcus:


robshakir:
Thanks Ebben -- applying this fix upstream.

PR #292, opened 2019-10-06
Add must statement for grpc listen-addresses

exa-arrcus:
- Prevent simultaneous configuration of IP/ANY

### DIFF
--- a/release/models/system/openconfig-system-management.yang
+++ b/release/models/system/openconfig-system-management.yang
@@ -104,6 +104,10 @@ module openconfig-system-management {
           }
         }
       }
+      default "ANY";
+      must "(not(../listen-addresses = 'ANY' and count(../listen-addresses) > 1))" {
+        error-message "'ANY' and IP addresses are mutually exclusive";
+      }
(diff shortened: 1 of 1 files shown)


robshakir:
I'm not inclined to take this one upstream, but am open to some discussion. We have one `must` statement in the `optical-transport` model, do you think this warrants something specific here, or can we include it in the documentation that these are mutually exclusive?

exa-arrcus:
Currently this is how we implement restrictions on `listen-addresses` in general either via augments or deviations.  If we left the model open then my expectation would be that everyone needs to either deviate to provide front end restrictions or implement backend validations (which means one has to deal w/ error-messaging during the transaction) - or in some cases... accidentally leave out all restriction logic which may produce undesirable side effects and bugs.

So imo this is probably best served within the data-model since the intention is pretty straightforward and cannot think of situations where both might be defined concurrently for any reason.  For quite a few other cases I would say `must` statements can be left out but I feel for this one, being loose can create more issues in the long run.

PR #310, opened 2019-11-18
AAA server list ordering modification

exa-arrcus:
- Change AAA server list to be ordered-by user

### DIFF
--- a/release/models/system/openconfig-aaa.yang
+++ b/release/models/system/openconfig-aaa.yang
@@ -249,6 +249,7 @@ module openconfig-aaa {
 
       list server {
         key "address";
+        ordered-by user;
         description
           "List of AAA servers";
 


robshakir:
@exa-arrcus I wonder here why we'd prefer this to be `ordered-by user` rather than explicitly have a preference for each server? I think that this has the advantage that it allows one to avoid implicit information in serialised data instances.

Thanks!
r.

exa-arrcus:
@robshakir - that is acceptable as well from an implementation pov.  I guess my questions with that approach would be
* consistency across modules on this approach.  There are other uses of `ordered-by user` among other modules (and AAA infact)
* treatment in the case where preferences are defined as equal.  This approach provides the advantage possibly in that an implementation can implement round-robin load-balancing in such scenarios

PR #364, opened 2020-06-24
LACP fallback mode feature

exa-arrcus:
Proposal for support of LACP fallback feature

```module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw oc-lag:aggregation
           +--rw oc-lag:lacp
              +--rw oc-lag:fallback
                 +--rw oc-lag:config
                 |  +--rw oc-lag:mode?      lacp-fallback-type
                 |  +--rw oc-lag:timeout?   uint16
                 +--ro oc-lag:state
                    +--ro oc-lag:mode?      lacp-fallback-type
                    +--ro oc-lag:timeout?   uint16
```

Various implementations:

EOS: https://www.arista.com/assets/data/pdf/user-manual/um-eos/Chapters/Port%20Channels%20and%20LACP.pdf
IOS-XR: https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5000/interfaces/61x/b-ncs5000-interfaces-configuration-guide-61x/b-ncs5000-interfaces-configuration-guide-61x_chapter_0101.html#task_844BB227554D4C42A96F1AF8D5AC477C
Junos: https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/force-up-edit-interfaces-qfx-series.html

### DIFF
--- a/release/models/interfaces/openconfig-if-aggregate.yang
+++ b/release/models/interfaces/openconfig-if-aggregate.yang
@@ -110,6 +110,23 @@ module openconfig-if-aggregate {
       defined and managed";
   }
 
+  typedef lacp-fallback-type {
+    type enumeration {
+      enum INDIVIDUAL {
+        description
+          "Fallback to individual ports.  All LAG members
+           will be forwarding when all the members stop
+           receiving LACP PDUs";
+      }
+      enum STATIC {
+        description
+          "Fallback to static LAG mode";
+      }
+    }
+    description
+      "Type to define the LACP fallback mode";
+  }
+
   // grouping statements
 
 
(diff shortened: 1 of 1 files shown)


robshakir:
Hi Ebben,

Thanks for the research here, and the PR. I'm supportive of adding this to the models.

I wonder whether it's better specified under `/lacp` though - since we already have LACP configuration and state there, and this is "how do we fallback when LACP fails". The same content would work - just in that different location. Thoughts?

Cheers,
r.

exa-arrcus:
Thx @robshakir - what you mention is actually a common dilemma in quite a bit of model design across various domains and contemplated the same as you suggest here - There is often the question of whether nodes should sit under an interface or within a domain with an interface ref + some config/state nodes and among implementations we often see inconsistency on these boundaries in general.  We also see this across various models within OC as well.

For this case, there are a few items that likely warrant a bit of further discussion

* The `/lacp/interfaces` subtree is really more centric around the LAG interface itself (e.g. ae, bond, bundle-ether, etc..) whereas fallback configuration could be as part of member physical interfaces as well depending
* All current implementations appear to rather conform to configuration/state directly as a subset of a physical or virtual (bundle) interface thus the translations/mappings are more difficult to conform for all while having to query separate subtrees for information related to an interface (the last part holds true among other model structures as well)

PR #375, opened 2020-08-31
Cache-Bust Badges

wenovus:


wenovus:
@aashaikh Neither GitHub support nor I could find the reason why the image is being cached. This workaround seems to work for now, so maybe we can just go with it.

wenovus:
Looks like the issue has resolved on its own.

aashaikh:
thanks @wenovus;   yes, it seems now unlikely that caching was the cause here, perhaps just a glitch with the transition.  We can keep an eye on it as we publish next round of updates.

PR #385, opened 2020-10-16
Sync from private.

Ruwanthi7:
Avoid empty lines.


mike-albano:
This appears to be an old PR. software-version was added as part of [revision 0.1.1](https://github.com/openconfig/public/blob/master/release/models/wifi/openconfig-ap-manager.yang#L46).

@Ruwanthi7 Let me know if there's any outstanding issues here.

PR #409, opened 2021-02-16
Fix CI

wenovus:
The major version wasn't being used correctly in `cloudbuild.yaml`.


robshakir (approved):
Thanks for the fix Wen.

PR #413, opened 2021-03-22
Add instance-id leaf to IS-IS config container

raviv-arista:
As per [RFC8202](https://tools.ietf.org/html/rfc8202), `instance-id:uint16` is used to uniquely identify an IS-IS instance. The existing key for ISIS instances is `instance:string`, which cannot be used for the instance ID TLV 7.


raviv-arista:
@googlebot I signed it!

PR #438, opened 2021-06-30
Fix for date/time pattern statement

migueberumenjr:
* (M) release/models/types/openconfig-yang-types.yang
    - Fix date-and-time description to match RFC3339


robshakir:
Thanks for this contribution -- we'll discuss this in the review meeting and come back with any comments.

PR #442, opened 2021-07-15
NHG-ID and NH-ID space management

akalluru1:
In some systems NHG & NH ID generation is internal so it would
result in mismatch between index pushed by the client and index
created by the system.

Adding a new leaf in the next-hop-group & next-hop AFT entry to
represent the client published value. This helps the client to
create an association between configured & extracted index.


akalluru1:
@googlebot I signed it!

robshakir:
Thanks for this contribution! I'll shepherd this request and am upstreaming it now.

sthesayi (approved):

PR #448, opened 2021-08-06
Fix extended community regex bugs

jbemmel:
The current regex would not allow target:4200000000:1234 or any ASN in the 4-byte private range of 4200000000..4289999999

The maximum 32-bit unsigned value is 10 digits: 4[0-1][0-9]{7} only allows 9

Fix: 4[0-1][0-9]{7} => 42[0-8][0-9]{7}|4[0-1][0-9]{8}


jbemmel:
@googlebot I signed it!

robshakir:
Thanks for this contribution. I'll shepherd this proposal  -- thanks for fixing it!

I've upstreamed this to the internal WG repo.

dplore:
Merged at #1024

PR #449, opened 2021-08-06
Add regression test case for 4-byte private ASN 4200000000

jbemmel:


robshakir:
Thanks for this change -- especially adding the tests. I've upstreamed this.

dplore:
merged as #449

PR #450, opened 2021-08-06
Add protobuf-metadata annotation.

robshakir:
```
 * (A) release/models/extensions/openconfig-metadata.yang
  - Add an annotation that can be used to store binary-marshalled
    protobuf contents within a YANG data tree.
 * (A) third_party/ietf/ietf-yang-metadata.yang
  - Import the YANG metadata annotation module from IETF. This
    module is vendored and subject to the IETF licensing.
```

### DIFF
--- a/third_party/ietf/ietf-yang-metadata.yang
+++ b/third_party/ietf/ietf-yang-metadata.yang
@@ -1,0 +1,84 @@
+module ietf-yang-metadata {
+
+  namespace "urn:ietf:params:xml:ns:yang:ietf-yang-metadata";
+
+  prefix "md";
+
+  organization
+    "IETF NETMOD (NETCONF Data Modeling Language) Working Group";
+
+  contact
+    "WG Web:   <https://datatracker.ietf.org/wg/netmod/>
+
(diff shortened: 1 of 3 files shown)


third_party/ietf/ietf-yang-metadata.yang
```diff
@@ -0,0 +1,84 @@
...
+               <mailto:kwatsen@juniper.net>
+
+     Editor:   Ladislav Lhotka
+               <mailto:lhotka@nic.cz>";
+
+  description
+    "This YANG module defines an 'extension' statement that allows
+     for defining metadata annotations.
+
+     Copyright (c) 2016 IETF Trust and the persons identified as
```
sthesayi, line 28: nit: Just wanted to confirm if the year should be updated to 2021?

sthesayi (approved):

third_party/ietf/ietf-yang-metadata.yang
robshakir: Since this is a vendored file we leave the copyright as per the upstream. Thanks.

robshakir:
This has been merged upstream.

PR #451, opened 2021-08-06
Support IP-in-IP encapsulation, resolution in alternate NI in AFT.

robshakir:
```
 * (M) yang/aft/*.yang
  - Support resolving a next-hop-group in a network instance other
    than the local one. This change fixes an existing leafref in
    a backwards compatible way (making it more permissive), and
    adds a new leaf referencing the network-instance name.
  - Support IP-in-IP encapsulation being described in the AFT model.
  - Add support for a per-entry metadata (up to 8 bytes) being reported
    with each AFT entry.

  - These changes upstream some changes from
    github.com/openconfig/gribi.
```

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -22,7 +22,15 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "0.6.0";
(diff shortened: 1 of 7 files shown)


robshakir:
@wenovus -- this is an interesting ygot bug that we might need to look at. In the case that we have a leaf that is called `metadata` - and we're also generating metadata fields in ygot, we end up with a name clash for the metadata field. I suspect we probably need to change the container-level metadata's struct field name.

sthesayi (approved):
The changes look good but we will need to fix ygot.

PR #452, opened 2021-08-06
Support referencing a non-local network-instance in PF action.

robshakir:
```
  * (M)yang/policy-forwarding/openconfig-pf-forwarding-policies.yang
  * (M)yang/policy-forwarding/openconfig-pf-interfaces.yang
  * (M)yang/policy-forwarding/openconfig-pf-path-groups.yang
  * (M)yang/policy-forwarding/openconfig-policy-forwarding.yang
   - Fix a reference in the policy-forwarding module where the
     network-instance reference was relative such that it allowed only
     the local network-instance's name to be referenced. Make this an
     absolute reference.
```

### DIFF
--- a/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
+++ b/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
@@ -260,7 +260,7 @@ submodule openconfig-pf-forwarding-policies {
 
     leaf network-instance {
       type leafref {
-        // this must be an absolute reference to allow another NHG to be
+        // this must be an absolute reference to allow another NI to be
         // referenced.
         path "/network-instances/network-instance/config/name";
       }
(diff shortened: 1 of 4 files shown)


release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -253,12 +260,9 @@ submodule openconfig-pf-forwarding-policies {
 
     leaf network-instance {
       type leafref {
-
-        // We are at:
-        // $NIROOT/policy-forwarding/policies/
-        // policy/rules/rule/action/config/
-        // network-instance
-        path "../../../../../../../../config/name";
+        // this must be an absolute reference to allow another NHG to be
```
akshaygattani, line 263: Rob - just a nit. In the comments, did we mean "another NI" instead of "another NHG"?

itbwtw (approved):
Verified that this does fix the support of referencing a non-local network-instance

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
robshakir: Yes, apologies. Fixed.

sthesayi (approved):

robshakir:
Merged upstream.

PR #460, opened 2021-08-11
Fix IS-iS area address regex

dan-lepage:
The existing one requires a trailing `.`, e.g. `49.0001.`; it should instead accept `49.0001`.

### DIFF
--- a/regexp-tests/openconfig-isis-types-test.yang
+++ b/regexp-tests/openconfig-isis-types-test.yang
@@ -1,0 +1,26 @@
+module openconfig-isis-types-test {
+  prefix "rt";
+  namespace "urn:openconfig-isis-types-test";
+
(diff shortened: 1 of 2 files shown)


wenovus:
Hi Dan,
Thanks for the change. There is some extra work if you don't mind to preserve backwards compatibility:
- Add tests for the other two `typedefs` as well.
- Remove anchors on the `pattern` statements (still keep them on `posix-pattern`).
- Remove `oc-ext:regexp-posix` on line 83.

As for doing local testing unfortunately the code is in a private repo.. maybe we can make it public one day.

wenovus:
Thanks! I'll leave someone else familiar with IS-IS to review the regexes and tests. Otherwise LGTM.

PR #463, opened 2021-08-17
Add ip-address-zoned union type for link-local neighbor addresses

mcarri01:
Similar to the existing typedef for ip-address, this new typedef
provides a union between existing ipv4-address-zoned and
ipv6-address-zoned types that are used by link-local peering.

Currently systems such as Arista's EOS support link-local peering.
In particular when dealing with BGP IPv6 link-local peering, the current
neighbor-address key is incompatible as it requires an ip-address type.
With this new union typedef, it may be possible to update the
neighbor-address key for BGP peer information to use this new
"ip-address-zoned" key instead to allow supporting link-local peers.
This change should be backwards compatible as it simply relaxes the
restrictions on the key type.

### DIFF
--- a/release/models/types/openconfig-inet-types.yang
+++ b/release/models/types/openconfig-inet-types.yang
@@ -31,7 +31,14 @@ module openconfig-inet-types {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.4.1";
+  oc-ext:openconfig-version "0.4.2";
+
+  revision "2021-08-17" {
+    description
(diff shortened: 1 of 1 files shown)


mcarri01:
@googlebot I signed it!

robshakir:
Thanks for the contribution, this LGTM. I'll shepherd this upstream.

mcarri01:
> 

Thanks Rob, appreciate. it. Just wanted to check in to see what the timeline would be for this to get shepherded upstream? Had honestly forgot about this PR and realized it was still open.

dplore:
I think this request may be obsoleted by #793 .   If you disagree, please feel free to request reopening this PR.

PR #474, opened 2021-09-10
State of Polarization Enhancement

abhinava:
Augmenting OpenConfig to include state and config parameters to track state of polarization data on optical-channel.


abhinava:
oops sorry.. Merged with the wrong master :)

Issue #478, opened 2021-09-15
Issue with when statement in openconfig-network-instance yang file


bhramidi:
We are seeing an issue with the when statement with the latest network instance yang file.

https://github.com/openconfig/public/blob/e9e3a82693d1f26c61d7fbf85b3b2d0418d4af9e/release/models/network-instance/openconfig-network-instance.yang#L743

 
Line 743: the when statement should ideally be ‘when “./config/identifier = oc-pol-types:OSPF’ instead of just “./config/identifier =‘OSPF’
Same applies for all protocols (identifier types: BGP, ISIS, PIM etc.

We are seeing the following failure with when statement as it was expecting OSPF alone.

```
Sending:

#1595
<nc:rpc xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0" message-id="urn:uuid:fd37f28d-ef3f-4d93-89b2-4324103a228b"><nc:edit-config>
    <nc:target>
      <nc:running/>
    </nc:target>
    <nc:config>
      <network-instances xmlns="http://openconfig.net/yang/network-instance">
        <network-instance>
          <name>default</name>
          <config>
            <name>default</name>
          </config>
          <protocols>
            <protocol>
              <identifier xmlns:oc-pol-types="http://openconfig.net/yang/policy-types">oc-pol-types:OSPF</identifier>
              <name>1</name>
              <config>
                <identifier xmlns:oc-pol-types="http://openconfig.net/yang/policy-types">oc-pol-types:OSPF</identifier>
                <name>1</name>
              </config>
              <ospfv2>
                <areas>
                  <area>
                    <identifier>1</identifier>
                    <config>
                      <identifier>1</identifier>
                    </config>
                    <interfaces>
                      <interface>
                        <id>GigabitEthernet3</id>
                        <config>
                          <id>GigabitEthernet3</id>
                        </config>
                      </interface>
                    </interfaces>
                  </area>
                </areas>
              </ospfv2>
            </protocol>
          </protocols>
        </network-instance>
      </network-instances>
    </nc:config>
  </nc:edit-config>
</nc:rpc>

##
Received message from host
<?xml version="1.0" ?>
<rpc-reply message-id="urn:uuid:fd37f28d-ef3f-4d93-89b2-4324103a228b" xmlns="urn:ietf:params:xml:ns:netconf:base:1.0" xmlns:nc="urn:ietf:params:xml:ns:netconf:base:1.0">
  <rpc-error>
    

    <error-type>application</error-type>
[...]

PR #482, opened 2021-09-22
Update openconfig-network-instance.yang

pullururajasekhar:
Fix proposal for the git issue https://github.com/openconfig/public/issues/478
   
Without adding the type oc-pol-types to the when statement, rpc reports failure due to existing when statement


pullururajasekhar:
> @googlebot I signed it!

@googlebot I signed it!

PR #503, opened 2021-10-14
Update YAML linter to match public directory structure.

robshakir:
```
commit 79af7e194ed7e79afb0e4d60225f236f947b7bfb
Author: Rob Shakir <robjs@google.com>
Date:   Thu Oct 14 12:02:36 2021 -0700

    Update YAML linter to match public directory structure.
```


wenovus (changes requested):
For scripts/create_concatenated_spec_file.sh

Can you do a stat on the spec files to make sure we're not validating an empty file?

I'm thinking
```
FILES="**/*/*/.spec.yml"
stat $FILES && cat $FILES | sed ...
```

wenovus (approved):
Thanks!

PR #504, opened 2021-10-16
Add health variables to components model.

robshakir:
```
commit 54c8618962211fe8334c45d531f79c1928689070
Author: Rob Shakir <robjs@google.com>
Date:   Sat Oct 16 09:19:36 2021 -0700

    Add telemetry-on-change annotation to health variables.
    
     * (M) release/yang/platform/openconfig-platform.yang
      - Add extension indicating that the health variables should be
        sent based on when they change.

commit 987f504239beecf7e6081f9ff276afd1382b1628
Author: Rob Shakir <robjs@google.com>
Date:   Sat Oct 16 09:17:06 2021 -0700

    Add component health indicators.
    
     * (M) release/yang/platform/openconfig-platform.yang
      - Per the definition of the gnoi.Healthz service in
        github.com/openconfig/gnoi - indicate whether a component is
        healthy or unhealthy following a check on the system. This
        allows the system to signal to external subscribers that the
        component is not in a healthy state such that further
        actions can be taken.
```

See the `gnoi.Healthz` [documentation](https://github.com/openconfig/gnoi/tree/master/healthz) for discussion of how these variables interact with the healthz service.

### DIFF
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -596,6 +596,7 @@ module openconfig-platform {
       }
       description
         "The status of the component, indicating its current health.";
+      oc-ext:telemetry-on-change;
     }
 
     leaf last-unhealthy {
@@ -604,6 +605,7 @@ module openconfig-platform {
         "The time at which the component as last observed to be unhealthy
         represented as nanoseconds since the Unix epoch. Unhealthy is defined
         as the component being in a DEGRADED or FAILED state.";
+      oc-ext:telemetry-on-change;
     }
 
     leaf unhealthy-count {
@@ -611,6 +613,7 @@ module openconfig-platform {
       description
         "The number of status checks that have determined this component
         to be in an unhealthy (DEGRADED or FAILED) state.";
+      oc-ext:telemetry-on-change;
     }
   }
 
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -534,6 +534,86 @@ module openconfig-platform {
     }
   }
 
+  grouping platform-health-top {
+    description
+      "Grouping containing health-related parameters.";
+
+    container health {
+      description
+        "The health of the component. The paramaters within this
+        container indicate the status of the component beyond whether
+        it is operationally up or down. When a signal is received
+        that a component is in an unhealth state the gNOI.Healthz
+        service can be used to retrieve further diagnostic information
+        relating to the component.
+
+        The contents of this directory relate only to the specific
+        component that it is associated with. In the case that child
+        components become unhealthy and this causes a parent component
+        to be unhealthy, the new unhealthy status should be reported at
+        both components, such that an interested system can take the
+        relevant actions (e.g., retrieve the Healthz output, or
+        apply mitigation actions)."
+      reference
+        "https://https://github.com/openconfig/gnoi/tree/master/healthz"
+
+      container state {
+        config false;
+        description
+          "Operational state parameters relating to component health.";
+        uses platform-health-state;
+      }
+    }
+  }
+
+  grouping platform-health-state {
+    description
+      "Operational state parmaeters relating to a platform component's
+      health.";
+
+    leaf status {
+      type enumeration {
+        enum HEALTHY {
+          description
+            "The component is in a HEALTHY state, and is operating
+            within the expected parameters.";
+        }
+        enum DEGRADED {
+          description
+            "The component is in a degraded state. That is to say that
+            it has not completely failed, but failures have occurred that
+            either reduce its ability to perform, or may cause it to reach
+            an unhealthy state should further failures occur.
+
+            An example of a degraded state may be a fan tray which has only
+            a subset of the fans it contains having failed.";
+        }
+        enum FAILED {
+          description
+            "The component is in a failed state, it is unhealthy and not
+            performing the function expected of it.";
+        }
+      }
+      description
+        "The status of the component, indicating its current health.";
+    }
+
+    leaf last-unhealthy {
+      type oc-types:timeticks64;
+      description
+        "The time at which the component as last observed to be unhealthy
+        represented as nanoseconds since the Unix epoch. Unhealthy is defined
+        as the component being in a DEGRADED or FAILED state.";
+    }
+
+    leaf unhealthy-count {
+      type uint64;
+      description
+        "The number of status checks that have determined this component
+        to be in an unhealthy (DEGRADED or FAILED) state.";
+    }
+  }
+
   grouping pcie-uncorrectable-errors {
     description
       "PCIe uncorrectable error statistics.";
@@ -1026,6 +1106,7 @@ module openconfig-platform {
         uses platform-component-properties-top;
         uses platform-subcomponent-ref-top;
         uses platform-anchors-top;
+        uses platform-health-top;
       }
     }
   }


robshakir:
@marcushines @sulrich - PTAL

release/models/platform/openconfig-platform.yang
```diff
@@ -534,6 +541,89 @@ module openconfig-platform {
...
+        "The status of the component, indicating its current health.";
+      oc-ext:telemetry-on-change;
+    }
+
+    leaf last-unhealthy {
+      type oc-types:timeticks64;
+      description
+        "The time at which the component as last observed to be unhealthy
+        represented as nanoseconds since the Unix epoch. Unhealthy is defined
+        as the component being in a DEGRADED or FAILED state.";
```
robshakir, line 614: Reword to say that this should be any state other than `HEALTHY`.

release/models/platform/openconfig-platform.yang
```diff
@@ -534,6 +541,89 @@ module openconfig-platform {
...
+          description
+            "The component is in a HEALTHY state, and is operating
+            within the expected parameters.";
+        }
+        enum DEGRADED {
+          description
+            "The component is in a degraded state. That is to say that
+            it has not completely failed, but failures have occurred that
+            either reduce its ability to perform, or may cause it to reach
+            an unhealthy state should further failures occur.
```
sulrich, line 593: a slight nit.  DEGRADED is still in an unhealthy state.  which is reinforced with the `last-unhealthy` treatment below.
potentially change to:

... or may cause it to reach a FAILED state should further failures or degradation occur.

release/models/platform/openconfig-platform.yang
```diff
@@ -534,6 +541,89 @@ module openconfig-platform {
...
+
+            An example of a degraded state may be a fan tray which has only
+            a subset of the fans it contains having failed.";
+        }
+        enum FAILED {
```
dplore, line 598: These do not align with gNOI healthz, probably they should?  

https://github.com/openconfig/gnoi/blob/3e9a99fe151c1801b9fe2fbd7b68ca0a066396fc/healthz/healthz.proto#L53-L57

earies: @robshakir - reference https://github.com/openconfig/gnoi/pull/65 as well as this is directly related

earies: @marcushines @robshakir 

The current gNOI healthz proto has the following enum definition
```
enum Status {
  STATUS_UNSPECIFIED = 0;   // The path doesn't support healthz
  STATUS_HEALTHLY = 1;      // The path is healthy
  STATUS_UNHEALTHY = 2;     // The path is unhealthy
}
```
Whereas the YANG enum mapping is as follows (since values are not specified)
```
HEALTHY = 0
DEGRADED = 1
FAILED = 2
```
I would suggest alignment here at a minimum and may want to consider tagging values in the YANG enum definition such as:

```
UNSPECIFIED = 0
HEALTHY = 1
DEGRADED = 2
FAILED = 3
```

release/models/platform/openconfig-platform.yang
```diff
@@ -534,6 +541,89 @@ module openconfig-platform {
     }
   }
 
+  grouping platform-health-top {
+    description
+      "Grouping containing health-related parameters.";
+
+    container health {
```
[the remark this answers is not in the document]
hellt: is it ok to have a discrepancy in the naming.
gnoi -> healthz
oc -> health

[right now](https://github.com/openconfig/gnoi/tree/6bddacd72ef8aa06304b887df414a43c8e63ff27/healthz#design) gnoic.healthz spec mentiones that `healthz` container is going to be present in OC

release/models/platform/openconfig-platform.yang
```diff
@@ -534,6 +541,89 @@ module openconfig-platform {
...
+      description
+        "The health of the component. The paramaters within this
+        container indicate the status of the component beyond whether
+        it is operationally up or down. When a signal is received
+        that a component is in an unhealth state the gNOI.Healthz
```
hellt, line 553: ...is in an unhealth**y** state...?

release/models/platform/openconfig-platform.yang
```diff
@@ -534,6 +541,89 @@ module openconfig-platform {
...
 
+  grouping platform-health-top {
+    description
+      "Grouping containing health-related parameters.";
+
+    container health {
+      description
+        "The health of the component. The paramaters within this
+        container indicate the status of the component beyond whether
+        it is operationally up or down. When a signal is received
```
hellt, line 552: Am I reading this correctly, that the only difference the `health` status conainer provides right now when compared to oper-status=up/down is the new state called "Degraded", which can indicate a partially operational status?

release/models/platform/openconfig-platform.yang
marcushines, line 548: i would prefer healthz

release/models/platform/openconfig-platform.yang
```diff
@@ -534,6 +541,89 @@ module openconfig-platform {
...
+      }
+    }
+  }
+
+  grouping platform-health-state {
+    description
+      "Operational state parmaeters relating to a platform component's
+      health.";
+
+    leaf status {
```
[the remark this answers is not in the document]
marcushines: is there a linkage between oper-status in all things vs. status

i feel like oper-status is very interface specific

release/models/platform/openconfig-platform.yang
marcushines: i think the status should align - as far as the value of degraded - if that in fact means to say - hey i am probably aggregating other things - so degraded means i am good but some of my subcomponents are not ... meh not sure i see the value - basically it becomes a very noisy signal for probably not much good

i.e. if a line card defines it's health around interface status (which it shouldn't) then it would be in a degraded state likely often but that actually doesn't reflect anything of value really

now if the linecard defines it's status around optic health - and there is an installed optic which is not reporting - that status degraded would make sense...

but again it wouldn't tell you anything about the actual linecard
vs say a switch fabric being off line and that linecard actually not having full capacity ... that would make alot of sense for degraded state

release/models/platform/openconfig-platform.yang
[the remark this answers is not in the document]
robshakir: I agree -- I think that the pattern here of having a separate enum makes sense to keep this generic.

release/models/platform/openconfig-platform.yang
robshakir: Done.

release/models/platform/openconfig-platform.yang
robshakir: I think given the ambiguity of `DEGRADED` I'll remove it. I haven't added `UNSPECIFIED` as generally we include this in the protobuf to handle the case that the `proto3` field is not set, and YANG doesn't have this implicit default value. Code generation creates it where required for the language.

release/models/platform/openconfig-platform.yang
robshakir: Fixed. Thanks.

release/models/platform/openconfig-platform.yang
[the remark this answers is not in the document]
robshakir: Fixed the naming to `healthz`.

release/models/platform/openconfig-platform.yang
robshakir: I'd imagine that it is possible to have such a case in systems that don't take auto-remediation of a particular case. For example, let's consider a disk that is full -- potentially, we might have `oper-status = ACTIVE` (given that we might not be able to offline such a disk if it was mounted as `/` for example), but 100% usage which would give us `status = UNHEALTHY`.

release/models/platform/openconfig-platform.yang
robshakir: Removed `DEGRADED` per discussion in other comments.

release/models/platform/openconfig-platform.yang
```diff
@@ -663,6 +669,80 @@ module openconfig-platform {
...
+      type enumeration {
+        enum HEALTHY {
+          description
+            "The component is in a HEALTHY state, and is operating
+            within the expected parameters.";
+        }
+
+        enum FAILED {
```
marcushines, line 717: enum Status {
  STATUS_UNSPECIFIED = 0;   // The path doesn't support healthz
  STATUS_HEALTHY = 1;       // The path is healthy
  STATUS_UNHEALTHY = 2;     // The path is unhealthy
}

why not unhealthly vs failed
also i think you do need to model unspecified?

marcushines (approved):

release/models/platform/openconfig-platform.yang
robshakir: We discussed and mentioned `UNSPECIFIED` will mean explicitly that the device has not checked the status of the component.

marcushines (approved):

dplore:
Minor suggestion here, but the platform model is getting large.   The health variables seem like they could be placed in a separate file and augment component state?  WDYT?

robshakir:
Sure, that's fine with me. I don't think this materially changes anything.

I still think that we should cut 1.0.0 of the `openconfig-platform` model though. I can do this in a separate PR.

dplore:
> Sure, that's fine with me. I don't think this materially changes anything.
> 
> I still think that we should cut 1.0.0 of the `openconfig-platform` model though. I can do this in a separate PR.

Agreed on all points.  The benefits are it makes for a smaller file and modularizes the models so implementations have another option /  cleaner way to express support on a per module (file) basis.

dplore:
This was presented at the openconfig operators meeting today with no specific comments for or against.

There was extensive discussion on the topic of alarms,  thresholds, and observation that healthz is also an alarm mechanism (a tree of boolean values indicating health of an entity).   This was inspired by several PR's which are related to thresholds and alarms.  We observe there are several different patterns for alarms/health indication from booleans and thresholds embedded as leaves within the relevant component to the generic (but opaque) /system/alarms model.   Each of these approaches has their own tradeoffs.  

OpenConfig could benefit from a more opinionated view on how alarms and thresholds should be implemented so we have a more predictable and consistent pattern.  However until we have rough consensus on a pattern to follow, I don't see a reason to block development.  

Therefore I will ask for last call for comments for this PR, to be merged in 2 weeks on Feb 7, 2023

release/models/platform/openconfig-platform-healthz.yang
```diff
@@ -0,0 +1,127 @@
...
+
+        The contents of this directory relate only to the specific
+        component that it is associated with. In the case that child
+        components become unhealthy and this causes a parent component
+        to be unhealthy, the new unhealthy status should be reported at
+        both components, such that an interested system can take the
+        relevant actions (e.g., retrieve the Healthz output, or
+        apply mitigation actions).";
+      reference
+        "https://https://github.com/openconfig/gnoi/tree/master/healthz";
```
hellt, line 62:
```suggestion
        "https://github.com/openconfig/gnoi/tree/main/healthz";
```

release/models/platform/openconfig-platform-healthz.yang
```diff
@@ -0,0 +1,127 @@
...
+        description
+          "Operational state parameters relating to component health.";
+        uses platform-health-state;
+      }
+    }
+  }
+
+  grouping platform-health-state {
+    description
+      "Operational state parmaeters relating to a platform component's
```
hellt, line 75:
```suggestion
      "Operational state parameters relating to a platform component's
```

release/models/platform/openconfig-platform-healthz.yang
```diff
@@ -0,0 +1,127 @@
...
+        represented as nanoseconds since the Unix epoch. Unhealthy is defined
+        as the component being in a state other than HEALTHY.";
+      oc-ext:telemetry-on-change;
+    }
+
+    leaf unhealthy-count {
+      type uint64;
+      description
+        "The number of status checks that have determined this component
+        to be in an unhealthy (DEGRADED or FAILED) state.";
```
hellt, line 116: we don't have these two statuses anymore
```suggestion
        to be in an unhealthy state.";
```

hellt (approved):

release/models/platform/openconfig-platform-healthz.yang
```diff
@@ -0,0 +1,127 @@
...
+            performing the function expected of it.";
+        }
+      }
+      description
+        "The status of the component, indicating its current health.";
+      oc-ext:telemetry-on-change;
+    }
+
+    leaf last-unhealthy {
+      type oc-types:timeticks64;
```
hellt, line 104: @robshakir 
the `timeticks64` type says it is a delta between two epochs - https://github.com/openconfig/public/blob/acd14524556aa4b95692558d0252cfa04c4c5ac5/release/models/types/openconfig-types.yang#L104

Shouldn't it rather use a UNIX time at the time the state change occurs?

hellt: I stand corrected, this type seems to assume the delta between the unix epoch and event time

release/models/platform/openconfig-platform-healthz.yang
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    leaf unhealthy-count {
+      type uint64;
+      description
+        "The number of status checks that have determined this component
+        to be in an unhealthy state.";
+      oc-ext:telemetry-on-change;
+    }
+  }
```
hellt, line 119: I wonder, should this counter increase when we use Healthz.Get and Healthz.Check RPC? Or should it instead only increase when the system detects the component unhealthy by doing an automatic check?

For example, if a component_A is in unhealthy state, should the counter increase when we do multiple Healthz.Get check targeting this component while it stays in the unhealthy state?

release/models/platform/openconfig-platform-healthz.yang
robshakir: The type is always between two epochs, but the description here clarifies that for this leaf this is since the unix epoch :)

release/models/platform/openconfig-platform-healthz.yang
robshakir: I think this should only be incremented when the device itself finds the component to be unhealthy, regardless of how many times a user triggers another check. We can clarify the leaf's description to be clearer if this is useful,

The justification for this is that it's probably not that useful to have a counter that says "you triggered 100 checks on this component and they are all unhealthy", but it is probably reasonable to say that "this component became unhealthy twice" (i.e., it transitioned between HEALTHY and UNHEALTHY twice), since this tells you that this is a new `UNHEALTHY` state, not the last one you have mitigated.

@marcushines - PTAL.

robshakir: Done :-)

dplore (approved):

PR #509, opened 2021-10-21
Remove 'exclude-filter' from telemetry model

earies:
* Remnants from original YANG subscribe RPC that did not carry over to
  gNMI so no support for dynamic subscriptions
* No definition of filter language for opaque string value w/ marking
  for possible removal added some time back


earies:
Leaving out semver changes in the case changes are aggregated prior to publish

robshakir:
I think there's a bigger conversation to have here - given that we have moved on quite significantly from the model that this describes. I'd propose that we look at what else needs to be removed from this module.

@gcsl - thoughts here?

PR #510, opened 2021-10-21
Remove top-level /bgp

earies:
* /bgp has been re-rooted under /network-instances for a few years with
  no intention to support standalone at the top-level
* This causes confusion for support and deviations are necessary for
  various implementations


robshakir:
Merging this upstream. Thanks @earies !

robshakir:
Merged upstream -- thanks Ebben.

PR #512, opened 2021-10-25
Field to enable EXTENDED_NEXTHOP_ENCODING negotiation

rcdomigan:
We would like to support forwarding of IPv4 routes with IPv6 next
hops (RFC 5549). Support generally needs to be enabled explicitly, and
this commit provides a flag with which to do so.


robshakir:
Hi @rcdomigan - this looks to duplicate #513 - but not cover the per-AFI/SAFI configuration. Can you perhaps review #513 and see if it looks good to you? Thanks.

rcdomigan:
I saw #513, I don't think I'm empowered to approve it, but I think it would satisfy my need. I meant to withdraw this PR earlier -- I'll close it now.

PR #519, opened 2021-11-11
Add dot1p leaf as configuration option on L2 conditions and terms

rpazyaquian:
### Original Problem

We need to support the class-of-service value on Ethernet packets as specified in dot1p.

### Proposed Solution

Add a `dot1p` leaf as a configuration option to `ethernet-header-config`, used in the `l2` container under the `ethernet-header-top` grouping.

### Note

This is my first contribution to the codebase and to Yang code in general, so I am still ramping up on proper dev practices. Please comment with any improvements I can make!

### DIFF
--- a/release/models/acl/openconfig-packet-match.yang
+++ b/release/models/acl/openconfig-packet-match.yang
@@ -27,7 +27,13 @@ module openconfig-packet-match {
     field is omitted from a match expression, the effect is a
     wildcard ('any') for that field.";
 
-  oc-ext:openconfig-version "1.3.1";
+  oc-ext:openconfig-version "1.3.2";
+
+  revision "2021-11-11" {
+    description
+      "Add dot1p configuation option to Ethernet.";
+    reference "1.3.2";
+  }
 
   revision "2021-06-16" {
     description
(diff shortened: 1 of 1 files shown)


release/models/acl/openconfig-packet-match.yang
```diff
@@ -118,6 +118,17 @@ module openconfig-packet-match {
       description
         "Ethertype field to match in Ethernet packets";
     }
+
+    leaf dot1p {
```
rpazyaquian, line 122: This was taken directly from https://github.com/openconfig/public/blob/9152b712c79d0fcd776ec4b8efc9f655f6c5b916/release/models/qos/openconfig-qos-elements.yang#L1311. I was unsure of how to reuse the other definition or whether I _should_, so I chose to just copy it. Is this acceptable, or is there a way to reduce repetition?

rpazyaquian:
Not sure how to interpret the failed checks. Is there something else I need to do?

rpazyaquian:
Closing this for now, since there are open questions re: use cases for L2 QoS-based classifications. If we require this in the future, we can certainly open the PR back up!

PR #522, opened 2021-11-18
Update sFlow and restructure top-level sampling container

earies:
* (A) release/models/sampling/openconfig-sampling.yang
  - Add sampling module with top-level container for sampling attachment
    point
* (M) release/models/sampling/openconfig-sampling-sflow.yang
  - Relocate attachment to augment top-level sampling module
  - Split sampling-rate to ingress/egress at global/interface level
  - Add support for polling-interval at global/interface level
  - Add support for setting IPv4/IPv6 agent-id
  - Add support for setting IPv4/IPv6 source-addresses
  - Add support for DSCP marking of sFlow PDUs per collector

Proposal notes:

Top-level structure
------------------------
Previously, the top-level `sampling` container was bound to sFlow under the
sflow namespace.  To allow for expansibility and modularity of other sampling
methods (e.g. NETFLOW, IPFix), the sampling container has been moved to it's
own module to define an attachment point for other modules to augment into.

Sampling rates
------------------------
The previous leaf of `sampling-rate` has been removed in favor of 2 new leafs
`ingress-sampling-rate` and `egress-sampling-rate`.  For implementations that
only support a single sampling rate (usually ingress), then a deviation can be
produced to not support egress however this proposal allows for both in an
expansible fashion.

References:
- Cisco IOS-XE: https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst1000/software/releases/15_2_7_e/configuration_guides/sys_mgmt/b_1527e_sys_mgmt_c1000_cg/configuring_sflow.pdf
- Juniper JUNOS: https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/sample-rate-edit-protocols-sflow.html

Polling interval
------------------------
A new leaf of `polling-interval` is added to both global and per-interface
level.  This is a widely supported node among implementations.

References:
- Arista EOS: https://www.arista.com/en/um-eos/eos-sflow#xx1152469
- Juniper JUNOS: https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/polling-interval-edit-protocols-sflow-ex-series.html
- Cisco IOS-XE: https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst1000/software/releases/15_2_7_e/configuration_guides/sys_mgmt/b_1527e_sys_mgmt_c1000_cg/configuring_sflow.pdf

Agent ID
------------------------
The `agent-id` is part of the sFlow specification in RFC3176 as
`sFlowAgentAddress` which is a IPv4 or IPv6 address embedded in the sFlow PDU
as the Agent address.  This leaf is not mutually exclusive and can be chosen
based off the address-family in use for the collector.

References:
- RFC3176: https://datatracker.ietf.org/doc/html/rfc3176
- Cisco IOS-XE: https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst1000/software/releases/15_2_7_e/configuration_guides/sys_mgmt/b_1527e_sys_mgmt_c1000_cg/configuring_sflow.pdf
- Juniper JUNOS: https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/agent-id-edit-protocols-sflow.html
- SONiC: https://github.com/Azure/SONiC/blob/master/doc/sflow/sflow_hld.md
- Pica8: https://docs.pica8.com/display/PicOS41sp/protocols+sflow+agent-id

Source Address
------------------------
The `source-address` leaf is removed in favor of specific IPv4/IPv6 variants.
This leaf is not mutually exclusive and can be used based off the
address-family in use for the collector (since you can have both IPv4 and IPv6
collectors concurrently)

QoS/DSCP Marking
------------------------
A new leaf `originated-qos-marking` is introduced per collector.  This
resembles other dial-out/response services defined within OpenConfig.


earies:
@robshakir @dplore - could one of you PTAL - thx

earies:
@robshakir @dplore - Just looping back to get a review on this - thx

earies:
@robshakir @dplore - Bumping again for review - thx

robshakir:
Thanks for the ping here.

release/models/sampling/openconfig-sampling-sflow.yang
```diff
@@ -58,15 +67,8 @@ module openconfig-sampling-sflow {
         for a specific interface.";
     }
 
-    leaf sampling-rate {
```
robshakir, line 61: I wonder whether it's better to have something like:

```yang

leaf rate-mode {
    type enumeration {
       enum SINGLE; // single rate for ingress and egress
       enum DIRECTIONAL; // rate for ingress/egress separately.
    }
}
```

and then leave this leaf along with `ingress-rate` and `egress-rate`. The concern I have with the current proposal is that it's rather implicit that you specify `ingress-rate` and this actually also implies egress rate too.

release/models/sampling/openconfig-sampling-sflow.yang
```diff
@@ -146,6 +148,13 @@ module openconfig-sampling-sflow {
         sFlow collector.  If uspecified, the collector destination
         is reachable in the default network instance.";
     }
+
+    leaf originated-qos-marking {
```
robshakir, line 152: Oops, apologies, I missed that this was also included here vs. my proposal in #656. I'll close that PR. Should this just be `dscp`?

Also, should this be configured per collector? It looks from the survey in #656 that this isn't the common support, and rather it is should be globally set for all collectors.

robshakir (changes requested):

release/models/sampling/openconfig-sampling-sflow.yang
earies: The only reason for the name was consistency to the only other published leaf today that represents the marking of sourced control/management traffic which falls in `openconfig-telemetry` today.  Since it is likely that the `openconfig-telemetry` model will evolve and iterate possibly spread to other domains, I'm fine w/ using slightly better terminology that can be brought across to other domains that lack this functionality but do exist in shipping implementations today.

Since IPv6 TC carries the 6-bits as "DSCP" the same as in IPv4, I'm fine changing this to just `dscp` - will push in next commit

release/models/sampling/openconfig-sampling-sflow.yang
```diff
@@ -222,25 +274,36 @@ module openconfig-sampling-sflow {
...
-        The allowable sampling rate range is generally a
-        property of the system, e.g., determined by the
-        capability of the hardware.";
+        "Sets the agent identifier for IPv6 PDUs.";
+      reference
+        "RFC 3176 - InMon Corporation's sFlow: A Method for
+        Monitoring Traffic in Switched and Routed Networks";
+    }
+
+    leaf source-address-ipv4 {
```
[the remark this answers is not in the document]
earies: That is correct - defined globally for all collectors

release/models/sampling/openconfig-sampling-sflow.yang
earies: The reason for the removal and replacement of ingress/egress separately is it seems most (all?) implementations support ingress only sampling or ingress/egress sampling - I'm not sure the `SINGLE` enum variant above applies but I'd have to see an implementation reference of a single rate applying to both ingress/egress simultaneously.

I was thinking in the case an implementation supports ingress only - then you could deviate out the egress rate as `not-supported`

e.g. https://www.arista.com/en/um-eos/eos-sflow#xx1152507

Welcome to commentary on above: @rolandphung 

In the case an implementation supports ingress + egress, both are available and can be configured independently with no expectation you must configure both

release/models/sampling/openconfig-sampling-sflow.yang
earies: Also as far as setting DSCP globally vs. collector - this is a common vendor neutral issue where I do see a split of implementation behaviors.  This can be moved up to a global configuration and for the implementations that support per-collector would then mean all collectors inherit the same global value.

release/models/sampling/openconfig-sampling-sflow.yang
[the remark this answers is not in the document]
robshakir: SGTM.

release/models/sampling/openconfig-sampling-sflow.yang
robshakir: I think this makes sense -- it seems unlikely to me that operationally one would need to specify different DSCP values for sFlow traffic based on collector.

robshakir (approved):
LGTM.

@dplore -- do you want to take a second look before we merge?

release/models/sampling/openconfig-sampling-sflow.yang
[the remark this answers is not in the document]
earies: @robshakir - Rethinking this one a bit - I've relocated the source addresses out of global configuration to each collector instance and aggregated under a single leaf node of type `oc-yang:ip-address`.  Since each collector in the model is network-instance aware, a single source address at the global level does not make much sense.

Commit: https://github.com/openconfig/public/pull/522/commits/ad4199f774e9037100f3921e1e1304932ec161bf

Tree excerpt:
```
module: openconfig-sampling
  +--rw sampling
     +--rw oc-sflow:sflow
        +--rw oc-sflow:collectors
           +--rw oc-sflow:collector* [address port]
              +--rw oc-sflow:config
                 +--rw oc-sflow:address?            oc-inet:ip-address
                 +--rw oc-sflow:port?               oc-inet:port-number
                 +--rw oc-sflow:source-address?     oc-inet:ip-address
                 +--rw oc-sflow:network-instance?   oc-netinst:network-instance-ref
```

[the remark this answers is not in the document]
dplore: Nice catch, probably the most common scenario is an sflow collector would use a source ip in a management vrf.

release/models/sampling/openconfig-sampling-sflow.yang
dplore: This has been updated to use a leaf name dscp which SGTM.

dplore (commented):
LGTM, pending comments from @rgwilton and @rolandphung regarding DSCP marking for sflow collector.  Will wait until up to July 9 for reply.  (2 weeks plus a couple days)

release/models/sampling/openconfig-sampling-sflow.yang
earies: @dplore - Actually - per this comment thread, commit https://github.com/openconfig/public/pull/522/commits/397a38c1af507d57476e8f2510af1ac1afd494c5 brought this to common global marking rather (where collectors would inherit the global marking)

```
module: openconfig-sampling
  +--rw sampling
     +--rw oc-sflow:sflow
        +--rw oc-sflow:config
        |  +--rw oc-sflow:dscp?   oc-inet:dscp
        +--rw oc-sflow:state
           +--rw oc-sflow:dscp?   oc-inet:dscp
```

rolandphung: DSCP for EOS is set at a global level: https://www.arista.com/en/um-eos/eos-user-security#:~:text=switch(config)%23-,sFlow%20qos%20dscp,-36

So this LGTM

release/models/sampling/openconfig-sampling-sflow.yang
rolandphung: Having a separate ingress/egress rate configuration sounds good, we only support ingress configuration only. This looks to be in the global configuration container as well which works well for us.

rolandphung (approved):

dplore (approved):

pphaal:
> The separation of sampling-rate has been removed in favor of 2 new leafs
> ingress-sampling-rate and egress-sampling-rate.

The [sFlow specification](https://sflow.org/developers/specifications.php) requires symmetric sampling rates when configuring bidirectional sampling. Adding support for bidirectional or egress sampling can be implemented with the addition of a direction attribute with ingress|egress|bidirectional values and reverting the ingress-sampling-rate/egress-sampling-rate settings back to the original sampling-rate setting. 

This proposed change would ensure that the agent configuration complies with the sFlow specification and prevent generation of non-conforming sFlow packet samples that would result in erroneous traffic reports from sFlow analyzers.

In the mean time, configuration verification code should flag situations where both ingress-sampling-rate and egress-sampling-rate are set and have different values.

PR #524, opened 2021-11-18
Initial commit of gNMI authorization policy proto.

morrowc:
Documentation (README.md) for the Authorization policy has been
included along with the initial protobuf definition.

### DIFF
--- a/release/authorization/README.md
+++ b/release/authorization/README.md
@@ -1,0 +1,366 @@
+# gNMI Authorization Protocol
+
+## Objective
+
+This proto and the reference code(to be delivered seperately) serve to describe
+an authorization framework for controlling which OpenConfig paths of a network
+device specific users can access using a gNMI micro-serivice. The authorization
+policy is initially intended to be deployed to a device, with the ability to
+define:
+
+*   Policy rules - each rule defines a single authorization policy.
+*   Groups of users - as a method to logically group users in the administrative
+    domain, for instance: operators or administrators.
+*   Users which are individual users as part of either rules or in groups.
+    Authentication information is not included in this Authorization
+    configuration.
+
+Policy rules are matched based on the best match for the authorization request,
+not the first match against a policy rule. Best match enables a configuration
+which permits a user or group access to particular OpenConfig paths while
+denying subordinate portions of the permitted paths, or the converse.
+
+Match rules permit a match against:
+
+*   User or Group (not both)
+*   an OpenConfig path
+
+An implicit deny is assumed, if there is no matching rule in the policy. Logging
+may be specified on a per-policy-rule basis as well as a default for the whole
+authorization policy.
+
+[OpenConfig paths](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#222-paths)
+are heirarchical, rooted at a defined "origin". OpenConfig may contain paths
+such as:
+
+```proto
+    /a/b/c/d
+```
+
+Paths may also have attributes associated with the path elements such as:
+
+```proto
+    /a/b[key=foo]/c/d
+```
+
+Attributes may be wildcarded in the policy, such as:
+
+```proto
+    /a/b[key=*]/c/d
+```
+
+Wildcards are only valid in policy rules when applied to attributes.
+Permitted use:
+
+```proto
+    /a/b[key=*]/c/d
+```
+
+Not permitted use:
+```proto
+    /a/b[key=foo]/*/d
+```
+
+## Bootstrap / Install Options
+
+System bootstrap, or install, operations may include an authorization policy
+delivered during bootstrap operations. It is suggested that the bootstrap
+process include the complete authorization policy so all production tools
+and services have immediate authorized access to finish installation and
+move devices into production in a timely manner.
+
+Using the Secure Zero Touch Provisioning (sZTP - RFC8572) process for
+bootstrap/installation provides a method to accomplish this delivery, and
+the delivery of all other bootstrap artifacts in a secure manner.
+
+## Conflict Resolution
+
+Policy evalyation should end with a single best match policy for the provided
+user / path. If there is more than one 'best match', an error must be logged
+and evaluation should return a failure.
+
+## An Example Authorization Protobuf
+
+```proto
+# proto-file: //experimental/users/morrowc/gnXi/proto/authorization.proto
+# proto-message: Policies
+version: "UUID-1234-123123-123123"
+created_on: 1234567890
+# Define 2 groups.
+group {
+  name: "family-group"
+  members { name: "stevie" }
+  members { name: "brian" }
+}
+group {
+  name: "test-group"
+  members { name: "crusty" }
+  members { name: "the-clown" }
+}
+# Action stevie to access /this/is/a/message_path
+policy {
+  id: "one"
+  log_level: LOG_NONE
+  path {
+    origin: "foo"
+    elem { name: "this" }
+    elem { name: "is" }
+    elem { name: "a" }
+    elem { name: "message_path" }
+  }
+  action: PERMIT
+  user { name: "stevie" }
+}
+# Action members of family-group to access /this/is/a/different/message_path
+policy {
+  id: "two"
+  log_level: LOG_BRIEF
+  path {
+    origin: "foo"
+    elem { name: "this" }
+    elem { name: "is" }
+    elem { name: "a" }
+    elem { name: "different" }
+    elem { name: "message_path" }
+  }
+  action: PERMIT
+  group { name: "family-group" }
+}
+# Demonstrate a key with an attribute defined.
+policy {
+  id: "key"
+  log_level: LOG_BRIEF
+  path {
+    origin: "foo"
+    elem { name: "this" }
+    elem { name: "is" }
+    elem { name: "a" }
+    elem {
+      name: "keyed"
+      key{
+        key: "Ethernet"
+        value: "1/2/3"
+      }
+    }
+    elem { name: "message_path" }
+  }
+  action: PERMIT
+  group { name: "test-group" }
+}
+# Demonstrate a key with a wildcard attribute.
+policy {
+  id: "wyld"
+  log_level: LOG_BRIEF
+  path {
+    origin: "foo"
+    elem { name: "this" }
+    elem { name: "is" }
+    elem { name: "a" }
+    elem {
+      name: "keyed"
+      key{
+        key: "Ethernet"
+        value: "*"
+      }
+    }
+    elem { name: "message_path" }
+  }
+  action: PERMIT
+  group { name: "family-group" }
+}
+# Demonstrate a key with a wildcard attribute and a user specific match.
+# The previous policy matches all family-group users and permits a command
+# path, the policy rule below specifically denies brian access to these paths.
+policy {
+  id: "wyld-stallions"
+  log_level: LOG_BRIEF
+  path {
+    origin: "foo"
+    elem { name: "this" }
+    elem { name: "is" }
+    elem { name: "a" }
+    elem {
+      name: "keyed"
+      key{
+        key: "Ethernet"
+        value: "*"
+      }
+    }
+    elem { name: "message_path" }
+  }
+  action: DENY
+  group { name: "brian" }
+}
+# Add a final rule which is an explicit deny rule.
+policy {
+  id: "explicit-deny"
+  log_level: LOG_FULL
+  action: DENY
+}
+```
+
+The example first policy rule:
+
+```proto
+# Action stevie to access /this/is/a/message_path
+policy {
+  id: "one"
+  log_level: LOG_NONE
+  path {
+    origin: "foo"
+    elem { name: "this" }
+    elem { name: "is" }
+    elem { name: "a" }
+    elem { name: "message_path" }
+  }
+  action: PERMIT
+  user { name: "stevie" }
+}
+```
+
+permits the singular user "stevie" to access the path:
+
+```shell
+    /this/is/a/message_path
+```
+
+Additionally, "stevie" is permitted access to all paths below the defined path,
+such as:
+
+```shell
+    /this/is/a/message_path/the
+    /this/is/a/message_path/the/one
+    /this/is/a/message_path/the/one/that
+    /this/is/a/message_path/the/one/that/knocks
+```
+
+The second policy rule:
+
+```proto
+# Action memberss of family-group to run /this/is/a/different/message_path
+policy {
+  id: "two"
+  log_level: LOG_BRIEF
+  path {
+    origin: "foo"
+    elem { name: "this" }
+    elem { name: "is" }
+    elem { name: "a" }
+    elem { name: "different" }
+    elem { name: "message_path" }
+  }
+  action: PERMIT
+  group { name: "family-group" }
+}
+```
+
+example policy permits members or the family-group access to a single path:
+
+```shell
+    /this/is/a/different/message_path
+```
+
+and all path elements beyond "message_path":
+
+```shell
+    /this/is/a/different/message_path/foo
+    /this/is/a/different/message_path/bar
+    /this/is/a/different/message_path/foo/baz/bing/boop
+```
+
+The third policy rule:
+
+```proto
+# Demonstrate a key with an attribute defined.
+policy {
+  id: "key"
+  log_level: LOG_BRIEF
+  path {
+    origin: "foo"
+    elem { name: "this" }
+    elem { name: "is" }
+    elem { name: "a" }
+    elem {
+      name: "keyed"
+      key{
+        key: "name"
+        value: "Ethernet1/2/3"
+      }
+    }
+    elem { name: "message_path" }
+  }
+  action: PERMIT
+  group { name: "test-group" }
+}
+```
+
+Permits access by the "test-group" users to the keyed path:
+
+```shell
+    /this/is/a/keyed[name=Ethernet1/2/3]/message_path
+```
+
+and all path elementas as beyond "message_path". The final policy rule:
+
+```proto
+# Demonstrate a key with a wildcard attribute.
+policy {
+  id: "wyld"
+  log_level: LOG_BRIEF
+  path {
+    origin: "foo"
+    elem { name: "this" }
+    elem { name: "is" }
+    elem { name: "a" }
+    elem {
+      name: "keyed"
+      key{
+        key: "name"
+        value: "*"
+      }
+    }
+    elem { name: "message_path" }
+  }
+  action: PERMIT
+  group { name: "family-group" }
+}
+```
+
+permits access by the "family-group" users to the keyed path, with no
+restrictions on the key values:
+
+```shell
+    /this/is/a/keyed[name=Ethernet1/2/3]/message_path
+    /this/is/a/keyed[name=POS3]/message_path
+    /this/is/a/keyed[name=Serial4/1]/message_path
+    /this/is/a/keyed[name=HSSI2]/message_path
+```
+
+Additionally, the path elements beyond "message_path" are available for access
+to this group as well.
+
+The wildcard character "*" (asterisk) is only able to be used as a value in
+keyed elements. The wildcard is only used to mask out all possible values but
+not portions of values, for instance:
+
+```shell
+    /this/is/a/keyed[name=*]/things - permitted usage of wildcard
+    /this/is/a/keyed[name=Ethernet1/*/3]/things - NOT permitted usage of wildcard
+```
+
+The policy rule:
+
+```proto
+# Add a final rule which is an explicit deny rule.
+policy {
+  id: "explicit-deny"
+  log_level: LOG_FULL
+  action: DENY
+}
+```
+
+provides an explcit deny for any request wich does not match any other policy
+rule. This rule also requests that the result be logged in full fidelity.
+
+Hopefully this illustrates the acceptable usage of the wildcard character in
+path elements of authorization policy rules.
--- a/release/authorization/authorization.proto
+++ b/release/authorization/authorization.proto
@@ -1,0 +1,115 @@
+// Authorization for gnXi operations.
+//
+// Define a list of policies, where the most specific policy is applied
+// on the device authorizing the action.
+//
+// Users may be referenced by user name, or as a group with a group name.
+// Groups are defined and managed in the authorization policy itself.
+//
+// Paths may be referenced in whole or in complete parts, ie:
+//   /interfaces/interface[name=Ethernet1/2/3]/state/counters
+//   /interfaces/interface[name=*]/state/oper-status
+//   /network-instances/network-instance/tables/table[protocol=BGP][address-family=*]
+//
+// Paths are gnmi.Path protobufs.
+//
+// The most specific match is returned for a request, this means the longest
+// path and the most specific user match (user preferred over group).
+//
+// Additionally, the policy maintains a timestamp of creation
+// and version number from the underlying version control system.
+// The version/timestamp are available to requestors to verify which version
+// of policy is being applied at the time of the request.
+//
+syntax = "proto3";
+
+package authorization.proto;
+
+import "third_party/openconfig/gnmi/proto/gnmi/gnmi.proto";
+
+option java_multiple_files = true;
+option go_api_flag = "OPAQUE_V0";
+
+// User is a singular username used only in the matching criteria.
+message User {
+  string name = 1;
+}
+
+// Group is a group of users, groups are a construct of the policy configuration
+// as a method to collect many users with the same authorizations together.
+message Group {
+  string name = 1;
+  repeated User members = 2;
+}
+
+// LogLevel is set for all RPCs (via default_log_level, or per-RPC specifically.
+enum LogLevel {
+  LOG_NONE = 0;   // Don't pollute the logs with high volume RPCs.
+  LOG_BRIEF = 1;  // RPC name and calling Principal name.
+  LOG_FULL = 2;   // RPC, Principal, Request, and Response.
+}
+
+// Action is the defined action for an AuthorizationPolicy.
+enum Action {
+  UNKNOWN = 0;
+  DENY = 1;
+  PERMIT = 2;
+}
+
+message AuthorizationPolicy {
+  // Policy Identifier, a unique string per policy/rule, eg: uuid.
+  string id = 1;
+  LogLevel log_level = 2;
+
+  oneof principal {
+    User user = 3;
+    Group group = 4;
+  }
+
+  // Path is the request path, longest prefix (by path elements from left
+  // to right).
+  gnmi.Path path = 5;
+
+  // Permit or deny the user/group access to the path specified.
+  Action action = 6;
+}
+
+message AuthorizationPolicies {
+  repeated AuthorizationPolicy policies = 1;
+  // Version string of this policy, a UUID4 or equivalent locally significant
+  // value.
+  string version = 2;
+  // Release created-on (in seconds since epoch) for this policy.
+  int64 created_on = 3;
+  // Default RPC log level, used only if an RPC has no defined logging level.
+  LogLevel default_log_level = 4;
+  // Groups of usernames collected for simplicity of policy expression.
+  repeated Group groups = 5;
+}
+
+// AuthorizationRequest contains a single user name and
+// gnmi Path being attempted. Data returned to an RPC Caller should
+// adhere to the policy.
+message AuthorizationRequest {
+  string user = 1;
+  gnmi.Path path = 2;
+}
+
+// AuthorizationReply returns the ack/nack for a single user request
+// as evaluated against the current policy, along with the policy id
+// and current policy version.
+message AuthorizationReply {
+  Action action = 1;
+  string id = 2;
+  string version = 3;
+}
+
+// VersionRequest requires no content, this is a simple status request.
+message VersionRequest {}
+
+// VersionReply returns only the version string and timestamp
+// of policy creation.
+message VersionReply {
+  string version = 1;
+  int64 created_on = 2;
+}


morrowc:
It looks like the yanglint error is from prior to my activity... based on the below snippet from it's output.

no_entry  wifi
no_entry  openconfig-access-points
heavy_dollar_sign  bash command
err : Schema node not found. (/openconfig-access-points:access-points/openconfig-access-points:access-point/openconfig-ap-interfaces:interfaces)
err : Invalid value "/openconfig-access-points:access-points/openconfig-access-points:access-point/openconfig-ap-interfaces:interfaces/openconfig-ap-interfaces:interface/openconfig-ap-interfaces:subinterfaces/openconfig-ap-interfaces:subinterface/openconfig-ap-interfaces:ipv4/openconfig-ap-interfaces:unnumbered" of "deviation".
err : Module "openconfig-ap-interfaces" parsing failed.
white_check_mark  openconfig-ap-manager
white_check_mark  openconfig-wifi-types

earies:
Currently, an authorization policy defines a `Path` with an `Action` (e.g. PERMIT/DENY) bound to a user or group however you may want users/groups that have privileges to either only read or edit specific subtrees.  Controlling at the RPC layer is too generic (e.g. Get() or Set()) so at a minimum I think the following needs to be considered here:

* A permission value bound to each policy (e.g. READ_ONLY, READ_WRITE)
* Permissions bound to a user or group

Currently, groups only have a member list of users and users only have a name identifier.  There could be consideration to expand users and groups to have attributes associated with them as well

The first could possibly be conveyed as something like:

```
message AuthorizationPolicy {
    string   id        = 1;
    LogLevel log_level = 2;

    oneof principal {
        User  user  = 3;
        Group group = 4;
    }

    gnmi.Path  path       = 5;
    Action     action     = 6;
    Permission permission = 7;
}

enum Permission {
    READ_ONLY  = 0;
    READ_WRITE = 1;
}
```

release/authorization/authorization.proto
```diff
@@ -0,0 +1,115 @@
...
+}
+
+message AuthorizationPolicy {
+  // Policy Identifier, a unique string per policy/rule, eg: uuid.
+  string id = 1;
+  LogLevel log_level = 2;
+
+  oneof principal {
+    User user = 3;
+    Group group = 4;
```
earies, line 66: Since these use the message types directly, this means that each principal within a policy contains the entire message construct each time.  For the case of a `Group`, while you can define these at the upper message layer of `AuthorizationPolicies`, the best reuse would be to only reference the User.name and Group.name only (especially if they are ever expanded with additional attributes)

morrowc: (I have no idea how to quote the original conversation here... either github sucks or I suck at github, I'm leaning on 2nd)

Eben: Agreed about the need for a RO/RW setting on the authorization policy.

For your second comment/suggestion i'm not sure I follow. As I was thinking about this originally the specification of a user or group in a policy (this policy applies to USER-X or GROUP-Y) was meant as a method to either: 
  "All of these users have this permission/policy"
or:
  "This one user has a specific policy/permission"

There's not a constraint on whether the username / groupname must be unique (between the two), though there could be. I'm probably misunderstanding your comment, however.

morrowc: fyi: Added the RO/RW bits to the proto, as a new ENUM and as a message element in  the AuthorizationPolicy message.

mint570: The second comment means to use "string group" instead of "Group group" IIUC. Technically, we can just fill the group name in "Group" here. But changing the type to string makes it more clear.

jsterne:
The readme is nice and detailed with lots of good illustrative examples - thx.

It might be good to clarify what "best match" is in the readme (it is in the proto though).

Is best match not a bit unusual though for these types of authorization rules ? Won't operators be more used to ordered lists of rules and find it easier to map their current approaches to something more like ACLs ? 

Here's an example of an ordered list of paths & actions that can give different access to subordinate portions (put the more specific matches first):

/this/is/a/message_path/the/one/that/knocks
permit

/this/is/a/message_path
deny

/this/is/a
permit

release/authorization/authorization.proto
morrowc: @mint570 ok, how would you differentiate the policy for
  user: Robert Oot - root
  group: root

or any other clash in user/group name uniqueness?
I didn't want to force the creator of the policy to have to make user/group names unique.
I suppose we COULD do that, it's not totally unreasonable, but it's not what I was thinking :)

mint570: The proto field name tells it is a user or a group.
BTW, shouldn't user and group be a repeated field? We can have multiple users and groups on the same path.

morrowc:
> The readme is nice and detailed with lots of good illustrative examples - thx.
> 
> It might be good to clarify what "best match" is in the readme (it is in the proto though).
> 

fair! I'll add the example set to the Readme.

> Is best match not a bit unusual though for these types of authorization rules ? Won't operators be more used to ordered lists of rules and find it easier to map their current approaches to something more like ACLs ?
> 

welp.. So, in practice, I have a slew of AAA rules, and I have to juggle these in odd ways to permit X  but not Y.
This is painful for me and prone to errors by others :( (and me, I make lots of errors!).

Taking the example:
  "Permit jsterne@ access to /foo/bar/baz but not /foo/bar/baz/beep but still permit /foo/bar/baz/beep/burp"
    (repeat this for a bunch of groups or users)

it's simpler to say (and think about, I think) this as a best-match sort of setup.

> Here's an example of an ordered list of paths & actions that can give different access to subordinate portions (put the more specific matches first):
> 
> /this/is/a/message_path/the/one/that/knocks permit
> 
> /this/is/a/message_path deny
> 
> /this/is/a permit

Agreed you can do this, it's complicated when you have  bunch of rules or conflicting sorts of requirements.
You are going to pay for this complexity somewhere, I thought that best match might instead be simpler to wrangle :)

I have some code that I planned to publish as a reference library to do this sort of evaluation, it's fairly straightforward and should help both creating and testing policies, and even could be used for actually evaluating the path/user/policy in a real system.

release/authorization/authorization.proto
morrowc: @mint570 I'm ready to believe you, so compiling the current proto:
  protoc --proto_path=. --go_out=. --go_opt=paths=source_relative  authorization.proto 

Gets me this snippet:
  https://gist.github.com/morrowc/542b0eacf4b7f3286210523793c28acd

I believe this code snippet, which did work before sending this PR, extracts the group/user content from a policy:
  https://gist.github.com/morrowc/d46fdc063eedb6cedf9bc77f75019ea3

This just, really, checks first:
  "Is there a user in the policy? and is that user the same as that which the policy evaluation is being performed for?"

Swapping the User/Group enum for just a string:
  https://gist.github.com/morrowc/022f4bfb2d8fda68019d330f5c03a5c5

The compiled bits are:
  https://gist.github.com/morrowc/e837b30e6713386d9dc5202dee542444

This seems like it will work out in the same manner in the code above, minus requesting the .GetName().
Cool, probably :) (see updated commit and thanks!)

For the repeated field conversation, internally I think the answer was really:
  It's simpler to do this as:
     "If you have many users with the same policy, make a group"
     "If you have many groups that need the same policy, make a group"
  in places where we've intentionally grouped things:
    "joe, jane, sam, sally all need this access"
  we've invariably ended up needing to split these as we add/remove access for
  one/some of the users in the list... Often this happens after we remove acesss for userX
  promptly breaking userY along the way.

Effectively it might be a little more verbose, but leaves the door open for easy and clear flexibility.

morrowc:
> > The readme is nice and detailed with lots of good illustrative examples - thx.
> > It might be good to clarify what "best match" is in the readme (it is in the proto though).
> 
> fair! I'll add the example set to the Readme.

I pulled some of the text from the proto definition into the README to explain
'best match'. Feel free to suggest changes to that text.

morrowc:
Seeing no other movement on this PR.. I'll go bug robjs@ to move it forward/merge it.

release/authorization/README.md
```diff
@@ -0,0 +1,370 @@
...
+and services have immediate authorized access to finish installation and
+move devices into production in a timely manner.
+
+Using the Secure Zero Touch Provisioning (sZTP - RFC8572) process for
+bootstrap/installation provides a method to accomplish this delivery, and
+the delivery of all other bootstrap artifacts in a secure manner.
+
+## Conflict Resolution
+
+Policy evalyation should end with a single best match policy for the provided
```
mint570, line 82: There can be certain ambiguous cases in the current proto. And also, we cannot treat multiple "best match" is due to invalid config. Here are some of the ambiguous cases:

1. What happy if two policy rules conflict with each other (same path, same user, same level but different action)? Should we honor the deny rule, the first rule, the last rule?
Example:
Rule 1: /a/b/c -> allow
Rule 2: /a/b/c -> deny

2. Each policy rule has a level (read_only or read_write), should we match on the exact level? (The examples given in this doc do not have level.)
Example 1:
Rule 1: /a/b/c -> read_only allow
Rule 2: /a/b/c/d -> read_write deny
If we have a read request on /a/b/c/d, should we match on rule 1 or rule 2? The expected behavior is that the user should be able to read /a/b/c/d.
Example 2:
Rule 1: /a/b/c -> read_only deny
Rule 2: /a/b/c/d -> read_write allow
If we have a read request on /a/b/c/d, should we match on rule 1 or rule 2? The expected behavior is that the user should be able to read /a/b/c/d.
Example 3:
Rule 1: /a/b/c -> read_only deny
Rule 2: /a/b/c -> read_write allow
Is the user allow to read /a/b/c? How about write?

3. From #2, what happen if a user matches both a user rule and a group rule?
Example 1:
Rule 1: /a/b/c -> user read_write deny
Rule 2: /a/b/c -> group read_only allow
If a user matches both rules, do they have read access on /a/b/c?

Example 2:
Rule 1: /a/b/c -> user read_write deny
Rule 2: /a/b/c -> group read_write allow
If a user matches both rules, do they have read access on /a/b/c? The intention of the rule is that the group has write access except one user.

4. We also have wildcard. What is "best match" mean?
Example 1:
Rule 1: /a/b[name=mgmt_intf]/c -> deny
Rule 2: /a/b[name=*]/c/d -> allow
If we have a request of /a/b[name=mgmt_intf]/c/d, it matches both rules. Rule 2 has longer prefix. However, the intention of the rule is that we deny access on the management interface (mgmt_intf).

Example 2:
Rule 1: /a/b[k1=*]/c/d[k2=v2]/e -> deny
Rule 2: /a/b[k1=v1]/c/d[k2=*]/e -> allow
If we have a request of /a/b[k1=v1]/c/d[k2=v2]/e, it matches both rules.

Example 3:
Rule 1: /a/b[k1=*][k2=v2]/c/d -> deny
Rule 2: /a/b[k1=v1][k2=*]/c/d -> allow
Multiple keys in the same path element is rare, but it is possible in the proto. A request of /a/b[k1=v1][k2=v2]/c/d will match both rules.

release/authorization/README.md
```diff
@@ -0,0 +1,370 @@
+# gNMI Authorization Protocol
+
+## Objective
```
mint570, line 3: The spec should also explains the different handling in read deny and write deny.
A single RPC request might involve multiple leaf paths or subtrees. Intuitively, for read request, if a leaf path is denied for read, that leaf path will not be returned to the user, but the user should get reply on other paths that they have read access. And the read RPC should succeed. For write request, if one of the path is denied in a write request, the whole RPC should fail with permission deny error.

morrowc: updated the README to include documentation regarding this point.

release/authorization/README.md
morrowc: > There can be certain ambiguous cases in the current proto. And also, we cannot treat multiple "best match" is due to invalid config. Here are some of the ambiguous cases:
> 
> 1. What happy if two policy rules conflict with each other (same path, same user, same level but different action)? Should we honor the deny rule, the first rule, the last rule?
>    Example:
>    Rule 1: /a/b/c -> allow
>    Rule 2: /a/b/c -> deny

For this case, I'd prefer that we error during ingestion of the config, actually.
This should force the operator to review/test their configuration prior to sending it to the device(s).

The other option to consider is to just follow a relatively standard:
  "Last write wins"

I think this will lead to ambiguity and troubleshooting problems, so I'm in favor of the first which is validation/testing prior to delivery to the device.

> 2. Each policy rule has a level (read_only or read_write), should we match on the exact level? (The examples given in this doc do not have level.)

I'll add some more detail to the examples. (I may steal your examples in fact!)

>    Example 1:
>    Rule 1: /a/b/c -> read_only allow
>    Rule 2: /a/b/c/d -> read_write deny
>    If we have a read request on /a/b/c/d, should we match on rule 1 or rule 2? 
>    The expected behavior is that the user should be able to read /a/b/c/d.

My intent is that the best match here is 'read_write for the longest path', so in your example the
2nd rule should match and cause a DENY to be returned. I bet we need to be able to separately
define READ and WRITE access, and not conjoin RW, I updated the proto definition to reflect that
decision/change.

>    Example 2:
>    Rule 1: /a/b/c -> read_only deny
>    Rule 2: /a/b/c/d -> read_write allow
>    If we have a read request on /a/b/c/d, should we match on rule 1 or rule 2?
>    The expected behavior is that the user should be able to read /a/b/c/d.

This rule set and request should end with a best match of RW /a/b/c/d, Rule 2.

>    Example 3:
>    Rule 1: /a/b/c -> read_only deny
>    Rule 2: /a/b/c -> read_write allow
>    Is the user allow to read /a/b/c? How about write?

This points out the need to separately define R and W and not conjoin RW, though I think
we need to permit setting one, or both in the same policy, otherwise there's the need
to specify the same rule 2x, one with read and one with write, for some usage patterns.

> 3. What happen if a user matches both a user rule and a group rule?
>    Example 1:
>    Rule 1: /a/b/c -> user read_write deny
>    Rule 2: /a/b/c -> group read_only allow
>    If a user matches both rules, do they have read access on /a/b/c?

The user rule is the most specific, so that rule would apply.
In this case user has RW access to /a/b/c

>    Example 2:
>    Rule 1: /a/b/c -> user read_write deny
>    Rule 2: /a/b/c -> group read_write allow
>    If a user matches both rules, do they have read access on /a/b/c?
>    The intention of the rule is that the group has write access except one user.

Also, in this case, user is denied RW access to /a/b/c, because the user rule is the most specific.

> 4. We also have wildcard. What is "best match" mean?
>    Example 1:
>    Rule 1: /a/b[name=mgmt_intf]/c -> deny
>    Rule 2: /a/b[name=*]/c/d -> allow
>
>    If we have a request of /a/b[name=mgmt_intf]/c/d, it matches both rules.
>    Rule 2 has longer prefix. However, the intention of the rule is that we deny
>    access on the management interface (mgmt_intf).

I believe the user would not have access to the mgmt_intf/c content, but would be able to view the /d level data, yes.
and for all other interfaces c/d would be available.

>    Example 2:
>    Rule 1: /a/b[k1=*]/c/d[k2=v2]/e -> deny
>    Rule 2: /a/b[k1=v1]/c/d[k2=*]/e -> allow
>    If we have a request of /a/b[k1=v1]/c/d[k2=v2]/e, it matches both rules.

Specified keys are a longer match than a wildcard (this I'll add to the documentation).
So in this case the user would be denied access.

>    Example 3:
>    Rule 1: /a/b[k1=*][k2=v2]/c/d -> deny
>    Rule 2: /a/b[k1=v1][k2=*]/c/d -> allow
>    Multiple keys in the same path element is rare, but it is possible in the proto.
>    A request of /a/b[k1=v1][k2=v2]/c/d will match both rules.

As above I think this request would be denied.
I think these both end as denials because the most specific path is the deny
path in both cases.

mint570: Seems like markdown remotes the "*" in my comment.
There are still some ambiguity. I suggest to have a section on detailed steps to apply the config in the other comment.

release/authorization/README.md
```diff
@@ -0,0 +1,407 @@
...
+*   Users which are individual users as part of either rules or in groups.
+    Authentication information is not included in this Authorization
+    configuration.
+
+Policy rules are matched based on the best match for the authorization request,
+not the first match against a policy rule. Best match enables a configuration
+which permits a user or group access to particular OpenConfig paths while
+denying subordinate portions of the permitted paths, or the converse.
+
+Best, or most specific, match is that which has the longest match to the
```
mint570, line 23: Let's add a section on what exactly is "best match" means.
Some suggestion text (just for reference):

- Authorization will build a tree with the configuration. If there is conflict in the config (with same path, same user, same level, but different action), then ...
- Authorization will take a user name, gnmi path, access level (read or write), and return a boolean (or match rule).
- Authorization will walk the tree with the given gnmi path. Record the matched rule. Final decision will be applied on the last matched rule.
- Matched rule means that the user must match the user name or group in the config node, and/or the access level must match in the config node.
- In key check, authorization will first check the specific key match subtree. If there is a match, return (instead of checking the wildcard subtree) EVEN if the wildcard subtree might return a longer matched rule. If there is no match, continue with the wildcard subtree.
- For multiple key check (just suggestion), authorization will sort the keys. Apply specific match with the keys in order.
- ...

release/authorization/authorization.proto
```diff
@@ -0,0 +1,125 @@
...
+// The version/timestamp are available to requestors to verify which version
+// of policy is being applied at the time of the request.
+//
+syntax = "proto3";
+
+package authorization.proto;
+
+import "github.com/openconfig/gnmi/proto/gnmi/gnmi.proto";
+
+option go_package = "github.com/openconfig/public/release/authorization";
```
earies, line 30: I'll assume if this is going to fall under gNSI then this content will likely make its way to a separate repo (following suit w/ gnmi, gnoi, etc..) and thus go_package name altogether?

release/authorization/authorization.proto
```diff
@@ -0,0 +1,125 @@
...
+// path and the most specific user match (user preferred over group).
+//
+// Additionally, the policy maintains a timestamp of creation
+// and version number from the underlying version control system.
+// The version/timestamp are available to requestors to verify which version
+// of policy is being applied at the time of the request.
+//
+syntax = "proto3";
+
+package authorization.proto;
```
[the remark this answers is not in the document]
morrowc: ok :) naming is easy, done.

release/authorization/authorization.proto
morrowc: probably, yes. Or, I don't really care where the package lands, someone in OC land can determine the best placement :) when they do we can update this path.

release/authorization/README.md
morrowc: oh :( it also changed my \* into \_ :( argh, markdown!

release/authorization/README.md
morrowc: @mint570 I like your additional text options. I added text and arranged the README to take it into account for most of your suggested bits. I shied away from the text about implementation details for 'checking specific keys then wildcards', and for sorting multiple key checks because they both sounds very much like implementation specific optimizations.

I hope the rewording brings the clarity you were after.
Happy to clarify more if required.

release/authorization/README.md
morrowc: note: backslash<star> works fine: \*

#sillymarkdownescapes

morrowc:
Ok, one more time around the watercooler with this, if it'd work better to chat over video/audio please let me know and I can set something up at a suitable time.

Alternately if we're ok with the current state, we can move this along as of thurs jan 6 2022.

jsterne:
Hi guys,

The association of users to groups is good. But I'm a little unsure of how some common router implementations would handle the fact that polices (aka rules) can be associated with either a group *or* a user.

Cisco IOS-XR, Nokia SR OS and JUNOS all seem to have the concept of command auth rules that apply to a group, and then users are associated with those groups. Command auth rules are all associated purely with the groups (not directly with a user).

### Cisco IOS-XR example (from https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r6-1/security/configuration/guide/b-syssec-cg-crs-61x/b-syssec-cg-crs-61x_chapter_010.html#task_1379897)
RP/0/RP0/CPU0:router(config)# taskgroup beta
RP/0/RP0/CPU0:router(config-tg)# task read bgp
...
RP/0/RP0/CPU0:router(config)# usergroup beta 
RP/0/RP0/CPU0:router(config-ug)# taskgroup beta 
...
RP/0/RP0/CPU0:router(config)# username user1 
RP/0/RP0/CPU0:router(config-un)# group beta // I changed this from sysadmin to illustrate better

### Nokia SR OS example (from https://infocenter.nokia.com/public/7750SR2110R1A/topic/com.sr.system.mgmt/html/security_cli.html?cp=26_1_5_7#akalaitz5iwmlo70wqs)
A:ALA-1>config>system>security# info
----------------------------------------------
...
            profile "ghost"
                default-action permit-all
                entry 1
                    match "configure"
                    action permit
                exit
                entry 2
                    match "configure service vprn <22>"
                    action read-only
                exit
                entry 3
                    match "show"
                exit
                entry 4
                    match "exit"
                exit
            exit
...
----------------------------------------------
A:ALA-1>config>system>security#
A:ALA-1>config>system>security# info
----------------------------------------------
...
    user "userA"
        ...        
        console
            member "ghost"
        exit
...

### JUNOS example (from https://www.juniper.net/documentation/us/en/software/junos/user-access/topics/topic-map/junos-os-access-privileges.html#id-understanding-junos-os-access-privilege-levels__d47158e1160)
user@host# show system login
class snmp-admin {
    permissions [ configure snmp snmp-control ];
}
user snmpuser {
    class snmp-admin;
    authentication {
        encrypted-password "$ABC123"; ## SECRET-DATA
    }
}

morrowc:
(NOTE: I reformatted your clips of config content.)

> The association of users to groups is good. But I'm a little unsure of how some common router implementations would handle the fact that polices (aka rules) can be associated with either a group _or_ a user.
> 
> Cisco IOS-XR, Nokia SR OS and JUNOS all seem to have the concept of command auth rules that apply to a group, and then users are associated with those groups. Command auth rules are all associated purely with the groups (not directly with a user).
> 

sure, so in the case of a conversion from 'open config' to 'vendor config', the
vendor config will likely have to handle: 
  "Oh, this is a rule for 1 user, I'll make a 1 user group called <uuid>,
    and put that user in the group, joy!"

right? In the end the 'user' and 'group' in either case is really just syntax and convenience
for the `operator`. If you are abstracting away the vendor config anyway (and all vendors
are already doing that anyway... arista/sysdb, XR/sysdb, JunOS/<whatever sysdb is called there)
then this problem isn't really a problem.

Additionally, it's not clear to me that vendor ideas about user/group/rules mean
much in this paradigm.

> ### Cisco IOS-XR example (from https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r6-1/security/configuration/guide/b-syssec-cg-crs-61x/b-syssec-cg-crs-61x_chapter_010.html#task_1379897)
>```shell 
> RP/0/RP0/CPU0:router(config)# taskgroup beta
> RP/0/RP0/CPU0:router(config-tg)# task read bgp ...
> RP/0/RP0/CPU0:router(config)# usergroup beta
> RP/0/RP0/CPU0:router(config-ug)# taskgroup beta ...
> RP/0/RP0/CPU0:router(config)# username user1
> RP/0/RP0/CPU0:router(config-un)# group beta // I changed this from sysadmin to illustrate better
>```
> 
> ### Nokia SR OS example (from https://infocenter.nokia.com/public/7750SR2110R1A/topic/com.sr.system.mgmt/html/security_cli.html?cp=26_1_5_7#akalaitz5iwmlo70wqs)
>```shell
> ## A:ALA-1>config>system>security# info
> ## ...
> profile "ghost"
> default-action permit-all
> entry 1
> match "configure"
> action permit
> exit
> entry 2
> match "configure service vprn <22>"
> action read-only
> exit
> entry 3
> match "show"
> exit
> entry 4
> match "exit"
> exit
> exit
> ...
> ## A:ALA-1>config>system>security#
> A:ALA-1>config>system>security# info
> ... user "userA" ... console member "ghost" exit ...
> ```
>
> ### JUNOS example (from https://www.juniper.net/documentation/us/en/software/junos/user-access/topics/topic-map/junos-os-access-privileges.html#id-understanding-junos-os-access-privilege-levels__d47158e1160)
> ```shell
> user@host# show system login
>  class snmp-admin {
>      permissions [ configure snmp snmp-control ];
>  }
>  user snmpuser {
>      class snmp-admin;
>      authentication {
>          encrypted-password "$ABC123"; ## SECRET-DATA
>      }
>  }
>```

morrowc:
Reminder to review comments/replies, one more day to act.. before revisions will require a whole new PR (which isn't a bad thing I suppose).

I have a pending discussion internally about version/created-on to get sorted, but I think the above comments/questions are all addressed.

morrowc:
@robshakir I think this is ready for review/etc.
(and perhaps this adds you as an official reviewer?)

morrowc:
Ping Rob @robshakir

hzhongmj:
Hi this is Hua Zhong from Arista Networks. I have a comment on the "action" field in the policy and overall model.

The natural model we can follow is filesystems (e.g., Linux/NTFS/APFS). It normally works with this:

- a policy is associated with the inode of the file or directory.
- if an inode has no policy, it recursively looks up in its parent inode.

This is inline with the "best match" model wrt longest match of the path name, so it's all good.

Then what's in a policy? You can actually check your computer for examples but it typically has the following:

user/group/everyone: Read/Write

For example, in my Document directory I have this:

<my-user-name>: RW
<everyone>: NO_PERMISSION

The lookup finds the best match within the policy (e.g., user name > everyone).

So the question is: why do we have permit/deny actions? Is it a leftover from the ordered ACLs? Should we remove it?

For more info on filesystem ACLs, see https://www.usenix.org/legacy/publications/library/proceedings/usenix03/tech/freenix03/full_papers/gruenbacher/gruenbacher_html/main.html

jsterne:
> The natural model we can follow is filesystems (e.g., Linux/NTFS/APFS).

Shouldn't we focus more on aligning with current router command authorization approaches (and RFC 6536 NACM) ? That seems to be more directly analogous to what were trying to achieve here, and what operators will be used to doing.

morrowc:
> > The natural model we can follow is filesystems (e.g., Linux/NTFS/APFS).
> 
> Shouldn't we focus more on aligning with current router command authorization approaches (and RFC 6536 NACM) ? That seems to be more directly analogous to what were trying to achieve here, and what operators will be used to doing.

With gRPC/gNOI/gNMI there's a little bit of a wrinkle in both the 6536 view AND the filesystem view, I think.
There's both a GET/SET distinction made prior to this proposed policy's view of the intended action, and a lack of granularity at that GET/SET layer. There are 2 separate but entwined systems for gRPC/gNOI/gNMI, in both the filesystem example and rfc6536 the single system manages the access controls.

hzhongmj:
> > > The natural model we can follow is filesystems (e.g., Linux/NTFS/APFS).
> > 
> > 
> > Shouldn't we focus more on aligning with current router command authorization approaches (and RFC 6536 NACM) ? That seems to be more directly analogous to what were trying to achieve here, and what operators will be used to doing.
> 
> With gRPC/gNOI/gNMI there's a little bit of a wrinkle in both the 6536 view AND the filesystem view, I think. 

Yeah, from RFC6536 section 3.4.4:

   6.   Process all rule-list entries, in the order they appear in the
        configuration [...]
   7.   For each rule-list entry found, process all rules, in order,

The action (permit/deny) makes sense in an ordered rule list, which most router implementations use. In a best-match model like the filesystem ACLs, it seems redundant.

If we use a hybrid model, where the policy is associated with the node (best match), but inside each policy we have an ordered list of rules, then it makes sense again. For example, if the policy says:

userA: RW
groupA: R
everyone: no access

In typical filesystems, the above entries have no order, it'd match in this order: user -> group -> everyone. So user A will get RW, user B inside groupA will get R, and everyone else will have no access.

But in an ordered list you'd do:

1. userA, op R or W, permit
2. groupA, op R, permit
3. deny

morrowc:
On Tue, 01 Feb 2022 21:03:57 +0000,
Hua Zhong ***@***.***> wrote:
> 
> [1  <text/plain; UTF-8 (7bit)>]
> > > > The natural model we can follow is filesystems (e.g., Linux/NTFS/APFS).
> > > 
> > > 
> > > Shouldn't we focus more on aligning with current router command authorization approaches (and RFC 6536 NACM) ? That seems to be more directly analogous to what were trying to achieve here, and what operators will be used to doing.
> > 
> > With gRPC/gNOI/gNMI there's a little bit of a wrinkle in both the 6536 view AND the filesystem view, I think. 
> 
> Yeah, from RFC6536 section 3.4.4:
> 
>    6.   Process all rule-list entries, in the order they appear in the
>         configuration [...]
>    7.   For each rule-list entry found, process all rules, in order,
> 
> The action (permit/deny) makes sense in an ordered rule list, which most router implementations use. In a best-match model like the filesystem ACLs, it seems redundant.
> 

Oh sorry, I think my mental model is something like:
  "Permit hua access to 'show config router bgp', specifically do not permit 'show config router bgp neighbor 1.2.3.4'"

Effectively: "Permit access, but hole-punch this one specific part of the config/etc away/out."

You could also think about this as: "deny all of the config, excpet 'show interface'"
You could fix either of these examples with an endlessly growing list of 'permit x, y, z" and never denying (but not permitting that specific thing)'.

That gets complicated and will be error prone as new things appear in the config over time (new peers in my bgp example).

release/authorization/README.md
```diff
@@ -0,0 +1,419 @@
...
+*   a defined KEY over a wildcard element in a keyed path.
+
+Authorization rules must be defined such that a single best match is possible.
+If the result of policy evaluation is more than one match, an error must be
+raised.
+
+Match rules permit a match against:
+
+*   User or Group (not both)
+*   an OpenConfig gNMI path
```
robshakir, line 42: I think this can read `a gNMI path` right, AFAICS you have no tie towards the schema necessarily being OpenConfig.

release/authorization/README.md
```diff
@@ -0,0 +1,419 @@
...
+Policy rules are matched based on the best match for the authorization request,
+not the first match against a policy rule. Best match enables a configuration
+which permits a user or group access to particular OpenConfig paths while
+denying subordinate portions of the permitted paths, or the converse, without
+regard to ordering of the rules in the configuration.
+
+## Best Match
+
+Authorization is performed for a singular user, gNMI path accessed, and access
+methodology (READ/WRITE). The result of an Authorization evaluation is an
```
robshakir, line 26: Is `READ`/`WRITE` sufficient here? I think that it could be, but wanted to understand how we expect that knits with the per-RPC security.  AFAICS, `READ` will correspond to both `Subscribe` and `Get`, and `WRITE` will correspond to `Set`. Is it intentional that there is no way to say `READ via Subscribe`?

Why might we care about this? `Get` on `/` might be extremely expensive for a device such that we want to prevent a user having `READ` access to `/` via Get, but `Subscribe` may be significantly cheaper and therefore allowed in the policy.  Do you envisage that there is a separate per-RPC policy somewhere that would allow that?

release/authorization/README.md
```diff
@@ -0,0 +1,419 @@
...
+*   User or Group (not both)
+*   an OpenConfig gNMI path
+*   an access method (READ or WRITE)
+
+An implicit deny is assumed, if there is no matching rule in the policy. Logging
+may be specified on a per-policy-rule basis as well as a default for the whole
+authorization policy.
+
+As a request is evaluated against the configured policy, a READ request for
+the configuration tree may traverse all of the tree. For portions of the tree
```
robshakir, line 50: All of the subtree of the request that is made and authorised, correct?

release/authorization/README.md
```diff
@@ -0,0 +1,419 @@
...
+for which the user has no access no data will be returned. A WRITE request
+which attempts to write to a denied subtree or element will return an error
+to the caller.
+
+[OpenConfig paths](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#222-paths)
```
robshakir, line 55: gNMI paths here.

release/authorization/README.md
```diff
@@ -0,0 +1,419 @@
...
+are heirarchical, and rooted at a defined "origin". OpenConfig may contain paths
```
robshakir, line 56:
```suggestion
are hierarchical, and rooted at a defined "origin". OpenConfig may contain paths
```

release/authorization/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+System bootstrap, or install, operations may include an authorization policy
+delivered during bootstrap operations. It is suggested that the bootstrap
+process include the complete authorization policy so all production tools
+and services have immediate authorized access to finish installation and
+move devices into production in a timely manner.
+
+Using the Secure Zero Touch Provisioning (sZTP - RFC8572) process for
+bootstrap/installation provides a method to accomplish this delivery, and
+the delivery of all other bootstrap artifacts in a secure manner.
```
robshakir, line 109: Do we have something here that describes how this is to be done, or are we leaving this as a future exercise?

release/authorization/README.md
```diff
@@ -0,0 +1,419 @@
...
+    /this/is/a/keyed[name=Serial4/1]/message_path
+    /this/is/a/keyed[name=HSSI2]/message_path
+```
+
+Additionally, the path elements beyond "message_path" are available for access
+to this group as well.
+
+The wildcard character "*" (asterisk) is only able to be used as a value in
+keyed elements. The wildcard is only used to mask out all possible values but
+not portions of values, for instance:
```
robshakir, line 395: Is it worth specifying whether keys can be omitted? In some cases, I think we allow `/this/is/a/keyed/message_path` to mean the same as `/this/is/a/keyed[name=*]/message_path`, I need to check whether the reference documentation included this case as I didn't immediately see it in https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md

release/authorization/authorization.proto
```diff
@@ -0,0 +1,124 @@
+// Authorization for gnXi operations.
```
robshakir, line 1: @dplore @chrisy @aashaikh 

Would appreciate your thoughts here -- I think that this is OK to be in `openconfig/public` given that it is a data model that corresponds to multiple different gN*I protocols, but I am not sure that we have precedent here. Would you suggest that we include this here - or potentially put it in a different repo (if so, it'd be good to discuss where).

I think that it should likely at least be in `release/proto/authorization`.

robshakir (changes requested):
Thanks for all the work on this @morrowc -- in general this looks good to me. I have a couple of questions around the doc. The proto generally looks in good shape.

Cheers,
r.

release/authorization/README.md
morrowc: Oh correct, I think I conflate OC/gNMI/gNOI .. oops :) fixed.

release/authorization/README.md
morrowc: We can add a SUBSCRIBE your example seems like a sane reason to do this. I think by the time the request gets to the authorization request you'd have to have brought along the method (GET(SUBSCRIBE)/SET) in order to tell if the authorization is for READ/WRITE.

Making the 'subscribe' explicit and not implicit in 'get' makes sense.
(added to proto)

release/authorization/README.md
morrowc: yes, fixed.

release/authorization/README.md
morrowc: I think this would fit into a bootstrap server implementation.

Effectively this is 1 more artifact to specify in the bootstrap process and delivery from the bootstrap server (or designated other endpoint/URL) to the device during bootstrapping.

release/authorization/README.md
morrowc: done

release/authorization/README.md
morrowc: This sounds like an implementation detail.. sort of :)

how about:
  "...in keyed elements, If the key is missing in a path a wildcard is assumed. The wildcard is only...."

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+This proto definition and the reference code(to be delivered seperately) serve
+to describe an authorization framework for controlling which gNMI paths of a
+network device users can access. The authorization policy is initially intended
+to be deployed to a device, with the ability to define:
+
+*   Policy rules - each rule defines a single authorization policy.
+*   Groups of users - as a method to logically group users in the administrative
+    domain, for instance: operators or administrators.
+*   Users which are individual users as part of either rules or in groups.
```
aashaikh, line 13: nit: rewrite for consistency as "Users - individuals referenced in rules or group definitions"

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+*   a defined KEY over a wildcard element in a keyed path.
+
+Authorization rules must be defined such that a single best match is possible.
+If the result of policy evaluation is more than one match, an error must be
+raised.
+
+Match rules permit a match against:
+
+*   User or Group (not both)
+*   an gNMI gNMI path
```
aashaikh, line 42: s/gNMI//

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+An implicit deny is assumed, if there is no matching rule in the policy. Logging
+may be specified on a per-policy-rule basis as well as a default for the whole
+authorization policy.
+
+As a request is evaluated against the configured policy, a READ / SUBSCRIBE
+request for the configuration tree may traverse all of the tree and subtrees.
+For portions of the tree for which the user has no access no data will be
+returned. A WRITE request which attempts to write to a denied gNMI path or
+element will return an error to the caller.
```
aashaikh, line 53: specify that it should be a "Permission Denied" gRPC error?

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+[gNMI paths](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-specification.md#222-paths)
+are hierarchical, and rooted at a defined "origin". gNMOpenConfigI may contain paths
+such as:
+
+```proto
+    /a/b/c/d
+```
+
+Paths may also have attributes associated with the path elements such as:
```
aashaikh, line 63: above we call these "keys" (e.g., `a defined KEY over a wildcard element in a keyed path`)-- maybe change above to "attributes" for consistency?

release/authorization/README.md
aashaikh: is using sZTP just a recommendation?  This section seems a bit outside the scope of the authorization model, though I don't have an issue with keeping it.

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+  name: "family-group"
+  members { name: "stevie" }
+  members { name: "brian" }
+}
+group {
+  name: "test-group"
+  members { name: "crusty" }
+  members { name: "the-clown" }
+}
+# Action stevie to access /this/is/a/message_path in a READ manner.
```
aashaikh, line 135: nit: ... access <path> with READ permission.

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+      key{
+        key: "Ethernet"
+        value: "*"
+      }
+    }
+    elem { name: "message_path" }
+  }
+  action: DENY
+  level: READ
+  group { name: "brian" }
```
aashaikh, line 234: this should be `user` not `group` ?

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+```proto
+# proto-file: //experimental/users/morrowc/gnXi/proto/authorization.proto
+# proto-message: Policies
+version: "UUID-1234-123123-123123"
+created_on: 1234567890
+# Define 2 groups.
+group {
+  name: "family-group"
+  members { name: "stevie" }
```
aashaikh, line 127: it seems more consistent with the docs above to use `user` as the keyword here instead of `members` (or `group` if composing groups of users).  my 2c

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+    /this/is/a/message_path/the
+    /this/is/a/message_path/the/one
+    /this/is/a/message_path/the/one/that
+    /this/is/a/message_path/the/one/that/knocks
+```
+
+The second policy rule:
+
+```proto
+# Action memberss of family-group to run /this/is/a/different/message_path
```
aashaikh, line 285: s/memberss/members

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+    /this/is/a/keyed[name=HSSI2]/message_path
+```
+
+Additionally, the path elements beyond "message_path" are available for access
+to this group as well.
+
+The wildcard character "*" (asterisk) is only able to be used as a value in
+keyed elements, if the keys are missing in a keyed path a wildcard is assumed.
+The wildcard is only used to mask out all possible values but not portions of
+values, for instance:
```
aashaikh, line 396: do we need to say anything about paths with multiple attributes?  presumably it's fine to wildcard any combination of keys in such a path?

release/authorization/README.md
```diff
@@ -0,0 +1,420 @@
...
+  level: READ
+  level: WRITE
+  action: DENY
+}
+```
+
+provides an explcit deny for any request wich does not match any other policy
+rule. This rule also requests that the result be logged in full fidelity.
+
+Hopefully this illustrates the acceptable usage of the wildcard character in
```
aashaikh, line 419: nit: can remove this sentence

release/authorization/authorization.proto
```diff
@@ -0,0 +1,127 @@
...
+// User is a singular username used only in the matching criteria.
+message User {
+  string name = 1;
+}
+
+// Group is a group of users, groups are a construct of the policy configuration
+// as a method to collect many users with the same authorizations together.
+message Group {
+  string name = 1;
+  repeated User members = 2;
```
aashaikh, line 41: per comment above, I think this could just be named `users`

release/authorization/authorization.proto
aashaikh: The placement seems a bit odd to me -- my preference would be to put it elsewhere.   Since we have a placeholder repo for gnsi already, that seems a more likely place for this.

release/authorization/README.md
morrowc: Yes, I think carl/marcus were pushing for some recommendation to be included for:
  "Ok, so you want to do this, how would you do this for even the earliest part of a device life cycle?"
  
  if you want to do thigns then, you don't have ssh nor https/tls, if you want to deliver in a safe method it's either over console (ick) or the bootstrap process. The only 'safe' bootstrap for remote devices is .. sztp.

release/authorization/authorization.proto
morrowc: sure, wasn't the gnsi there though a different thing entirely?
(I have no care where this lands, to be clear... this was as good as any starting point)

release/authorization/README.md
morrowc: done

release/authorization/README.md
morrowc: done

release/authorization/README.md
morrowc: done

release/authorization/README.md
morrowc: I think you mean 'above to "keys" for consistency'

or I am not parsing this correctly :)

release/authorization/README.md
morrowc: done.

(why can't vim do this for me.. it probably can! I probably disabled it because it's annoying! :) )

release/authorization/README.md
morrowc: I think this came up previously, I don't see why we'd limit the number of keyed parts in a rule.
The existing text doesn't talk about one or many  keyed parts, so I'd just leave it as is.

release/authorization/README.md
morrowc: done

release/authorization/authorization.proto
morrowc: done

release/authorization/authorization.proto
aashaikh: since that repo is empty, and this proto package is called `gnsi.authorization`, it seems like it could go there seamlessly :-)

morrowc: ah! ok, I can move this (I think) pretty directly to that part of the tree.
Wait on robjs/chrisy/dplore? or just move and pay later? :)

release/authorization/README.md
morrowc: done

release/authorization/authorization.proto
morrowc: Ddi the GNSI repository rename to GNPSI perhaps?
  openconfig/gnpsi/README.md:
  "Have single gNSI RPC's in the service and include typing in response"

morrowc:
bleep blop bloop - for those following along at home, this moved to:
  https://github.com/openconfig/gnsi/pull/1

release/authorization/authorization.proto
aashaikh: thanks -- I see the PR in the gnsi repo now.

PR #526, opened 2021-11-29
Add OSPF stub area configuration options

stmosher:
This PR allows for the configuration of OSPF area types of stub, totally-stubby, and not-so-stubby (nssa) and their associated parameters.

For instance, the below would configure area "1" as an OSPF stub area.
`ospfv2:
  areas:
    area:
    - config:
        identifier: 1
      identifier: 1
      stub-options:
        stub:
          config:
            enabled: True
            default-information-originate: True
        totally-stubby:
          config:
            enabled: False
            default-information-originate: False
        nssa:
          config:
            enabled: False
            no-summary: False
            default-information-originate: False
`
I believe these are configuration that users of OSPFv2 need. Please let me know if you have any questions or need anything else.

### DIFF
--- a/release/models/ospf/openconfig-ospfv2-area.yang
+++ b/release/models/ospf/openconfig-ospfv2-area.yang
@@ -25,6 +25,12 @@ submodule openconfig-ospfv2-area {
 
   oc-ext:openconfig-version "0.2.2";
 
+  revision "2021-11-29" {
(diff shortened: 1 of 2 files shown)


stmosher:
This will solve issue: https://github.com/openconfig/public/issues/530

Issue #253, opened 2019-07-24
Add prefix-limit-exceeded state leaf to prefix-limit


vcheny:
prefix-limit-exceeded would indicate if a bgp neighbor has exceeded prefix-limit for given aft-safi.

e.g.

/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/afi-safis/afi-safi/ipv4-unicast/prefix-limit/state/

PR #532, opened 2021-12-02
Updates to BGP prefix-limits

earies:
* (M) release/models/bgp/openconfig-bgp.yang
* (M) release/models/bgp/openconfig-bgp-common-structure.yang
* (M) release/models/bgp/openconfig-bgp-common.yang
* (M) release/models/bgp/openconfig-bgp-global.yang
* (M) release/models/bgp/openconfig-bgp-neighbor.yang
* (M) release/models/bgp/openconfig-bgp-peer-group.yang
* (M) release/models/bgp/openconfig-bgp-common-multiprotocol.yang
  - Change 'prefix-limit' to 'prefix-limit-accepted' to indicate
    post-policy processing
  - Add 'prefix-limit-received' container
  - Add state boolean node 'prefix-limit-exceeded'
  - Add state uint16 node 'remaining-restart-time' to signal remaining
    time in seconds if the prefix-limit is tripped in accordance to the
    restart-timer
  - Change 'restart-timer' to uint16 from decimal64

# Prefix-limit container restructuring

To account for the difference between accepted (post-policy) and received
prefix-limits, this proposal introduces another container and renames the old
container appropriately.  The rename of the old container `prefix-limit` could
be reverted here if there is strong preference to keep backwards compatibility
in favor of a less descriptive name (which is also used in various shipping
implementations)

Each variant of prefix-limits contains the same YANG nodes for both
config/state.

In addition, a state node named `remaining-restart-time` is introduced to
signal the remaining time in seconds if the prefix-limit is tripped in
accordance to the restart-timer.

The `restart-timer` node is changed from a decimal64 to uint16 type to closely
match implementations (decimal types still remain throughout protocol module
sets likely still need to be addressed as supporting subsecond granularity is
uncommon/unimplemented)

References:

* Arista EOS: https://eos.arista.com/bgp-primer-for-internet-peering/#2145_Maximum_routes_and_maximum_accepted_routes
* Juniper JUNOS:
  https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/prefix-limit-edit-protocols-bgp.html
  https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/accepted-prefix-limit-edit-protocols-bgp.html


earies:
Addresses issue raised in #253

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -367,10 +374,10 @@ submodule openconfig-bgp-common-multiprotocol {
     description
       "Grouping for configuration common to all AFI,SAFI";
 
-    container prefix-limit {
+    container prefix-limit-accepted {
```
dplore, line 377: My preference is to leave this container as-is.  But keeping the clarification in the description is a good idea

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -24,7 +24,14 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "6.0.0";
+  oc-ext:openconfig-version "8.0.0";
```
dplore, line 27: note: there is a 7.0.0 update which has been accepted previously but not published to public yet.

This can now become a minor revision since if comment below on prefix-limit is accepted, there is no breaking change.

dplore:
Note change 'restart-timer' to uint16 from decimal64 is a breaking change.  Known BGP implementations don't support sub-second restart-timers.

dplore:
LGTM

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
earies: Address via: https://github.com/openconfig/public/pull/532/commits/d53f9c098ecf4c9cf34c82233294e3bc30048991

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
earies: Keeping the version at 8.0.0 for current alignment

earies:
@dplore - let me know if there is any other comments or if this PR can be accepted - thx

dplore:
No other changes needed, I'll work to get this merged

dplore (approved):
LGTM, I will close this PR and sync with the private models repo

akshaygattani:
Couple of model change suggestions:

1) The remaining-restart-time can be specified in terms of time epochs (time when event happened) as opposed to in terms of "time remaining" until next event. Otherwise, we will need to find a way to keep streaming updates on the "time remaining" absent actual session state change. As an example, this is how the client receives the peer last established time and based on this can compute the "time elapsed since peer established"

231 leaf last-established {
232 type oc-types:timeticks64;
233 description
234 "This timestamp indicates the time that the
235 BGP session last transitioned in or out of the Established
236 state. The value is the timestamp in nanoseconds relative to
237 the Unix Epoch (Jan 1, 1970 00:00:00 UTC).
238
239 The BGP session uptime can be computed by clients as the
240 difference between this value and the current time in UTC
241 (assuming the session is in the ESTABLISHED state, per the
242 session-state leaf).";
243 }

we recommend that we similarly use type "oc-types:timeticks64" here, and the leaf could be either a) last-limit-exceed-idle-time or b) next-restart-time. Either could be used by the client to compute remaining-restart-time. (b) seems slightly preferable in that it will not have to use the restart-timer config for computation.

2) The restart-timer and remaining-restart-timer (or next-restart-time) doesn't need to be a per neighbor/per AFI-SAFI parameter. Unlikely that we would want different restart-timers for exceeding different AFI-SAFI limits in config. Further, once the session is IDLED, the remaining-restart-timer (or next-restart-time) becomes a property of the session and all AFI/SAFIs share that.

The restart-time can be moved to:

bgp/neighbors/timers/config/
and the remaining-restart-timer (or next-restart-time) can be moved to:
bgp/neighbors/timers/state
We would have to name the leafs:

limit-exceed-restart-time
and
limit-exceed-next-restart-time
(or similar) for clarity.

Another option is to nest another "prefix-limit" container under:
bgp/neighbors/timers/config
bgp/neighbors/timers/state
and add the "limit-exceed-restart-time" and "limit-exceed-next-restart-time" leafs under these new containers

earies:
> Couple of model change suggestions:
> 
>     1. The remaining-restart-time can be specified in terms of time epochs (time when event happened) as opposed to in terms of "time remaining" until next event. Otherwise, we will need to find a way to keep streaming updates on the "time remaining" absent actual session state change. As an example, this is how the client receives the peer last established time and based on this can compute the "time elapsed since peer established"
> 
> 
> 231 leaf last-established { 232 type oc-types:timeticks64; 233 description 234 "This timestamp indicates the time that the 235 BGP session last transitioned in or out of the Established 236 state. The value is the timestamp in nanoseconds relative to 237 the Unix Epoch (Jan 1, 1970 00:00:00 UTC). 238 239 The BGP session uptime can be computed by clients as the 240 difference between this value and the current time in UTC 241 (assuming the session is in the ESTABLISHED state, per the 242 session-state leaf)."; 243 }
> 
> we recommend that we similarly use type "oc-types:timeticks64" here, and the leaf could be either a) last-limit-exceed-idle-time or b) next-restart-time. Either could be used by the client to compute remaining-restart-time. (b) seems slightly preferable in that it will not have to use the restart-timer config for computation.

This is acceptable however want to note this is a design item that should be taken into account across all domains for consistency and client expectation.

@dplore @robshakir - if we lock in on this approach, this should be addressed in other domains as well (e.g. isis `remaining-hold-time` and `remaining-lifetime` on quick glance)

As far as (a) vs. (b): I'm fine w/ (b) where the timestamp is set to the future.  When the value is initialized can trigger an ON_CHANGE and remaining time calculations can be performed in either scenario.

>     2. The restart-timer and remaining-restart-timer (or next-restart-time) doesn't need to be a per neighbor/per AFI-SAFI parameter. Unlikely that we would want different restart-timers for exceeding different AFI-SAFI limits in config. Further, once the session is IDLED, the remaining-restart-timer (or next-restart-time) becomes a property of the session and all AFI/SAFIs share that.
> 
> 
> The restart-time can be moved to:
> 
> bgp/neighbors/timers/config/ and the remaining-restart-timer (or next-restart-time) can be moved to: bgp/neighbors/timers/state We would have to name the leafs:
> 
> limit-exceed-restart-time and limit-exceed-next-restart-time (or similar) for clarity.
> 
> Another option is to nest another "prefix-limit" container under: bgp/neighbors/timers/config bgp/neighbors/timers/state and add the "limit-exceed-restart-time" and "limit-exceed-next-restart-time" leafs under these new containers

If we go with `next-restart-time` per above then I would question if we need to distinguish the leaf to add a "reason" into the name but rather this can be derived from other surrounding nodes.  That way this leaf can be used generically as to when the next session establishment retry will take place.

Agreed that once we trip the threshold in this case - the property of the `next-restart-time` timestamp is one of the session so this can be moved up a level out of per AFI/SAFI.

Per AFI/SAFI configuration of thresholds is still applicable and will remain - is a logical OR on MP-BGP session

The state leaf `prefix-limit-exceeded` will remain as is as this is the indicator as to which prefix-limit policy was tripped.

The `restart-timer` config/state leaf can be supported per AFI/SAFI in JUNOS however I do agree that in operational reality it is unlikely to carry a different policy between the 2.  If we're coming to a middle ground here due to implementation differences then I'm also open to moving this up to the per-neighbor level with the result being this value is applied to all AFI/SAFI combinations on that session

dplore:
@robshakir agreed to respond to these comments

dplore:
Seems we are all okay with 
- remaining-restart-time can be specified in terms of time epochs (time when event happened) as opposed to in terms of "time remaining" until next event 
- use type "oc-types:timeticks64" and  leaf  next-restart-time
- timer and remaining-restart-timer (or next-restart-time) doesn't need to be a per neighbor/per AFI-SAFI parameter

Can you make these changes @earies ?

earies:
Pushing these changes shortly - comments inline
 
>     * remaining-restart-time can be specified in terms of time epochs (time when event happened) as opposed to in terms of "time remaining" until next event

This will be accomplished rather by a leaf named `last-prefix-limit-exceeded` of type `oc-types:timeticks64` directly at the neighbor level
 
>     * use type "oc-types:timeticks64" and  leaf  next-restart-time
 
The above leaf accomplishes this as the timestamp when any prefix-limit under that neighbor is breached.  The next 'restart-time' can then be calculated off of that vs. setting a time in the future

>     * timer and remaining-restart-timer (or next-restart-time) doesn't need to be a per neighbor/per AFI-SAFI parameter

The `restart-time` leaf will be moved up under the neighbor timer hierarchy instead of per AFI/SAFI.  As for the "remaining" time, this will be covered by above

earies:
Relevant tree outputs

_NOTE: The use of `uint16` vs. `decimal64`.  The use of `decimal64` should be deprecated for most timer nodes as implementations generally do not support sub-second timers to be specified_

```
  +--rw network-instances
        +--rw protocols
              +--rw bgp
                 +--rw neighbors
                    +--rw neighbor* [neighbor-address]
                       +--rw timers
                          +--rw config
                          |  +--rw connect-retry?                    decimal64
                          |  +--rw hold-time?                        decimal64
                          |  +--rw keepalive-interval?               decimal64
                          |  +--rw minimum-advertisement-interval?   decimal64
                          |  +--rw restart-time?                     uint16
                          +--ro state
                             +--ro connect-retry?                    decimal64
                             +--ro hold-time?                        decimal64
                             +--ro keepalive-interval?               decimal64
                             +--ro minimum-advertisement-interval?   decimal64
                             +--ro restart-time?                     uint16
                             +--ro negotiated-hold-time?             decimal64
```

```
  +--rw network-instances
        +--rw protocols
              +--rw bgp
                 +--rw neighbors
                    +--rw neighbor* [neighbor-address]
                       +--ro state
                          +--ro last-prefix-limit-exceeded?   oc-types:timeticks64
```

```
  +--rw network-instances
        +--rw protocols
              +--rw bgp
                 +--rw neighbors
                    +--rw neighbor* [neighbor-address]
                       +--rw afi-safis
                             +--rw ipv4-unicast
                                +--rw prefix-limit
                                   +--rw config
                                   |  +--rw max-prefixes?            uint32
                                   |  +--rw prevent-teardown?        boolean
                                   |  +--rw warning-threshold-pct?   oc-types:percentage
                                   +--ro state
                                      +--ro max-prefixes?            uint32
                                      +--ro prevent-teardown?        boolean
                                      +--ro warning-threshold-pct?   oc-types:percentage
                                      +--ro prefix-limit-exceeded?   boolean
```

```
  +--rw network-instances
        +--rw protocols
              +--rw bgp
                 +--rw neighbors
                    +--rw neighbor* [neighbor-address]
                       +--rw afi-safis
                             +--rw ipv4-unicast
                                +--rw prefix-limit-received
                                   +--rw config
                                   |  +--rw max-prefixes?            uint32
                                   |  +--rw prevent-teardown?        boolean
                                   |  +--rw warning-threshold-pct?   oc-types:percentage
                                   +--ro state
                                      +--ro max-prefixes?            uint32
                                      +--ro prevent-teardown?        boolean
                                      +--ro warning-threshold-pct?   oc-types:percentage
                                      +--ro prefix-limit-exceeded?   boolean
```

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -403,6 +410,28 @@ submodule openconfig-bgp-common-multiprotocol {
           "State information relating to the prefix-limit for the
           AFI-SAFI";
         uses bgp-common-mp-all-afi-safi-common-prefix-limit-config;
+        uses bgp-common-mp-all-afi-safi-common-prefix-limit-state;
+      }
+    }
+    container prefix-limit-received {
```
robshakir, line 416: (nit) linebreak above here.

release/models/bgp/openconfig-bgp-neighbor.yang
```diff
@@ -305,6 +312,20 @@ submodule openconfig-bgp-neighbor {
...
+
+    leaf last-prefix-limit-exceeded {
+      type oc-types:timeticks64;
+      description
+        "This timestamp indicates the time that the BGP session last
+        violated a configured max prefix-limit for any AFI/SAFI combination
+        on the session.  The value is the timestamp in nanoseconds relative
+        to the Unix Epoch (Jan 1, 1970 00:00:00 UTC).
+
+        If the 'prefix-limit-exceeded' node is set to true for any AFI/SAFI
```
robshakir, line 324: Is this strictly true? I expect that if:

 * prefix-limit is tripped
 * connection-retry expires, and the local speaker sends an `OPEN`
 * for any reason the session is not re-established

then this leaf's value is not set to a new value (since we didn't hit max-prefix) and so the sentence is not correct.

It might be worth specifying when this value gets set to `nil`/deleted again -- should it be reset when a session is re-established? I'd imagine we don't necessarily need to say this because it'd be possible to tell whether a max-prefix event happened in the context of this session by looking at when the session was established.

robshakir (changes requested):

release/models/bgp/openconfig-bgp-neighbor.yang
earies: Taking the above workflow - lets assume we start from an established BGP session

* Now a prefix-limit is breached for any AFI/SAFI on the session - lets say IPv6 Unicast
* The `prefix-limit-exceeded` node for `neighbors/neighbor/afi-safis/afi-safi/ipv6-unicast/state` is set to `true` *AND* this leaf `last-prefix-limit-exceeded` has the breach time recorded
* connection retry now expires
* The _current status_ `prefix-limit-exceeded` leaf should be reset to false prior to attempting session re-establishment
* The `last-prefix-limit-exceeded` could still remain - This can always retain the last time it was tripped or can be completely reset at each new session establishment and is only ever set/updated during a breach event

If we have consensus on the above, I'll readjust the wording here to convey.  The `prefix-limit-exceeded` node is meant to convey the current state of the AFI/SAFI on the session which would only ever hold a true value for period of "last-prefix-limit-exceeded" + "restart-time" then flip back to false/reset during new session establishment.  

As far as resetting/deleting the `last-prefix-limit-exceeded` node vs. retaining a timestamp that could be many session establishments in the past (if sessions were not breached but rather reset for any other reason) - I'd be open to thoughts here.  On one hand, reset/delete holds less cross session state (which can be problematic for the implementation) but if both of these leafs are reset as soon as the retry timer expires, the history of the node state is lost (but should be gathered by other means of offline telemetry data/local logs, etc..)

dplore: Regarding reset of last-prefix-limit-exceeded, it seems there is no clear right or wrong answer.  From an operational 
 point of view, it seems useful to only update last-prefix-limit-exceeded when an event occurs and NOT reset it.  This makes it similar to timestamps such as 'last-carrier-transition' for interfaces which is commonly treated this same way which also provides useful operational information.   Preserving this can help give insight into issues where there is momentary lapse of streaming telemetry.

dplore (changes requested):
Will approve pending description change to clarify when to set prefix-limit-exceeded and last-prefix-limit-exceeded

release/models/bgp/openconfig-bgp-neighbor.yang
```diff
@@ -305,6 +312,23 @@ submodule openconfig-bgp-neighbor {
...
+        on the session, then the next restart time for the session can be
+        calculated as this value plus the configured 'restart-time' under
+        the neighbor timers.
+
+        This value should be retained across established sessions and is only
+        set/updated during a prefix-limit-exceeded event.";
```
dplore, line 330: Possibly even more clear? 

```suggestion
        set/updated when prefix-limit-exceeded transitions from false to true.";
```

earies: ack thx - addressed in https://github.com/openconfig/public/pull/532/commits/7ce1d279d92c56dfdaa29640e8c4d82d3e04d0ac

release/models/bgp/openconfig-bgp-neighbor.yang
earies: ack - addressed in https://github.com/openconfig/public/pull/532/commits/67c7c0e9948f96b6ff80ba521cd4f72b901a1a04

earies:
Adjusted commit message:

```
Updates to BGP prefix-limits

* (M) release/models/bgp/openconfig-bgp.yang
* (M) release/models/bgp/openconfig-bgp-common-structure.yang
* (M) release/models/bgp/openconfig-bgp-common.yang
* (M) release/models/bgp/openconfig-bgp-global.yang
* (M) release/models/bgp/openconfig-bgp-neighbor.yang
* (M) release/models/bgp/openconfig-bgp-peer-group.yang
* (M) release/models/bgp/openconfig-bgp-common-multiprotocol.yang
  - Add 'prefix-limit-received' container
  - Add state boolean node 'prefix-limit-exceeded'
  - Add 'last-prefix-limit-exceeded' leaf at neighbor state level to
    indicate the last time a prefix-limit was breached
  - Change 'restart-timer' to 'restart-time', change type to uint16 from 
    decimal64 and relocate to session/neighbor level vs. AFI/SAFI
```

dplore:
Thanks for the clean commit message!

PR #537, opened 2021-12-11
Introduce BIOS identity for component type

earies:
* (M) release/models/platform/openconfig-platform-types.yang
  - Add BIOS identity as valid component type

# BIOS identity

Introduce identity `BIOS` that derives from `SOFTWARE_COMPONENT_MODULE` so that
a given component type related to BIOS/UEFI can be categorized correctly

The identity name is meant to cover legacy BIOS as well as UEFI


dplore:
LGTM

dplore (approved):

dplore:
Thanks for your contribution @earies

PR #540, opened 2021-12-16
Add p4rt interface/node-id relationship

earies:
  - Introduce leafref of the parent integrated-circuit to an interface

# Interface 'node-id' relationship

Introduce a leafref of the parent integrated-circuit `node-id` to the interface
to create a unique association for P4 programming.

### DIFF
--- a/release/models/p4rt/openconfig-p4rt.yang
+++ b/release/models/p4rt/openconfig-p4rt.yang
@@ -26,7 +26,13 @@ module openconfig-p4rt {
     The P4RT protocol specification is linkde from https://p4.org/specs/
     under the P4Runtime heading.";
 
-  oc-ext:openconfig-version "0.2.0";
+  oc-ext:openconfig-version "0.3.0";
+
+  revision 2021-12-15 {
+    description
+      "Add interface parent IC node identifier";
+    reference "0.3.0";
+  }
 
   revision 2021-07-20 {
     description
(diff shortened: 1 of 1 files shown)


dplore:
@earies can you explain the motivation for this? Is it to support multi-asic devices?

earies:
@dplore correct - Current p4rt model augments to both components (integrated-circuit) and interface both w/ user-defined identifiers.  The intention is that each integrated-circuit will be tagged with a user-defined identifier as well as an interface.

Now the relationship is static and provided by the system of interface -> IC mapping however these are merely user-defined values to explicitly set the intent via configuration.  The other option is that this is purely derived state from the system and not user-defined.

For user-defined - the system would need to validate and ensure the mappings within the bounds of the implementation

robshakir:
Ebben -- not quite sure that I am following here.

Today, we expect that there is a static interface to IC mapping (for physical interfaces) - so configuring both at the IC and the interface level gives this tuple. We decided against both being at the interface level because there is an expectation of one connection per IC in P4RT, and this seemed to keep the semantics for the P4RT protocol cleaner.

rsgcp:
Hi Ebben, there is another way to get the parent ASIC node-id is via /components/component hierarchy for interface. The interface parent points to ASIC. And ASIC has the node-id defined in it. Wouldn't that suffice? Thanks.

earies:
> Hi Ebben, there is another way to get the parent ASIC node-id via /components/component hierarchy for interface. The interface parent points to ASIC. And ASIC has the node-id defined in it. Wouldn't that suffice? Thanks.

@robshakir @rsgcp @dplore - that would suffice (assuming you are referring to below) however the spec currently would point to a component of type `PORT` whereas the node-id would be defined under a component of type `INTEGRATED_CIRCUIT`

```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--ro state
           +--ro oc-port:hardware-port?   -> /oc-platform:components/component/name
```

dplore:
Here's the p4rt model which declares itself as a parent of interfaces and has the node-id for the integrated circuit

earies:
Per offline discussion - closing this PR

Documentation updates will be followed up w/ subsequent PR to address component/subcomponent relationships so that the `hardware-port` reference can be walked back to a parent `INTEGRATED_CIRCUIT` to resolve the `node-id`

PR #541, opened 2021-12-24
Route programming state module

dplore:
Initial proposal to model the state of routes in a network instance to hardware programming.

Note the model only defines leaves for routing programming errors

### DIFF
--- a/release/models/route-programming/openconfig-route-programming-state.yang
+++ b/release/models/route-programming/openconfig-route-programming-state.yang
@@ -157,6 +157,7 @@ module openconfig-route-programming-state {
 
     leaf reason {
       type enumeration {
+        "The IPv4 or IPv6 prefix that the route state corresponds to.";
         enum INVALID {
           description
             "The reason for failure was that route, or its dependencies, was
@@ -172,6 +173,8 @@ module openconfig-route-programming-state {
             "The programming failure occurred for an unknown reason.";
         }
       }
+      description
+        "The reason for failure";
     }
   }
 
--- a/release/models/route-programming/openconfig-route-programming-state.yang
+++ b/release/models/route-programming/openconfig-route-programming-state.yang
@@ -1,0 +1,184 @@
+module openconfig-route-programming-state {
+  prefix "oc-rtstate";
+
+  namespace "http://openconfig.net/yang/route-programming";
+
+  import openconfig-extensions { prefix "oc-ext"; }
+  import openconfig-network-instance { prefix "oc-ni"; }
+  import openconfig-inet-types { prefix "oc-inet"; }
+  import openconfig-types { prefix "oc-types"; }
+
+  organization
+    "OpenConfig working group";
+
+  contact
+    "OpenConfig working group
+    www.openconfig.net";
+
+  description
+    "This module provides detailed information as to the programming
+    state of routes within a particular network instance. It can be used to
+    track where there are routing programming errors on a device.";
+
+  oc-ext:regexp-posix;
+  oc-ext:catalog-organization "openconfig";
+  oc-ext:origin "openconfig";
+
+  oc-ext:openconfig-version "0.1.0";
+
+  revision "2021-10-16" {
+    description
+      "Initial revision.";
+    reference "0.1.0";
+  }
+
+  grouping route-state-top {
+    description
+      "Top-level grouping for route state in OpenConfig.";
+
+    container route-programming {
+      config false;
+      description
+        "The programming status of routes within the network-instance. Route
+        programming reports are divided into a set of sub-categories:
+
+         - failed programming -- these routes are those that were attempted to be
+           programed to the hardware, but the programming operation failed.
+         - stale programming -- these routes are those that are programmed into
+           hardware, but some failure occurred in updating to a new state.
+         - drop programming - these routes are explicitly programmed into hardware
+           to point to a destination that discards packets.";
+
+      container failed-routes {
+        description
+          "Surrounding container for the list of routes that fail programming.";
+
+        list failed {
+          key "prefix";
+
+          description
+            "A prefix that was attempted to the programmed into hardware, but the
+            programming operation failed.";
+
+          leaf prefix {
+            type leafref {
+              path "../state/prefix";
+            }
+            description
+              "Reference to the prefix that keys the failed list.";
+          }
+
+          container state {
+            description
+              "Operational state parameters relating to a failed programming
+              operation.";
+            uses route-state-common;
+          }
+        }
+      }
+
+      container stale-routes {
+        description
+          "Surrounding container for the list of routes that are currently in
+          a stale state.";
+
+        list stale {
+          key "prefix";
+
+          description
+            "A prefix that is currently installed in hardware, but a subsequent
+            operation to update its programming failed - such that the entry in
+            hardware is stale.";
+
+          leaf prefix {
+            type leafref {
+              path "../state/prefix";
+            }
+            description
+              "Reference to the prefix that keys the stale list.";
+          }
+
+          container state {
+            description
+              "Operational state parameters relating to a stale route.";
+            uses route-state-common;
+          }
+        }
+      }
+
+      container drop-routes {
+        description
+          "Surrounding container for the list of routes that are currently in
+          a drop state.";
+
+        list drop {
+          key "prefix";
+
+          description
+            "A prefix that is currently installed in hardware but with an explicit
+            instruction that it should discard packets that are destined towards
+            it.";
+
+          leaf prefix {
+            type leafref {
+              path "../state/prefix";
+            }
+            description
+              "Reference to the prefix that keys the drop list.";
+          }
+
+          container state {
+            description
+              "Operational state parameters relating to a drop route.";
+            uses route-state-common;
+          }
+        }
+      }
+    }
+  }
+
+ grouping route-state-common {
+    description
+      "Common state parameters that correspond to a particular route
+      type.";
+
+    leaf prefix {
+      type oc-inet:ip-prefix;
+      description
+        "The IPv4 or IPv6 prefix that the route state corresponds to.";
+    }
+
+    leaf time {
+      type oc-types:timeticks64;
+      description
+        "Time that the route was programmed, or the programming attempt was made
+        expressed as nanoseconds since the Unix epoch.";
+    }
+
+    leaf reason {
+      type enumeration {
+        enum INVALID {
+          description
+            "The reason for failure was that route, or its dependencies, was
+            deemed to be invalid.";
+        }
+        enum HW_PROGRAMMING_FAILED {
+          description
+            "The reason for the failure was that the hardware programming
+            failed.";
+        }
+        enum UNKNOWN {
+          description
+            "The programming failure occurred for an unknown reason.";
+        }
+      }
+    }
+  }
+
+  augment "/oc-ni:network-instances/oc-ni:network-instance" {
+    description
+      "Augment the network-instance model with the route-programming container.";
+
+    uses route-state-top;
+  }
+}
--- a/release/models/route-programming/openconfig-route-programming-state.yang
+++ b/release/models/route-programming/openconfig-route-programming-state.yang
@@ -157,7 +157,6 @@ module openconfig-route-programming-state {
 
     leaf reason {
       type enumeration {
-        "The IPv4 or IPv6 prefix that the route state corresponds to.";
         enum INVALID {
           description
             "The reason for failure was that route, or its dependencies, was
--- a/release/models/route-programming/openconfig-route-programming-state.yang
+++ b/release/models/route-programming/openconfig-route-programming-state.yang
@@ -160,6 +160,14 @@ module openconfig-route-programming-state {
         monitoring subsystem expressed as nanoseconds since the Unix epoch.";
     }
 
+    leaf dest-component {
+      type leafref {
+        path "/oc-platform:components/oc-platform:component";
+      }
+      description
+        "The destination component for the route programming";
+     }
+ 
     leaf reason {
       type enumeration {
         enum INVALID {
--- a/release/models/route-programming/openconfig-route-programming-state.yang
+++ b/release/models/route-programming/openconfig-route-programming-state.yang
@@ -42,10 +42,15 @@ module openconfig-route-programming-state {
         "The programming status of routes within the network-instance. Route
         programming reports are divided into a set of sub-categories:
 
-         - failed programming -- these routes are those that were attempted to be
-           programed to the hardware, but the programming operation failed.
-         - stale programming -- these routes are those that are programmed into
-           hardware, but some failure occurred in updating to a new state.
+         - failed programming -- A prefix which is not present in the AFT is
+           attempted to be added into the AFT but failed.  Traffic destined for
+           this prefix will not be matched.
+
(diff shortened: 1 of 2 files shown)


earies:
A few initial comments/observations:
* This module is called out specifically as `-state` which while true does not follow the design pattern of other modules (even though some others consist mainly of r/o nodes).  If this is going to be a module all to itself, it might be best to generalize the nomenclature for any unknown future proofing
* Should this rather exist as attributes to prefixes in the AFT model vs. it's own separate constructs dedicated to route programming?  Sitting within the AFT domain leverages existing structures however targeting from a client requires a filter to a non-key node in what could be an extremely large list (e.g. more expensive operation)
* In chassis/multi-component based systems - you could have hw programming errors only occur on 1 or a subset of components.  Per the current model, this would have to be abstracted in a way that does not provide visibility into precisely where the source of the error occurred.

release/models/route-programming/openconfig-route-programming-state.yang
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    leaf time {
+      type oc-types:timeticks64;
+      description
+        "Time that the route was programmed, or the programming attempt was made
+        expressed as nanoseconds since the Unix epoch.";
+    }
+
+    leaf reason {
```
earies, line 158: As mentioned in some general observations, a programming error could occur on one component in a multi-component system but there is currently no signal into such.  While this reason gives a generic code to the client, does it make sense to include an opaque leaf to signal additional information helpful to the client?  The other option is fully modeled structures to represent components.

release/models/route-programming/openconfig-route-programming-state.yang
```diff
@@ -0,0 +1,184 @@
...
+
+          container state {
+            description
+              "Operational state parameters relating to a stale route.";
+            uses route-state-common;
+          }
+        }
+      }
+
+      container drop-routes {
```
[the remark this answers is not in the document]
dplore: Thanks for this good comment.  I believe the original intent of this container was to represent errors.  Can you describe the use case(s) for an intentionally dropped route?  I ask because maybe there is another list that should be added for intentionally dropped routes.

release/models/route-programming/openconfig-route-programming-state.yang
```diff
@@ -0,0 +1,184 @@
...
+      type.";
+
+    leaf prefix {
+      type oc-inet:ip-prefix;
+      description
+        "The IPv4 or IPv6 prefix that the route state corresponds to.";
+    }
+
+    leaf time {
+      type oc-types:timeticks64;
```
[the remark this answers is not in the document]
dplore: I will update the description to say the timestamp represents the time the route programming state is detected

release/models/route-programming/openconfig-route-programming-state.yang
dplore: Do you mean for example, in a two ASIC system, route programming could succeed on one ASIC and fail on a second?    Are you suggesting something like:
```
leaf dest-component {
  type leafref {
    path "/oc-platform:components/oc-platform:component";
  }
  description
    "The destination component for the route programming";
 }
```

earies: Correct - a reference to a component at least gives an indication as to where the failure was sourced.  An additional leafref as you have above in `route-state-common` seems appropriate for such

release/models/route-programming/openconfig-route-programming-state.yang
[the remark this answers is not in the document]
mayukh288: The most common use case would where we configure static null routes where we want to deliberately drop the traffic rather than having it get routed through the default/some less specific route.

release/models/route-programming/openconfig-route-programming-state.yang
mayukh288: Just to confirm, if the publishing logic doesn't quite know the component, the expectation would be for it to just skip populating this leaf correct? ( As I understand non-key leaf attributes are by default optional )

mayukh288:
> A few initial comments/observations:
> 
> * This module is called out specifically as `-state` which while true does not follow the design pattern of other modules (even though some others consist mainly of r/o nodes).  If this is going to be a module all to itself, it might be best to generalize the nomenclature for any unknown future proofing
> * Should this rather exist as attributes to prefixes in the AFT model vs. it's own separate constructs dedicated to route programming?  Sitting within the AFT domain leverages existing structures however targeting from a client requires a filter to a non-key node in what could be an extremely large list (e.g. more expensive operation)

Having these being in a separate tree seems better. It separates the error reporting state from the AFT per se. As you said having it separate also makes it easier for a client to just subscribe for this error state without subscribing to the entire AFT. ( at least the routes collection within it ie )

> * In chassis/multi-component based systems - you could have hw programming errors only occur on 1 or a subset of components.  Per the current model, this would have to be abstracted in a way that does not provide visibility into precisely where the source of the error occurred.

earies:
> Having these being in a separate tree seems better. It separates the error reporting state from the AFT per se. As you said having it separate also makes it easier for a client to just subscribe for this error state without subscribing to the entire AFT. ( at least the routes collection within it ie )

As we then start to take the approach of detaching some data from existing trees into new subtrees for specific purposes and organization, I would propose some global structure of classification otherwise we are going to end up with many domains coming in with no proper anchor points that could have potential for different design patterns.

So far - we have no configuration related to this proposal but we don't want to limit such and if we concentrate on the purpose of this domain - it is currently to provide a view/reporting into "anomaly" or "error" state for route programming.

It then might make sense to consider:
* `route-programming` to be a r/w container with sub-containers of config/state where we are only concentrating on state today (config wise could very well be the future enablement/disablement and other characteristics of route-programming reporting in general)
* Possibly nest this underneath a more descriptive domain such as `anomalies`, `errors`, etc.. (either as a parent or child to `route-programming`).  If at a parent level then you have a consistent path to target errors or anomalies (much like log messages) then targeting domains - the opposite means you target specific domains for the related error/anomaly data.

release/models/route-programming/openconfig-route-programming-state.yang
earies: You would have the choice of either
* not populating the leaf
* deviating the leaf altogether

The first case allows for populating when you know and not when you do not.  The latter covers non-support altogether

General rule of thumb across r/o data being optional is what you describe however this should be documented at a minimum for OC gRPC services wrt. client/server expectations

dplore:
> > Having these being in a separate tree seems better. It separates the error reporting state from the AFT per se. As you said having it separate also makes it easier for a client to just subscribe for this error state without subscribing to the entire AFT. ( at least the routes collection within it ie )
> 

@mayukh288 is there a specific route programming method being used that we want to report errors on?   p4rt?  gRIBI? Something else?

dplore (commented):
Requesting change for an optional destination component.

release/models/route-programming/openconfig-route-programming-state.yang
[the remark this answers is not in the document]
mayukh288: Yes that sounds about right. That use will get captured by the collection though as it is defined now. ( Since the packets for just prefixes will get dropped in the data plane )

mayukh288:
> > Having these being in a separate tree seems better. It separates the error reporting state from the AFT per se. As you said having it separate also makes it easier for a client to just subscribe for this error state without subscribing to the entire AFT. ( at least the routes collection within it ie )
> 
> As we then start to take the approach of detaching some data from existing trees into new subtrees for specific purposes and organization, I would propose some global structure of classification otherwise we are going to end up with many domains coming in with no proper anchor points that could have potential for different design patterns.
> 
> So far - we have no configuration related to this proposal but we don't want to limit such and if we concentrate on the purpose of this domain - it is currently to provide a view/reporting into "anomaly" or "error" state for route programming.
> 
> It then might make sense to consider:
> 
> * `route-programming` to be a r/w container with sub-containers of config/state where we are only concentrating on state today (config wise could very well be the future enablement/disablement and other characteristics of route-programming reporting in general)
> * Possibly nest this underneath a more descriptive domain such as `anomalies`, `errors`, etc.. (either as a parent or child to `route-programming`).  If at a parent level then you have a consistent path to target errors or anomalies (much like log messages) then targeting domains - the opposite means you target specific domains for the related error/anomaly data.

Giving this some high level structure makes sense, I would think the hierarchy Errors --> Route-Programming would be better than the other way round.

mayukh288:
> > > Having these being in a separate tree seems better. It separates the error reporting state from the AFT per se. As you said having it separate also makes it easier for a client to just subscribe for this error state without subscribing to the entire AFT. ( at least the routes collection within it ie )
> 
> @mayukh288 is there a specific route programming method being used that we want to report errors on? p4rt? gRIBI? Something else?

If I understood the question, you were asking whether it makes sense to further sub-divide route-programming into various sub containers ( one for each specific method of programming?). I would think not, regardless of the various mechanism(s) to program the routes, they all end up in one logical FIB, so having just one container to report all route-programming errors should be fine.

dplore:
> > 
> > @mayukh288 is there a specific route programming method being used that we want to report errors on? p4rt? gRIBI? Something else?
> 
> If I understood the question, you were asking whether it makes sense to further sub-divide route-programming into various sub containers ( one for each specific method of programming?). I would think not, regardless of the various mechanism(s) to program the routes, they all end up in one logical FIB, so having just one container to report all route-programming errors should be fine.

Thanks for this explanation.   Openconfig is very protocol and component oriented.  To put this request into that context, I think we can state "We want to report telemetry for a single logical FIB (component)".    

Today we have AFT representing a logical FIB entity.  To make subscribing to these errors efficient, perhaps we can come up with a way for a client subscribe to just the errors from a given AFT instance.

Can we take advantage of a similar pattern we as in the  integrated circuit and associated [pipeline error counters](https://github.com/openconfig/public/blob/3c6b98f1573fb9c3e16ae97d5bce46c0d51f6d82/release/models/platform/openconfig-platform-pipeline-counters.yang#L90)

sulrich:
> Thanks for this explanation. Openconfig is very protocol and component
> oriented. To put this request into that context, I think we can state "We want
> to report telemetry for a single logical FIB (component)".
> 
> Today we have AFT representing a logical FIB entity. To make subscribing to
> these errors efficient, perhaps we can come up with a way for a client
> subscribe to just the errors from a given AFT instance.

just to be clear on this discussion and the conclusions here. 

the current state of the PR reflects the above conclusion and an operator can
subscribe to a specific network-instance in order to constrain the reporting. so
i think things are square here. 

albeit w/o some of the additional hierarchy that's been discussed in some of the
review.

> 
> Can we take advantage of a similar pattern we as in the integrated circuit and
> associated [pipeline error counters](https://github.com/openconfig/public/blob/3c6b98f1573fb9c3e16ae97d5bce46c0d51f6d82/release/models/platform/openconfig-platform-pipeline-counters.yang#L90)

wrt the above ... is there still a desire to extend this reporting to the
component (ASIC) level?  this seems like a significant addition and a deviation
from the current model.

release/models/route-programming/openconfig-route-programming-state.yang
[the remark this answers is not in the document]
dplore: Ok, counting routes/RIB entries which are created and forwarding to a next hop such as null should be counted as successful routes programmed and added to the drop-routes/prefix list.  

If a 'null route' cannot be programmed, then it should be added to the failed-routes/prefix list.

Will update the description accordingly

release/models/route-programming/openconfig-route-programming-state.yang
```diff
@@ -0,0 +1,186 @@
...
+
+      container failed-routes {
+        description
+          "Surrounding container for the list of routes that fail programming.";
+
+        list failed {
+          key "prefix";
+
+          description
+            "A prefix that was attempted to the programmed into hardware, but the
```
xw-g, line 60: I"m trying to understand the difference between `failed-routes` and `stale-routes`.

Is this PR trying to say the following?
* If a prefix is **not** in the FIB and being programmed into FIB but failed, categorize it into `failed-routes`. Traffic will **not** be matched to the prefix in FIB, because it's not installed yet into the FIB.
* If a prefix is already in the FIB and being updated but failed, categorize it into `stale-routes`. Traffic will still be following the *old* next-hop for forwarding. The AFT streaming will still populate the **old** next-hop for the prefix.

dplore: Yes, your wording is perhaps more obvious

release/models/route-programming/openconfig-route-programming-state.yang
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    leaf time {
+      type oc-types:timeticks64;
+      description
+        "Represents the time the route programming state change was detected by the 
+        monitoring subsystem expressed as nanoseconds since the Unix epoch.";
+    }
+
+    leaf dest-component {
```
xw-g, line 163: If `des-component` is for providing more error details for a modular chassis, should we consider:
1. make it a list? For example, multiple linecard could have programming failures. 
2. move the `reason` under each component, because each component might have different failure reasons? 

Another approach is
1. Remove `dest-component`.
2. Make `reason` a container where we can add a new filed to store string messages. The field can then be populated by devices to provide more info in terms of which component failed the programming? 

I feel like the alternative is more flexible, because we can always add back `dest-component` later if needed? 

cc: @earies

release/models/route-programming/openconfig-route-programming-state.yang
```diff
@@ -42,10 +42,15 @@ module openconfig-route-programming-state {
...
+         - failed programming -- A prefix which is not present in the AFT is
+           attempted to be added into the AFT but failed.  Traffic destined for
+           this prefix will not be matched.
+
+         - stale programming -- A prefix is already in the AFT is requested to
+           be updated but failed.  Traffic destined for this prefix will be
+           forwarded to the old next-hop.  AFT telemetry should continue to
+           reflect the old next-hop for the prefix.
+
          - drop programming - these routes are explicitly programmed into hardware
```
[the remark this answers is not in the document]
mayukh288: Sorry for the delay in replying. The next-hop state should already be able to represent this via the interface-ref. The interface-ref for intentional drop next-hop(s) should be something like Null0. 

Your question whether we need to additionally capture this fact here in the route-programming model is still valid though. My opinion is that having a collated container which has all the routes which are in error state or dropping traffic is still useful.

While reading this over I realized that we are only capturing the intentional drops in this sub container. We can also have un-intentional drops, where the system programmed a drop route although the FIB did not specify that. Where is that captured @dplore ?

akalluru1:
can there be programming errors for other types of routes under `afts` container, for ex: MPLS/labeled routes, ethernet entries? If so, would it make sense to place these failed routes, drop routes & etc... under the respective `aft` containers, i.e., `ipv4-unicast`, `ipv6-unicast`, `mpls` & etc...

mayukh288:
> can there be programming errors for other types of routes under `afts` container, for ex: MPLS/labeled routes, ethernet entries? If so, would it make sense to place these failed routes, drop routes & etc... under the respective `aft` containers, i.e., `ipv4-unicast`, `ipv6-unicast`, `mpls` & etc...

--> Having a separate subtree for the error reporting is better in my opinion. It allows clients to clearly subscribe to this subtree if they are interested. If required we can create containers for the mpls routes also under this subtree

dplore:
Superceded by #725

PR #542, opened 2021-12-29
Expectations for the `openconfig` origin.

robshakir:
```
 * (M) docs/implementation-guidance/openconfig-gnmi-guidelines.md
  - This change adds a document that covers the expectations for the
    openconfig origin in gNMI. These clarifications allow for
    simplifications to be made for some functionality.
```

This document is to be checked in as an addendum to the models. PTAL :-)


robshakir:
@dplore Could you PTAL here, I'd like to check this in if we can as it's been referenced a few times recently.

doc/implementation-guidance/openconfig-gnmi-guidelines.md
```diff
@@ -0,0 +1,101 @@
...
+ 
+This set of simplifying rules allow for separation of users from the exact
+module structure, whilst still supporting extensibility of the OpenConfig tree.
+We observe that asserting thse constraints avoids alternate approaches, such as
+defining a single huge module for the entire tree, as some implementations do -
+which causes significant difficulty in consuming the model for specific use
+cases which make use of only a subset of nodes. This single module approach is
+taken by some implementations, e.g.,
+[`nokia-conf`](https://github.com/nokia/7x50_YangModels/blob/master/latest_sros_20.7/nokia-conf.yang),
+and [`configuration`](https://github.com/Juniper/yang/blob/master/17.1/17.1R1/configuration.yang)
```
earies, line 92: Note that the approach of a single native `configuration` module within JUNOS was refactored a few years back to a slightly more modular augment approach - https://github.com/Juniper/yang/tree/master/21.3/21.3R1/junos/conf

dplore, line 92:
```suggestion
and [`junos-conf`](https://github.com/Juniper/yang/tree/master/21.3/21.3R1/junos/conf)
```

dplore (approved):
LGTM, one small suggestion as recommended by @earies to update the junos reference link

doc/implementation-guidance/openconfig-gnmi-guidelines.md
```diff
@@ -0,0 +1,101 @@
...
+  prefixing this name with a module namespace.
+* Base OpenConfig modules and augmentations MUST comply with this requirement.
+  * We expect that a simple means to achieve this is to create vendor-specific
+    node names for augmentations - e.g., `vendorFoo/config/new-config-leaf`.
+* RFC7951-formatted JSON MAY omit the module name prefixes, as described in the
+  [specification](https://tools.ietf.org/html/rfc7951#section-4).
+ 
+This set of simplifying rules allow for separation of users from the exact
+module structure, whilst still supporting extensibility of the OpenConfig tree.
+We observe that asserting thse constraints avoids alternate approaches, such as
```
sulrich, line 86: s/thse/these/

robshakir: Fixed, thanks.

doc/implementation-guidance/openconfig-gnmi-guidelines.md
robshakir: Thanks - I've removed the direct references here, since I'm not sure they were critical.

robshakir: Removed these references as stated above.

robshakir (commented):
Thanks for the comments, I've updated the document.

cfyo:
Is this saying then that if I have a multi-vendor platform and want to perform a gNMI subscription to /interfaces/interface to each but if one vendor has created a vendor-specific node name for augmentations for that path I need to do a different gNMI subscription depending on the vendor with that addition.  And if all vendors have one then each vendor has a slightly different gNMI subscription that must be used.  This seems to defeat the agnostic vision of OC and operators now will have to keep track of not only OC path but potential vendor specific nodes used in it.   I would suggest requiring the vendor must use the same vendor-specific node name across all OC paths at minimum so it doesn't have to become more complicated.

dplore:
> Is this saying then that if I have a multi-vendor platform and want to perform a gNMI subscription to /interfaces/interface to each but if one vendor has created a vendor-specific node name for augmentations for that path I need to do a different gNMI subscription depending on the vendor with that addition. And if all vendors have one then each vendor has a slightly different gNMI subscription that must be used. This seems to defeat the agnostic vision of OC and operators now will have to keep track of not only OC path but potential vendor specific nodes used in it. I would suggest requiring the vendor must use the same vendor-specific node name across all OC paths at minimum so it doesn't have to become more complicated.

Hi @cfyo,  this PR formalizes a requirement:
```
when `origin` YANG extension is set to `openconfig` and in gNMI when the `origin` field is set to 
`openconfig` or an empty string, all paths MUST be unique when considering only the node 
name, and **NOT** prefixing this name with a module namespace.
```

The intent of this clarification is to safely permit referencing OC paths without using prefixes.

With this requirement in place, if/when a vendor augments an openconfig model using a vendor proprietary model, the vendor will need to present unique paths in order to coexist with openconfig or their model may be in conflict with openconfig.  

Unless the model is checked into the openconfing/public repository, it is not openconfig compliant.  If a vendor deviates or augments openconfig the augment is a vendor model, not openconfig.    

We don't want vendor augments to openconfig.  Instead we should encourage a PR to be submitted to add the desired models to openconfig.   But yang defines augment and deviate as features, so it is possible to 'extend' openconfig with proprietary models.

cfyo:
@dplore can you provide some examples of what the gNMI command will look like with the openconfig origin

wenovus:
Hi I'm wondering if we should specify, for the "openconfig" origin, whether to **always** omit the module names from the path when streaming gNMI? If so shall we add it?

Another thing we might add is:
"anything that is in the OpenConfig origin is expected to pass the OpenConfig linter checks (OpenConfig style guide), including augments"

PR #544, opened 2021-12-30
Packet mirroring model

dplore:
New model for configuring packet mirroring 
```
module: openconfig-mirror
  +--rw mirroring
     +--rw sessions
        +--rw session* [name]
           +--rw config
           |  +--rw name?                  string
           |  +--rw mirror-action?         mirror-action
           |  +--rw interfaces*            -> /oc-if:interfaces/interface/name
           |  +--rw direction?             enumeration
           |  +--rw acl?                   -> /oc-acl:acl/acl-sets/acl-set/name
           |  +--rw source-address?        oc-inet:ip-address
           |  +--rw destination-address?   oc-inet:ip-address
           |  +--rw network-instance?      oc-ni:network-instance-ref
           |  +--rw max-capture-size?      uint32
           |  +--rw max-size?              uint32
           +--ro state
              +--ro name?                  string
              +--ro mirror-action?         mirror-action
              +--ro enabled?               boolean
              +--ro interfaces*            -> /oc-if:interfaces/interface/name
              +--ro direction?             enumeration
              +--ro acl?                   -> /oc-acl:acl/acl-sets/acl-set/name
              +--ro source-address?        oc-inet:ip-address
              +--ro destination-address?   oc-inet:ip-address
              +--ro network-instance?      oc-ni:network-instance-ref
              +--ro max-capture-size?      uint32
              +--ro max-size?              uint32
```

## Implementations

References for packet mirroring  with filtering to remote destination

Cisco IOS XR
https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/Interfaces/73x/configuration/guide/b-interfaces-config-guide-cisco8k-r73x/configuring-traffic-mirroring.html#Cisco_Concept.dita_aef1adcd-7dd5-44f2-9f60-10fc37a99b05

Juniper JunOS
https://kb.juniper.net/InfoCenter/index?page=content&id=KB35163&actp=METADATA

Arista EOS:
https://eos.arista.com/eos-4-15-3f/gre-filtered-mirroring/


dplore:
References for packet mirroring  with filtering to remove destination

Cisco IOS XR
https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/Interfaces/73x/configuration/guide/b-interfaces-config-guide-cisco8k-r73x/configuring-traffic-mirroring.html#Cisco_Concept.dita_aef1adcd-7dd5-44f2-9f60-10fc37a99b05

Juniper JunOS
https://kb.juniper.net/InfoCenter/index?page=content&id=KB35163&actp=METADATA

Arista EOS:
https://eos.arista.com/eos-4-15-3f/gre-filtered-mirroring/

release/models/mirror/openconfig-mirror.yang
```diff
@@ -0,0 +1,165 @@
...
+        "Which action is performed when a packet is selected for mirroring";
+    }
+
+    leaf name {
+      type string;
+      description
+        "Name of mirror session.";
+    }
+
+    leaf source-interface {
```
rolandphung, line 76: Unsure for the other vendors, but for Arista boxes, there could be multiple interfaces assigned to one mirror session.

On a side note, each interface could have an independent direction and acl.

release/models/mirror/openconfig-mirror.yang
```diff
@@ -0,0 +1,165 @@
...
+          description
+            "Mirror packets in both the transmit and receive directions.";
+        }
+      }
+      default BOTH;
+      description
+        "Direction of packet flow through the source interface to mirror.";
+    }
+
+    leaf acl {
```
rolandphung, line 108: For Arista boxes, ACL is only applied to one of the following: ipv4, ipv6 or macs, not all.

release/models/mirror/openconfig-mirror.yang
```diff
@@ -0,0 +1,165 @@
...
+      type leafref {
+        path "/oc-acl:acl/oc-acl:acl-sets/oc-acl:acl-set/oc-acl:name";
+      }
+      description
+        "ACL used for filtering packets prior to mirroring.";
+    }
+
+    leaf destination-address {
```
rolandphung, line 116: I think there needs to be a source-address for GRE headers as well

release/models/mirror/openconfig-mirror.yang
```diff
@@ -0,0 +1,165 @@
...
+        type oc-inet:ip-address;
```
rolandphung, line 117: The set of IP addresses (source and destination) should reference a single network-instance.

release/models/mirror/openconfig-mirror.yang
```diff
@@ -0,0 +1,165 @@
...
+      "Operational state data for packet mirroring sessions";
+      // TODO: Add state variables
+  }
+
+
+  grouping mirroring-top {
+    description
+      "Grouping for the packet mirroring sessions.";
+
+    container mirror-sessions {
```
rolandphung, line 135: Is this collection missing a nested `mirror-session` list collection with a leaf reference to config/name as the key?

rolandphung: Maybe this could be broken into something like `/mirroring/sessions/session[name]/config`. I realize now that `mirror-sessions` made me imply that it was supposed to be more than one. Breaking it down like this allows for configurations that can be applied across all mirroring sessions.

earies:
Some initial comments/feedback:

* Various implementations do not implement port-mirroring in this fashion thus this caters closer to 1 specifically.  Implementation differences can vary greatly here much like is seen in various other domains.
* For remote mirroring, there are generally 2 methods to implement encap
  * Inline encap defining the tunnel attributes directly within the domain (as is this PR)
  * Tunnel interfaces where policies are defined separately (e.g. ACL) with specific actions to direct mirrored traffic
* In addition, this specifies the mirror source (or ACL defining interesting traffic) directly within a domain.  An alternative is to leverage ACLs applied to interfaces that co-exist for alternate purposes to set an action of 'mirror' for a given ACE.

If an implementation supports the inverse as described above, then not only would this present various constructs unaccounted for (e.g. tunnel interface creation, ACL definition) but also issues with co-existence of any other domains that translate to an underlying ACL applied to an interface

dplore:
Open question,  should enable/disable be defined as a gnoi rather than configuration?    If the intent is that an enabled mirror session is ephemeral and should not be persisted across reboots, then the mirror enable/disable should be moved into a gnoi action instead of config.   I appreciate any feedback on this.

release/models/mirror/openconfig-mirror.yang
dplore: Refactored to create a list of mirror-sessions and added interfaces as a leaf-list to each mirror session

release/models/mirror/openconfig-mirror.yang
dplore: added network instance reference

release/models/mirror/openconfig-mirror.yang
dplore: refactored mirror-sessions to be a list.

release/models/mirror/openconfig-mirror.yang
dplore: done

release/models/mirror/openconfig-mirror.yang
dplore: It is fair for implementations to return an error for acl-types which are not supported.  I will add a note reflecting this.

vishnureddybadveli:
References for packet mirroring HPE/ARUBA:

https://techhub.hpe.com/eginfolib/networking/docs/switches/5820x-5800/5998-7395r_nmm_cg/content/441755847.htm

release/models/mirror/openconfig-mirror.yang
```diff
@@ -0,0 +1,241 @@
...
+  grouping mirror-file-config {
+    description
+      "Configuration data for files used to store packets mirrored to a
+       file.";
+
+    leaf max-file-size {
+      type uint32;
+      when "../mirror-action = 'FILE'";
+      description
+        "Maximum size of the file used to store mirrored packets.  The
```
xavier-contreras, line 86: Should specify the unit of measure for file size?  I assume bytes?

xavier-contreras:
Left a minor comment on the yang.

On the topic of `if the intent is that an enabled mirror session is ephemeral and should not be persisted across reboots`, I believe a common use case is to also port mirror to an interface going to an IDS for example for security fingerprinting, and that is typically persistent.

release/models/mirror/openconfig-mirror.yang
```diff
@@ -0,0 +1,241 @@
...
+  // identity statements
+  // typedef statements
+  typedef mirror-action {
+    type enumeration {
+      enum FILE {
+        description
+          "Captured packets are stored in a local file on the
+           device in pcap format.";
+      }
+      enum GRE {
```
LimeHat, line 64: I'm wondering if this is the best approach. 

There could be other encapsulations (e.g. GUE, VXLAN) that can share a set of properties (source-ip, network-instance, etc).

Additionally, the blanket "GRE" enum does not cover different parameters that can be associated with a GRE tunnel and/or GRE flavors. For example, it can be l2 over gre or l3 over gre. GRE tunnel itself can have additional parameters such as the presence of the `key` field in the header.

Perhaps it would be more convenient to use enum `tunnel` or `remote` and then create another grouping that can be used to define the tunnel parameters.

LimeHat (changes requested):

release/models/mirror/openconfig-mirror.yang
```diff
@@ -0,0 +1,241 @@
...
+  // extension statements
+  oc-ext:regexp-posix;
+  oc-ext:catalog-organization "openconfig";
+  oc-ext:origin "openconfig";
+
+  // feature statements
+  // identity statements
+  // typedef statements
+  typedef mirror-action {
+    type enumeration {
```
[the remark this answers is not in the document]
dplore: I'll add a port option.

Regarding `mirror-action` I am ok to change this typdef name.  I didn't use "destination" because that overlapped with a destination IP address (if using the remote option).  But maybe that is not too confusing, since `destination-address` is used for the IP addr.   But maybe it's ok:

* `/mirroring/sessions/session/destination` - PORT, FILE, REMOTE
* `/mirroring/sessions/session/destination-address` - the destination IP address for REMOTE

Any other suggestions for renaming mirror-action?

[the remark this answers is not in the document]
rolandphung: Maybe `mirror-redirection`?

PR #551, opened 2022-01-05
2022-01-05: Sync models and public repos

dplore:
Includes change:
* Add support for multiple errors per IC subsystem block #1068 models, #505 public

### DIFF
--- a/release/models/platform/openconfig-platform-pipeline-counters.yang
+++ b/release/models/platform/openconfig-platform-pipeline-counters.yang
@@ -65,9 +65,16 @@ module openconfig-platform-pipeline-counters {
     5 blocks, is to have the abililty to receive all drop counters from
     all 5 blocks, for example, with one request.";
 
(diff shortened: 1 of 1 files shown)


robshakir:
To merge these, we try and keep a standard format commit message of the form:

```
short description of change

 * (A) an-added-file.yang
   - description of added file
 * (D) a-removed-file.yang
  - description of removed file
 * (M) modified-file.yang
  - description of changes
```

In this workflow there's a need to do `git commit --amend` (you'll need to create a new branch for this, since amending causes conflicts). [this script](https://github.com/openconfig/models/wiki/CommitMessageScript) auto-creates the template to fill out for a commit if you're creating one afresh.

One thing I was looking into is how to enforce this before a commit is accepted upstream, but I didn't find a decent way to do it with squash and merge (which we use).

dplore:
Fixed in PR #552

Issue #184, opened 2018-10-02
network-instance-ref path should have namespaces


iwanb:
In openconfig-network-instance there is this typedef:
```
  typedef network-instance-ref {
    type leafref {
      path "/network-instances/network-instance/config/name";
    }
    description
      "A re-usable type that can be referenced within other
       modules that references a network instance.";
  }
```

Which is used in openconfig-aft-network-instance:
```
grep network-instance-ref aft/* -C 2
aft/openconfig-aft-network-instance.yang-
aft/openconfig-aft-network-instance.yang-    leaf network-instance {
aft/openconfig-aft-network-instance.yang:      type oc-ni:network-instance-ref;
aft/openconfig-aft-network-instance.yang-      description
aft/openconfig-aft-network-instance.yang-        "The network-instance within which the next-hop should be resolved.
--
aft/openconfig-aft-network-instance.yang-
aft/openconfig-aft-network-instance.yang-    leaf origin-network-instance {
aft/openconfig-aft-network-instance.yang:      type oc-ni:network-instance-ref;
aft/openconfig-aft-network-instance.yang-      description
aft/openconfig-aft-network-instance.yang-        "If the AFT entry was imported from another network instance (e.g., it
```

But the leafref path is resolved in the namespace where it is used, not where it is defined. It's not very clear in the Yang 1.0 RFC but the 1.1 says explicitly:
```
   o  If the "path" statement is defined within a typedef, the context
      node is the leaf or leaf-list node in the data tree that
      references the typedef.
```

I found it using yanglint from libyang, which gives this error when trying to load openconfig-aft-network-instance:
`err : No resolvents found for leafref "/network-instances/network-instance/config/name". (/openconfig-network-instance:network-instances/openconfig-network-instance:network-instance/openconfig-network-instance:afts/openconfig-network-instance:next-hops/openconfig-network-instance:next-hop/openconfig-network-instance:state/network-instance)`

So AFAIK the path should be:
`path "/oc-netinst:network-instances/oc-netinst:network-instance/oc-netinst:config/oc-netinst:name";`

jacossio:
Would it be possible to consider a fix for this issue? @OpenConfigBot

robshakir:
Sure - per the CI, we observe that pyang, goyang and other toolchains do not see that this is an issue -- as long as the prefix resolution works in the correct way (since the context node may use different prefixes than the value defined in the `openconfig-network-instance`) module, then this should be a no-op change.

PR #557, opened 2022-01-19
Add prefix qualifiers to `network-instance-ref`.

robshakir:
```
   - Add prefixes to network-instance-ref typedef.
```

Checking against CI to address #184.


earies:
Looks like we need an approval here still before merge - this LGTM - @robshakir @dplore - PTAL thx

dplore:
Hi @earies, can you resolve the merge conflict?

earies:
@dplore @robshakir - Just noticed this is still pending - can we get this resolved/merged?  Thx

robshakir:
Done, re-running the CI here.

It's not entirely clear to me that this is required though.

RFC6020 says:

> [5.4](https://datatracker.ietf.org/doc/html/rfc6020#section-5.4).  Resolving Grouping, Type, and Identity Names
>
>   Grouping, type, and identity names are resolved in the context in
>   which they are defined, rather than the context in which they are
>   used.

Given that this is the case, then the `typedef` here is defined in the context of `openconfig-network-instance` that does not need this prefixing, such that I don't think toolchains that require it are fully implementing this correctly.

earies:
> Done, re-running the CI here.
> 
> It's not entirely clear to me that this is required though.
> 
> RFC6020 says:
> 
> > [5.4](https://datatracker.ietf.org/doc/html/rfc6020#section-5.4).  Resolving Grouping, Type, and Identity Names
> > Grouping, type, and identity names are resolved in the context in
> > which they are defined, rather than the context in which they are
> > used.
> 
> Given that this is the case, then the `typedef` here is defined in the context of `openconfig-network-instance` that does not need this prefixing, such that I don't think toolchains that require it are fully implementing this correctly.

Thx @robshakir and yes per that section in both 1.0 and 1.1 spec, I am also leaning to this being a common toolchain resolver issue now that likely needs to be dug into further.

robshakir:
@earies thanks! I was just wondering whether we should just add a `TODO` and submit this.

It seems to me that if an implementation does implement the above logic, then `oc-netinst` is alwayhs resolved in the context of `openconfig-network-instance`. To this end, it should be harmless to have this be included.

The thing that I am not sure is whether implementations are robust to this -- e.g., if I do:

```yang
module a {
  prefix "a";
  typedef a-ref {
    type leafref { path "/a:a"; }
  }
  leaf-list a { type string; }
}
```

and then

```yang
module b {
   prefix b;
   import a { prefix not-a; }
   leaf q { type not-a:a-ref; }
}
```

then in this case, an implementation that interprets `a-ref` in the context of `a` will work, but one that uses the context of `b` will not work correctly.

ygot/goyang appear to support this case fine, but I'm not sure about all toolchains.

robshakir:
Which major toolchain is it that doesn't work with the current setup?

earies:
> Which major toolchain is it that doesn't work with the current setup?

So far you can observe this behavior in `libyang` (yanglint 2.0.221) and `yangson` (1.4.13)

```
module a {
  yang-version "1";
  namespace "urn:a";
  prefix a;

  typedef instance-ref {
    type leafref {
      path "/a/b/name";
    }
  }

  container a {
    list b {
      key name;
      leaf name { type string; }
    }
  }
}
```
```
module c {
  yang-version "1";
  namespace "urn:c";
  prefix c;

  import a {
    prefix a;
  }

  container c {
    leaf instance {
      type a:instance-ref;
    }
  }
}
```

```
{
    "a:a": {
        "b": [
            {
                "name": "TEST1"
            }
        ]
    },
    "c:c": {
        "instance": "TEST1"
    }
}
```

yangson:
```
yangson.exceptions.InvalidLeafrefPath: c:instance
```

libyang:
```
libyang err : Not found node "a" in path. (Schema location "/c:c/instance".)
```

robshakir:
Ah, the fix was reverted in libyang. There was a discussion [here](https://github.com/CESNET/libyang/issues/1082). My interpretation is not necessarily the same here, but as with the wonderful nature of rfc6020 or rfc7950 there seem to be many places where different interpretations are possible.

Given that this doesn't, to our CI's knowledge, break tooling that we run CI against, and should fix libyang+yangson -- I suggest we submit it. WDYT?

earies:
> Ah, the fix was reverted in libyang. There was a discussion [here](https://github.com/CESNET/libyang/issues/1082). My interpretation is not necessarily the same here, but as with the wonderful nature of rfc6020 or rfc7950 there seem to be many places where different interpretations are possible.
> 
> Given that this doesn't, to our CI's knowledge, break tooling that we run CI against, and should fix libyang+yangson -- I suggest we submit it. WDYT?

Agreed on clarifying and fixing the interpretation in consistent fashion across toolchains as well as merging this patch set for now.

This has been done a handful of times already throughout the model sets:
* https://github.com/openconfig/public/pull/598
* https://github.com/openconfig/public/pull/611

robshakir:
Sold and merged, thanks for the review @earies!

PR #562, opened 2022-01-25
Add support for SFP28 and SFP56 form factor identities

LeonGWang:
SFP28 and SFP56 transceiver form-factor identities are missing from
OpenConfig.


LeonGWang:
Closed in favor of #692 so I can sync the master branch.

PR #572, opened 2022-01-31
Remove outdated list of models in documentation.

xw-g:
Proposes to remove the outdated model list and descriptions, because they are already represented by the file & folder structure of `release/models`


robshakir (approved):
LGTM, I don't think this list is useful any more and just creates maintenance overhead.

Merging this change directly here because we do not mirror this file from upstream.

PR #573, opened 2022-02-02
Add a summary document explaining interface/subinterface differences

robshakir:
When discussing device implementations for telemetry where a system maintains a "default" subinterface - there have been some different interpretations as to how to map telemetry to the entities that exist there. This document provides some implementation guidance on this matter.

```
  * (A) doc/reference/interface_subinterface.md
    - Add a document that covers the way that systems that have an
      implied default subinterface should handle telemetry data. This
      addresses an implementation decision discussed with external
      developers.
```


doc/interface_subinterface.md
```diff
@@ -0,0 +1,88 @@
...
+- Represent the state of the physical or logical interface - whether or not it
+  is subdivided.
+
+### Systems with a 'Default' Subinterface
+
+For systems that allow configuration directly onto a parent interface that
+cannot be configured directly in OpenConfig (e.g., IP address information) -
+this configuration should be mapped to a subinterface with index 0. This
+subinterface should have the encapsulation of the 'default' subinterface
+specified if this is relevant - in most cases, the encapsulation iss is
```
dplore, line 58:
```suggestion
specified if this is relevant - in most cases, the encapsulation is
```

doc/interface_subinterface.md
```diff
@@ -0,0 +1,88 @@
...
+   42.
+ - Represent state of the logical subinterface only, i.e., for example there
+   may be separate `oper-status` between subinterfaces.
+
+For systems that have this default mapping of subinterface 0 to a parent
+interface, the configuration of separate IPv4 and IPv6 subinterfaces with no
+specified encapsulations is invalid - since in this case there is ambiguity as
+to which OpenConfig `subinterface` entities map to the "default" subinterface's
+config on the device. (i.e., OpenConfig `subinterface` 4 and 6 configured on
+`GigabitEthernet42` would potentially both map into configuration and state
```
dplore, line 85: Is this meant to refer to index 4 and index 6?  Or ipv4 and ipv6?

robshakir: Index 4 and 6 -- let me look at the wording. :-)

robshakir:
ptal, i believe this is now ready for merge.

doc/interface_subinterface.md
```diff
@@ -0,0 +1,88 @@
...
+
+For systems that have this default mapping of subinterface 0 to a parent
+interface, the configuration of separate IPv4 and IPv6 subinterfaces with no
+specified encapsulations is invalid - since in this case there is ambiguity as
+to which OpenConfig `subinterface` entities map to the "default" subinterface's
+config on the device. (i.e., OpenConfig `subinterface` index 4 and 6 configured
+on `GigabitEthernet42` would potentially both map into configuration and state
+directly under `GigabitEthernet` in the native schema). Conversely, a single
+subinterface that specifies both IPv4 and IPv6 addresses is valid, since this
+unambiguously maps to the "default" subinterface on these systems.
```
dplore, line 88: I found this last part a little difficult to follow.  Does this cover the intent?  Or is there some case I am missing?

```suggestion
 - When IPv4 and IPv6 are configured on the default interface, both must be 
    treated as attached to index 0.
```

dplore (changes requested):

doc/interface_subinterface.md
```diff
@@ -0,0 +1,88 @@
...
+**Contributors**: robjs<sup>†</sup>, ayhchan<sup>†</sup>  
+<sup>†</sup> @google.com
+**April 2019**
+
+## Problem Statement
+The OpenConfig interfaces model provides two constructs for interfaces:
+
+* Interfaces - which represent individual physical or logical interfaces which
+  may or may not be subdivided on a system. A physical interface in OpenConfig
+  *cannot* have a Layer 3 address (i.e., IPv4, or IPv6). For example, the
```
earies, line 11: I think there should be some coverage of expectations for `routed-vlan` and `tunnel` interfaces here as well.

In these situations, the normal `subinterface` IPv4/IPv6 hierarchies are rooted directly at the interface with no subinterface concepts currently.  For implementations that support this still under subinterfaces (e.g. JUNOS), what is the expectation then for interface naming since you have a subinterface index that needs to be accounted for?

The only current solution I see is to concatenate the interface/subinterface into the interface name now for these logical interfaces which diverges protocol from physical interfaces.

Essentially it is the reverse of the problem statement described here where some implementations support direct IP configuration at the base interface level

robshakir: This seems a reasonable clarification to make. I can add to this document to cover this.

The concept that was intended here in the design was that logical interfaces that are within the `/interfaces` hierarchy, we would not allow subdivision of these logical interfaces any further (since there didn't seem to be a case whereby one would operationally want to subdivide such interfaces rather than create a new one).

When you say "diverges protocol from physical interfaces" -- isn't this something that must be true for a logical interface? (i.e., if I have a virtual interface that corresponds to a L3 interface on a VLAN, then this interface might be coupled to multiple physical interfaces, so probably needs to be separate, the same is likely true where we have tunnels that are not associated with physical interfaces.)

earies: > When you say "diverges protocol from physical interfaces" -- isn't this something that must be true for a logical interface?

Not all - I was just pointing out there are 2 classes of interfaces (routed-vlan (SVI/BVI/IRB/VLAN, etc..) + tunnel) that break from the interface/subinterface concept in the modeling when maybe they shouldn't have.  Other logical interfaces, such as a loopback interface in some implementations is represented as just `lo` (which falls into the category of assuming there is an index `0`) or the full `loX.Y` (e.g. `lo0.0`, `lo0.1`, etc..).  In the prior `lo` and `lo:0` are actually different interfaces...

Another concern as a side effect due to the divergence of these 2 classes of interfaces from the rest (which other logicals fall into such as loopbacks) is anywhere that references any constructs such as `oc-if:interface-ref-common` are subject to an incorrect assumption since the construct doesn't exist.

I don't recall the history on this item but is there any reason that routed-vlan and tunnel interfaces just didn't follow the same consistent pattern here where the same protocol could be followed if subinterfaces didn't actually exist (e.g. use index 0 placeholder)?

rgwilton: I would guess the question is one of resources, e.g., does this effectively mean that you end up needing to "interface/sub-interface like" objects to represent every routed-vlan/tunnel where the state of both of these objects is the same.

Of course the IETF model solves this the other way.  Every sub-interface *is* an interface and hence you only need the feature/config binding to interface objects and then it applies equally well to all types of interface (i.e., physical interfaces, VLAN sub-interfaces, LAGs, tunnels, etc).

ElodinLaarz:
Reviewed in OC Community Meeting May 7th, 2026:

Concern is that this PR has not merged yet and is being referenced. If we want to use this (or something derived from this) as the SOT, it should be reviewed and merged in so we avoid referencing an old, open PR.

PR #583, opened 2022-02-10
fix a spelling error of the retransmission-queue-length leaf

awebsters:
Minor spelling error within ospfv2-area-interface-neighbor-state grouping.


robshakir (changes requested):
Thanks for the change -- since this covers a leaf name change, could you make it `0.4.0` please?

Thanks!
r.

robshakir:
Thanks for the fix here -- this LGTM. Going ahead and merging this change.

robshakir (approved):

PR #584, opened 2022-02-11
Proposal for simple key authentication support in IS-IS model

nikhilgArista:
Add two new leaves 'auth-mode' and 'auth-password' under
isis/levels/level/authentication/ and
isis/interfaces/interface/authentication

auth-mode leaf specifies the type of the authentication which
can be either TEXT or MD5. The auth-password leaf specifies
the authentication key used in the IS-IS PDUs.

### DIFF
--- a/release/models/isis/openconfig-isis.yang
+++ b/release/models/isis/openconfig-isis.yang
@@ -18,6 +18,7 @@ module openconfig-isis {
   import openconfig-segment-routing { prefix "oc-sr"; }
   import openconfig-bfd { prefix "oc-bfd"; }
   import openconfig-keychain { prefix "oc-keychain"; }
+  import openconfig-keychain-types { prefix "oc-keychain-types"; }
 
   // Include submodules:
   // IS-IS LSP is the LSDB for IS-IS.
@@ -53,7 +54,13 @@ module openconfig-isis {
             +-> { levels config }
             +-> { level adjacencies }";
 
-  oc-ext:openconfig-version "0.8.0";
+  oc-ext:openconfig-version "0.9.0";
+
+  revision "2022-03-01" {
+    description
+      "Add simple authentication key support.";
+    reference "0.9.0";
+  }
 
   revision "2022-02-24" {
     description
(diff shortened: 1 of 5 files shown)


release/models/isis/openconfig-isis.yang
```diff
@@ -286,6 +292,25 @@ module openconfig-isis {
...
+    leaf auth-mode {
+      type identityref {
+        base oc-isis-types:AUTH_MODE;
+      }
+      description
+        "The type of authentication used in the IS-IS Hello packets.
+
+        This leaf along with the sibling leaf 'auth-password' can be used
+        to configure the simple key authentication.";
+    }
```
dplore, line 304: Commenting on behalf of @robshakir
Do we need to have another leaf that tells us whether we are using the keychain or the simple password here? I think that it'd be possible to do so - since if someone were to populate both keychain and auth-password it might be unclear which is expected to be used.

Of course we could just say that this isn't a valid combination -- if so, let's update the description to ensure that this is noted.

release/models/isis/openconfig-isis.yang
```diff
@@ -1397,6 +1422,26 @@ module openconfig-isis {
         packets, overriding the value of the enabled leaf in this context.";
     }
 
+    leaf auth-mode {
```
dplore, line 1425: Commenting on behalf of @robshakir

Perhaps make a common grouping that is included in both places here, rather than having the repetition?

dplore (changes requested):

release/models/isis/openconfig-isis.yang
nikhilgArista: Yeah it's a good idea to have an explicit leaf to tell the type. I have added a leaf 'auth-type'. It seems there already was an identityref defined for auth-type although not used anywhere. It had three values - NONE, SIMPLE_KEY, KEYCHAIN. I have removed the NONE from it as it would be confusing with the enabled leaf. For no authentication one can just set the 'enabled' leaf to false.

release/models/isis/openconfig-isis.yang
nikhilgArista: I have now moved the simple authentication leaves into a separate grouping.

release/models/keychain/openconfig-keychain-types.yang
```diff
@@ -35,12 +41,6 @@ module openconfig-keychain-types {
     "Base identify to define the type of authentication";
   }
 
-  identity NONE {
-    base AUTH_TYPE;
-    description
-    "NO authentication is used";
-  }
-
```
dplore, line 43: I understand the desire to remove this leaf, but considering backwards compatibility, is it necessary to remove this identity?

nikhilgArista: Today there is no leaf using this type 'AUTH_TYPE' so I think there is no backward compatibility problem.

release/models/isis/openconfig-isis.yang
```diff
@@ -1366,8 +1419,8 @@ module openconfig-isis {
       default false;
       description
         "When this leaf is set to true, authentication of IS-IS PSNP, CSNP and
-        LSP packets is enabled using the authentication details specified in
-        the keychain in the sibling leaf.
+        LSP packets is enabled using the specified authentication details in
+        the sibling leaves.
 
         The simbling 'disable-<type>' leaves can be used to override the value
```
robshakir, line 1425: `s/simbling/sibling/` - could you fix this whilst making changes to this file please? Thanks!

release/models/keychain/openconfig-keychain-types.yang
robshakir: I think that this is unreferenced, as you say, so I'm OK with removing it.

robshakir (changes requested):

release/models/keychain/openconfig-keychain-types.yang
dplore: I was thinking of the use case where config generation software in the community has been written to utilize this leaf would be affected by its removal.  FWIW, I scanned our code and we are not using this identity, so from a selfish point of view, I am ok with removing it as well. Marking resolved.

dplore (approved):
Please resolve robshakir comment and merge conflicts then I can also give approval.

release/models/isis/openconfig-isis.yang
nikhilgArista: Fixed

PR #585, opened 2022-02-16
Add support for SFP_DD and DSFP form factor identities

LeonGWang:
SFP_DD and DSFP transceiver form factor identities are missing from
OpenConfig.

### DIFF
--- a/release/models/optical-transport/openconfig-transport-types.yang
+++ b/release/models/optical-transport/openconfig-transport-types.yang
@@ -22,7 +22,13 @@ module openconfig-transport-types {
     "This module contains general type definitions and identities
     for optical transport models.";
 
-  oc-ext:openconfig-version "0.18.1";
+  oc-ext:openconfig-version "0.19.0";
+
+  revision "2023-07-24" {
+    description
+      "Add SFP_DD and DSFP form factor identities.";
+    reference "0.19.0";
+  }
 
   revision "2023-02-08" {
     description
@@ -865,6 +871,22 @@ module openconfig-transport-types {
     description
       "Small form-factor pluggable transceiver supporting up to
       50 Gb/s signal";
+  }
+
+  identity SFP_DD {
+    base TRANSCEIVER_FORM_FACTOR_TYPE;
+    description
+      "SFP-DD electrical interfaces will employ 2 lanes that operate up
+      25 Gbps NRZ modulation or 56 Gbps PAM4 modulation, providing
+      solutions up to 50 Gbps or 112 Gbps PAM4 aggregate";
+    reference "http://sfp-dd.com";
+  }
+
+  identity DSFP {
+    base TRANSCEIVER_FORM_FACTOR_TYPE;
+    description
+      "A transceiver implementing the DSFP Transceiver specification";
+    reference "https://dsfpmsa.org/";
   }
 
   identity XFP {


release/models/optical-transport/openconfig-transport-types.yang
```diff
@@ -669,6 +675,20 @@ module openconfig-transport-types {
       up to 16 Gb/s signals, including 10 GbE and OTU2";
   }
 
+  identity SFP_DD {
+    base TRANSCEIVER_FORM_FACTOR_TYPE;
+    description
+      "SFP DD pluggable transceiver supporting up to 100 Gb/s signal
+      with 2x50G physical channels.";
```
jqjqlee, line 682: Actually, physical channel corresponds to optical lane. SFP-DD form factor does not confine the implementation of optical lanes. For example, a SFP-DD 100GBASE-DR1 module may have a single 100G optical lane. This problem in description also applies to existing QSFP28 and added DSFP. For added identities since now, we should stop using "physical channels" in description to avoid confusion. "electrical interface" or "electrical lane" is suggested. What do you think? @yawei-yin @ejbrever

ejbrever:
@thofmeister could you please help review this?

dplore:
@ejbrever can you follow up with @thofmeister offline and get back to us?  There is some interest to complete this PR.  Thank you!

release/models/optical-transport/openconfig-transport-types.yang
LeonGWang: Can I get a consensus and an example of the new description (using "electrical interface/lane")?

LeonGWang:
Hi all, this PR is stalled. Could we let this PR merge as a band-aid solution? I am happy to change the descriptions of the new identities when a new consensus is agreed upon.

Many thanks for considering my request.

release/models/optical-transport/openconfig-transport-types.yang
dplore: How about
```
SFP-DD electrical interfaces will employ 2 lanes that operate up to 25 Gbps NRZ modulation or 56 Gbps PAM4 modulation, providing solutions up to 50 Gbps or 112 Gbps PAM4 aggregate
```

ref: http://sfp-dd.com/

@ejbrever what do you think?

release/models/optical-transport/openconfig-transport-types.yang
```diff
@@ -867,6 +873,20 @@ module openconfig-transport-types {
...
+  }
+
+  identity DSFP {
+    base TRANSCEIVER_FORM_FACTOR_TYPE;
+    description
+      "Dual Small form-factor pluggable transceiver supporting up to
+      100 Gb/s signal with 2x50G physical channels.";
```
dplore, line 887: This one has many different electrical options.  How about just pointing to the reference specification?

```suggestion
      "A transceiver implementing the DSFP Transceiver specification.";
      reference "https://dsfpmsa.org/";
```

release/models/optical-transport/openconfig-transport-types.yang
LeonGWang: Thanks for the example @dplore! I have updated the description according to your suggestion.

release/models/optical-transport/openconfig-transport-types.yang
LeonGWang: Thanks @dplore, `DSFP` description is also updated following your example.

ejbrever (approved):
LGTM

release/models/optical-transport/openconfig-transport-types.yang
dplore: I think this makes sense now.  LGTM.

release/models/optical-transport/openconfig-transport-types.yang
dplore: LGTM, thanks

dplore (approved):

PR #586, opened 2022-02-16
Add VLAN ID to ingress tracking values.

greg-dennis:


release/models/ate/openconfig-ate-flow.yang
```diff
@@ -23,7 +24,13 @@ module openconfig-ate-flow {
     stream of packets whose definition is outside of the context of this
     module.";
 
-  oc-ext:openconfig-version "0.2.0";
+  oc-ext:openconfig-version "0.2.1";
```
robshakir, line 27: Can you make this `0.3.0` since this is not a backwards compatible change?

greg-dennis: Done

robshakir:
Thanks Greg -- we'll merge this tomorrow as we adjust the primary SoT for the models.

PR #593, opened 2022-02-23
Add Hello PDU padding type to IS-IS global configuration.

rpazyaquian:
### Original Problem

We need to be able to disable IS-IS Hello padding on a global level. This requires specifying the padding type for IS-IS Hello PDUs in global IS-IS configuration. This is currently only available on an interface level.

### Proposed Solution

Add a `hello-padding` leaf of type `oc-isis-types:hello-padding-type` to `isis-global-config`, set to `STRICT` by default.


dplore (approved):
LGTM

rpazyaquian:
Thank you! Anything else I need to do to get this merged in?

dplore:
Noting else needed, thanks!  Merging

Issue #594, opened 2022-02-25
Import Policy in OSPF Configuration


sbarguil:
I would be usefull to include the `apply-policy` container into the `/network-instances/network-instance/protocols/protocol/ospfv2/global` configuration. This will allow to import routes learnt from other protocols into OSPF whitout using the table connections, providing more control, flexibiity in the protocols configuration.

The functionality will be equivalent to some well know config attributes like:

- **Huawei**:
ospf [ process-id ]
import-route { limit limit-number | { bgp [ permit-ibgp ] | direct | unr | rip [ process-id-rip ] | static | isis [ process-id-isis ] | ospf [ process-id-ospf ] } [ cost cost | type type | tag tag | route-policy route-policy-name ] * }

- **Nokia**:
ospf
import policy-name [policy-name]: 
__The import route policy name, up to a maximum of 5. Allowed values are any string up to 32 characters long composed of printable, 7-bit ASCII characters. If the string contains special characters (#, $, spaces, etc.), the entire string must be enclosed within double quotes. The specified names must already be defined.__

This feature is already supported in Openconfig for the BGP protocol under the peer and Neighboor configuration:

````
module: openconfig-network-instance
  +--rw network-instances
     +--rw network-instance* [name]
        +--rw name                       -> ../config/name
        +--rw protocols
           +--rw protocol* [identifier name]
              +--rw identifier          -> ../config/identifier
              +--rw name                -> ../config/name
              +--rw config
              +--rw bgp
              |  +--rw global
              |  +--rw neighbors
              |  |  +--rw neighbor* [neighbor-address]
              |  |     +--rw apply-policy
              |  |     |  +--rw config
              |  |     |  |  +--rw import-policy*           -> /oc-rpol:routing-policy/policy-definitions/policy-definition/name
              |  |     |  |  +--rw default-import-policy?   default-policy-type
              |  |     |  |  +--rw export-policy*           -> /oc-rpol:routing-policy/policy-definitions/policy-definition/name
              |  |     |  |  +--rw default-export-policy?   default-policy-type
              |  |     |  +--ro state
              |  |     |     +--ro import-policy*           -> /oc-rpol:routing-policy/policy-definitions/policy-definition/name
              |  |     |     +--ro default-import-policy?   default-policy-type
              |  |     |     +--ro export-policy*           -> /oc-rpol:routing-policy/policy-definitions/policy-definition/name
              |  |     |     +--ro default-export-policy?   default-policy-type
              |  +--rw peer-groups
[...]

PR #595, opened 2022-03-01
Add policy support #594

sbarguil:
To solve #594

### DIFF
--- a/release/models/ospf/openconfig-ospfv2-area-interface.yang
+++ b/release/models/ospf/openconfig-ospfv2-area-interface.yang
@@ -25,7 +25,13 @@ submodule openconfig-ospfv2-area-interface {
     "This submodule provides OSPFv2 configuration and operational
     state parameters that are specific to the area context";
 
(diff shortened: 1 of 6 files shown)


robshakir:
i don't think the way that this PR proposes to make this change is compatible with the modelling approach we have for import policies today.

the import policies that are specified in bgp are defined per-neighbour or group, and are not intended to be able to import routes from other protocols. the design for doing that is through table-connections. i.e., to do something like redistribute from is-is to bgp, one must create a table connection between INSTALL_PROTOCOL BGP and INSTALL_PROTOCOL ISIS that then makes these routes available. some systems do this 'by default' but others do not and need 'redistribute XXX' type configuration, so we had to figure whether this should be explicit or not since these are two fundamentally different approaches.

i believe what you're trying to do here is to rely on implicit connections and then define the import policy as to /what/ makes it into ospf, is that correct? i think we need to consider how that can co-exist with the table connections without creating two different ways to do the same thing.

PR #598, opened 2022-03-06
Add prefix qualifier to keychain-ref

earies:
* (M) release/models/keychain/openconfig-keychain.yang
    - Add prefixes to keychain-ref typedef

Similar issue to https://github.com/openconfig/public/pull/557, the leafref
must be qualified when used and referenced from alternate schema locations


dplore (approved):
LGTM

PR #599, opened 2022-03-08
User-defined defaults and description updates for list keys

earies:
    - Update network-instance name description to provide implementation
      guidance on default name for DEFAULT_INSTANCE should OpenConfig
      schemas not be used for configuration but rather for state only as
      well as suggested name
    - Update protocol instance name with suggested value as well as
      default name should OpenConfig schemas not be used for
      configuration but rather for state only
    - Update grpc server name with suggested value as well as default
      name should OpenConfig schemas not be used for configuration but
      rather for state only

This PR is to kick off the concept of embedding implementation suggestions (and
defaults where applicable) for where user-defined fields are used as list keys.
Most notably starting in areas where users have a hard time selecting an
appropriate value and where an implementation may only support single instances
in a multi-instance capable model structure.

The notion of implementation guidance and default values also caters to
scenarios where a client may want to consume OpenConfig modeled state data
without using OpenConfig modeled configuration.  These suggestions and defaults
enable the implementation to pack known values into list keys to formulate
compliant paths.

### DIFF
--- a/release/models/network-instance/openconfig-network-instance.yang
+++ b/release/models/network-instance/openconfig-network-instance.yang
@@ -47,7 +47,14 @@ module openconfig-network-instance {
     virtual switch instance (VSI). Mixed Layer 2 and Layer 3
     instances are also supported.";
 
-  oc-ext:openconfig-version "0.16.2";
+  oc-ext:openconfig-version "0.16.3";
+
+  revision "2022-03-07" {
+    description
+      "Description updates for DEFAULT_INSTANCE implementation
+      guidance and default value/guidance for protocol instances";
+    reference "0.16.3";
+  }
 
   revision "2021-11-17" {
     description
@@ -1135,8 +1142,10 @@ module openconfig-network-instance {
     leaf name {
       type string;
       description
-        "An operator-assigned unique name for the forwarding
-        instance";
+        "An operator-assigned unique name for the network instance.
+        If the operator does not designate a name for the instance of
+        type 'DEFAULT_INSTANCE' (e.g. config), the implementation
+        should use the name of 'DEFAULT' (e.g. state).";
     }
 
     leaf type {
@@ -1146,12 +1155,14 @@ module openconfig-network-instance {
       description
         "The type of network instance. The value of this leaf
         indicates the type of forwarding entries that should be
-        supported by this network instance. Signalling protocols
-        also use the network instance type to infer the type of
-        service they advertise; for example MPLS signalling
-        for an L2VSI network instance would infer a VPLS service
-        whereas a type of L2PTP would infer a VPWS (pseudo-wire)
-        service";
+        supported by this network instance. Signalling protocols also
+        use the network instance type to infer the type of service
+        they advertise; for example MPLS signalling for an L2VSI
+        network instance would infer a VPLS service whereas a type of
+        L2PTP would infer a VPWS (pseudo-wire) service.
+
+        An implementation must support only a single network-instance
+        of type 'DEFAULT_INSTANCE'.";
     }
 
     leaf enabled {
@@ -1204,8 +1215,16 @@ module openconfig-network-instance {
 
     leaf name {
       type string;
+      default "DEFAULT";
       description
-        "A unique name for the protocol instance";
+        "A unique name for the protocol instance.
+
+        If the operator does not designate a name for the protocol
+        instance (e.g. config), the implementation should use the
+        name of 'DEFAULT' (e.g. state).  In addition, for
+        implementations that support single protocol instances, the
+        default value is recommended for consistency and uniqueness
+        per protocol instance.";
     }
 
     leaf enabled {
--- a/release/models/system/openconfig-system-grpc.yang
+++ b/release/models/system/openconfig-system-grpc.yang
@@ -22,9 +22,16 @@ module openconfig-system-grpc {
     to be included in the list.";
 
 
-  oc-ext:openconfig-version "0.1.1";
+  oc-ext:openconfig-version "0.1.2";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
+
+  revision "2022-03-07" {
+    description
+      "Description and default value updates for grpc-server
+      implementation guidance.";
+    reference "0.1.2";
+  }
 
   revision "2021-06-16" {
     description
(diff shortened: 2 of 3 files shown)


release/models/network-instance/openconfig-network-instance.yang
```diff
@@ -1146,12 +1155,14 @@ module openconfig-network-instance {
...
-        whereas a type of L2PTP would infer a VPWS (pseudo-wire)
-        service";
+        supported by this network instance. Signalling protocols also
+        use the network instance type to infer the type of service
+        they advertise; for example MPLS signalling for an L2VSI
+        network instance would infer a VPLS service whereas a type of
+        L2PTP would infer a VPWS (pseudo-wire) service.
+
+        An implementation must support only a single network-instance
+        of type 'DEFAULT_INSTANCE'.";
```
dplore, line 1165: This makes sense to me.  I struggle to find a way to programmatically define this constraint in yang.  I'd like to see some comment from the community if this constraint is acceptable.  In my limited knowledge, devices only support one default VRF/AFT.

earies: If we think we want to implement this as an actual model enforcement, we can apply a `must` constraint at the top-level `network-instances` container like:

```
    must "not(count(./network-instance/config[type='oc-netinst-types:DEFAULT_INSTANCE']) > 1)" {
      error-message "Only a single network-instance of type DEFAULT_INSTANCE is permitted";
    }
```

robshakir: I'd propose that this is where we use feature profiles/tests against the schema to implement these kind of constraints. We've generally stuck with the philosophy that the YANG here is constrained to defining the schema, and not necessarily complex inter-field validation, since we then end up mandating that there is complex XPATH parsing.

I think there's also concerns here (for me) that we make OpenConfig more brittle by doing this, since we start to try and describe more different constraints in the schema, that might not be valid for every implementation. (I do not think that this is one, but think there are cases where this will be more blurry.)

release/models/network-instance/openconfig-network-instance.yang
```diff
@@ -1204,8 +1215,16 @@ module openconfig-network-instance {
...
     leaf name {
       type string;
+      default "DEFAULT";
       description
-        "A unique name for the protocol instance";
+        "A unique name for the protocol instance.
+
+        If the operator does not designate a name for the protocol
+        instance (e.g. config), the implementation should use the
+        name of 'DEFAULT' (e.g. state).  In addition, for
```
robshakir, line 1224: So, to understand here -- if we were to have a device come up with no configuration, it should use `DEFAULT` as the value, in the case that a client changes this name to `MY_FANCY_DEFAULT_INSTANCE_NAME` then I expect that there would be deletes for that name, and then readvertisement of all state under that new path. Is that a reasonable expectation?

release/models/system/openconfig-system-grpc.yang
```diff
@@ -101,9 +108,16 @@ module openconfig-system-grpc {
...
       description
         "The name of the gRPC server instance that is running on
-        the local system.";
+        the local system.
+
+        If the operator does not designate a name for the protocol
+        instance (e.g. config), the implementation should use the
+        name of 'DEFAULT' (e.g. state).  In addition, for
+        implementations that support a single gRPC server instance,
+        the default value is recommended for consistency.";
```
robshakir, line 120: Does this assume that there is a default gRPC server running on the system regardless of config at boot time?

robshakir (commented):
Thanks for suggesting this Ebben, I definitely think its worth discussing.

One area that I think we probably should discuss is the relationship between some of these defaults and the config used at boot time. There is an implication in some of the comments that some of these values are mutable post-initialisation, which I think has some implementation implications that we might want to consider.

I'm fine, in general, though with an approach that says "an implementor is recommended to use this name as a default value".

release/models/network-instance/openconfig-network-instance.yang
dplore: My opinion: +1, I like the approach of using of code to implement 'branching logic constraints' vs. in the schema.

earies: agreed - will leave this out of schema constraints but are we in agreement of the wording used to give guidance and thus backend/client enforcement?  Loosely allowing the possibility of > 1 instances of type `DEFAULT_INSTANCE` is a recipe for disaster especially given other constraints and assumptions throughout the model sets imo

release/models/network-instance/openconfig-network-instance.yang
earies: More so for the case of non-OpenConfig modeled configuration (e.g. native) or some domain underneath `network-instances` is implied natively is when the default values would be utilized.

For example - native configuration of BGP in an implementation but a client is looking to consume state data according to OpenConfig modeling.  Since various user-defined keys must be configured to give context to formulate the path, these defaults are essentially "helpers" in addition to guidance/recommendations for consistency should one configure via OpenConfig and an implementation only supports a single instance.

Since these are still user-defined (that part is not changing) then yes a client could change the name or initially provision one to their liking (should the implementation support).  The implementation would then need to migrate and adapt to the new keys which in reality brings up operational considerations of in-flight migrations and expectations of this magnitude (but outside the scope of this PR)

release/models/system/openconfig-system-grpc.yang
earies: No - a `grpc-server` still needs to be instantiated by way of configuration.  A default value here does not imply the list member is always active with any downstream default/mandatory attributes

Default values on list keys are a bit unique because it must be defined to be instantiated so it does not solve that case

But rather it is similar to above in that:
* It is used as recommendation/guidance for clients to use to provision a single instance implementation (or the main list member) - driving consistency outside the model to *some* free-form string values where it may make sense
* It is used as a default name should a grpc-server endpoint be provisioned by way of non-OpenConfig (e.g. native) so that state context can be fulfilled to formulate a valid path/instance should a client still be interested in say subscribing to `/system/grpc-servers`

dplore: SGTM

release/models/network-instance/openconfig-network-instance.yang
dplore: SGTM

release/models/network-instance/openconfig-network-instance.yang
dplore: SGTM

dplore (approved):

dplore:
This was discussed with the operators group today.  Feedback was we should make this a major revision and point out in this PR that this is may be a backwards incompatible change for some implementations due to what could be a NEW default value of a network instance name (for example if a vendor currently has some default name string such as "BASE" and this change requires  “DEFAULT”

earies:
Thx @dplore - sounds reasonable to me.  Will update the version and revision descriptions to reflect this across the models

dplore (approved):

PR #600, opened 2022-03-09
Add BGP neighbor state confederation support

yaonanliang:
Add `confederation` leaf to BGP neighbor state container to indicate whether the peer is configured in the same confederation of the local BGP peer.

### DIFF
--- a/release/models/optical-transport/openconfig-transport-types.yang
+++ b/release/models/optical-transport/openconfig-transport-types.yang
@@ -22,14 +22,20 @@ module openconfig-transport-types {
     "This module contains general type definitions and identities
     for optical transport models.";
 
-  oc-ext:openconfig-version "0.17.0";
+  oc-ext:openconfig-version "0.17.1";
+
+  revision "2022-12-05" {
+    description
+      "Fix trailing whitespace";
+    reference "0.17.1";
+  }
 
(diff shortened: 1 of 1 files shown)


release/models/bgp/openconfig-bgp-neighbor.yang
```diff
@@ -30,7 +30,14 @@ submodule openconfig-bgp-neighbor {
     "This sub-module contains groupings that are specific to the
     neighbor context of the OpenConfig BGP module.";
 
-  oc-ext:openconfig-version "8.0.0";
+
+  oc-ext:openconfig-version "9.0.0";
```
robshakir, line 34: Does this need to be a major revision change? If it's just a new leaf addition then it can be `8.1.0` I think.

robshakir (changes requested):

robshakir:
Please sign the CLA (either being added to the corporate CLA your org has signed, or individually if your organisation permits this) for us to move this forward.

Thanks!
r.

release/models/bgp/openconfig-bgp-neighbor.yang
yaonanliang: Updated. Thanks.

PR #602, opened 2022-03-16
Add utilization leaves for memory blocks in pipeline-counters model

aredmon8551:
(M) release/models/platform/openconfig-pipeline-counters.yang
   -Add leaf for each memory attribute to indicate utilization

This proposal seeks to add three new leaves to the OpenConfig pipeline-counters model to represent ASIC memory utilization as relative values.

The primary motivation behind this proposal is that these leaves provide a more vendor and architecture agnostic measurement of memory utilization, and offload the often complex logic of deriving an accurate utilization of memory to the implementor (vendor). Lookup, nexthop, and ACL memories vary greatly across vendors and implementations. Each can be implemented on a variety (and combination) of memory types (e.g., TCAM, SRAM, various types of DRAM) and using a wide variety of search approaches and data structures. For instance, Juniper's express family of ASICs uses several different memories, search algos, and data structures just for ACL/filter programming which makes representing usage in entries or bytes difficult and ultimately low signal.

Several vendor implementations today yield relative values - EOS does so for both ACL and lookup memory (usage/util metrics):
https://eos.arista.com/introduction-to-managing-eos-devices-platform-specific-monitoring-and-troubleshooting/

IOS-XR also vends relative utilization values per lookup table (page 126):
https://www.ciscolive.com/c/dam/r/ciscolive/us/docs/2018/pdf/BRKARC-3000.pdf

### DIFF
--- a/release/models/platform/openconfig-platform-pipeline-counters.yang
+++ b/release/models/platform/openconfig-platform-pipeline-counters.yang
@@ -65,9 +65,15 @@ module openconfig-platform-pipeline-counters {
     5 blocks, is to have the abililty to receive all drop counters from
     all 5 blocks, for example, with one request.";
 
-  oc-ext:openconfig-version "0.2.1";
+  oc-ext:openconfig-version "0.3.0";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
+
+  revision "2022-03-16" {
+    description
+      "Add percent utilization leaves for lookup, nexthop, and ACL memories.";
+    reference "0.3.0";
+  }
 
   revision "2022-01-19" {
     description
@@ -504,6 +510,12 @@ module openconfig-platform-pipeline-counters {
         "The amount of memory used in the lookup subsystem.";
     }
 
+    leaf lookup-memory-utilization {
+      type oc-types:percentage;
+      description
+        "Percent utilization of overall lookup memory.";
+    }
+
     leaf nexthop-memory {
       type uint64;
       units bytes;
@@ -516,6 +528,12 @@ module openconfig-platform-pipeline-counters {
       units bytes;
       description
         "The amount of nexthops memory used in the lookup subsystem.";
+    }
+
+    leaf nexthop-memory-utilization {
+      type oc-types:percentage;
+      description
+        "Percent utilization of overall nexthop memory.";
     }
 
     leaf acl-memory-total-entries {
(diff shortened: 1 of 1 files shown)


aredmon8551:
@dplore PTAL, this is PR559 but re-done due to merge conflict.

dplore:
Thanks for the contribution @aredmon8551.  These are for convenience since it is trivial to derive from the existing memory and usage leaves, right?  I would like to hear from the community if these are important and have solicited feedback from rwilton@cisco.com and amittal@arista.com.

sulrich:
@anuraagmittal @rolandphung

aredmon8551:
Ack, thanks @dplore - to be more blunt than my initial paragraph/readme: this is necessity, not convenience.  The units of measurement and modeling for these memories does not map to complex implementations.

dplore:
> Ack, thanks @dplore - to be more blunt than my initial paragraph/readme: this is necessity, not convenience. The units of measurement and modeling for these memories does not map to complex implementations.

acl-memory-used-bytes / acl-memory-total-bytes  = acl-memory-utilization ?

Or are you saying that underlying hardware cannot be represented as total bytes and bytes used, but that it is possible to publish an abstract utilization percentage instead?

aredmon8551:
@dplore I would recommend that, though my proposal here is to invite the least amount of change necessary to allow Juniper to support this model.

The current units of measurement do pose challenges for more complex ASIC architectures - I would recommend making this model less opinionated on what units to measure ASIC memories in and simply using relative units (percent).

aredmon8551:
@dplore - we are unfortunately blocked on implementing this state model until these changes are accepted.  Can you let me know if this proposal is acceptable?  Thanks!

dplore:
Another proposal is pending that is similar but offers additional schema.  @anuraagmittal @rolandphung

aredmon8551:
Is there a PR for that proposal that we can review?

rolandphung:
It's a WIP, I should have it ready by next week.

aredmon8551:
> It's a WIP, I should have it ready by next week.

Thank you.

robshakir:
@dplore @rsgcp @xw-g @aredmon8551 -- it seems generally that these changes that remain in the PR (i.e., additions of percentage utilisation) are orthogonal to the wider questions. Can we move ahead with this change?

dplore:
@aredmon8551 can Juniper support what is specified in #624 ?  Will review with you as we would prefer to have one way to expose integrated circuit resource utilization.

aredmon8551:
Believe we can support #624 though I do see value in updating this model to be flexible enough to support a broader variety of network ASICs/future-proof things a bit.

robshakir:
to update this thread, some discussion as to exactly what the right split of data is ongoing -- such that we're stalled a little here. @dplore is leading that work.

Issue #348, opened 2020-05-02
Use of decimal64 type in BGP models


exa-arrcus:
There are various leafs within the BGP models that make use of the `decimal64` YANG type

e.g.

```
  grouping bgp-common-neighbor-group-timers-config {
    description
      "Config parameters related to timers associated with the BGP
      peer";

    leaf connect-retry {
      type decimal64 {
        fraction-digits 2;
      }
      default 30;
      description
        "Time interval in seconds between attempts to establish a
        session with the peer.";
    }

    leaf hold-time {
      type decimal64 {
        fraction-digits 2;
      }
      default 90;
      description
        "Time interval in seconds that a BGP session will be
        considered active in the absence of keepalive or other
        messages from the peer.  The hold-time is typically
        set to 3x the keepalive-interval.";
      reference
        "RFC 4271 - A Border Gateway Protocol 4, Sec. 10";
    }

    leaf keepalive-interval {
      type decimal64 {
        fraction-digits 2;
      }
      default 30;
      description
        "Time interval in seconds between transmission of keepalive
        messages to the neighbor.  Typically set to 1/3 the
        hold-time.";
    }

    leaf minimum-advertisement-interval {
      type decimal64 {
        fraction-digits 2;
      }
      default 30;
      description
        "Minimum time which must elapse between subsequent UPDATE
        messages relating to a common set of NLRI being transmitted
        to a peer. This timer is referred to as
        MinRouteAdvertisementIntervalTimer by RFC 4721 and serves to
        reduce the number of UPDATE messages transmitted when a
        particular set of NLRI exhibit instability.";
      reference
        "RFC 4271 - A Border Gateway Protocol 4, Sec 9.2.1.1";
    }
  }
```

However this leaves an effective range of `-92233720368547758.08..92233720368547758.07`

1. Why are these decimals at all?  Are there any implementations supporting decimals for these types of values?
2. By default this allows negative numbers which do not apply in any such cases

I'm curious the reasoning behind why these are not just standard unsigned integers as every implementation is likely going to have to deviate and reel this back in to acceptable ranges at a minimum?

robshakir:
Hey Ebben,

1. This was a debate point at the time we originally put together the model. There was an assertion that some software implementations would want to be able to support sub-second granularity here - and hence this was a `decimal64`. I think that we should re-survey whether this useful going forward.

2. Yes, these should have a range that must be >0 for sure.

Cheers,
r.

exa-arrcus:
[...]

PR #604, opened 2022-03-19
Replacement of BGP decimal64 types

earies:
* (M) release/models/bgp/openconfig-bgp.yang
* (M) release/models/bgp/openconfig-bgp-common-structure.yang
* (M) release/models/bgp/openconfig-bgp-common.yang
* (M) release/models/bgp/openconfig-bgp-global.yang
* (M) release/models/bgp/openconfig-bgp-neighbor.yang
* (M) release/models/bgp/openconfig-bgp-peer-group.yang
* (M) release/models/bgp/openconfig-bgp-common-multiprotocol.yang
  - Deprecate and migrate decimal64 types for various BGP timers

### Summary

Since the initial versions of the BGP models, various timer types were declared
as 2 fraction-digit `decimal64` types however it has been noted that no known
shipping implementations support decimal/float types for these node values if
configurable whatsoever.  In addition, the `decimal64` type is a signed type
and negative values are not permitted for these timers.

This has resulted in implementations to either:

* Deviate the leaf as `not-supported`
* Deviate the leaf to change the type
* Support the leaf as-is but implement translation logic to drop the digits and
  thus precision as well as enforce backend validation at their discretion

This addresses the long-standing issue previously raised at:
https://github.com/openconfig/public/issues/348

Each leaf in this change set is either currently not supported by an
implementation or mismatched and noted.  The replaced datatypes chosen are all
of type `uint16` corresponding to the valid ranges per implementation which is
likely to correspond directly to the underlying type.  Note that since this is
a type change, this is backwards incompatible and thus a major revision is
incremented.

### `connect-retry`

References:

* Arista EOS: https://github.com/aristanetworks/yang/blob/master/EOS-4.27.2F/release/openconfig/models/bgp/arista-bgp-deviations.yang#L206-L216
* Cisco IOS-XR: https://github.com/YangModels/yang/blob/main/vendor/cisco/xr/751/cisco-xr-openconfig-network-instance-deviations.yang#L1378

### `hold-time`

References:

* Juniper JUNOS: https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/hold-time-edit-protocols-bgp.html
* Arista EOS: https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1118565
* Cisco IOS-XR: https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r3-9/routing/command/reference/rr39crs1book_chapter1.html#wp959794130

### `keepalive-interval`

References:

* Arista EOS: https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1118565
* Cisco IOS-XR: https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r3-9/routing/command/reference/rr39crs1book_chapter1.html#wp959794130

### `minimum-advertisement-interval`

References:

* Juniper JUNOS: https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/out-delay-edit-protocols-bgp.html
* Cisco IOS-XR: https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/routing/command/reference/rr40xr12kbook_chapter1.html#wp2027236550
* Arista EOS: https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1123712

### `negotiated-hold-time`

Same references as `hold-time`

### `stale-routes-time`

References:

* Juniper JUNOS: https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/stale-routes-time-edit-protocols-bgp.html
* Cisco IOS-XR: https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r3-9/routing/command/reference/rr39crs1book_chapter1.html#wp1302250876
* Arista EOS: https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1116676

### DIFF
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -24,7 +24,13 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "8.0.0";
+  oc-ext:openconfig-version "9.0.0";
+
+  revision "2022-03-19" {
+    description
+      "Transition decimal64 types to uint16 for various BGP timers";
+    reference "9.0.0";
+  }
 
   revision "2021-12-01" {
     description
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -21,7 +21,13 @@ submodule openconfig-bgp-common-structure {
     "This sub-module contains groupings that are common across multiple BGP
     contexts and provide structure around other primitive groupings.";
 
-  oc-ext:openconfig-version "8.0.0";
+  oc-ext:openconfig-version "9.0.0";
+
+  revision "2022-03-19" {
+    description
+      "Transition decimal64 types to uint16 for various BGP timers";
+    reference "9.0.0";
+  }
 
   revision "2021-12-01" {
     description
--- a/release/models/bgp/openconfig-bgp-common.yang
+++ b/release/models/bgp/openconfig-bgp-common.yang
@@ -24,7 +24,13 @@ submodule openconfig-bgp-common {
     may be application to a subset of global, peer-group or neighbor
     contexts.";
 
-  oc-ext:openconfig-version "8.0.0";
+  oc-ext:openconfig-version "9.0.0";
+
+  revision "2022-03-19" {
+    description
+      "Transition decimal64 types to uint16 for various BGP timers";
+    reference "9.0.0";
+  }
 
   revision "2021-12-01" {
     description
@@ -131,9 +137,7 @@ submodule openconfig-bgp-common {
       peer";
 
     leaf connect-retry {
-      type decimal64 {
-        fraction-digits 2;
-      }
+      type uint16;
       default 30;
       description
         "Time interval in seconds between attempts to establish a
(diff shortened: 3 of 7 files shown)


robshakir (approved):
I concur this is the right thing to do -- we'll review one more time with the ops community, but I support this change.

@s19nal @oscargdd @dplore -- PTAL.

earies:
> According to the Cisco IOS-XR link quoted, Keepalive and Hold time appear to use uint32 ? Keepalive: "Range is 0 to 4294967295." Hold-Time: "Values are 0 or a number in the range from 3 to 4294967295."
> 
> Do we want to set a blanket uint32 for all instead of a uint16?

Thx @s19nal - that does appear to be the case according to IOS-XR docs however the hold-time is actually a 2-byte uint16 field sent in the `OPEN` message (https://datatracker.ietf.org/doc/html/rfc4271#section-4.2) and the keepalive is always a fraction of this value.

Looking at IOS-XR published native models, you will see that while these are produced using `uint32` types, they are restricted to the `uint16` ranges so imo, we don't need to account for this (and the documentation appears incorrect now as a result)

https://github.com/YangModels/yang/blob/main/vendor/cisco/xr/761/Cisco-IOS-XR-um-router-bgp-cfg.yang#L4155-L4184

robshakir:
Thanks for looking at this @s19nal.

@dplore - are you OK with this change? If so, please go ahead and merge.

dplore:
Thanks @earies !

PR #605, opened 2022-03-21
Adding config leaf for BGP port

fabrizio2210:
In some rare cases, the BGP port could differ from the standard 179.  
This configuration leaf allows us to connect to a peer to a different port than 179.
To maintain a backward compatibility, I put 179 as default.

Please advise if I need to edit other files.

### DIFF
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -24,12 +24,12 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "9.0.0";
+  oc-ext:openconfig-version "8.1.0";
 
   revision "2022-03-21" {
     description
       "Add BGP port";
-    reference "9.0.0";
+    reference "8.1.0";
   }
 
   revision "2021-12-01" {
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -24,7 +24,13 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "8.0.0";
+  oc-ext:openconfig-version "9.0.0";
+
+  revision "2022-03-21" {
+    description
+      "Add BGP port"; 
+    reference "9.0.0";
+  }
 
   revision "2021-12-01" {
     description
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -28,7 +28,7 @@ submodule openconfig-bgp-common-multiprotocol {
 
   revision "2022-03-21" {
     description
-      "Add BGP port"; 
+      "Add BGP port";
     reference "9.0.0";
   }
 
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -21,7 +21,13 @@ submodule openconfig-bgp-common-structure {
     "This sub-module contains groupings that are common across multiple BGP
     contexts and provide structure around other primitive groupings.";
 
-  oc-ext:openconfig-version "8.0.0";
+  oc-ext:openconfig-version "9.0.0";
+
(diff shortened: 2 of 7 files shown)


earies:
Few quick comments:

* Definitely _rare_ in most shipping implementations.  OpenConfig models are constructed geared towards current use-cases according to shipping implementations of which this is likely not to be in many (if any) hardware based products
* This is a new leaf addition that is infact backwards compatible thus only a minor version would be incremented

My personal opinion to this is if it were to be added, then nearly every implementation would need to provide a deviation.  Do you have references to any shipping implementations that might support this capability?

fabrizio2210:
I agree this could be just minor version should be bumped up (e.g. 8.1.0), but I was misled by previous version bumps where also for minor (I think) changes a major version was used.
Is 8.1.0 correct as proposed version?

My proposal starts from a new requirement for Espresso SDN (https://blog.google/products/google-cloud/making-google-cloud-faster-more-available-and-cost-effective-extending-sdn-public-internet-espresso/). In Espresso SDN the BGP speaker are very versatile and can listen to an arbitrary port. I do not expect, in the short term, any vendor to produce a router that can listen to a different port than 179, although BGP relies on TCP and, by design, TCP can have an arbitrary port.

fabrizio2210:
Any other doubts or comments?

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -24,7 +24,13 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "8.0.0";
+  oc-ext:openconfig-version "9.0.0";
```
dplore, line 27: Hi, can you set the version number to 8.1.0 since this is merely adding a leaf and is not a breaking change?

fabrizio2210: Done!

fabrizio2210:
I don't understand why now the `pyangbind` workflow is complaining:
```
messages:{path:"/workspace/release/models/network-instance/openconfig-network-instance.yang" line:294 code:"XPATH_NODE_NOT_FOUND1" type:"warning" level:4 message:'node "openconfig-network-instance::config" is not found in "openconfig-network-instance::evpn"'}
messages:{path:"/workspace/release/models/network-instance/openconfig-network-instance.yang" line:324 code:"XPATH_NODE_NOT_FOUND1" type:"warning" level:4 message:'node "openconfig-network-instance::config" is not found in "openconfig-network-instance::config"'}
messages:{path:"/workspace/release/models/network-instance/openconfig-network-instance.yang" line:342 code:"XPATH_NODE_NOT_FOUND1" type:"warning" level:4 message:'node "openconfig-network-instance::config" is not found in "openconfig-network-instance::state"'}
messages:{path:"/workspace/release/models/interfaces/openconfig-if-ethernet.yang" line:13 code:"UNUSED_IMPORT" type:"warning" level:4 message:'imported module "openconfig-platform-types" not used'}
FATAL: pyangbind cannot build module that pyang has found errors with.
INFO: encountered (, 'XPATH_NODE_NOT_FOUND1', ('openconfig-network-instance', 'config', 'openconfig-network-instance', 'evpn'))
```

dplore:
Hi @wenovus can you take a look at the failing pyangbind lint checking for this PR?

wenovus:
@dplore Fixed at https://github.com/openconfig/public/pull/619, the issue is because the newest version of pyang (which the CI system uses automatically) added a new check that was triggered, and since it's a warning and we don't check warnings in the pyang checker, pyangbind (a more sensitive tool) complained instead.

dplore:
Rerunning checks

dplore:
@fabrizio2210 can you rebase to include Wen's fix at #619 ?  Then we can approve and merge

fabrizio2210:
Done! Now no warnings were arised.

robshakir:
Quagga allows this to be set with `--bgp_port_num` (https://www.nongnu.org/quagga/docs/quagga.html), GoBGP added an augmentation for this in their config schema (https://github.com/osrg/gobgp/blob/master/docs/sources/configuration.md).

I think it's definitely more common in BGP implementations that are decoupled from a NOS -- but given multiple OSS daemons support it, I'm OK with this being merged.

@dplore -- WDYT here?

dplore:
I am also ok with it, the use case makes sense.  Thanks for providing two references for implementations Rob!

dplore (approved):

fabrizio2210:
Can someone merge it?  
I don't think I have write permissions.

dplore:
Got it, thanks @fabrizio2210

PR #609, opened 2022-03-29
Add support of the packet length and several protocols filter

sbarguil:
Add support of the packet length and several protocols filter. 
Add ICMP field in IPv4 and IPv6 packets

### DIFF
--- a/release/models/acl/openconfig-packet-match.yang
+++ b/release/models/acl/openconfig-packet-match.yang
@@ -27,11 +27,17 @@ module openconfig-packet-match {
     field is omitted from a match expression, the effect is a
     wildcard ('any') for that field.";
 
-  oc-ext:openconfig-version "1.3.1";
+  oc-ext:openconfig-version "1.4.0";
+
+  revision "2022-03-29" {
+    description
+      "Add ICPM Fields for filtering.";
+    reference "1.4.0";
+  }
 
   revision "2021-06-16" {
     description
-      "Remove trailing whitespace";
+      "Remove trailing whitespace.";
     reference "1.3.1";
   }
 
--- a/release/models/acl/openconfig-packet-match.yang
+++ b/release/models/acl/openconfig-packet-match.yang
@@ -244,7 +244,21 @@ module openconfig-packet-match {
         leaf is specified. It is not valid to specify both 'dscp' and 'dscp-set together.'";
     }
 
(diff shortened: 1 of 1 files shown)


sbarguil:
Cames from https://github.com/openconfig/models/pull/988

release/models/acl/openconfig-packet-match.yang
```diff
@@ -432,4 +454,55 @@ module openconfig-packet-match {
...
+      description
+        "Also known as control messages.";
+      reference
+        "RFC 792: Internet Control Message Protocol
+         RFC 4443: Internet Control Message Protocol (ICMPv6)
+                   for Internet Protocol Version 6 (IPv6)
+                   Specification.";
+    }
+    leaf icmp-code {
+      type uint8;
```
robshakir, line 497: Should we have this as an `enumeration`? We have an internal model that has the definitions of the ICMPv4 and ICMPv6 types that I can propose as an addition if this makes sense.

robshakir (changes requested):
Per the usual requests -- can we add some references to >= 2 implementations that support this behaviour? I believe that they exist. I'd just like to make sure that we have this clearly documented for each PR.

release/models/acl/openconfig-packet-match.yang
```diff
@@ -27,11 +27,17 @@ module openconfig-packet-match {
     field is omitted from a match expression, the effect is a
     wildcard ('any') for that field.";
 
-  oc-ext:openconfig-version "1.3.1";
+  oc-ext:openconfig-version "1.4.0";
+
+  revision "2022-03-29" {
+    description
+      "Add ICPM Fields for filtering.";
```
sourcequench, line 34: s/ICPM/ICMP

release/models/acl/openconfig-packet-match.yang
```diff
@@ -432,4 +454,55 @@ module openconfig-packet-match {
...
+        uses ip-icmp-header-fields-config;
+      }
+    }
+  }
+
+  grouping ip-icmp-header-fields-config {
+    description
+      "Collection of ICMP header fields that can be
+       used to set up a match filter.";
+    leaf icmp-type {
```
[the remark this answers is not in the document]
robshakir: Addressed in the forked PR.

PR #610, opened 2022-03-29
Prefix qualification for AFT absolute paths

earies:
* (M) release/models/aft/openconfig-aft-common.yang
  * (M) release/models/aft/openconfig-aft-ethernet.yang
  * (M) release/models/aft/openconfig-aft-ipv4.yang
  * (M) release/models/aft/openconfig-aft-ipv6.yang
  * (M) release/models/aft/openconfig-aft-mpls.yang
  * (M) release/models/aft/openconfig-aft-pf.yang
  * (M) release/models/aft/openconfig-aft.yang
    - Add prefixes to next-hop-group leafref
    - Use common typedef for next-hop-group-network-instance

Similar issues to:
* https://github.com/openconfig/public/pull/598
* https://github.com/openconfig/public/pull/557

Absolute paths to alternate schema locations must be qualified

PR #611, opened 2022-03-29
Prefix qualification for AFT network-instance absolute paths

earies:
* (M) release/models/aft/openconfig-aft-common.yang
* (M) release/models/aft/openconfig-aft-ethernet.yang
* (M) release/models/aft/openconfig-aft-ipv4.yang
* (M) release/models/aft/openconfig-aft-ipv6.yang
* (M) release/models/aft/openconfig-aft-mpls.yang
* (M) release/models/aft/openconfig-aft-pf.yang
* (M) release/models/aft/openconfig-aft.yang
* (M) release/models/aft/openconfig-aft-network-instance.yang
  - Relocate network-instance specific next-hop-group leafs to the AFT
    network-instance model
  - Fix prefixing for network-instance specific leafrefs

Adjustment to previous https://github.com/openconfig/public/pull/610

And addressing similar issues as found in:
* https://github.com/openconfig/public/pull/598
* https://github.com/openconfig/public/pull/557

This involves relocating next-hop-group leafs that have absolute paths to
network-instances to the AFT network-instance model to prevent circular
dependencies and segment network-instance specifics to its own module.

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -22,7 +22,15 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "0.9.0";
+  oc-ext:openconfig-version "1.0.0";
+
+  revision "2022-03-29" {
+    description
+      "Relocate next-hop-group/next-hop-group-network-instance
+      from openconfig-aft-common to resolve absolute path
+      leafref specific to network-instances";
+    reference "1.0.0";
+  }
 
   revision "2021-12-09" {
     description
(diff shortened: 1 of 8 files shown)


release/models/aft/openconfig-aft-common.yang
```diff
@@ -22,7 +22,15 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "0.9.0";
+  oc-ext:openconfig-version "1.0.0";
```
robshakir, line 25: Do we want to cut `1.0.0` here -- it seems reasonable, but maybe needs more consideration -- or just do `0.10.0`?

robshakir (dismissed):
Thanks for this Ebben, LGTM in general -- WDYT on the 1.0.0 readiness?

release/models/aft/openconfig-aft-common.yang
earies: Since `1.0.0` has since been published (and a few other PRs are in flight touching this domain) - this PR is now updated to reflect `2.0.0`.  This is a backwards incompatible change thus the major rev should be incremented here

dplore:
@earies can you provide an example of the circular dependency that would happen if we left these leaves in aft-common?

earies:
> @earies can you provide an example of the circular dependency that would happen if we left these leaves in aft-common?

Sure - if the `next-hop-group` and `next-hop-group-network-instance` leafrefs are merely updated in place to add the prefix to become fully-qualfied then `openconfig-network-instance` needs to be imported resulting in a circular dependency since `openconfig-network-instance` imports the AFT domain

```
../aft/openconfig-aft-common.yang:15: error: circular dependency for module "openconfig-network-instance"
```

Thus, the reason for relocation to a module that is not part of the import set

earies:
@dplore @robshakir - let me know if this looks good or if anything else is needed before merge

akalluru1:
@earies @dplore @robshakir It seems like this change inadvertently removed following attributes:
```/network-instances/network-instance[name=*]/afts/ipv4-unicast/ipv4-entry[prefix=*]/state/next-hop-group
/network-instances/network-instance[name=*]/afts/ipv4-unicast/ipv4-entry[prefix=*]/state/next-hop-group-network-instance
/network-instances/network-instance[name=*]/afts/ipv6-unicast/ipv6-entry[prefix=*]/state/next-hop-group
/network-instances/network-instance[name=*]/afts/ipv6-unicast/ipv6-entry[prefix=*]/state/next-hop-group-network-instance
/network-instances/network-instance[name=*]/afts/mpls/label-entry[label=*]/state/next-hop-group
/network-instances/network-instance[name=*]/afts/mpls/label-entry[label=*]/state/next-hop-group-network-instance```
can someone please look into it?

Issue #606, opened 2022-03-22
Update oc-macsec model to use oc-keychain model


m26singhvi:
Currently, openconfig macsec model -  public/release/models/macsec/openconfig-macsec.yang has a keychain container for configuring keys. 

Recently, a keychain model was published here - https://github.com/openconfig/public/blob/master/release/models/keychain/openconfig-keychain.yang 

which allows configuration of keys as well - most of it is pretty similar to what is present in Macsec model already. 

It will be good if the macsec model can remove out the key-chain container and refer to the key-chain model defined globally.

dplore:
Thanks for the comment!  To help speed this along, I encourage you to submit a pull request to make the changes you would like to see.  Thank you!

m26singhvi:
Thanks for taking a look at this. I will soon submit a pull request for this.

PR #614, opened 2022-04-07
 Update oc-macsec model to use oc-keychain model #606

m26singhvi:
Remove key-chain container from macsec openconfig model completely
 and use global key-chain model instead.

### DIFF
--- a/release/models/macsec/openconfig-macsec.yang
+++ b/release/models/macsec/openconfig-macsec.yang
@@ -7,6 +7,7 @@ module openconfig-macsec {
   import openconfig-interfaces { prefix oc-if; }
   import openconfig-macsec-types { prefix macsec-types; }
   import openconfig-yang-types { prefix oc-yang; }
+  import openconfig-keychain { prefix "oc-keychain"; }
 
   organization
     "OpenConfig working group";
@@ -17,10 +18,16 @@ module openconfig-macsec {
     "This module defines configuration and state data for
      MACsec IEEE Std 802.1AE-2018.";
 
-  oc-ext:openconfig-version "0.2.0";
+  oc-ext:openconfig-version "1.0.0";
   oc-ext:regexp-posix;
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
+
+  revision "2022-04-28" {
+    description
+      "Use global key chain model.";
+    reference "1.0.0";
+  }
 
   revision "2020-05-01" {
     description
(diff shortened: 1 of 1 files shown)


dplore (changes requested):
Hi, please rebase/merge with latest master to pick up a change which will fix the pyangbind check.

release/models/macsec/openconfig-macsec.yang
```diff
@@ -17,11 +18,17 @@ module openconfig-macsec {
     "This module defines configuration and state data for
      MACsec IEEE Std 802.1AE-2018.";
 
-  oc-ext:openconfig-version "0.2.0";
+  oc-ext:openconfig-version "0.3.0";
```
dplore, line 21: Need to bump major revision number since this is a breaking change,  ie:  (1.0.0 instead of 0.3.0)

dplore (changes requested):
Need to do version bump, otherwise, LGTM

release/models/macsec/openconfig-macsec.yang
m26singhvi: Bumped up the version to 1.0.0 at couple of places and updated the revision date as well.

m26singhvi:
> Need to do version bump, otherwise, LGTM

Bumped up the version to 1.0.0 at couple of places and updated the revision date as well.

dplore:
Hi,  the miscellaneous checks CI check shows the following submodules still need updating:

⛔  submodule versions must match the belonging module's version
module set openconfig-rib-bgp is at 0.8.0 (openconfig-rib-bgp-tables.yang), non-matching files: openconfig-rib-bgp-attributes.yang (0.7.0), openconfig-rib-bgp-shared-attributes.yang (0.7.0), openconfig-rib-bgp-table-attributes.yang (0.7.0)

m26singhvi:
> 

Yes, I noticed that error but I don't think it is related to my changes. I started seeing that error after I did a rebase once. In any case, I will check more about the error.

m26singhvi:
> Hi, the miscellaneous checks CI check shows the following submodules still need updating:
> 
> ⛔ submodule versions must match the belonging module's version module set openconfig-rib-bgp is at 0.8.0 (openconfig-rib-bgp-tables.yang), non-matching files: openconfig-rib-bgp-attributes.yang (0.7.0), openconfig-rib-bgp-shared-attributes.yang (0.7.0), openconfig-rib-bgp-table-attributes.yang (0.7.0)

This has got fixed in the upstream, so I went ahead and did a rebase.

dplore (approved):
LGTM, thank you for your contribution!

m26singhvi:
> LGTM, thank you for your contribution!

Thanks for the review !

m26singhvi:
Do I need to raise a merge request for this somewhere ?

dplore:
Thank you for your contribution @m26singhvi !

PR #615, opened 2022-04-14
Add mike-albano as OWNER for /release/models/wifi

dplore:
Although this is redundant to the current codeowners, this explicitly assigns Mike as a codeowner for wifi models.  This is a  new format the community would like to use with the goal of increasing the velocity of reviews by delegating reviewer and write privileges to members in the community.


aashaikh (approved):
LGTM - thanks!

PR #616, opened 2022-04-15
Add counter collection into pipeline counters

rolandphung:
(M) release/models/platform/openconfig-pipeline-counters.yang
- Add list of counters keyed by name for each block of packet and drop containers

This proposal is to add a collection of counters into the pipeline-counter module for each IC subsystem block.

The motivation of this proposal is to provide flexibility in publishing platform specific counters that do not fit into existing counters defined within `pipeline-counters`. Every vendor may have multiple platforms with different and unique implementations. Providing a place to put counters that do not fit currently defined leafs will allow for more information and better real time analysis of the devices.

Example of one block:
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:packet
                 +--ro oc-ppc:interface-block
                    +--ro oc-ppc:counters
                       +--ro oc-ppc:counter* [name]
                          +--ro oc-ppc:name     -> ../state/name
                          +--ro oc-ppc:state
                             +--ro oc-ppc:name?    string
                             +--ro oc-ppc:count?   oc-yang:counter64
```

### DIFF
--- a/release/models/platform/openconfig-platform-pipeline-counters.yang
+++ b/release/models/platform/openconfig-platform-pipeline-counters.yang
@@ -65,14 +65,20 @@ module openconfig-platform-pipeline-counters {
     5 blocks, is to have the abililty to receive all drop counters from
     all 5 blocks, for example, with one request.";
 
-  oc-ext:openconfig-version "0.2.1";
+  oc-ext:openconfig-version "0.3.0";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
+
+  revision "2022-04-13" {
+    description
+      "Add generic pipeline counters per block";
+    reference "0.3.0";
+  }
 
   revision "2022-01-19" {
     description
       "Fixed typo for aggregate field.";
-    reference "0.2.1";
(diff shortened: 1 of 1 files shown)


robshakir:
Roland, thanks for this -- is this intended to coexist with #624?

rolandphung:
Yes. Counters and utilization should work together, but the review/changes should be independent of each other.

dplore:
Speaking for the open config operators group: The open config operators have reviewed this model proposal and believes it is not suitable for addition to openconfig.  Initially there was concern over the use of arbitrary strings to publish a list of integrated circuit (IC) counters.   The use of strongly strongly typed data was then debated.  Upon further review it was acknowledged that integrated circuits are not industry standard designs and that even a strongly typed model which exposes features that are unique to the vendor of the IC could not likely follow the charter of openconfig being vendor neutral.   

Speaking only my own behalf:  It's my recommendation to instead define a non-openconfig model (vendor model) which exposes the integrated circuit counters.

PR #617, opened 2022-04-15
Updates to route preference/metrics

earies:
    - Change internal/external route preference to unrestricted uint32
  * (M) release/models/local-routing/openconfig-local-routing.yang
    - Remove of top-level /local-routes as it has been re-rooted under
      network-instance protocols for a number of major releases
    - Description update to static route metric
    - Addition of static route preference (administrative distance)
    - Addition of aggregate route preference/metric

### Change IS-IS internal/external route preference datatype

The current implementation of a restricted uint8 type restricts some
implementations disallowing for the value '0' as well as values > 255.  This
expands the data-type to support full values up to the limit of the uint32
type.  If an implementation is limited to lesser values, then deviations can be
produced.

NOTE: Since data-types and permitted ranges differ across various
implementations, the suggestion is to loosen the datatype vs. implement
restrictions that do not accomodate all implementations.

References:

* Juniper JUNOS: https://www.juniper.net/documentation/en_US/junos/topics/reference/general/routing-protocols-default-route-preference-values.html
* Cisco IOS-XE: https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/iproute_isis/configuration/xe-3s/irs-xe-3s-book.pdf

### Local Routing refactoring

First is the removal/deprecation of `/local-routing` in favor of `/network-instances` based configuration/state.

In addition is the ability to configure preference (admin distance) and metrics for both static and aggregate routes.

References:

* Juniper JUNOS: https://www.juniper.net/documentation/us/en/software/junos/static-routing/topics/ref/statement/aggregate-edit-routing-options.html
                 https://www.juniper.net/documentation/us/en/software/junos/static-routing/topics/ref/statement/metric-edit-routing-options.html
* Arista EOS: https://www.arista.com/en/um-eos/eos-ipv4#xx1144639
* Cisco IOS-XR: https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r3-9/routing/configuration/guide/xr12krc39_chapter8.html#task_1011694

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -10,6 +10,7 @@ submodule openconfig-aft-common {
   import openconfig-mpls-types { prefix "oc-mplst"; }
   import openconfig-policy-types { prefix "oc-pol-types"; }
   import openconfig-aft-types { prefix "oc-aftt"; }
+  import openconfig-evpn-types { prefix "oc-evpn-types"; }
 
   organization
     "OpenConfig working group";
@@ -22,11 +23,17 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "0.10.0";
+  oc-ext:openconfig-version "0.11.0";
 
-  revision "2022-04-15" {
+  revision "2022-05-10" {
     description
       "Add preference/metric to IPv4/IPv6 AFT entries.";
+    reference "0.11.0";
+  }
+
+  revision "2022-01-26" {
+    description
+      "Add vni-label and tunnel-src-ip-address properties under next-hops";
     reference "0.10.0";
   }
 
@@ -148,6 +155,7 @@ submodule openconfig-aft-common {
 
           uses aft-common-entry-nexthop-state;
           uses aft-labeled-entry-state;
+          uses aft-evpn-entry-state;
         }
 
         container ip-in-ip {
@@ -233,6 +241,27 @@ submodule openconfig-aft-common {
         forwarding entry. This leaf is applicable only to next-hops
         which include MPLS label information, and its value typically
         corresponds to the RSVP-TE LSP name.";
+    }
+  }
+
+  grouping aft-evpn-entry-state {
+    description
+      "Operational state for evpn related information in forwarding entry";
+
+    leaf vni-label {
+      type oc-evpn-types:evi-id;
+      description
+        "Where applicable, the next hop label representing the virtual
+        network identifier (VNI) for the forwarding entry. This leaf is
+        applicable only to next-hops which include VXLAN encapsulation
+        header information";
+    }
+
+    leaf tunnel-src-ip-address {
+      type oc-inet:ip-address;
+      description
+        "Where applicable this represents the tunnel source ip address.
+        For VXLAN this represents the source VTEP ip address";
     }
   }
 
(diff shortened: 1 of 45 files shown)


release/models/aft/openconfig-aft-common.yang
```diff
@@ -342,6 +348,18 @@ submodule openconfig-aft-common {
         "The protocol from which the AFT entry was learned.";
     }
 
+    leaf preference {
```
robshakir, line 351: Today, the AFT model exposes only the selected routes -- and must because there's no ability to have >1 entry with the same key. I'd just like the clarify that the intention is to return just the admin distance/preference of the AFT entry that is active, we're not trying to change the original semantics here.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -342,6 +348,18 @@ submodule openconfig-aft-common {
...
+      type uint32;
+      description
+        "The administrative distance (preference) of the AFT entry.";
+    }
+
+    leaf metric {
+      type uint32;
+      description
+        "The metric (or cost) of the AFT entry.";
```
robshakir, line 360: This and the preference seem a little questionable to me -- what's the use case for having this here vs. having it in the LSDB of the protocol that would populate it?

robshakir (commented):
Thanks for the changes Ebben.

`local-routing` changes LGTM, that seems fairly clear. I'm not sure I quite understand the motivation for the `metric` and `preference` leaves in the AFT model. I think there's also some consideration here that this could be introduced into gRIBI (if we allow these fields) where I suspect things might become significantly more complex (we deliberately designed gRIBI to have one admin distance/preference for the whole protocol thus far).

release/models/aft/openconfig-aft-common.yang
earies: That's correct - one part is the returned state but the other intention was since this model is imported into the stub model that generates the proto IDL for gRIBI programming that it could be utilized to "set" the preference and/or metric without modifying the gRIBI APIs directly.

If however as you mention in the other comment that gRIBI is to have a single preference/metric combo that is protocol wide and not per entry, then this would solely then be used for state representation however the driver behind this was for gRIBI.

If gRIBI is going to have a single preference/distance then we will want to discuss how this is driven (likely `/system/` schema config) and preferred order since various implementations can differ in preference/admin-distance treatment (e.g. after static/connected but prior to protocols, etc..)

release/models/aft/openconfig-aft-common.yang
earies: As mentioned in the other comment, primary driver is gRIBI programming/flexibility but that aside, much like `show route` and `show route extensive|detail` across implementations - it is common to display entries (all or preferred) with their respective preference/metric

release/models/aft/openconfig-aft-common.yang
earies: @robshakir - any additional thoughts or concerns on this?

earies: Per discussion - removed AFT related parameters.  Will bring in gRIBI preference related via separate PR.  Updated in latest commit

release/models/aft/openconfig-aft-common.yang
earies: Removed AFT related preference/metric in latest commit and adjusted PR header

dplore:
The net changes are:
- Add configure preference (admin distance) for aggregate routes.
- Expand preference value from uint8 to uint32

dplore (approved):
Thanks for your contribution!

PR #618, opened 2022-04-19
Modified leaf to leaf-list for asn in set-as-path-prepend.

shimangkumar:
as-path-prepend have asn as of type leaf is current model. Since it is of type leaf, it is restricting to have only one kind of asn present in yang. Modified the asn from type leaf to leaf-list so that mutiple distinct ASN can be added in as-path-prepend.

### DIFF
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -775,7 +775,7 @@ module openconfig-bgp-policy {
       "Top-level grouping for the AS path prepend action";
 
     container set-as-path-prepend {
-      when "../set-as-path-prepend-type/config/asn-prepend-type = SINGLE";
+      when "../set-as-path-prepend-type/config/asn-prepend-type='SINGLE'";
       description
         "Action to prepend the specified AS number to the AS-path a
         specified number of times";
@@ -828,7 +828,7 @@ module openconfig-bgp-policy {
       "Top-level grouping for the AS path prepend list action";
 
     container set-as-path-prepend-list {
-      when "../set-as-path-prepend-type/config/asn-prepend-type = MULTIPLE";
+      when "../set-as-path-prepend-type/config/asn-prepend-type='MULTIPLE'";
       description
         "Action to prepend the specified AS numbers to the AS path";
 
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -813,7 +813,7 @@ module openconfig-bgp-policy {
     leaf asn {
       type oc-inet:as-number;
       description
-       "AS number added in the list for AS path prepend."
+       "AS number added in the list for AS path prepend.";
     }
 
   }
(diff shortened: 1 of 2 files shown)


release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -671,17 +678,17 @@ module openconfig-bgp-policy {
...
-        should be between 1 and the maximum supported by the
-        implementation.";
+        leaf-list to the AS path. If no value is specified by the
+        asn leaf-list, the local AS number of the system is used.
+        The value should be between 1 and the maximum supported by
+        the implementation.";
     }
 
-    leaf asn {
+    leaf-list asn {
```
robshakir, line 687: I don't think that this model conforms to implementations that are out there today - since you're allowing for the case that if the `leaf-list` is specified to be `[15169 4445]` then `repeat-n` is `2` then you append `15169 4445 15169 4445`. 

Can you please provide documentation as to implementations that support this today? If it's just that you would like to support prepending more than just one ASN - consider having a new `asns` leaf-list that does not break backwards compatibility, and does not support the `repeat-n` functionality.

robshakir:
Thanks for the contribution -- please find some comments inline.

release/models/bgp/openconfig-bgp-policy.yang
shimangkumar: @robshakir  Yes we do support multiple asn with repeat-n functionality. As a result if ` 15169 4445 ` is specified with **repeat** of  2 append will be ` 15169 4445 15169 4445.`

Please find the attached document :
`multi-asn prepend support `
https://www.arista.com/en/support/toi/eos-4-24-1f/14501-as-path-statement-enhancement
`usage of repeat with multi-asn prepend support`
https://www.arista.com/en/support/toi/eos-4-24-2f/14548-route-map-as-path-repeat-option

Since multi-asn support is there, thought to modify the leaf to leaf-list.

robshakir: Could you please check that this is commonly supported in another implementation too? I'm OK with this change if it is, otherwise I think we need to discuss it further. Thanks.

shimangkumar: I proposed the new change to supoort multiple AS number values in AS path prepend.
There will be 2 new container added,
1. Contianer to select AS path prepend type, this will have a leaf node  asn-prepend-type of type enum SINGLE and MULTIPLE.
2. If leaf node is SINGLE set-as-path-prepend container is selected which will have two leaves ASN and repeat-n. To repeat the same AS number specified number of times in AS path prepend.
2. If leaf node is MULTIPLE, a newly added container set-as-path-prepend-list will be selected. Newly added container set-as-path-prepend-list is having a list is selected
3. To make the containert conditional between set-as-path-prepend and set-as-path-prepend-list when statement is used.
4. set-as-path-prepend-list container will have a list in it, list will have two leaves index and asn. index will be the key of list.

shimangkumar: Please take another look, i modified the same instead of leaf-list.

PR #620, opened 2022-04-20
Add dropped-octets to openconfig-qos-interfaces.  Removed vlan part of request.

cfyo:
Request addition of dropped-octets to path: /qos/interfaces/interface/output/queues/queue/state to replace use of vendor proprietary SNMPs.

References:
Juniper: jnxCosQstatTotalRedDropBytes 
       https://www.juniper.net/documentation/en_US/northstar4.3.0/topics/concept/northstar-data-collection-via-snmp-4.2.html
Cisco: cbQosCMDropByte64
     https://www.cisco.com/c/en/us/support/docs/ip/simple-network-management-protocol-snmp/119031-technote-router-00.html
     https://iphostmonitor.com/mib/oids/CISCO-CLASS-BASED-QOS-MIB/cbQosCMDropByte64.html


release/models/qos/openconfig-qos-interfaces.yang
```diff
@@ -321,6 +321,12 @@ submodule openconfig-qos-interfaces {
       description
         "Number of packets dropped by the queue due to overrun";
     }
+ 
+    leaf dropped-octets {
```
robshakir, line 325: I think we initially excluded the octet counter here as it wasn't clear that implementations count the size of packets that they dropped, that decision was a long time ago though, are there implementations with this today? (I'm happy to include it.)

release/models/vlan/openconfig-vlan.yang
```diff
@@ -117,7 +117,28 @@ module openconfig-vlan {
   grouping vlan-state {
```
robshakir, line 117: Given that these leaves are in the `config` container above, it seems like we shouldn't need to redefine them? I see them in both containers today:

```
        +--rw vlans
        |  +--rw vlan* [vlan-id]
        |     +--rw vlan-id    -> ../config/vlan-id
        |     +--rw config
        |     |  +--rw vlan-id?   oc-vlan-types:vlan-id
        |     |  +--rw name?      string
        |     |  +--rw status?    enumeration
        |     +--ro state
        |     |  +--ro vlan-id?   oc-vlan-types:vlan-id
        |     |  +--ro name?      string
        |     |  +--ro status?    enumeration
```

Was there some other reason to redefine them?

release/models/vlan/openconfig-vlan.yang
```diff
@@ -140,7 +161,15 @@ module openconfig-vlan {
     description
       "TPID opstate for dot1q-enabled interfaces";
 
-    // placeholder
+    leaf tpid {
```
robshakir, line 164: As above, I think this is already defined based on it being within the `config` container.

robshakir (changes requested):

release/models/qos/openconfig-qos-interfaces.yang
cfyo: Yes, I'd like to include it.

release/models/vlan/openconfig-vlan.yang
cfyo: Ok

release/models/vlan/openconfig-vlan.yang
cfyo: OK

release/models/qos/openconfig-qos-interfaces.yang
robshakir: This sounds good to me. Can you link to these MIBs in the description of the PR please? (To demonstrate support for them across implementations.)

cfyo: @robshakir  done, please review.

dplore (approved):
LGTM, thank you for the contribution

dplore:
I will merge this request in two weeks on Sep 27, pending any objections from the community.

PR #623, opened 2022-04-21
Add power-admin-state for fabrics and supervisors

itbwtw:
The current model supports power management for linecards through path `/components/component/linecard/config/power-admin-state`, but not for fabrics and supervisors. This PR supports them by adding the following path:
```
(type=FABRIC) /components/component[name=<fabric_module>]/fabric/config/power-admin-state
(type=CONTROLLER_CARD) /components/component[name=<supervisor_module>]/controller-card/config/power-admin-state
```
Added YANG paths:
```
  +--rw components
     +--rw component* [name]
        +--rw fabric
        |  +--rw config
        |  |  +--rw oc-fabric:power-admin-state?   oc-platform-types:component-power-type
        |  +--ro state
        |     +--ro oc-fabric:power-admin-state?   oc-platform-types:component-power-type
        +--rw controller-card
           +--rw config
           |  +--rw oc-ctrl-card:power-admin-state?   oc-platform-types:component-power-type
           +--ro state
              +--ro oc-ctrl-card:power-admin-state?   oc-platform-types:component-power-type
```

Documents showing that two vendors have implemented the component power management feature:

1. Cisco: https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst6500/ios/15-4SY/config_guide/sup6T/15_3_sy_swcg_6T/power.html

2. Juniper: https://www.juniper.net/documentation/us/en/software/junos/chassis/topics/ref/command/request-chassis-fpc.html

3. Arista: https://www.arista.com/en/um-eos/eos-switch-administration-commands#prompt         See section "power enable module" directly above


release/models/platform/openconfig-platform-fabric.yang
```diff
@@ -0,0 +1,88 @@
...
+  oc-ext:openconfig-version "0.1.0";
+
+  revision "2022-04-20" {
+    description
+      "Initial revision";
+    reference "0.1.0";
+  }
+
+  // OpenConfig specific extensions for module metadata.
+  oc-ext:regexp-posix;
```
robshakir, line 35: I believe that this can be removed now -- we've made all the patterns have their own extension for POSIX patterns, and this module has none within it.

release/models/platform/openconfig-platform-fabric.yang
```diff
@@ -0,0 +1,88 @@
...
+
+  // typedef statements
+
+  // grouping statements
+
+  grouping fabric-config {
+    description
+      "Configuration data for fabric components";
+
+    leaf power-admin-state {
```
robshakir, line 53: Should we make this a common definition somewhere? It could probably be a `typedef` in `openconfig-platform-types`.

robshakir (commented):
Thanks - LGTM, but a couple of proposed clarifications.

release/models/platform/openconfig-platform-fabric.yang
itbwtw: Thanks for the catch. I'll remove this from both files - openconfig-platform-fabric.yang and openconfig-platform-controller-card.yang

release/models/platform/openconfig-platform-fabric.yang
itbwtw: Thanks for the catch. I'll remove this from both files -- openconfig-platform-fabric.yang and openconfig-platform-controller-card.yang

release/models/platform/openconfig-platform-fabric.yang
itbwtw: I'm not sure if I understand this correctly -- did you mean `grouping` rather than `typedef`? I  can define the `grouping` in `openconfig-platform-types` so that we can code re-use it in all three modules: linecard, fabric and controller-card

cfyo: If this is going to be added shouldn't there be other components that it should cover to be consistent such as for transceiver and power-supply that both have an "enabled" for power on/off and potentially others.  Would it make sense if we want to add this to add it under the /components/component/state path?

release/models/platform/openconfig-platform-linecard.yang
```diff
@@ -23,7 +23,14 @@ module openconfig-platform-linecard {
     "This module defines data related to LINECARD components in
     the openconfig-platform model";
 
-  oc-ext:openconfig-version "0.1.2";
+  oc-ext:openconfig-version "0.1.3";
```
dplore, line 26: due to deletion of leaf (breaking change), this is now a new, major release, so version should become "1.0.0"

release/models/platform/openconfig-platform-types.yang
```diff
@@ -231,6 +237,18 @@ module openconfig-platform-types {
     }
   }
 
+  grouping component-power-management {
+    description
+      "Common grouping for managing compoent's power";
```
dplore, line 242:
```suggestion
      "Common grouping for managing component power";
```

dplore (changes requested):
Couple of minor comments.  Per [openconfig contribution guide](https://github.com/openconfig/public/blob/master/doc/external-contributions-guide.md)  p lease also reference two vendor implementations for enable/disable of component power.

dplore:
For reviewing convenience: 
```module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw fabric
        |  +--rw config
        |  |  +--rw oc-fabric:power-admin-state?   oc-platform-types:component-power-type
        |  +--ro state
        |     +--ro oc-fabric:power-admin-state?   oc-platform-types:component-power-type
        +--rw controller-card
           +--rw config
           |  +--rw oc-ctrl-card:power-admin-state?   oc-platform-types:component-power-type
           +--ro state
              +--ro oc-ctrl-card:power-admin-state?   oc-platform-types:component-power-type
```

release/models/platform/openconfig-platform-fabric.yang
dplore: @cfyo I think this is not a bad idea, but is perhaps some minor scope creep for this PR.    The other defined OC component types I think this might make sense for are fan and power supply?   For power supply, it might be more appropriate to support an admin-config leaf for 'enabled/disabled'.   Transceiver also has enabled/disabled which probably controls its electrical power?   May I suggest we address that in a separate PR so we can keep this one moving forward?

release/models/platform/openconfig-platform-types.yang
```diff
@@ -239,7 +239,7 @@ module openconfig-platform-types {
 
   grouping component-power-management {
     description
-      "Common grouping for managing compoent's power";
+      "Common grouping for managing compoent power";
```
dplore, line 242:
```suggestion
      "Common grouping for managing component power";
```

dplore:
@itbwtw another PR was merged in ahead and incremented the version, so you will also need to bump up the version here as well.  Thanks!

dplore:
Thanks @itbwtw .  So the last step is to reference two vendor implementations of this feature.  You can paste in URL's to documentation simply showing that vendors have a feature to power off linecards which aligns with the intent of this openconfig.  This should be done for features like this that are typically well established capabilities, per  [openconfig contribution guide](https://github.com/openconfig/public/blob/master/doc/external-contributions-guide.md)

itbwtw:
Sorry for taking so long finding the required documents. Here are the documents showing that two vendors have implemented the component power management feature:

1. Cisco: https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst6500/ios/15-4SY/config_guide/sup6T/15_3_sy_swcg_6T/power.html

2. Juniper: https://www.juniper.net/documentation/us/en/software/junos/chassis/topics/ref/command/request-chassis-fpc.html

rolandphung:
@dplore PTAL

brianneville:
The proposed CLI command `[no] power enable module Supervisor<num>` will only work for powering-down the standby supervisor.

Attempting to use this command to power down the active supervisor will result in an error and instruct you to use the `reload` command to reload the supervisor instead:
```
(s2)(config)#no power enable module Supervisor2
% Powering off the current supervisor is not supported. Use 'reload power' instead.
```
The `reload power` command however will reload the module, rather than just powering it down.

Should this power-admin-status leaf have any purpose on the active supervisor, or should we only support it on the standby supervisor?      
 I don't think it would make sense to support it on the active supervisor, but I'm interested to see if anyone else has thoughts on this. 

Even if this leaf was to be supported on the active supervisor but make use of the `reload` command, there would still be the issue of the race condition between the GNMI server (running on the active supervisor) and the reload command. That is, a GNMI client who is subscribed may not receive updates on this leaf anyway, as the reload may occur before the GNMI server gets to send out an update.

dplore:
Yes, to operationally request a switchover from primary to secondary (gNOI) is not in scope for this PR.

If we want a controller card to be powered off it seems reasonable to reject it if the controller card is primary.  The operator would be required to perform a switch over action first.  [This description](https://github.com/openconfig/public/pull/623/files#diff-f787893a2471f7ec1785fa010e4981f339dc68a144cffb802df9afcdf02a30b5R1061)  could be updated to reflect it.  WDYT @itbwtw

dplore:
2 week last call for comments on this change, to be merged on Aug17.  PTAL @earies @rgwilton @hellt

rolandphung:
@dplore is this ready to merge?

dplore (approved):

PR #624, opened 2022-04-21
Add resouce utilization to components

rolandphung:
* (A) release/models/platform/openconfig-platform-common.yang
  - Add platform-utilization models
  - Add platform-utilization into chassis and integrated circuit
* (M) release/models/platform/openconfig-platform-linecard.yang
  - Add platform-utilization into linecard

This proposal is trying to model the utilization of hardware without adding further specific leaves within pipeline-counters.

The goals of this proposal is the same as #602, defining a vendor and architecture agnostic way to represent memory usage or hardware table entries. Unlike #602, this avoids defining specific a specific leaf within the pipeline counter which may not be an accurate representative of the current usage especially when each vendor has different complex implementation and approaches that usually cannot be represented by just one leaf.

On a separate note, the utilization container could be a better place to put the thresholds (#506) configuration and alarms. `utilization` already associates with a specific component and `resource` can break down further into separate features/tables.

`pyang` output of `/components/component/integrated-circuit/utilization`:
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
           +--rw utilization
              +--rw resources
                 +--rw resource* [name]
                    +--rw config
                    |  +--rw name?   string
                    +--rw state
                       +--rw name?                  string
                       +--rw used?                  uint64
                       +--rw used-percent?          oc-types:percentage
                       +--rw committed?             uint64
                       +--rw free?                  uint64
                       +--rw max-limit?             uint64
                       +--rw high-watermark?        uint64
                       +--rw last-high-watermark?   oc-types:timeticks64
```

### DIFF
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -1,0 +1,152 @@
+submodule openconfig-platform-common {
+
+  yang-version "1";
+
+  belongs-to openconfig-platform {
+    prefix "oc-platform";
+  }
+
+  import openconfig-extensions { prefix oc-ext; }
+  import openconfig-types { prefix oc-types; }
+
+  organization "OpenConfig working group";
+
+  contact
+    "OpenConfig working group
+    www.openconfig.net";
+
+  description
+    "This modules contains common groupings that are used in multiple
+    components within the platform module.";
+
+  oc-ext:openconfig-version "0.1.0";
+
+  revision "2022-04-15" {
+    description
+      "Initial revision with platform utilization.";
+    reference "0.1.0";
+  }
+
+  // extension statements
+
+  // feature statements
+
+  // identity statements
+
+  // typedef statements
+
+  // grouping statements
+
+  grouping platform-utilization-top {
+    description
+      "Top level of utilization.";
+
+    container utilization {
+      description
+        "Utilization of the component.";
+
+      container resources {
+        description
+          "Enclosing container for the resources in this component.";
+
+        list resource {
+          key "name";
+          description
+            "List of resources, keyed by resource name.";
+
+          leaf name {
+            type leafref {
+              path "../config/name";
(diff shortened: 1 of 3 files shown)


aredmon8551:
LGTM, thanks Roland.

robshakir:
This seems a reasonable proposal -- one question to clarify, do you see this replacing the pipeline counters model, or is it expected that this gives more granular breakdowns of the utilisation?

The original motivation for the pipeline counters model was absolutely to abstract away some of the details of the underlying implementation to make the telemetry actionable. One concern that I have with this model is that I don't actually have any clue as the consumer of the telemetry what the name is or means. This ends up meaning that whilst I can create a nice graph from it, I can't actually trigger any repair action without a map that tells me what the name means and the severity of it. I think that I'm generally supportive for having these implementation specific counters exposed but only if we also have a means to expose more generic counters that allow us to determine from an operational perspective some particular repair action.

rolandphung:
This should not replace the pipeline counters model. We should be able to work with both broadly defined leaves to discern something has gone wrong, while we have these specific utilization telemetry to tell what exactly has gone wrong. However, the underlying hardware resources are usually dynamically reserved/split/shared, making it very hard to determine what value is representative of the usage with only a single leaf.

robshakir:
Great, this SGTM :-) I will review the proposal.

dplore:
After some discussion is seems percentage == used / used + free and is desirable to remove from this PR.  

This could be revisited in a future PR regarding alarms, for example in #506 if desired.

dplore (changes requested):
After some discussion is seems percentage == used / used + free and is desirable to remove from this PR.

dplore:
Thank you @rolandphung ,  this is queued for review at May 3rd openconfig operators meeting

dplore (approved):

PR #625, opened 2022-04-23
Add p4RT id to component/chassis/

dplore:
(M) release/models/p4rt/openconfig-p4rt.yang

Adds a p4rt id leaf for the chassis component.  This is intended to be used as a controller defined  identifier which can be related to  interface/oc-p4rt:id and integrated-circuit/oc-p4rt:node-id

```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw chassis
        |  +--rw config
        |  |  +--rw oc-p4rt:id?   uint32
        |  +--ro state
        |  |  +--ro oc-p4rt:id?   uint32
```


release/models/platform/openconfig-platform-chassis.yang
```diff
@@ -0,0 +1,79 @@
...
+
+  grouping chassis-config {
+    description
+      "Config that corresponds to the global hardware package";
+
+    leaf chassis-id {
+      type uint32;
+      description
+        "A unique identifier representing the global device.  This is
+        intended to be used as a root of a tree of component ids";
```
earies, line 53: Could you elaborate a bit more on the intention here?  Since this is tied to the `CHASSIS` component, in single chassis situations, I'll assume only ever a single value here per network-element but how would this be used in multi-chassis scenarios?  The component name would likely distinguish this as well.

What do you mean by "root of a tree of component ids" ?

release/models/platform/openconfig-platform-chassis.yang
```diff
@@ -0,0 +1,79 @@
...
+  // augment statements
+
+  augment "/oc-platform:components/oc-platform:component/" +
+    "oc-platform:chassis/oc-platform:config" {
+    description
+      "Augment chassis component config.";
+    uses chassis-config;
+  }
+
+    augment "/oc-platform:components/oc-platform:component/" +
```
earies, line 68: Indentation looks slightly off here

release/models/platform/openconfig-platform-chassis.yang
```diff
@@ -0,0 +1,79 @@
...
+
+  // data definition statements
+
+  // augment statements
+
+  augment "/oc-platform:components/oc-platform:component/" +
+    "oc-platform:chassis/oc-platform:config" {
+    description
+      "Augment chassis component config.";
+    uses chassis-config;
```
earies, line 65: Should this be gated by a `when` statement when the component is of type `CHASSIS` ? (This question goes for all containers that should only be used/visible if the component is of a certain type)

release/models/platform/openconfig-platform-chassis.yang
dplore: On behalf of Yang Zhang:  The intention here is for config-gen to have the ability to write a chassis-id number (uint32) into the switch, and for controller to read the value out. In a multi chassis situation, the config-gen will write different chassis-id numbers into different chassises. Similar to the /components/component/integrated-circuit/state/node-id which is for chip level id.

release/models/platform/openconfig-platform-chassis.yang
dplore: On behalf of Yang Zhang: I do not have a specific restriction for when statement here. If we do not have a gate for /components/component/integrated-circuit/state/node-id, then it is fine for chassis-id here too.

From Darren: I am also ok without a when statement for this.  This is an augment of oc-platform:chassis so perhaps it is implicit that chassis-id is only a child of chassis?
  +--rw components
     +--rw component* [name]
        +--rw chassis
        |    +--rw config
        |    |  +--rw oc-platform-chassis:chassis-id?   uint32

release/models/platform/openconfig-platform-chassis.yang
cfyo: These seems to conflict with the "id" that already exists at the top level: /components/component/state/id and there is a type in this state container as well that would be CHASSIS in this instance.  How will the "chassis-id" under /component/component/chassis/state be different?

release/models/platform/openconfig-platform-chassis.yang
```diff
@@ -0,0 +1,79 @@
...
+
+  // typedef statements
+
+  // grouping statements
+
+  grouping chassis-config {
+    description
+      "Config that corresponds to the global hardware package";
+
+    leaf chassis-id {
```
robshakir, line 49: Is there a use of this outside of P4RT? If there isn't, then I suggest that we should specify it (like we do with `node-id`) in the `openconfig-p4rt` module rather than here. [reference](https://github.com/openconfig/public/blob/master/release/models/p4rt/openconfig-p4rt.yang)

I think this makes it clearer what this is used for, and gives more context vs. being a generic "ID".

release/models/platform/openconfig-platform-chassis.yang
robshakir: ACK to @cfyo here, I think that the intent here is specifically for P4RT implementations, moving this under the p4rt module, and giving it a name that differentiates it for this use seems like a good idea.

robshakir (changes requested):

release/models/platform/openconfig-platform-chassis.yang
ghostli123: In OC standard, the "/components/component/state/id" is a string type (http://ops.openconfig.net/branches/models/master/docs/openconfig-platform.html). 

In Arista switch, below is some state/id query results, basically echo the name in the path. 

`{
        "Path": "components/component[name=Ucd9032-3]/state/id",
        "values": {
          "components/component/state/id": "Ucd9032-3"
        }
      },
      {
        "Path": "components/component[name=SwitchAsic1]/state/id",
        "values": {
          "components/component/state/id": "SwitchAsic-1"
        }
      }`

The one we requested (config/chassis-id and state/chassis-id) here is different from the state/id in several aspects: 1) the type is uint32; 2) it is writable and readable; 3) it is used to indicate which node a physical trunk is associated to. Since a trunk can cross multiple chips, we have to associate the trunk with a chassis node (instead of the chip node) for controller applications, and assign the `chassis node id` (no such an id in the switch now) for trunk association here.

release/models/platform/openconfig-platform-chassis.yang
ghostli123: Thanks for the suggestion, Rob! TBH, even for `node-id`, it is used beyond openconfig-p4rt in WBB and gPINS. For each singleton port, we associate it with a `node-id`, so that it could be processed by corresponding SFE node (chip node) when there is any update about the singleton port. Then the update event will be passed to upper-layer applications (e.g., TM). 

For example, eth1 is controller by chip1 in the physical switch. Then the port status of eth1 turns to UP, this info will be sensed by gNMI client (as we subscribe to interface state path). Based on switch hierarchy (interface->port->chip), the controller knows that eth1 belongs to chip1 (node-id is 999), and distributes this port state update along with other info (i.e., a high-level representation) to SFE chip1 node, based on the `node-id`. Then the SFE chip1 node is responsible for dispatching the high-level representation to upper layer applications. 

The chassis-id serves the same purpose here, but it is used for chassis node. Since p4rt does not have a sense of chassis, it might not be appropriate to move under `openconfig-p4rt`. But I do not have strong objection for moving if we want to keep consistence.

release/models/platform/openconfig-platform-chassis.yang
cfyo: Hi @ghostli123 - thanks for the response - a few things: 1) you could put an integer in the string and convert it  or propose modifying the type like making it a union like /components/component/type 2)  you could propose adding the id to the /components/component/config/id to make it writable 3) what other vendors have this same issue?

ghostli123: Hi @cfyo, thanks for the suggestion! 

1) it depends on the vendor support. this makes the vendor support harder.

2) Same above. It would be easier for vendor to add a new leaf rather than updating the existing logic. (I do not have strong objection, but changing the existing OC path representation would take much longer time for vendor to adapt the change, compared with a new independent leaf).

3) I just have Arista box to check out. I assume that the implementation of other vendors follow the same standard regarding state/id path right now.

dplore: What if we introduce a new, unique leaf for all components that is a user defined, configurable id.  This would make it consistent across all the components.  For example:  
/components/component/config/id-userdef
or
/components/component/config/description (this has precedent as a rw string in many other places)

@ghostli123 what do you think?  

If you think this might work for your use case, we need you to explain to the community in more detail what the use case is.   Why is it useful for the id to be a configurable integer?   Why does the read only, vendor defined id string not satisfy the use case?  

Maybe the above detail will help us choose a more intuitive name for the leaf.

If the use case is very unusual and not applicable, we may be better off doing a private augment that is not in openconfig.

ghostli123: Hi @dplore, for sure, it works for me. If the leaf can be written, and there is a corresponding state node to read, for the chassis id, that is good enough for our use case. Thanks for the suggestion!

release/models/platform/openconfig-platform-chassis.yang
robshakir: I think my preference for this to be somewhere other than just in the base platform model is that this is very much a SDN concept, I'm not sure that I've seen it be used in other approaches. Even if we take @dplore's approach with a configurable ID I think we should probably put it in some form of extension module.

dplore: Moved this to the p4rt module as it is relevant to the use case

release/models/platform/openconfig-platform-chassis.yang
dplore: no longer using this file so resolving

release/models/platform/openconfig-platform-chassis.yang
dplore: moved to p4rt model so resolving this thread.  Please feel free to open comments on the new commit

release/models/p4rt/openconfig-p4rt.yang
```diff
@@ -90,18 +106,16 @@ module openconfig-p4rt {
...
-  augment "/oc-if:interfaces/oc-if:interface/oc-if:state" {
-    description
-      "Add interface-specific applied configuration for P4RT.";
-
-    uses p4rt-interface-config;
+    leaf id {
+      type uint32;
+      description
+        "A unique identifier representing the global hardware device.  This is
+        intended to be used as a root of a tree of p4rt ids";
```
rolandphung, line 117:
```suggestion
        intended to be used as a root of a tree of P4RT ids";
```

release/models/p4rt/openconfig-p4rt.yang
```diff
@@ -145,4 +177,23 @@ module openconfig-p4rt {
 
     uses p4rt-ic-config;
   }
+
+  augment "/oc-platform:components/oc-platform:component/" +
+    "oc-platform:chassis/oc-platform:config" {
+    description
+      "Add p4rt chassis config to platform chassis component.";
```
rolandphung, line 184:
```suggestion
      "Add P4RT chassis config to platform chassis component.";
```

release/models/p4rt/openconfig-p4rt.yang
```diff
@@ -145,4 +177,23 @@ module openconfig-p4rt {
...
+    uses p4rt-chassis-config;
+  }
+
+  augment "/oc-platform:components/oc-platform:component/" +
+    "oc-platform:chassis/oc-platform:state" {
+    description
+      "Add p4rt chassis state to platform chassis component.";
```
rolandphung, line 191:
```suggestion
      "Add P4RT chassis state to platform chassis component.";
```

release/models/p4rt/openconfig-p4rt.yang
```diff
@@ -90,18 +106,16 @@ module openconfig-p4rt {
     }
   }
 
-  augment "/oc-if:interfaces/oc-if:interface/oc-if:config" {
+  grouping p4rt-chassis-config {
     description
-      "Add interface-specific intended configuration for P4RT.";
+      "Config regarding p4rt use cases which corresponds to the global device hardware package";
```
rolandphung, line 111:
```suggestion
      "Config regarding P4RT use cases which corresponds to the global device hardware package";
```

rolandphung (changes requested):

rolandphung (approved):

ghostli123 (approved):
Thanks for the implementation!

release/models/p4rt/openconfig-p4rt.yang
```diff
@@ -47,12 +53,22 @@ module openconfig-p4rt {
...
   identity P4RT {
-    base "oc-grpc:GRPC_SERVICE";
-    description
-      "P4RT: P4 Runtime (P4RT) Service.";
+      base "oc-grpc:GRPC_SERVICE";
+      description
+        "P4RT: P4 Runtime (P4RT) Service.";
   }
 
+  // typedef statements
```
robshakir, line 68: Not required?

release/models/p4rt/openconfig-p4rt.yang
```diff
@@ -90,18 +106,16 @@ module openconfig-p4rt {
...
-
-  augment "/oc-if:interfaces/oc-if:interface/oc-if:state" {
-    description
-      "Add interface-specific applied configuration for P4RT.";
-
-    uses p4rt-interface-config;
+    leaf id {
+      type uint32;
+      description
+        "A unique identifier representing the global hardware device.  This is
```
robshakir, line 116: "global hardware device" is interesting wording. Given that this is `chassis-id` should this just be "the overall set of forwarding resources that are controlled by this physical or logical chassis" (to handle multi-chassis)?

robshakir (commented):
Thanks for constraining this to the P4RT module, I think this is the only place it has relevance.

release/models/p4rt/openconfig-p4rt.yang
robshakir: Also, suggest that this description should include the word "P4RT" somewhere :-)

dplore: Perhaps:

> An identifier used for the root of oc-p4rt:node-id.  It is used to indicate which oc-p4rt:node-id's belong to the same device. ";

robshakir (approved):

dplore:
Thanks Rob, will cover this in the Aug 30 Openconfig Operators meeting

dplore:
This was covered in Aug 30 meeting without objection.  Will merge on Sep 2nd. (14 days after [commits](https://github.com/openconfig/public/pull/625/commits/0711e904e77a2f33962806a45be92ad99e5bafe2)  which moved this to the p4rt model)  if there are no further changes requested.

ghostli123:
Thanks for the help, Darren!

Issue #626, opened 2022-04-24
CLEI code for components


ejbrever:
I'd like to propose adding a leaf within the component state (https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform.yang#L318) for the CLEI code.

The CLEI code is an industry standard globally unique identifier for telecom components :
https://en.wikipedia.org/wiki/Common_Language_Information_Services

earies:
SGTM - I'd suggest just opening a PR to add the relevant node

For JUNOS, you can reference https://www.juniper.net/documentation/us/en/software/junos/interfaces-ethernet/topics/ref/command/show-chassis-hardware.html

dplore:
Seems reasonable and in scope to platform models to me.

ejbrever:
Just noting that most (if not all) optical platforms also support CLEI codes. A couple references:
Cisco: https://www.cisco.com/c/en/us/support/docs/optical-networking/ons-15454-sonet-multiservice-provisioning-platform-mspp/46247-15454-cleicodes.html
Nokia: https://documentation.nokia.com/html/365-372-324R10.0/index.html?i=365-372-324R10.0-qzx79

PR #628, opened 2022-04-25
CLEI-code

ejbrever:
https://github.com/openconfig/public/issues/626

@dplore @earies Can you please review?

### DIFF
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,7 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.18.0";
+  oc-ext:openconfig-version "0.19.0";
+
+  revision "2022-07-28" {
+    description
+      "Add container for controller card component";
+    reference "0.19.0";
+  }
 
   revision "2022-07-11" {
     description
@@ -1062,6 +1068,25 @@ module openconfig-platform {
         config false;
         description
           "Operational state data for software module components";
+      }
+    }
+
+    container controller-card {
+      description
+        "Data for controller card components, i.e., for components
+        with type=CONTROLLER_CARD";
+
+      container config {
+        description
+          "Configuration data for controller card components. Note that disabling
+          power to the primary supervisor should be rejected, and the operator is
+          required to perform a switchover first.";
+      }
+
+      container state {
+        config false;
+        description
+          "Operational state data for controller card components";
       }
     }
   }
--- a/release/models/platform/.spec.yml
+++ b/release/models/platform/.spec.yml
@@ -11,8 +11,10 @@
     - yang/platform/openconfig-platform-cpu.yang
     - yang/platform/openconfig-platform-ext.yang
     - yang/platform/openconfig-platform-software.yang
+    - yang/platform/openconfig-platform-fabric.yang
     - yang/platform/openconfig-platform-pipeline-counters.yang
     - yang/platform/openconfig-platform-integrated-circuit.yang
+    - yang/platform/openconfig-platform-controller-card.yang
   build:
     - yang/platform/openconfig-platform.yang
     - yang/platform/openconfig-platform-common.yang
@@ -24,6 +26,8 @@
     - yang/platform/openconfig-platform-ext.yang
     - yang/platform/openconfig-platform-cpu.yang
     - yang/platform/openconfig-platform-software.yang
+    - yang/platform/openconfig-platform-fabric.yang
     - yang/platform/openconfig-platform-pipeline-counters.yang
     - yang/platform/openconfig-platform-integrated-circuit.yang
+    - yang/platform/openconfig-platform-controller-card.yang
   run-ci: true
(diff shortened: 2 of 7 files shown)


earies:
LGTM

robshakir:
@ejbrever -- I see that there is one implementation linked from the issue specified. Could we link others here too please?

release/models/platform/openconfig-platform.yang
```diff
@@ -407,6 +413,14 @@ module openconfig-platform {
         (field replaceable unit)";
     }
 
+    leaf clei-code {
+      type string;
+      description
+        "Common Language Equipment Identifier (CLEI) code of the
```
robshakir, line 419: Is there a reference for CLEI? It'd be good to know whether there is more than just a `string` format for it, but I wasn't sure where to look it up :-)

robshakir (changes requested):

release/models/platform/openconfig-platform.yang
earies: It is covered as part of ANSI ATIS-0300213.2014 however these are not freely open accessible documents

https://www.commonlanguage.com/resources/commonlang/productshowroom/showroom/equip_id/carriers/eqpt_ja_501_i1.pdf

```
1. What is a Common Language Equipment Identification (CLEI) Code?
CLEI Codes are globally unique, 10-character alphanumeric intelligent codes that
identify equipment in a structured naming format. There is a one-to-one relationship
between a CLEI Code and a Manufacturer’s Product Code (part number including
the hardware version).
```

So it could likely be scoped if necessary but unsure if all implementations conform to this tightly

cfyo: Hi @earies - if the CLEI code has a relationship with the manufacturer's product code and there is already a "part-no"  leaf and a "hardware-version" leaf that can be used to identify the CLEI code in the provider's backend inventory/asset management system.  It is sort of like the CLLI code that is covered by the "location" leaf.

dplore: This sounds sensible, but it seems it is non trivial to automatically derive the CLEI code from part-no leaf and hardware-version when the device could provide it directly.  I am not a subject matter expert on CLEI data, so my opinion is not very informed.

ejbrever: On the config side that makes sense to me as if we wanted to configure a card type for instance, part-no and potentially hardware-version are all that would be needed. We wouldn't want to have an additional config leaf which could lead to unnecessary configuration permutations. I definitely think we should push back on having CLEI be used for config for that reason.

In this case, this is just a state leaf. Some operators do use CLEI codes and having to figure out this mapping on the operator side seems like work we can remove by letting the device's report this value. Please let me know if this makes sense.

ejbrever:
@robshakir sure, I linked a couple other implementations in the issue as well that were publicly accessible. I believe all optical vendors (at least the ones I have worked with) support CLEI codes, although most of the documentation is not publicly accessible.

osquitun:
I would then also question about CLLI to allow configuration for location that is related to CLEI based on Telcordia closed specifications.  These originated prior to the divesture of A&T and are largely US telco requirements that pertain to a small subset of providers and deal with inventory and asset management that are best left to other backend systems as these attributes are of practical use by operations and teams dealing with the management and operational state of the device.  The vendors selling to these telco provides are required to provide CLEI like larger router and optical vendors who also still use outdated TL1 protocols for management and configuration.  Can these be provided, yes, but should they as the industry moves forward with increasing virtualization where these do no apply?

ejbrever:
@osquitun, I am okay either way with CLLI as well. CLEI and CLLI I think also present two different problem spaces for me.

With CLLI, from the device's perspective, this is just a random string as it doesn't know or care where it physically is located. You may have to configure it, but ultimately it doesn't drive any meaningful changes on the device.

With CLEI, I do worry if this is offered as a config leaf that it might get used one day by an implementation to drive configuration. It is also a special piece of information that the platform knows already and that operators would not know without digging into the platform's documentation.

For those reasons, I am okay with both of these ending up in OpenConfig. However, CLEI should only be a state leaf, where CLLI should be config and state.

* Side note...I have no intentions of adding CLLI into this PR :)

dplore:
@ejbrever can you reference at least two vendors that support providing CLEI data?  Ideally you could provide a link to documentation showing these vendors support this information.

Also note, you'll need to rebase/merge this PR to catch it up with recent contributions.

ejbrever:
@dplore sure. Most (if not all) optical platforms support CLEI codes. A couple references:
Cisco: https://www.cisco.com/c/en/us/support/docs/optical-networking/ons-15454-sonet-multiservice-provisioning-platform-mspp/46247-15454-cleicodes.html
Nokia: https://documentation.nokia.com/html/365-372-324R10.0/index.html?i=365-372-324R10.0-qzx79

dplore:
LGTM, but need a version bump:

⛔  submodule versions must match the belonging module's version
module set openconfig-platform is at 0.17.0 (openconfig-platform.yang), non-matching files: openconfig-platform-common.yang (0.16.0)

PR #630, opened 2022-04-29
Add convergence tracking stats to ATE flow model

jasdeep-hundal:
Adds leaves to expose statistics related to protocol convergence testing (eg. the time taken for BGP routes to be installed and used for forwarding.)


robshakir (approved):
This looks good to me :-) Thanks!

PR #632, opened 2022-04-29
Add a leaf to be able to add routing-mac-address to the system.

rsgcp:
This leaf will add the ability to add a new mac-addresses to the system. In particular, it will allow the ability to configure a routing-mac-address. Any packets destined to the routing-MAC-address will be sent through the routing pipeline of the system, as opposed to just being switched at layer-2. Most routers today allocate a MAC address from the block of MAC address that the system owns for this purpose, like the Integrated Routing and Bridging (IRB) MAC. This configuration will allow adding an additional routing MAC address that does not necessarily belong to the block of MAC address that system owns.


release/models/system/openconfig-system.yang
```diff
@@ -266,6 +272,18 @@ module openconfig-system {
         information such as the current system date and time, uptime,
         last login timestamp, etc.";
     }
+
+    leaf my-mac-address{
```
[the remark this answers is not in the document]
rsgcp: Thanks Ebben. Not sure why lint did not catch that.

We discussed about the possibility of multiple MAC address, but weren't sure of a use case where controller may want several different ones. So left it singular. We can always extend it if someone comes up with a case where controller managed multiple MAC addresses are needed.

release/models/system/openconfig-system.yang
```diff
@@ -266,6 +272,18 @@ module openconfig-system {
...
+      type oc-yang:mac-address;
+      description
+        "MAC address to be added to my-mac-address table. Packets with
+	the dest-mac as this mac-address should be routed by the switch.
```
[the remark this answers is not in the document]
rsgcp: Replace <tab> with spaces. Should align now. Reworded description.

rsgcp:
@robshakir @dplore @earies  gentle reminder to finish review. Should be a quick one. Thanks.

release/models/system/openconfig-system.yang
```diff
@@ -266,6 +272,17 @@ module openconfig-system {
         information such as the current system date and time, uptime,
         last login timestamp, etc.";
     }
+
+    leaf my-mac-address {
```
robshakir, line 276: Should we call this `local-mac-address` or similar? I'm not sure whether we ever refer to a device owning things with `my` anywhere in the models.

release/models/system/openconfig-system.yang
```diff
@@ -266,6 +272,17 @@ module openconfig-system {
...
+      type oc-yang:mac-address;
+      description
+        "MAC address to be added to the my-mac-address table, in addition to
```
robshakir, line 279: Is there a reference to this table's existence generically across system implementations? If not, it would be worth defining what "my" means. I assume that you are specifying something that is treated as a local MAC owned by the controller cards of the device.

release/models/system/openconfig-system.yang
robshakir, line 276: Do we ever expect that there is other configuration of this type? I'm not quite sure that `/system/config` is the best place for this to go - maybe something like `/system/mac-addr/config/local`, you could add other MAC-related configuration here.

earies: +1 to above comment - if this is only ever singular (as previously mentioned) and system wide just simply `/system/config/mac-address` or `/system/config/local-mac-address` should suffice.  Would recommend against first person pronoun here.

release/models/system/openconfig-system.yang
dplore: +1 to move leaf to `/system/config/mac-address`  and also request a more elaborate description for this leaf.  The description should help answer questions like:

* are packets sourced by the controller card intended to use this mac address as a source mac address?
* are packets ingress to the device with a dest mac matching `/system/config/mac-address`  expected to be  punted to the controller card?  
* are there any other functions affected by this mac address?

What is a mac address table?  Do we need to refer or define a mac address table, and/or Is this meant to be an ARP table?

dplore (commented):
+1 to move leaf to /system/config/mac-address 

Also request a more elaborate description for this leaf. The description should help answer questions like:

Are packets sourced by the controller card intended to use this mac address as a source mac address?

are packets ingress to the device with a dest mac matching /system/config/mac-address expected to be punted to the controller card?

are there any other functions affected by this mac address?

What is a mac address table? Do we need to refer or define a mac address table, and/or Is this meant to be an ARP table?

release/models/system/openconfig-system.yang
rsgcp: Added clarification for these. Thanks.

release/models/system/openconfig-system.yang
```diff
@@ -908,6 +914,49 @@ module openconfig-system {
...
 
+  grouping system-macaddr-routing-config {
+    description
+      "Configuration data for system's routing MAC addresses.";
+
+    leaf routing-mac {
+      type oc-yang:mac-address;
+      description
+        "Any packets destined to this MAC address must be routed by the system,
+	in addition to any other MAC addresses that system may already have been
```
robshakir, line 925: Please fix indentation here - it should match the `"Any` line above.

release/models/system/openconfig-system.yang
```diff
@@ -908,6 +914,49 @@ module openconfig-system {
...
   }
 
+  grouping system-macaddr-routing-config {
+    description
+      "Configuration data for system's routing MAC addresses.";
+
+    leaf routing-mac {
+      type oc-yang:mac-address;
+      description
+        "Any packets destined to this MAC address must be routed by the system,
```
robshakir, line 924: I'm a little confused at this statement -- should we say that these MACs MUST be accepted by the system and forwarded per normal operations? I think this is what the rest of the paragraph says, but it'd be good to be very explicit.

release/models/system/openconfig-system.yang
robshakir, line 925:
```suggestion
	in addition to any other MAC addresses that the system may already have been
```

release/models/system/openconfig-system.yang
```diff
@@ -908,6 +914,49 @@ module openconfig-system {
...
+	in addition to any other MAC addresses that system may already have been
+	using for that purpose. Packets destined to this MAC address are expected
+	to go through ingress lookup and result in packet being routed by the
+	switch. It is not expected to be used as the source MAC address of the
```
robshakir, line 928: I think that we should probably say that they should be forwarded according with the result of that lookup -- if the action is to drop, then they should be dropped.

robshakir (changes requested):

release/models/system/openconfig-system.yang
```diff
@@ -914,6 +920,52 @@ module openconfig-system {
...
+
+    leaf routing-mac {
+      type oc-yang:mac-address;
+      description
+        "Any packets destined to this MAC address must be sent through the
+        routing pipeline by the system. This MAC address is used to identify
+        routed packets in addition to any other destination MAC addresses,
+        for instance Integrated Routing and Briding MAC or L3 interface MAC,
+        that the system may already have been using to perform routing.
+            It is not expected that this MAC address will be used as the
```
robshakir, line 935:
```suggestion

         It is not expected that this MAC address will be used as the
```

release/models/system/openconfig-system.yang
```diff
@@ -914,6 +920,52 @@ module openconfig-system {
...
+        any packets generated by the system, or a MAC address used in ARP
+        response. This MAC address may not be allocated from the block of
+        MAC address that system owns. For instance, it's allocation could
+        be managed by an external controller.";
+    }
+  }
+
+  grouping system-macaddr-top {
+    description
+      "Top-level grouping for configuration and state of system's MAC addresses";
```
robshakir, line 946:
```suggestion
      "Top-level grouping for configuration and state of system's MAC addresses.";
```

(nit) can we add periods to the end of these sentences here and below.

robshakir (changes requested):

rsgcp (commented):
Changes incorporated.

rsgcp (commented):
Changes incorporated.

rsgcp (commented):
Changes incorporated.

rsgcp:
Merging virtual mac address changes.

release/models/system/openconfig-system.yang
rsgcp: Addressed all of these in new commit.

release/models/system/openconfig-system.yang
```diff
@@ -914,6 +920,52 @@ module openconfig-system {
...
+
+    leaf routing-mac {
+      type oc-yang:mac-address;
+      description
+        "Any packets destined to this MAC address must be sent through the
+        routing pipeline by the system. This MAC address is used to identify
+        routed packets in addition to any other MAC addresses that the system
+        may already have been using to perform routing.
+
+         It is not expected that this MAC address will be used as the
```
robshakir, line 935:
```suggestion
        It is not expected that this MAC address will be used as the
```

(nit)

robshakir (approved):
this looks ok to go for me, but i'd like some commentary from the wider group - @rgwilton @hellt @aredmon8551 @rolandphung @earies PTAL.

dplore:
@rsgcp will update the PR name and PR description with the use case and references to vendor documentation regarding integrated routing and bridging (IRB) as this leaf is closely related to IRB functionality.

rolandphung (approved):

earies (approved):

rsgcp:
Hi @robshakir @dplore, I have updated the description and header. Please take a look and merge to the mainline. Thanks.

dplore:
Thanks for the contribution @rsgcp !

Issue #629, opened 2022-04-27
Add Authentication in VRRP configuration


sbarguil:
A router will ignore incoming VRRP packets from routers that do not have the same authentication configuration for a VRRP group. 
VRRP commonly has three authentication schemes:

- No authentication
- Plain text authentication
- MD5 authentication

VRRP packets will be rejected in any of the following cases:

- The authentication schemes differ on the router and in the incoming packet.
- MD5 digests differ on the router and in the incoming packet.
- Text authentication strings differ on the router and in the incoming packet.

PR #636, opened 2022-05-03
629- Update authentication for VRRP

sbarguil:


sbarguil:
#629

robshakir:
@sbarguil I think that this was the PR that you mentioned where there needed to be some discussion of what we did in IS-IS, is that right? Could you remind me of the question :-) Thanks.

dplore:
@sbarguil can you please provide the following:
reference to at least 2 vendor implementations of this feature (urls to documentation is great)

For convenience for reviewers, can you provide the following
tree output (for example from pyang -f tree) before and after so we can easily visualize the difference?  

Thanks for your help making it faster/easier to review!

sbarguil:
> @sbarguil can you please provide the following: reference to at least 2 vendor implementations of this feature (urls to documentation is great)
> 
> For convenience for reviewers, can you provide the following tree output (for example from pyang -f tree) before and after so we can easily visualize the difference?
> 
> Thanks for your help making it faster/easier to review!

Hi @dplore, here are the references:

- Cisco configuration guide for VRRP: [Link](https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/ipapp_fhrp/configuration/15-mt/fhp-15-mt-book/fhp-vrrp.html#GUID-3931AB2F-738A-4BE6-B06E-E46AB30A33FA)
- Juniper VRRP config guide (only for ipv4): [Link](https://www.juniper.net/documentation/us/en/software/junos/high-availability/topics/task/vrrp-authentication-for-ipv4-configuring.html)
- Nokia VRRP authentication (only for ipv4):[Link]( https://documentation.nokia.com/html/0_add-h-f/93-0073-HTML/7750_SR_OS_Router_Configuration_Guide/VRRP-Intro.html#:~:text=VRRP%20message%20authentication%20is%20applicable,transmitting%20a%20VRRP%20advertisement%20message.)

robshakir:
w.r.t the fact that IS-IS has both `SIMPLE_KEY` within a keychain and `TEXT`, this was intentional since we noted that it wasn't possible to map some implementations to the keychain model (and there tended to be support for both). As implementations catch up to supporting keychain, we could choose to deprecate the configuration directly within the IS-IS model.

If there are implementations that (in the VRRP context) map to just keychain, then this seems reasonable not to have the same approach that we have in IS-IS and just use keychain.

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -819,6 +828,61 @@ module openconfig-if-ip {
     }
   }
 
+  grouping ip-vrrp-simple-key-authentication-config {
+    description
+      "This grouping defines VRRP simple authentication config.";
+
+    leaf auth-mode {
```
robshakir, line 835: @dplore Doesn't this leaf do the same job as `auth-type` that you'd mentioned? (maybe it was added after you commented.)

PR #645, opened 2022-05-10
Fix submodule versioning from commit 3a5269e

earies:
Fix submodule versioning from commit 3a5269e - CI/CD pipeline currently reporting errors as a result


dplore (approved):
Thanks for your fix.  These made it past earlier CI checks somehow.

PR #646, opened 2022-05-12
Add state-synced container under afts container

akalluru1:
AFT streaming over gNMI is an eventually consistent system.
When the device updates an entry it is usually expected to
stream an update to the client within a vert short amount
of time (few milliseconds). So, Telemetry collector or a
controller that parse the AFT doesn't have a consistent
snapshot, or overall versioned copy of AFT with the device
at any point of time.

In certain failure modes like device boot up, gNMI daemon
failure and device/routing engine stateful switchover
Telemetry collector or a controller need a flag to
determine whether it is in consistent with the device or
not such that it can a corrective action when needed.
A device would set this leaf or flag to indicate to the
client that AFT data/view is consistent.

Following Tree shows new container `state-synced` & newly added elements under it:
```  
+--rw network-instances
        +--ro afts
           +--ro ipv4-unicast
           +--ro ipv6-unicast
           +--ro state-synced
           |  +--ro state
           |     +--ro ipv4-unicast?   boolean
           |     +--ro ipv6-unicast?   boolean
```

### DIFF
--- a/release/models/aft/openconfig-aft-state-synced.yang
+++ b/release/models/aft/openconfig-aft-state-synced.yang
@@ -40,7 +40,7 @@ submodule openconfig-aft-state-synced {
         default false;
         description
           "State synced signal indicating consistent device snapshot of
-          IPv4 unicast AFT entries. Before setting this flag to True
+          IPv4 unicast AFT entries. Before setting this flag to true
           next-hop-groups and next-hops AFT entries, associated with
           ipv4-unicast AFT entries, are expected to be consistent with
           device snapshot.";
--- a/release/models/aft/openconfig-aft-state-synced.yang
+++ b/release/models/aft/openconfig-aft-state-synced.yang
@@ -18,7 +18,7 @@ submodule openconfig-aft-state-synced {
 
   oc-ext:openconfig-version "2.1.0";
 
-  revision "2022-05-12" {
+  revision "2022-06-09" {
     description
       "Add state-synced container under afts.";
     reference "2.1.0";
--- a/release/models/aft/openconfig-aft-state-synced.yang
+++ b/release/models/aft/openconfig-aft-state-synced.yang
@@ -52,7 +52,7 @@ submodule openconfig-aft-state-synced {
         description
           "State synced signal indicating consistent device snapshot of
           IPv6 unicast AFT entries. Before setting this flag to true
-          next-hop-groups & next-hops AFT entries, associated with
+          next-hop-groups and next-hops AFT entries, associated with
           ipv6-unicast AFT entries, are expected to be consistent with
           device snapshot.";
       }
--- a/release/models/aft/openconfig-aft-state-synced.yang
+++ b/release/models/aft/openconfig-aft-state-synced.yang
@@ -41,7 +41,7 @@ submodule openconfig-aft-state-synced {
         description
           "State synced signal indicating consistent device snapshot of
           IPv4 unicast AFT entries. Before setting this flag to True
-          next-hop-groups & next-hops AFT entries, associated with
+          next-hop-groups and next-hops AFT entries, associated with
           ipv4-unicast AFT entries, are expected to be consistent with
           device snapshot.";
       }
--- a/release/models/aft/openconfig-aft-state-synced.yang
+++ b/release/models/aft/openconfig-aft-state-synced.yang
@@ -1,4 +1,4 @@
-submodule openconfig-aft-fib-ready {
+submodule openconfig-aft-state-synced {
   belongs-to "openconfig-aft" {
     prefix "oc-aft";
   }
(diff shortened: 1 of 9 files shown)


xw-g:
@akalluru1, thanks for the contribution. The use case / motivation makes sense. A few comments: 

1. Is it practical, or needed, to differentiate the `fib_ready` state of each category (e.g. `ip-unicast` in your current PR suggestion), instead of just a single flag/state for the whole AFT table?. If we do think a state flag is needed per category/types, then probably we need to add more clarifications, e.g., if `ip-unicast=true` but `next-hop-group=false`, which one to trust? Should we wait?  
2. Do we have a better name than `fib_ready`? `fib_ready` might make ppl think it's "ready" for forwarding, etc.? How about something like `state_synced`?

xw-g:
cc: @earies

akalluru1:
@xw-g Thank you so much for the review! Appreciate your help. Replies to your comments below:

1. The intention behind current suggestion `ip-unicast` was that in future we can add another flag for `mpls` AFT. Arista EOS maintains different state synced flags for IP FIB (for IPv4 & IPv6) & MPLS LFIB tables as one table represents IP forwarding & another represents label forwarding. We don't need or intend to have a flag for each AFT category or container.
2. Sure, we will discuss this internally and modify the container name to `state_synced` as suggested.

akalluru1:
one small clarification on the name, you mean `state-synced`, i.e., using `-` (hyphen) instead of `_`(underscore) as this seems to be the pattern or syntax.

xw-g:
Thanks, @akalluru1. 

1.  If the intention currently only comes from one vendor's implementation details (maybe other vendors do the same), it sounds like we want to take more thought/discussion to make sure the proposal is vendor neutral?  Another ways to ask the question is: is it reasonable/ok to have only one flag for the AFT table(regardless IP or MPLS) in the case of possible failures? 
2. Yes, thanks, I meant `state-synced` as an option for consideration. 

@earies anyone from your team can help comment?

akalluru1:
1. It makes sense. We will wait for the update.
2. I'll update the diff with the suggestion later today.

robshakir:
I think here it makes more sense to have a separate `ipv4-unicast` and `ipv6-unicast` flag for synchronisation if we decide they need to be separate. The current approach means that if an implementation has a way to determine separately for each, it cannot be easily represented - whereas if the two are coupled, it's relatively trivial to set two leaves to `true` together. 

I echo @xw-g's question here -- how are NH + NHG covered here? Are they assumed to be consistent always?

Is there a reason to put this in a separate container? It seems like `afts/state/ipv[46]-unicast-synced` would be sufficient.

akalluru1:
Thanks for your comment @robshakir! I'll add 2 flags, i.e., `ipv4-unicast` & `ipv6-unicast` instead of a single flag `ip-unicast`. NHG + NH are assumed to be consistent once these flags are enabled only after the relevant NHGs & NHs, referred by IPv[46] routes, are updated. Currently `state` container is not present under `afts` container so that was the reason to add a new container name `state-synced`. If you think it's better to just name this new container as `state` please let me know I can rename it.

robshakir:
Thanks -- please ping me again once you've made that change.

I think it would be good to have each of the `ipv4-unicast` and `ipv6-unicast` flags' description state that the NHG and NHs must be consistent before they are set. 

I'm fine with `state-synced/state/ipv[46]-unicast` as a name.

akalluru1:
@robshakir I've made the change as you suggested. please take a look. Thanks!

release/models/aft/openconfig-aft-state-synced.yang
```diff
@@ -0,0 +1,61 @@
...
+          ipv4-unicast AFT entries, are expected to be consistent with
+          device snapshot.";
+      }
+
+      leaf ipv6-unicast {
+        type boolean;
+        default false;
+        description
+          "State synced signal indicating consistent device snapshot of
+          IPv6 unicast AFT entries. Before setting this flag to True
```
robshakir, line 54:
```suggestion
          IPv6 unicast AFT entries. Before setting this flag to true
```

release/models/aft/openconfig-aft-state-synced.yang
```diff
@@ -0,0 +1,61 @@
...
+          next-hop-groups & next-hops AFT entries, associated with
```
robshakir, line 55:
```suggestion
          next-hop-groups and next-hops AFT entries, associated with
```

release/models/aft/openconfig-aft-state-synced.yang
```diff
@@ -0,0 +1,61 @@
...
+        "Operational state parameters relating to the state
+        synced signals of various AFTs.";
+
+      leaf ipv4-unicast {
+        type boolean;
+        default false;
+        description
+          "State synced signal indicating consistent device snapshot of
+          IPv4 unicast AFT entries. Before setting this flag to True
+          next-hop-groups & next-hops AFT entries, associated with
```
robshakir, line 44:
```suggestion
          next-hop-groups and next-hops AFT entries, associated with
```

release/models/aft/openconfig-aft.yang
```diff
@@ -232,6 +241,27 @@ module openconfig-aft {
         uses aft-ethernet-structural;
       }
 
+      container state-synced {
+        description
+          "AFT streaming over gNMI is an eventually consistent system.
```
robshakir, line 246:
```suggestion
          "In some cases AFT streaming (e.g., over gNMI) is an eventually consistent system.
```

Minor editorial change because gNMI is not intended to be the only supported transport for OpenConfig models (albeit at this point it is the defacto standard around telemetry).

release/models/aft/openconfig-aft.yang
```diff
@@ -232,6 +241,27 @@ module openconfig-aft {
...
+          When the device updates an entry it is usually expected to
+          stream an update to the client within a vert short amount
+          of time (few milliseconds). So, Telemetry collector or a
```
robshakir, line 249:
```suggestion
          of time (few milliseconds). Given this is the casee, a telemetry collector or a
```

release/models/aft/openconfig-aft.yang
```diff
@@ -232,6 +241,27 @@ module openconfig-aft {
...
+          controller that parse the AFT doesn't have a consistent
+          snapshot, or overall versioned copy of AFT with the device
+          at any point of time.
```
robshakir, line 252:
```suggestion
          at any specific point of time.
```

robshakir (approved):
LGTM, very small editorial points on wording and then we're good to merge.

robshakir:
@akalluru1 - can you merge master into your fork so that the CI runs with the update ygot workflow please?

akalluru1:
@robshakir I've pulled in a change or synced from openconfig:master Is that what you were referring to?

robshakir:
Yep, exactly -- kicking off the CI.

dplore (approved):
LGTM, thank you

PR #653, opened 2022-06-08
Implement acceptable interface identifier range for P4RT

earies:
  - Add range statement to interface identifier in conformance to P4RT
    specification v1.3.0
  - Description cleanup to map OpenConfig terminology to P4RT
    specification terminology

Per https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-translation-of-port-numbers, the 'SDN Port' is the equivalent attribute to the OpenConfig P4RT interface 'id'.  This change limits the acceptable uint32 range for configurable values per specification.

### DIFF
--- a/release/models/p4rt/openconfig-p4rt.yang
+++ b/release/models/p4rt/openconfig-p4rt.yang
@@ -72,7 +72,18 @@ module openconfig-p4rt {
 
         Note that this identifier is used only when a numeric reference
         to the interface is required, it does not replace the unique
-        name assigned to the interface.";
+        name assigned to the interface.
+
+        Ranges:
+
+        0x00000000: Unspecified/Invalid
+        0x00000001: Minimum Port Value
+        0xFFFFFEFF: Maximum Port Value
+
+        Reserved:   0xFFFFFFF0 - 0xFFFFFFFF
+        0xFFFFFFFA: Recirculate
+        0xFFFFFFFD: CPU
+        ";
       reference
         "P4 Runtime Specification v1.3.0
         https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-translation-of-port-numbers";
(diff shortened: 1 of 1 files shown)


robshakir (approved):
This LGTM. One question we could have is whether to change the type here to have enumerated values for the special cases, but given that these are actually invalid values - then I think I prefer this change as-is.

@dplore @rsgcp @xw-g -- any objections?

earies:
> This LGTM. One question we could have is whether to change the type here to have enumerated values for the special cases, but given that these are actually invalid values - then I think I prefer this change as-is.
> 
> @dplore @rsgcp @xw-g -- any objections?

Thx @robshakir  - yeah, if all options were available for config or state of all possible values in a different construct/location then I would say yes but since this restriction is on the interface `id` itself it's locked into the range for 'SDN Ports'

dplore:
Does an SDN controller ever need to address the SDN_PORT_RECIRCULATE or SDN_PORT_CPU?  Reading from the spec it seems like this is explicitly defined and permitted?  Seems like the change should permit those values.

earies:
> Does an SDN controller ever need to address the SDN_PORT_RECIRCULATE or SDN_PORT_CPU? Reading from the spec it seems like this is explicitly defined and permitted? Seems like the change should permit those values.

This restriction is solely for the interface `id` node - e.g. to define an identifier that falls within the valid SDN_PORT range

RECIRCULATE and CPU would not be values ever assigned to a interface in this context

dplore:
Ok, I think you are saying /interfaces/interface is not expected to include  CPU or RECIRCULATE entities and therefore we don't need to allow those IDs?

earies:
> Ok, I think you are saying /interfaces/interface is not expected to include  CPU or RECIRCULATE entities and therefore we don't need to allow those IDs?

Correct

dplore:
Hi @wenovus we are seeing ygot CI check failing here?

wenovus:
> Hi @wenovus we are seeing ygot CI check failing here?  

Is the PR in sync with master?

dplore:
@earies please rebase/merge with master to pick up changes that should fix the CI issue.  :)

earies:
> @earies please rebase/merge with master to pick up changes that should fix the CI issue. :)

thx @dplore @wenovus - branch sync'd and passing CI now

dplore (approved):

PR #655, opened 2022-06-10
added units and clarifying text for timestamps.

sulrich:
- (M) release/models/platform/openconfig-platform.yang
  - added units and clarifying text for the following leafs
    - last-switchover-time
    - last-reboot-time

### DIFF
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,7 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.16.0";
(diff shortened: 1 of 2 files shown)


release/models/platform/openconfig-platform.yang
```diff
@@ -472,11 +478,14 @@ module openconfig-platform {
 
     leaf last-switchover-time {
       type oc-types:timeticks64;
+      units "nanoseconds";
```
robshakir, line 481: Technically, I think this is now stated in the type (https://github.com/openconfig/public/blob/master/release/models/types/openconfig-types.yang#L106), but I am OK with restating it here.

robshakir (approved):
LGTM, thanks for the change.

PR #657, opened 2022-06-15
Fix ygot CI

wenovus:


robshakir:
Did this fix the CI at this point, Wen, or is there something more we should look at?

robshakir:
Looks like this is fixed at this point, do you still want to submit this? (I couldn't tell whether the change of GCB base image was needed here or not.)

wenovus:
> Looks like this is fixed at this point, do you still want to submit this? (I couldn't tell whether the change of GCB base image was needed here or not.)
No it's not fixed, the GCB trigger somehow got disconnected between yesterday and this morning. I just messaged you offline about this issue.

robshakir (approved):

wenovus:
@robshakir @dplore ready to merge now.

PR #662, opened 2022-06-28
Remove invalid openconfig-if-types references

earies:
* (D) release/models/interfaces/openconfig-if-types.yang
    - Deletion of unused openconfig-if-types
  * (M) release/models/interfaces/openconfig-if-aggregate.yang
  * (M) release/models/bfd/openconfig-bfd.yang
    - Removal of invalid when statements to unused identities

`openconfig-if-types` was introduced as a module to potentially deprecate the
use of `iana-if-type` however was never seen to completion for the transition.
Any references to statements depending on this module were incorrect as the
interface `type` cannot use the identities defined within.  This patch set
removes any references to `openconfig-if-types` for the current model set and
can be revisited if this path is still of interest at a later date.


dplore:
This looks like a good cleanup item to me.  `openconfig-if-types` is not referenced in any other models.

dplore (approved):

dplore (approved):

PR #664, opened 2022-07-07
Do not submit: Test yanglint2

wenovus:


wenovus:
So it looks like yanglint 2.0.221 (note: not 2.0.112) fixes the issue. I'm content just waiting for the next Debian package to be released instead of trying to build our own Debian package. I will revert yanglint to v1 again for the present.

hellt:
or you can use the container image I built to close the gap between code release and binaries

PR #666, opened 2022-07-08
Add initial OpenConfig model for gRPC Tunnel feature

earies:
* (A) release/models/system/openconfig-system-grpc-tunnel.yang
    - Add OpenConfig model for gRPC Tunnel feature

References:
* https://github.com/openconfig/grpctunnel/
* https://github.com/openconfig/grpctunnel/blob/master/doc/grpctunnel_design.md
* https://github.com/openconfig/grpctunnel/blob/master/proto/tunnel/tunnel.proto

```
module: openconfig-system
  +--rw system
     +--rw oc-sys-grpc-tun:grpc-tunnel
        +--rw oc-sys-grpc-tun:config
        |  +--rw oc-sys-grpc-tun:retry-interval?   uint64
        +--ro oc-sys-grpc-tun:state
        |  +--ro oc-sys-grpc-tun:retry-interval?   uint64
        +--rw oc-sys-grpc-tun:servers
        |  +--rw oc-sys-grpc-tun:server* [name]
        |     +--rw oc-sys-grpc-tun:name      -> ../config/name
        |     +--rw oc-sys-grpc-tun:config
        |     |  +--rw oc-sys-grpc-tun:name?                 string
        |     |  +--rw oc-sys-grpc-tun:enable?               boolean
        |     |  +--rw oc-sys-grpc-tun:host?                 oc-inet:host
        |     |  +--rw oc-sys-grpc-tun:port?                 oc-inet:port-number
        |     |  +--rw oc-sys-grpc-tun:transport-security?   boolean
        |     |  +--rw oc-sys-grpc-tun:certificate-id?       string
        |     |  +--rw oc-sys-grpc-tun:source-address?       oc-inet:ip-address
        |     |  +--rw oc-sys-grpc-tun:network-instance?     oc-netinst:network-instance-ref
        |     |  +--rw oc-sys-grpc-tun:dscp?                 oc-inet:dscp
        |     |  +--rw oc-sys-grpc-tun:target*               target-type
        |     +--ro oc-sys-grpc-tun:state
        |        +--ro oc-sys-grpc-tun:name?                 string
        |        +--ro oc-sys-grpc-tun:enable?               boolean
        |        +--ro oc-sys-grpc-tun:host?                 oc-inet:host
        |        +--ro oc-sys-grpc-tun:port?                 oc-inet:port-number
        |        +--ro oc-sys-grpc-tun:transport-security?   boolean
        |        +--ro oc-sys-grpc-tun:certificate-id?       string
        |        +--ro oc-sys-grpc-tun:source-address?       oc-inet:ip-address
        |        +--ro oc-sys-grpc-tun:network-instance?     oc-netinst:network-instance-ref
        |        +--ro oc-sys-grpc-tun:dscp?                 oc-inet:dscp
        |        +--ro oc-sys-grpc-tun:target*               target-type
        +--rw oc-sys-grpc-tun:target-options
           +--rw oc-sys-grpc-tun:config
           |  +--rw oc-sys-grpc-tun:pattern*         enumeration
           |  +--rw oc-sys-grpc-tun:custom-string?   string
           |  +--rw oc-sys-grpc-tun:delimeter?       string
           +--ro oc-sys-grpc-tun:state
              +--ro oc-sys-grpc-tun:pattern*         enumeration
              +--ro oc-sys-grpc-tun:custom-string?   string
              +--ro oc-sys-grpc-tun:delimeter?       string
```


earies:
@dplore @robshakir PTAL

dplore:
@csl to review

gwizdms:
Would a better choice be to integrate into openconfig-telemetry.yang?

earies:
@gwizdms - `openconfig-telemetry` imo is in need of an overhaul to the point where it would likely be deprecated - a PR will be cut soon that will expose more granularity to gRPC services outside of this model that will cover much of the dynamic subscriptions (as well as other RPCs and services such as gNOI, gRIBI, P4RT, etc..).  The persistent subscriptions portion of the telemetry model is not precise enough and leads to proprietary implementation today (With varying nodes having never been implemented and supported).

At the same time, `grpctunnel` is not limited to "telemetry" so I think it would be incorrect to attach it to that domain as it is more so a generic call-home method to transport any inbound TCP services

mirovr:
Ebben,
would it be possible to extend target-type with the configurable port? 
Often, the implemetations allow to specify custom port for servers, like NETCONF, gRPC and etc.
In addition, I believe it would be nice to have an option for target-type being custom string. This allows to use grpc-tunnel also for custom services. for instance, telnet is not specified in this proposal....

thanks!
miro

earies:
> Ebben, would it be possible to extend target-type with the configurable port? Often, the implemetations allow to specify custom port for servers, like NETCONF, gRPC and etc. In addition, I believe it would be nice to have an option for target-type being custom string. This allows to use grpc-tunnel also for custom services. for instance, telnet is not specified in this proposal....
> 
> thanks! miro

@mirovr - For your first point, this was a consideration but the issue turns into a bit of unnecessary duplication of specifying the target port (both at the listening TCP service and at this tunnel mapping layer) and maintaining synchronization between the two.  In reality the listening service port specification is good enough and the implementation can maintain the dynamic mapping.  If a user updates say SSH to listen on tcp/2222 vs. tcp/22 within the SSH service configuration, this would be good enough that the tunnel target parameters need not be updated to match.  Essentially this is abstracted away from user-complexity...

For the second - this is a good point since the `.proto` is a _very_ loose API from this standpoint.  The `TargetType` enum is more or less a placeholder for what *should* be populated into a string field (https://github.com/openconfig/grpctunnel/blob/master/proto/tunnel/tunnel.proto#L79-L82).  For this, I think a middle ground approach would be to keep the "well defined" enum variants as they are but make this a union with a string specifying that the string value be loose and negotiated out of band - what do you think?

mirovr:
@earies - thanks for your response. I understand the reasoning and it seems OK to me.
thanks!

miro

release/models/system/openconfig-system-grpc-tunnel.yang
```diff
@@ -0,0 +1,348 @@
...
+    "OpenConfig working group";
+  contact
+    "www.openconfig.net";
+
+  description
+    "This module adds configuration and operational state relating to gRPC
+    tunnel (https://github.com/openconfig/grpctunnel/) to invoke one or
+    more dialout TCP over gRPC tunnels to bypass situations in which
+    direct dialin gRPC communications are not feasible.";
+
```
robshakir, line 22: (nit) extra blank line.

release/models/system/openconfig-system-grpc-tunnel.yang
```diff
@@ -0,0 +1,348 @@
...
+          "gRIBI target type";
+      }
+      enum P4RT {
+        value 9559;
+        description
+          "P4 Runtime target type";
+      }
+    }
+    description
+      "Sets the TCP based protocols to be tunneled over the gRPC tunnel";
```
robshakir, line 74: (grammar nit) can we add periods to the end of these description sentences please (throughout the PR).

release/models/system/openconfig-system-grpc-tunnel.yang
```diff
@@ -0,0 +1,348 @@
...
+        "The IPv4/IPv6 address or hostname of the tunnel server";
+    }
+
+    leaf port {
+      type oc-inet:port-number;
+      description
+        "The port number of the tunnel server";
+    }
+
+    leaf transport-security {
```
robshakir, line 110: We've discussed this a bit in the past -- should we ever allow TLS to be turned off for this application? The 'lab testing' thing is "convenient" but it's also relatively easy to just generate self-signed certificates.

Do we ever expect that someone would run this without TLS in production?

release/models/system/openconfig-system-grpc-tunnel.yang
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+    leaf certificate-id {
+      type string;
+      description
+        "Name of the certificate that is associated with client (e.g.
+        mutual authentication).  The certificate ID is provisioned
+        through other interfaces to the device, such as the gNOI
+        certificate management service.  If this leaf is provisioned and
+        transport-security is enabled, mutual authentication is implied.
```
robshakir, line 126: Could you expand on the choice to implicitly turn on mutual authentication here? It seems like we might want to have this toggled individually -- i.e., I may use some other form of authentication for the session and not mutual cert auth whilst still wanting TLS.

release/models/system/openconfig-system-grpc-tunnel.yang
```diff
@@ -0,0 +1,348 @@
...
+    leaf dscp {
+      type oc-inet:dscp;
+      description
+        "DSCP marking of packets generated by the gRPC tunnel client";
+    }
+
+    leaf-list target {
+      type union {
+        type target-type;
+        type string;
```
robshakir, line 156: If there is a string here, what does it mean? In the cases above it seems to imply the TCP ports that will be plumbed through -- in the case where there is just a string wouldn't it be better to have `target-type` rather be an identity and then expand on this with the protocols that are not defined in the base model?

release/models/system/openconfig-system-grpc-tunnel.yang
```diff
@@ -0,0 +1,348 @@
...
+      }
+    }
+  }
+
+  grouping target-options-config {
+    description
+      "Grouping for gRPC tunnel client target options configuration and
+      operational state";
+
+    leaf-list pattern {
```
robshakir, line 213: I may be missing something here, but I'm not clear what pattern this refers to. I didn't find any reference to these options (and what these various values mean) in the `tunnel.proto`. Is there some extension that's being used here, or something that I forgot about? thanks.

robshakir (changes requested):
@gcsl @jxx-gg could you PTAL at this model for configuring the device side of a gRPC tunnel connection? thanks!

release/models/system/openconfig-system-grpc-tunnel.yang
earies: Addressed in commit https://github.com/openconfig/public/pull/666/commits/fae9ac3b3a62b798acba53272400175c2e927690

release/models/system/openconfig-system-grpc-tunnel.yang
earies: Addressed in commit https://github.com/openconfig/public/pull/666/commits/fae9ac3b3a62b798acba53272400175c2e927690

release/models/system/openconfig-system-grpc-tunnel.yang
earies: This actually presents a unique case where there could be more justification that just lab/testing in that an operator could choose to disable TLS on the tunnel but enforce for the grpc-server connection (or vice-versa) if you only want a single layer of TLS

release/models/system/openconfig-system-grpc-tunnel.yang
earies: This actually accommodates @mirovr's query which is valid in that grpc-tunnel is meant to tunnel TCP applications however per the grpc-tunnel specification there is only a small number of TCP applications defined (what is brought over to here into the typedef `target-type`) - in theory you could target any TCP application that may or may not be worthy of being defined in OpenConfig modules as enums or identities.

We could move this to an identityref and add identities (instead of enum variants) each time but this is churn that may not scale and would not accomodate any non-vendor neutral TCP applications.

Open to suggestions - this gives a looser option for users where the strict enum/identity method is precise and limiting

release/models/system/openconfig-system-grpc-tunnel.yang
earies: I would put this in the bucket of a helpful addition for abstraction to the spec.

The `target-options` here are to give a customizable and variable based approach to populating the `target` string field in the protobuf message - a unique value to distinguish the tunnel connection

https://github.com/openconfig/grpctunnel/blob/master/proto/tunnel/tunnel.proto#L78

`target-options` brings in a `pattern`, `custom-string` and `delimeter`

The `pattern` defines an ordered leaf-list of of variables that can be derived from the system or user-defined to pack into the string.  For instance if `[ HOSTNAME VENDOR CUSTOM ]` are defined then the client would populate a `custom-string` of say `ID=2022` then protobuf `target` string might be constructed as `pr1.sea1:VendorA:ID=2022`

If the user populates the `delimeter` field (e.g. `\\` then it can override the implementations default delimeter) to produce a string such as `pr1.sea1\\VendorA\\ID=2022`

This also accounts for the simpler case where the user _always_ wants to define the `target` string statically thus the configuration would consist of `pattern = [ CUSTOM ]` with a `custom-string = "Whatever you wish goes here"`

release/models/system/openconfig-system-grpc-tunnel.yang
```diff
@@ -0,0 +1,348 @@
...
+    }
+    description
+      "Sets the TCP based protocols to be tunneled over the gRPC
+      tunnel.";
+    reference
+      "OpenConfig gRPC Tunnel:
+      https://github.com/openconfig/grpctunnel/blob/master/proto/tunnel/tunnel.proto";
+  }
+
+  grouping tunnel-server-config {
```
gcsl, line 80: This seems to be a parallel definition of what was created in proto, here.
https://github.com/openconfig/grpctunnel/blob/master/cmd/target/proto/config/target_config.proto

Do we want to have per-leaf, redudant configuration definitions between proto and YANG?

release/models/system/openconfig-system-grpc-tunnel.yang
gcsl: I'm reluctant to add this option as it might inadvertently send a signal that encryption is reasonable to disable.  I realize there is an overhead associated but that seems a worthwhile security requirement to me.

earies: Currently, this toggle exists for each `grpc-server` as well - default to `true` with the same blurb

If this should be removed to not promote disabling transport security whatsoever, I would then suggest that the grpctunnel specification mandate encryption (as well as network-element based gRPC service endpoints for gNMI, gRIBI, gNOI, etc..  if it will be removed for grpc-servers as well)

release/models/system/openconfig-system-grpc-tunnel.yang
earies: Essentially it provides the flexibility for both - an on-box dynamic generator based on runtime parameters or the ability to insert a completely user-defined string (or a mix of both).

Reducing this down to a string and taking away the complexity is also fine (might still be worthwhile to define supported characters, max string length, etc.. otherwise we'll see divergence among implementors)

@gwizdms - any additional comments here?

PR #667, opened 2022-07-11
Switchover ready

rolandphung:
  - Add boolean switchover-ready leaf

Add an indication for when redundant components are ready to failover.

Relevant pyang output:
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--ro state
           +--ro switchover-ready?   boolean
```

References:
* Arista `show redundancy status`
  * https://www.arista.com/en/um-eos/eos-supervisor-redundancy
* Juniper `show system switchover`
  * https://www.juniper.net/documentation/us/en/software/junos/high-availability/topics/ref/command/show-system-switchover.html
* Cisco `show redundancy states`
  * https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst9400/software/release/16-9/command_reference/b_169_9400_cr/high_availability_commands.html#reference_t1d_ppp_dgb

### DIFF
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,7 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.17.0";
+  oc-ext:openconfig-version "0.18.0";
+
+  revision "2022-07-11" {
+    description
+      "Add switchover ready";
+    reference "0.18.0";
+  }
 
   revision "2022-06-10" {
     description
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -515,7 +515,7 @@ module openconfig-platform {
       type boolean;
       description
         "For components that have redundant roles, this reports a value
-        of that indicates if the secondary component is ready to
+        that indicates if the secondary component is ready to
         failover in the case of component failure.";
     }
   }
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -515,12 +515,12 @@ module openconfig-platform {
       type boolean;
       description
         "For components that have redundant roles, this reports a value
-        that indicates if the secondary component is ready to
-        failover in the case of component failure.
+        that indicates if the component is ready to support failover.
 
-        The components with a redundant-role of primary should reflect
-        the overall system's switchover status. For example, two
-        supervisors in a device should both report the same value.";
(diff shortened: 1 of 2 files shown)


release/models/platform/openconfig-platform.yang
```diff
@@ -504,6 +510,14 @@ module openconfig-platform {
         value is the timestamp in nanoseconds relative to the Unix Epoch
         (Jan 1, 1970 00:00:00 UTC).";
     }
+
+    leaf switchover-ready {
+      type boolean;
+      description
+        "For components that have redundant roles, this reports a value
+        of that indicates if the secondary component is ready to
```
dplore, line 518:
```suggestion
       that indicates if the secondary component is ready to
```

dplore (changes requested):
Small nit on the description text

dplore (approved):

dplore:
@earies @rgwilton for any comments

aredmon8551:
Clarification - should this RPC be available on both supervisors/routing-engines, or only the backup/redundant one?

dplore:
IMO, primary and secondary should both support the leaf.  

It's also good to get a consistent answer whether querying the primary or secondary CONTROLLER_CARD (to use OC language). It seems the primary should be 'ready' if the secondary is also 'ready'.    Does that make sense?  We can add description to clarify this.

rolandphung:
Agreed. I think it makes more sense in a scenario where there are more than 2 redundant components. The primary stating that it is "ready" is signaling that at least one component is ready to failover.

Let me see how to word this into the description.

release/models/platform/openconfig-platform.yang
```diff
@@ -504,6 +510,18 @@ module openconfig-platform {
...
+    leaf switchover-ready {
+      type boolean;
+      description
+        "For components that have redundant roles, this reports a value
+        that indicates if the secondary component is ready to
+        failover in the case of component failure.
+
+        The components with a redundant-role of primary should reflect
+        the overall system's switchover status. For example, two
+        supervisors in a device should both report the same value.";
```
dplore, line 523:
```suggestion
        "For components that have redundant roles, this reports a value
        that indicates if the component is ready to support failover.

        The components with a redundant-role  should reflect the overall 
        system's switchover status.  For example, two supervisors in a 
        device, one as primary and the other as secondary, should both 
        report the same value.";
```

dplore (changes requested):
Suggested clarification

dplore (approved):

dplore:
Holding until July 22 (2 weeks after PR submitted) for any additional community feedback

robshakir (approved):

PR #672, opened 2022-07-12
Modify ntp enabled description to reflect true and false case

mlaouane:


dplore (approved):
Thanks for the contribution!

PR #673, opened 2022-07-18
Correct BFD diagnostic codes per RFC5880/6428

earies:
* (M) release/models/bfd/openconfig-bfd.yang
    - Add missing codes and adjust value mappings according to RFC

Previously, the enum variants for BFD diagnostic codes missed value (3) per
RFC5880.  This PR addresses the missing diagnostic code, correcting the
subsequent variant values and adding the reserved value (9) per RFC6428.

### DIFF
--- a/release/models/bfd/openconfig-bfd.yang
+++ b/release/models/bfd/openconfig-bfd.yang
@@ -26,7 +26,13 @@ module openconfig-bfd {
     "An OpenConfig model of Bi-Directional Forwarding Detection (BFD)
     configuration and operational state.";
 
-  oc-ext:openconfig-version "0.2.4";
+  oc-ext:openconfig-version "0.3.0";
+
+  revision "2022-07-18" {
+    description
+      "Correct diagnostic codes per RFC5880/6428";
+    reference "0.3.0";
+  }
 
   revision "2022-06-28" {
     description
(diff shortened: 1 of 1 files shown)


release/models/bfd/openconfig-bfd.yang
```diff
@@ -138,48 +144,61 @@ module openconfig-bfd {
           "The BFD echo function failed - echo packets have not been
           received for the required period of time.";
       }
-      enum FORWARDING_RESET {
+      enum NEIGHBOR_SESSION_DOWN {
         value 3;
+        description
+          "The neighbor has signaled the session down.";
+      }
```
robshakir, line 151: Can you please add a blank like between each of these `enum` values? Not sure how we missed this originally. Thanks.

release/models/bfd/openconfig-bfd.yang
```diff
@@ -138,48 +144,61 @@ module openconfig-bfd {
...
+      enum FORWARDING_RESET {
+        value 4;
         description
           "The forwarding plane in the local system was reset - such
           that the remote system cannot rely on the forwarding state of
           the device specifying this error code.";
       }
       enum PATH_DOWN {
-        value 4;
```
robshakir, line 149: Per the other change, do these values correspond to the spec? Is it better for us to just remove them all?

robshakir (changes requested):

release/models/bfd/openconfig-bfd.yang
earies: These values do now correspond to the spec but yes - we could just remove them all and state the design pattern is to never assign variant values but pay very close attention to ordering if auto assigned values do infact have side effects

https://datatracker.ietf.org/doc/html/rfc5880#section-4.1
https://datatracker.ietf.org/doc/html/rfc6428#section-5

release/models/bfd/openconfig-bfd.yang
earies: Between each enum variant definition or between the value and description?

Looking at all other models, there are no additional lines between either ... Do you have an example in any models?

robshakir: Ah, apologies, yes - please ignore this. I had misparsed these as identities.

dplore:
I am ok with the changes, but as this is technically a yang breaking change due to the enum value changes, should the major version be incremented?  

As long as we specify all or no values for the enums I think we are ok.  (ie: we should not introduce enums where some values are specified and other are not).

PR #676, opened 2022-07-25
Add software-version leaf to `system/state`

### LINKED ISSUE #675: Vendor-neutral software version leaf
I am interested in adding an OpenConfig leaf that contains software version pertaining to the primary supervisor/routing engine of a router.  This is useful because it will allow a vendor-neutral way to query this commonly desired property.  

As of now, a `software-version` leaf only exists under `components`, but component names and/or the tree structure may vary by vendor, meaning one has to iterate over all components to find one that is of `OPERATING_SYSTEM` type to identify the desired `software-version` leaf.  A dedicated leaf would make this much easier.

What do you think?  Would `openconfig/system/state/software-version` make sense for the leaf location and name?

tjburrows:
Add openconfig/system/state/software-version 

Fixes #675

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -46,7 +46,13 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.14.0";
+  oc-ext:openconfig-version "0.15.0";
+
+  revision "2022-07-25" {
+    description
+      "Add system software version.";
+    reference "0.15.0";
+  }
 
   revision "2022-07-20" {
     description
@@ -346,6 +352,15 @@ module openconfig-system {
           "This timestamp indicates the time that the system was last
           restarted.  The value is the timestamp in nanoseconds relative
           to the Unix Epoch (Jan 1, 1970 00:00:00 UTC).";
+    }
+
+    leaf software-version {
+        type string;
+        description
+          "Operating system version of the currently active controller
+          of the device.  It is required that this value matches the
+          value of the state/software-version leaf in the component
+          of type OPERATING_SYSTEM.";
     }
   }
 


release/models/system/openconfig-system.yang
```diff
@@ -341,6 +347,12 @@ module openconfig-system {
           restarted.  The value is the timestamp in nanoseconds relative
           to the Unix Epoch (Jan 1, 1970 00:00:00 UTC).";
     }
+
+    leaf software-version {
+        type string;
+        description
+          "Operating system version of the primary routing engine.";
```
robshakir, line 354: i think this should not be the "primary" routing engine, but maybe "the currently active controller of the device" since this is more generic. if this were to describe a linux device it would be whatever device is currently responding to the gNMI agent afaict.

note that there's some discussion to have here that i'd like to prompt: this represents a simplification of the user querying this data, at the cost of denormalising -- today one needs to walk the set of components to find the OS that corresponds to the currently active routing engine, which requires more query logic. we should consider as a wider group whether this is the right trade-off to be making.

dplore: I've added this to the openconfig operators group agenda for comment there.  Also appreciate public comments as Rob indicated.

tjburrows: Thanks for reviewing this.  I am suggesting adding this as a second option for describing the software version, which is higher level / abstracted away from the components, so one does not have to be component-specific if they want to know this information.

cfyo: Why is just software-version being requested here, is there a use case?  If it makes sense to include software-version wouldn't you also want os-name and/or manufacturer unless that is expected to be part of the software-version string?

robshakir: #675 covers some of the discussion here. As discussed, this is denormalising the models -- since one should be able to get this from the components that are of type `OPERATING_SYSTEM` within `/components/component`.

The challenge from a consumer perspective is that having this stored there means you need to walk the list of the components to find the component, and then subsequently read the version from it. This would have the additional information.

The question for the group here is whether we feel that the use case of "what version of the OS is this device running" justifies the denormalisation of the model.

dplore: OC Community has a suggestion to require this leaf value should be the exact same as component type=`OPERATING_SYSTEM`.

tjburrows: I have updated the description to include this requirement, as well as the phrase "currently active controller of the device".

dplore: Thanks! Looks like you also need to fast forward / resync your fork to master to resolve a conflict.

tjburrows: I fast forwarded, and merged my three commits into one.

release/models/system/openconfig-system.yang
```diff
@@ -46,7 +46,13 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.14.0";
+  oc-ext:openconfig-version "0.14.1";
+
+  revision "2022-07-25" {
+    description
+      "Add system software version.";
+    reference "0.14.1";
```
dplore, line 54:
```suggestion
    reference "0.15.0";
```

Since we are adding a leaf (vs. fixing a minor error/omission) we increment the minor version.

dplore (changes requested):
Just one minor change to the semantic version, otherwise looks good to me.  @earies I believe you had a comment on this which would be good to share here

release/models/system/openconfig-system.yang
tjburrows: I updated the reference version on this line, as well as the `oc-ext:openconfig-version` to both say `0.15.0`

dplore (approved):
This LGTM,  I will represent for you on the Sep 13 OC operator's meeting.  If you would like to present, I can forward the invite to you!

PR #677, opened 2022-07-26
Add File Synchronization for redundancy

rolandphung:
* (A) release/models/system/openconfig-system-redundancy.yang
  * Add OpenConfig model for file synchronization
* (M) release/models/system/openconfig-system.yang

This proposal introduces a set of models which represent files that synchronize between two controller cards or supervisors.  Synced files can be specified as arbitrary files from within the system.

pyang output:
```
module: openconfig-system                                                      
  +--rw system
     +--rw redundancy
        +--rw file-sync
           +--rw files
              +--rw file* [path]
                 +--rw path      -> ../config/path
                 +--rw config
                 |  +--rw path?   string
                 +--ro state
                    +--ro path?             string
                    +--ro status?           enumeration
                    +--ro last-sync-time?   oc-types:timeticks64
```

Reference:
- Arista
  - `show redundancy file-replication`
    - https://www.arista.com/en/um-eos/eos-switch-booting-commands#xx1122125
  - `file-replication` configuration example
```
redundancy
   file replication
      path flash:/random-file
      path flash:/another-file
```
- Cisco `rsync`
  - https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/system-management/73x/b-system-management-cg-8k-73x/m-synch-file-systems.html?dtid=osscdc000283#Cisco_Concept.dita_2134690f-1fa9-44a9-8ccb-b62836abaeff
- Juniper `commit synchronize`
  - https://www.juniper.net/documentation/us/en/software/junos/cli/topics/topic-map/synchronising-routing-engines.html

### DIFF
--- a/release/models/system/openconfig-system-redundancy.yang
+++ b/release/models/system/openconfig-system-redundancy.yang
@@ -58,17 +58,17 @@ module openconfig-system-redundancy {
 
     leaf status {
       type enumeration {
-        enum unsynchronized {
+        enum UNSYNCHRONIZED {
           description
             "The file is being synchronized or the synchronization has yet to be started.";
         }
 
-        enum synchronized {
+        enum SYNCHRONIZED {
           description
             "The file was successfully synchronization.";
         }
 
-        enum failed {
+        enum FAILED {
           description
             "The file synchronization has failed.";
         }
--- a/release/models/system/openconfig-system-redundancy.yang
+++ b/release/models/system/openconfig-system-redundancy.yang
@@ -1,0 +1,156 @@
+module openconfig-system-redundancy {
+
+  yang-version "1";
+
+  // namespace
+  namespace "http://openconfig.net/yang/system/redundancy";
+
+  prefix "oc-sys-red";
+
+  // import some basic types
+  import openconfig-extensions { prefix oc-ext; }
+  import openconfig-types { prefix oc-types; }
+
+  // meta
+  organization "OpenConfig working group";
+
+  contact
+    "OpenConfig working group
+    www.openconfig.net";
+
+  description
+    "This module defines the configuration and operational state data
+    related to redundant systems, such as two supervisors in a device.";
+
+  oc-ext:openconfig-version "0.1.0";
+
+  revision "2022-07-20" {
+    description
+      "Initial revision with file sync";
+    reference "0.1.0";
+  }
+
+  // extension statements
+
+  // feature statements
+
+  // identity statements
+
+  // typedef statements
+
+  // grouping statements
+
+  grouping redundancy-file-sync-config {
+    description
+      "Configuration grouping for file synchronization.";
+
+    leaf path {
+      type string;
+      description
+        "The system file path of the file to be synched to another
+        redundant system.";
+    }
+  }
+
+  grouping redundancy-file-sync-state {
+    description
+      "Operational grouping for file synchronization.";
+
+    leaf status {
+      type enumeration {
+        enum unsynchronized {
+          description
+            "The file is being synchronized or the synchronization has yet to be started.";
+        }
+
+        enum synchronized {
+          description
+            "The file was successfully synchronization.";
+        }
+
+        enum failed {
+          description
+            "The file synchronization has failed.";
+        }
+      }
+      default unsynchronized;
+      description
+        "The enum reports the synchronization status of the file.";
+    }
+
+    leaf last-sync-time {
+      type oc-types:timeticks64;
+      units "nanoseconds";
+      description
+        "This reports the time of the last successful synchronization of the file.
+        The value is the timestamp in nanoseconds relative to the Unix Epoch
+        (Jan 1, 1970 00:00:00 UTC).";
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


aredmon8551:
@rolandphung - can you elaborate on what files are candidates for synchronization here?  Is it limited to configuration files, or would it include scripts, arbitrary files, etc. as well?  The documentation seems to indicate this is limited to configuration files.
https://www.arista.com/en/um-eos/eos-supervisor-redundancy?searchword=eos%20routing%20information%20protocol

rolandphung:
> @rolandphung - can you elaborate on what files are candidates for synchronization here? Is it limited to configuration files, or would it include scripts, arbitrary files, etc. as well? The documentation seems to indicate this is limited to configuration files. https://www.arista.com/en/um-eos/eos-supervisor-redundancy?searchword=eos%20routing%20information%20protocol

This feature is not limited to configuration files but scripts and arbitrary files as well. The configuration in Arista switches would look as follow:
```
redundancy
   file replication
      path flash:/random-file
      path flash:/another-file
```

I've updated the original comment to reflect this.

dplore:
@rgwilton @hellt for comment

release/models/system/openconfig-system-redundancy.yang
```diff
@@ -0,0 +1,156 @@
...
+    container file-sync {
+      description
+        "Top-level container for file synchronization.";
+
+      container files {
+        description
+          "This contains the list of files to synchronize between
+          multiple systems.";
+
+        list file {
```
robshakir, line 104: i wonder whether this is the right way to model this -- it allows arbitrary specification of synchronisation, yes, but it means that the payload is actually not vendor neutral. it's going to be specific to the target that you have.

if we wanted to think about this as a more vendor neutral model, then i think it would be preferable to consider the different types of files that we would want to synchronise, e.g.,:

 * configuration files
 * os images (although i am not sure that this should be something that isn't done through os.Install)

is there an operational use case that is being considered here?

i'd like to get other operator's feedback on this as well.

robshakir (changes requested):

release/models/system/openconfig-system-redundancy.yang
rgwilton: Understanding the use case would be interesting here.  I would mostly expect devices to abstract this away.  I.e., the network software would take care of synchronizing appropriate files between the active and standby RPs.

aredmon8551: Agreed @rgwilton - some implementations implicitly perform this behavior, or rely on a commit or RPC to trigger it - for instance Junos EVO configurations are synchronized between RE/RP's by default (when commits are issued on the primary RE).  Scripts, arbitrary files (e.g., a MD5 hash file, a config snippet, or patch), and software images are treated differently and will likely vary across implementations in their treatment - not actually aware of what IOS-XR or SR-Linux do here.

rolandphung: > is there an operational use case that is being considered here?

I think there's two parts here.

The first is providing telemetry of the system important files that must be synchronize in order to function after a failover occurs (configuration, scripts, private keys). Agreeing with @rgwilton that the software should be taking care of this by providing the list of files, but in this case the software should provide some feedback if something goes wrong, this being one way to do so. The only thing I can think of abstracting out the possible content would be the main configuration file and private keys.

The second is allowing the user to add their own files/scripts to this list for their own custom features to function though failover without first party support of said feature.

> os images (although i am not sure that this should be something that isn't done through os.Install)

I also disagree that os images should be synchronized (by default), generally that should be separated in order to perform hitless upgrades. Otherwise extra additional steps are needed to disable sync and perform the upgrade and re-enable sync.

robshakir: I think this is not so much about what we think is the philosophically best implementation, but what is supported.

We discussed this across the operators group yesterday, and there were a few observations (cc: @dplore @chrisy @marcushines):
 
* the `rsync` functionality linked appears to be a one-off command rather than a process that continuously synchronises files.
* the Arista docs link appears to be to a show command, is this configuration supported today?

I am not really in favour of having this be around the filesystem paths for fundamentals of the system, if we agree that this configuration is required (which today it does not seem to be agreed), I would like to see an approach that abstracts this away from implementation-specific paths and rather lets an implementation choose how to synchronise things like "current configuration".

If there is a use case and support for other files being synchronised (@rgwilton @aredmon8551, it'd be good to hear about what the implementations you're familiar with support), then I think we could add this as an extension to these 'common' types.

rolandphung: > the Arista docs link appears to be to a show command, is this configuration supported today?

Yes, though the doc is lacking (I have requested official documentation for it), there is configuration that is possible to be added per file. The example would be as follow:
```
Switch(config)# redundancy
Switch(config-redundancy)# file replication
Switch(config-redundancy-repl)# <file path A>
Switch(config-redundancy-repl)# <file path B>
```
File path may include a directory in which it would recursively sync the files underneath. It's also worthwhile to note that the a trigger for re-sync is whenever a file/folder was modified.

PR #681, opened 2022-07-30
Added last-calculated-sr-traffic leaf

mryuezhang:


release/models/segment-routing/openconfig-rsvp-sr-ext.yang
```diff
@@ -379,9 +385,17 @@ module openconfig-rsvp-sr-ext {
...
+        utilisation and segment routing traffic was performed. 
+        The value is expressed relative to the Unix Epoch 
+        (Jan 1, 1970 00:00:00 UTC).";
+    }
+    
+    leaf last-calculated-sr-traffic {
+      type uint64;
+      units "kbps";
+      description
+        "The last calculated value of the Segment Routing traffic.";
```
dplore, line 398: Is this the same as "last-calculated-sr-utilisation" except that it is specifying the units?    

Should last-calculated-sr-utilisation instead be updated to specify units "kbps".  (Apologies if I am missing reference to segment routing)

dplore (commented):
Thanks for the pull request.  Please see the CONTRIBUTING.md file for instructions on how to get the CLA signed and for requests for at least two references to vendor implementation of the change you're proposing.

release/models/segment-routing/openconfig-rsvp-sr-ext.yang
robshakir: Yes, it looks like it -- I think the original issue raised here was an artifact of the fact that the name of this leaf got updated in review, and we didn't catch it.

One observation - although it pains me - is that this is the British English spelling of utilisation, I think the standard across the models today is American English, so we maybe should update the leaf name.

PR #682, opened 2022-08-01
Define default behavior for user creation when no role is defined

aredmon8551:
* (M) release/models/system/system-aaa.yang
* (M) release/models/system/system-aaa-radius.yang
* (M) release/models/system/system-aaa-tacacs.yang

 - Update role to be mandatory attribute of user configuration

This is to address the undefined behavior when a system user is created but no role is specified.  

1. Junos/EVO require a role and do not allow a user to be created without one.
2. EOS allows the creation of a user without specifying a role.  The default role is 'network-operator',
with a privilege level of 1. (https://www.arista.com/en/um-eos/eos-user-security#xx1347271)
3. Can't find documentation for IOS-XR, but NX-OS behaves in a similiar manner to EOS and assigns a 
default role of network-operator to a user if a role isn't provided at time of creation (role of 'operator').

###Arista EOS###
eos-oc1#show run | in username test
eos-oc1(config)#username test nopassword 
eos-oc1#show run | in  username test
username test nopassword
eos-oc1#show users accounts 
user: admin
       role: network-admin
       privilege level: 15
user: test
       role: network-operator
       privilege level: 1

###Cisco NX-OS###
D-SW1# show run | in "username test"
D-SW1# conf t
Enter configuration commands, one per line.  End with CNTL/Z.
D-SW1(config)# username test
warning: password for user:test not set. S/he may not be able to login
D-SW1(config)# exit
D-SW1# show run | in "username test"
username test password 5 !  role network-operator
D-SW1# show user-account test
user:test
        this user account has no expiry date
        roles:network-operator

### DIFF
--- a/release/models/system/openconfig-aaa-types.yang
+++ b/release/models/system/openconfig-aaa-types.yang
@@ -71,8 +71,7 @@ module openconfig-aaa-types {
   identity SYSTEM_DEFINED_ROLES {
     description
       "Base identity for system_defined roles that can be assigned
-      to users.  Default of SYSTEM_ROLE_READ_ONLY is used if role
-      is not specified during user creation.";
+      to users.";
   }
 
   identity SYSTEM_ROLE_ADMIN {
--- a/release/models/system/openconfig-aaa-types.yang
+++ b/release/models/system/openconfig-aaa-types.yang
@@ -22,7 +22,14 @@ module openconfig-aaa-types {
     "This module defines shared types for data related to AAA
     (authentication, authorization, accounting).";
 
-  oc-ext:openconfig-version "0.4.1";
+  oc-ext:openconfig-version "0.5.0";
+
+  revision "2022-07-29" {
+    description
+      "Add two new SYSTEM_DEFINED_ROLES: SYSTEM_ROLE_OPERATOR
+      and SYSTEM_ROLE_READ_ONLY.";
+    reference "0.5.0";
+  }
 
   revision "2018-11-21" {
     description
@@ -71,7 +78,8 @@ module openconfig-aaa-types {
   identity SYSTEM_DEFINED_ROLES {
     description
       "Base identity for system_defined roles that can be assigned
-      to users.";
+      to users.  Default of SYSTEM_ROLE_READ_ONLY is used if role
+      is not specified during user creation.";
   }
 
   identity SYSTEM_ROLE_ADMIN {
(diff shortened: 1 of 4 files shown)


release/models/system/openconfig-aaa-types.yang
```diff
@@ -82,6 +90,22 @@ module openconfig-aaa-types {
       on the device.";
   }
 
+  identity SYSTEM_ROLE_OPERATOR {
+    base SYSTEM_DEFINED_ROLES;
+    description
+      "Built-in role that allows the equivalent of network-operator
+      permission for all read-only and operational commands
+      on the device.";
```
dplore, line 98: Do we require this role to be a system defined role?   FWIW, Access to operational commands is expected to be covered with [gNSI/authz](https://github.com/openconfig/gnsi/tree/main/authz)

release/models/system/openconfig-aaa-types.yang
```diff
@@ -82,6 +90,22 @@ module openconfig-aaa-types {
...
+  }
+
+  identity SYSTEM_ROLE_READ_ONLY {
+    base SYSTEM_DEFINED_ROLES;
+    description
+      "Built-in role that allows for the execution of read-only
+      commands on the device.  Lacks permissions to operational
+      or configuration commands.";
```
dplore, line 106:
```suggestion
      "Built-in role that allows for the access to state paths.  
      Lacks permissions to operational (gNOI) and lacks
      write access to config paths.";
```

release/models/system/openconfig-aaa-types.yang
aredmon8551: @dplore yes, it would require an equivalent role in each implementation - looking through the lone vendor I didn't investigate, it looks like SR-Linux may not have any default roles beyond that of 'admin'.
https://infocenter.nokia.com/public/SRLINUX213R1A/index.jsp?topic=%2Fcom.srlinux.configbasics%2Fhtml%2Fconfigb-security.html


"4.2.2. Local users
Local users are users configured within the SR Linux itself. By default, SR Linux supports a single local user, named admin; other local users can be added as necessary."

The alternative may be to mandate providing a role upon user creation to avoid undefined behavior.

dplore: I like making user role mandatory when creating a user to avoid undefined behavior.

Regarding system defined roles, to keep things minimalistic I think we should stick with only one system defined role,  that role being "SYSTEM_ROLE_ADMIN".      With this in mind can you remove the additional two roles?

earies: @dplore Are you then suggesting the only change required then is to tag the `role` leaf as `mandatory true;` then ?

Since this is a union with a string, we're likely to then see usage of each implementations system defined role/group/classes and/or user-defined that must still be done via alternate provisioning (e.g. native schema or side channel) but then at least enforcing any user defined must carry with it a role of some kind.

dplore: Yes, I believe we should only tag the `role` leaf as `mandatory true;`

I am open to other feedback if this causes some issue.

I don't see a good solution around string at this moment.  Users should be able to define their own custom names for roles.  In addition, a NOS/OpenConfig device could implement additional roles, even by default.   They are just not mandatory.

robshakir: Note that whilst some implementations will use gNSI/authz, I do not believe that this is going to be the only user specification that we need to support in the OpenConfig models.

That note aside, I do tend to agree that making the role mandatory here is a good move -- the action of creating a user without a role is somewhat implicit behaviour.

For these roles, I don't think we have somewhere in the model that they are defined, this is likely something that does need some thought for non-gNSI users.

aredmon8551:
Moved to https://github.com/openconfig/public/pull/687

PR #686, opened 2022-08-09
Adding control over sending of IS-IS CSNP packets on P2P interfaces

adamsimpson1:
Adding a single new leaf to control (optionally disable) the periodic transmission of CSNP PDUs on point-to-point interfaces


dplore:
Thanks for the contribution, before we can review and merge, please do take care of the CLA and we can proceed.

PR #691, opened 2022-08-17
Clarify next-hop ip-address description

Ankur19:
This change is because of a recent regression found in hierarchical route installation test.

As per the latest expectation, for recursive next-hop-groups, AFT should return the flattened next-hop instead of the direct next-hop.

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,13 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "2.2.0";
+  oc-ext:openconfig-version "2.2.1";
+
+  revision "2022-08-17" {
+    description
+      "Clarify next-hop ip-address description.";
+    reference "2.2.1";
+  }
 
   revision "2022-06-16" {
     description
@@ -289,7 +295,9 @@ submodule openconfig-aft-common {
     leaf ip-address {
       type oc-inet:ip-address;
       description
-        "The IP address of the next-hop system.";
+        "The IP address of the next-hop system. This leaf represents the
+        ip-address of the flattened next-hop representation in case of
+        recursion or indirection at the next-hop-group level";
     }
 
     leaf mac-address {
--- a/release/models/aft/openconfig-aft-ethernet.yang
+++ b/release/models/aft/openconfig-aft-ethernet.yang
@@ -20,7 +20,13 @@ submodule openconfig-aft-ethernet {
     "Submodule containing definitions of groupings for the abstract
     forwarding tables for Ethernet.";
 
-  oc-ext:openconfig-version "2.2.0";
+  oc-ext:openconfig-version "2.2.1";
+
+  revision "2022-08-17" {
+    description
+      "Clarify next-hop ip-address description.";
+    reference "2.2.1";
+  }
 
   revision "2022-06-16" {
     description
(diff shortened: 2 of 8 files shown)


release/models/aft/openconfig-aft-common.yang
```diff
@@ -289,7 +295,9 @@ submodule openconfig-aft-common {
     leaf ip-address {
       type oc-inet:ip-address;
       description
-        "The IP address of the next-hop system.";
+        "The IP address of the next-hop system. This leaf represents the
+        ip-address of the flattened next-hop representation in case of
+        recursion or indirection at the next-hop-group level";
```
robshakir, line 300: I'm not sure I understand this clarification in description. But let's start with what we're trying to achieve. 

In the case that we have recursion, e.g.,:

```
2001:db8:1::/48 -> 2001:db8:cafe::1
2001:db8:cafe::/48 -> 2001:db8:beef::1
```

I expect that the AFT entry for `2001:db8:1::/48` is fine to have `2001:db8:cafe::1` as a next-hop, and need the client to do recursion onto another route to determine how to do this, rather than needing to "flatten" to have `2001:db8:beef::1`.

If we do this flattening then there is a necessity update significantly more entries when the next-hop for `2001:db8:cafe::/48` changes - since now all resulting routes need to change. This will have significant scaling implications in systems with many routes and few next-hops.

The general aim has been to avoid needing recursion of NHGs -- since this got quickly into how different implementations choose to group NHGs for re-use in specific hardware. I think (from our conversation) we are saying that we don't want to "flatten" something here - we just want to expand NHGs to represent all constituent NHs that are within them with the relevant weight. Thus, it seems like we probably want some description here in the NHG part of the model rather than here.

akalluru1: @robshakir As per our discussions AFT Telemetry/export would represent final flattened/resolved information, similar to the data present in hardware, used by the controller. Controller doesn't need the indirect next-hops. In this context @Ankur19 thought it would be helpful to update the comment.

I agree with Rob that we may not need to update the comment here, in its current form, since the same YANG file would be used to generate gribi protofiles too & could give a wrong impression that always program resolved information.

@robshakir what do you think of following text:

```The IP address of the next-hop system. In telemetry/export this leaf
      represents the ip-address of the flattened next-hop representation
      in case of recursion.```

robshakir: @akalluru1 I'm not sure that flattening recursion is something that we've discussed; and indeed when we iterated to this AFT model vs. a more direct one (prefix->NHs) one of the concerns was the volume of updates that would be driven by an individual NH change.

I'd like to discuss this in the context of generic AFT telemetry:

 * generally, consumers MAY be interested in the fact that there are routes that are dependent upon each other, if one use case is not, then that's OK but removing information (e.g., there is loss of which protocol installed the route) seems a challenge for general usability.
 * the update volume is going to increase significantly in this case, i.e., 1M routes -> NHG1 -> 2001:db8::1 and 2001:db8::2, 2001:db8::1/128 via fe80::1%eth2 -- if you change the 2001:db8::1 route and don't allow recursion, you need to generate telemetry updates for every one of the 1M routes which is going to cause significant scaling concerns.

I think there may be some misinterpretation here. IMHO:

 * we need to allow recursion between routes
 * we do not want to allow NHGs to recurse to NHGs -- i.e., there shouldn't be a concept where we have NHG1 which references NHG2+NHG3 which then reference NHs. This is essentially encoding some "composite" next-hop into the model - in this case the "flattening" I suggest is that we have NHG1 contain the NHs that are in NHG2 and NHG3 - with the relevant weights. There has been some discussion that this might mean the quantisation is not perfect, but I believe that this is worth it as a tradeoff.

robshakir: (To be clear, I don't agree with the proposed wording, because I don't think this is the right design.)

PR #692, opened 2022-08-18
Add support for SFP28 and SFP56 form factor identities

LeonGWang:
SFP28 and SFP56 transceiver form factor identities are missing from
OpenConfig.

The successor of #562, which I had to close because I based the PR on the master branch.


ejbrever (approved):
Looks good to me.

Also adding @thofmeister if he has any thoughts.

dplore (approved):

PR #649, opened 2022-06-06
Manage prefix list and extend ACLs to match prefix lists in source/destination

oscargdd:
The proposal is to extend the ACL model to be able to match source/destination address vs a prefix list (currently only individual matching is possible). In adition, the prefix-lists are maintained. Note that the prefix lists are different from the routing policiy prefix lists. 

Network operators maintain sets of IP prefixes that are related to each other, e.g., deny-lists or accept-lists that are associated with those provided by a VPN customer, or to protect interfaces at PE nodes.  These lists are maintained and  manipulated by security expert teams. Hence, the operational use case is to be able to handle sets (prefix, ports..) independently and reference them from prefix lists. You can also see https://datatracker.ietf.org/doc/draft-dbb-netmod-acl/ that documents the use case. 

It is a functionality that we are using today in live deployments and it's required to automate with Openconfig the ACL management.

Tree of the proposal (updated to current version)

```
module: openconfig-acl-defined-sets
  +--rw defined-sets
     +--rw ipv4-prefix-sets
     |  +--rw ipv4-prefix-set* [name]
     |     +--rw name      -> ../config/name
     |     +--rw config
     |     |  +--rw name?          string
     |     |  +--rw description?   string
     |     |  +--rw prefix*        oc-inet:ipv4-prefix
     |     +--ro state
     |        +--ro name?          string
     |        +--ro description?   string
     |        +--ro prefix*        oc-inet:ipv4-prefix
     +--rw ipv6-prefix-sets
     |  +--rw ipv6-prefix-set* [name]
     |     +--rw name      -> ../config/name
     |     +--rw config
     |     |  +--rw name?          string
     |     |  +--rw description?   string
     |     |  +--rw prefix*        oc-inet:ipv6-prefix
     |     +--ro state
     |        +--ro name?          string
     |        +--ro description?   string
     |        +--ro prefix*        oc-inet:ipv6-prefix
     +--rw port-sets
        +--rw port-set* [name]
           +--rw name      -> ../config/name
           +--rw config
           |  +--rw name?          string
           |  +--rw description?   string
           |  +--rw port*          oc-pkt-match-types:port-num-range
           +--ro state
              +--ro name?          string
              +--ro description?   string
              +--ro port*          oc-pkt-match-types:port-num-range
```

The sets are called within the ACLs. The updated ACL module tree is:
```
+--rw acl
     +--rw config
     +--ro state
     |  +--ro counter-capability?   identityref
     +--rw acl-sets
 ...
[...]

PR #693, opened 2022-08-18
Add support for IPV4/IPV6 source/destination address-set and source/destination port-set.

LeonGWang:
IPv4 and IPv6 protocols could allow for multiple source and destination address prefix defined.

Multiple source and destination ports should also be allowed.

This change is backwards compatible and has precedence in 'dscp'
and 'dscp-set'.

### DIFF
--- a/release/models/acl/openconfig-packet-match.yang
+++ b/release/models/acl/openconfig-packet-match.yang
@@ -27,7 +27,14 @@ module openconfig-packet-match {
     field is omitted from a match expression, the effect is a
     wildcard ('any') for that field.";
 
-  oc-ext:openconfig-version "1.3.1";
+  oc-ext:openconfig-version "1.4.0";
+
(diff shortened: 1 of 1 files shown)


oscargdd:
Hi LeonGWang,

   I think the functionality is interesting and needed. Can you indicate in which platforms it is currently available? Please also check PR https://github.com/openconfig/public/pull/649 were aditionally I have proposed to manage the list of addresses (prefixes) separately and point to a defined set.

sulrich:
@LeonGWang this would benefit from adding src/dst port sets as well.

LeonGWang:
> @LeonGWang this would benefit from adding src/dst port sets as well.

@sulrich I pushed a new version that includes `source-port-set` and `destination-port-set`

LeonGWang:
Closing this PR because #649 has merged already.

PR #699, opened 2022-08-30
Add update-generation container in bgp global

shimangkumar:
Added a new container update-generation to configure update
generation message parameters under bgp global. update-generation
conatiner have one leaf wait-fib-install of boolean type and
default value as false. wait-fib-install is used to postpone the
route advertisment untill installed in FIB.

### DIFF
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -24,7 +24,13 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "9.1.0";
+  oc-ext:openconfig-version "9.2.0";
+
+  revision "2022-08-30" {
+    description
+      "Added update generation support.";
+    reference "9.2.0";
+  }
 
   revision "2022-05-21" {
     description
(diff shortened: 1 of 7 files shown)


akalluru1:
@dplore @robshakir

dplore:
I will bring to the openconfig operators meeting for Sep 20, 2022

earies:
Can we first follow contributor guidelines and reference other implementation support along with proper explanation and justification?

shimangkumar:
Added a new container update-generation uunder bgp global tree.
```
bgp
 |
 | --- global
       |
       | --- update-generation
             |
             | --- config
                   |
                   | --- wait-fib-install
  
```
                        
 update-generation container will have leaf related to update message generation.
 
 In this container currently added one leaf **wait-fib-install of type boolean.**
 
 **wait-fib-install** is related to postpone update generation until route is not installed in fib.
 
 Here is the command snippet related to configuration of wait-fib-install:
```
(config)#router bgp 1
(config-router-bgp)#?
  ...
  update                    Configure BGP update generation
  ...
(config-router-bgp)#update ?
  ...
  wait-install          Synchronize update generation with route programming in the FIB
(config-router-bgp)#update wait-install ?
  batch-size  batch size for FIB route acknowledgements
  <cr>
(config-router-bgp)#update wait-install batch-size ?
  <0-10000>  batch size
(config-router-bgp)#update wait-install batch-size 1000
(config-router-bgp)#exit
(config)#sh run sec router bgp
router bgp 1
   update wait-install batch-size 1000
(config)#
```

We need to add a leaf only for update wait-install keyword.

dplore:
Thanks @shimangkumar , can you please reference at least two implementations from different companies?  It is most helpful if you can add URL references to documentation describing the feature.

PR #701, opened 2022-09-01
CLEI-code - second attempt

ejbrever:
This is a carryover of https://github.com/openconfig/public/pull/628 which was left outstanding too long. There are no changes from that other than it is using an up to date branch that isn't causing issues :)

Most (if not all) optical platforms support CLEI codes. A couple references:
Cisco: https://www.cisco.com/c/en/us/support/docs/optical-networking/ons-15454-sonet-multiservice-provisioning-platform-mspp/46247-15454-cleicodes.html
Nokia: https://documentation.nokia.com/html/365-372-324R10.0/index.html?i=365-372-324R10.0-qzx79

### DIFF
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -20,7 +20,13 @@ submodule openconfig-platform-common {
     "This modules contains common groupings that are used in multiple
     components within the platform module.";
 
-  oc-ext:openconfig-version "0.19.0";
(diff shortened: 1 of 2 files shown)


ejbrever:
@dplore would you mind LGTM'ing again? It's all the same as before (https://github.com/openconfig/public/pull/628), but in a new PR since the old one got old and I couldn't figure out how to fast forward the branch :)

dplore (approved):

PR #703, opened 2022-09-12
Adding control over sending of IS-IS CSNP packets on P2P interfaces. …

nokia1adam:
Change Scope
Strictly speaking it is not necessary to send periodic CSNP PDUs on point-to-point interfaces (i.e. after the adjacency has been established). Some implementations support a configuration option to disable these periodic transmissions, to save resources. This change introduces a new leaf to support this configuration option.

Platform Implementations
To be provided


nokia1adam:
Already addressed by PR #712

PR #704, opened 2022-09-12
Add next-hop-group/next-hop-group-network-instance to network-instance's spec files

wenovus:
Currently they don't show up in the [documentation](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html), but they should be there.


robshakir (approved):
Thanks for this!

PR #711, opened 2022-09-20
[WiFi] Replace uptime counter with timestamp

xavier-contreras:
### Change Scope

This changes the leaf "uptime" under openconfig-ap-manager.yang from a uint32 counter to a new leaf of "up-time" which will be a timeticks64.  This is to make telemetry ingestion less churn heavy.  The intent of the leaf does not change.

This change is not backwards compatible as a leaf is replaced.

### Platform Implementations

oc-types:timeticks64 is used throughout other models and it should be trivial to implement.

### DIFF
--- a/release/models/wifi/openconfig-ap-manager.yang
+++ b/release/models/wifi/openconfig-ap-manager.yang
@@ -13,6 +13,7 @@ module openconfig-ap-manager {
   import openconfig-yang-types { prefix oc-yang; }
   import openconfig-inet-types { prefix oc-inet; }
   import openconfig-wifi-types { prefix oc-wifi; }
+  import openconfig-types { prefix oc-types; }
 
   // Meta
   organization "OpenConfig working group";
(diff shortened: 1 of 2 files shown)


jsnyder81:
Should we move this to the new release model: https://github.com/openconfig/public/blob/master/doc/releases.md ?

If we follow this, would should aim to mark `uptime` as deprecated and introduce `up-time`.

dplore:
> Should we move this to the new release model: https://github.com/openconfig/public/blob/master/doc/releases.md ?
> 
> If we follow this, would should aim to mark `uptime` as deprecated and introduce `up-time`.

Agreed, this is a good example where we can deprecate the leaf to make this change non-breaking.

xavier-contreras:
> > Should we move this to the new release model: https://github.com/openconfig/public/blob/master/doc/releases.md ?
> > If we follow this, would should aim to mark `uptime` as deprecated and introduce `up-time`.
> 
> Agreed, this is a good example where we can deprecate the leaf to make this change non-breaking.

Fixed -- guidance did not exist prior to this PR, good to see it documented.  I also updated the version to minor increase since this is no longer a breaking change.

jsnyder81:
LGTM

mike-albano (approved):
LGTM

PR #715, opened 2022-09-22
Add support for vendor specific hardware drop counters

rolandphung:
- (M) release/models/platform/openconfig-platform-pipeline-counters.yang

### Change Scope

This request creates a location in /components/component/integrated-circuit/pipeline-counters/drop where vendors may populate platform specific drop counters. The goal is to define a process for vendors to create and publish their internal counters and aggregate the corresponding counters. With counters categorized with severity, the switch can aggregate based on severity while retaining counter transparency.

More details in the [original proposal](https://docs.google.com/document/d/1bLbeG8Wp1vkaxQg4hCF-1G2XJLSCpjcwMD8eftByM2Y/edit).

There are two major differences from the original proposal that was made:
- No vendor type was needed as this could be abstracted in a generic `vendor` container. Each vendor should augment this container in order to add the vendor and platform name.
- Adverse/Congestion/Packet Processing counter types are not needed, instead it is split between their individual containers.

#### Vendor independent section
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:drop
                 +--ro oc-ppc:state
                    +--ro oc-ppc:adverse-aggregate?             oc-yang:counter64
                    +--ro oc-ppc:congestion-aggregate?          oc-yang:counter64
                    +--ro oc-ppc:packet-processing-aggregate?   oc-yang:counter64
                 +--ro oc-ppc:vendor
```

#### Vendor dependent section
Example vendor augment
```
augment "/components/component/integrated-circuit/pipeline-counter/drop/vendor" {
  container <vendor name> {
    container <platform name> {
      uses pipeline-vendor-drop-containers;
    }
  }
}
```

Example pyang
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:drop
                 +--ro oc-ppc:vendor
                    +--ro <vendor name>
                      +--ro <platform name>
                        +--ro oc-ppc:adverse
                           +--ro oc-ppc:state
                        +--ro oc-ppc:congestion
                           +--ro oc-ppc:state
                        +--ro oc-ppc:packet-processing
                           +--ro oc-ppc:state
```

### Platform Implementations

 * Cisco ASR 9000: [link to documentation](https://community.cisco.com/t5/service-providers-knowledge-base/asr9000-xr-troubleshooting-packet-drops-and-understanding-np/ta-p/3126715)
 * Cisco Nexus 5000: [link to documentation](https://www.cisco.com/c/en/us/support/docs/switches/nexus-5000-series-switches/116172-technote-product-00.html)
 * Juniper: [link to documentation](https://www.juniper.net/documentation/us/en/software/junos/junos-overview/topics/concept/using-show-commands-for-packet-drops.html)
 * Arista:  `show hardware counter drop`
   * A sample of possible counters are listed in the Appendix A of the [original proposal](https://docs.google.com/document/d/1bLbeG8Wp1vkaxQg4hCF-1G2XJLSCpjcwMD8eftByM2Y/edit#heading=h.7035it1ttmrt)

### DIFF
--- a/doc/vendor_counter_guide.md
+++ b/doc/vendor_counter_guide.md
@@ -14,7 +14,8 @@ Each implementor should augment `/components/component/integrated-circuit/pipeli
 
 If these aggregate counters are implemented, the sum of the vendor-specific counters must match the aggregate counters.
 
-If an integrated-circuit has a vendor-specific packet drop counter which cannot differentiate between packet-processing, congestion and adverse drops, then that counter should still be exposed as a vendor-specific packet-processing counter with an appropriate description in the vendor's augmentation.   The packet-processing-aggregate counter should be incremented in this scenario as expected above.  Such a counter is undesirable as it means this hardware cannot meet the goal for identifying adverse packet drops in the ASIC, but it is better not to ruin the fidelity of the adverse-aggregate drop counter with noise of intended packet drops.
+If an integrated-circuit has a vendor-specific packet drop counter which cannot differentiate between packet-processing, congestion and adverse drops, then that counter should still be exposed as a vendor-specific packet-processing counter with an appropriate description in the vendor's augmentation.   The `packet-processing-aggregate` counter should be incremented in this scenario as expected above.  Such a counter is undesirable as it means this hardware cannot meet the goal for identifying adverse packet drops in the ASIC, but it is better not to ruin the fidelity of the `adverse-aggregate` drop counter with noise of intended packet drops.
+
 ## Example
 
 This following is a sample augmentation file.
--- a/doc/vendor_counter_guide.md
+++ b/doc/vendor_counter_guide.md
@@ -6,12 +6,15 @@ This document provides the guidelines for the vendor-specific portions of openco
 
 ## Usage: Vendor-specific pipeline drop counter
 
-Each implementor should augment `/components/component/integrated-circuit/pipeline-counter/drop/vendor` with their own vendor and platform containers. Within the platform container, that container may use the utility grouping `oc-ppc:pipeline-vendor-drop-containers` that provides the adverse/congestion/packet-processing specific containers. For each set of adverse/congestion/packet-processing counters augmented into `oc-ppc:pipeline-vendor-drop-containers`, the sum of the counters should be included in the values of the aggregate leaves:
+Each implementor should augment `/components/component/integrated-circuit/pipeline-counter/drop/vendor` with their own vendor and platform containers. The naming of the platform container may consist of the platform name, ASIC family, or a combination of both platform and ASIC family. Within the platform container, that container may use the utility grouping `oc-ppc:pipeline-vendor-drop-containers` that provides the adverse/congestion/packet-processing specific containers. For each set of adverse/congestion/packet-processing counters augmented into `oc-ppc:pipeline-vendor-drop-containers`, the sum of the counters should be included in the values of the aggregate leaves:
 
-- Counters within  `.../pipeline-counter/drop/vendor/<vendor>/<platform>/adverse/state` aggregate into `.../pipeline-counter/drop/state/adverse-aggregate`
-- Counters within  `.../pipeline-counter/drop/vendor/<vendor>/<platform>/congestion/state`  aggregate into `.../pipeline-counter/drop/state/congestion-aggregate`
-- Counters within  `.../pipeline-counter/drop/vendor/<vendor>/<platform>/packet-processing/state` aggregate into `.../pipeline-counter/drop/state/packet-processing-aggregate`
+- Counters within `.../pipeline-counter/drop/vendor/<vendor>/<platform>/adverse/state` aggregate into `.../pipeline-counter/drop/state/adverse-aggregate`
+- Counters within `.../pipeline-counter/drop/vendor/<vendor>/<platform>/congestion/state`  aggregate into `.../pipeline-counter/drop/state/congestion-aggregate`
+- Counters within `.../pipeline-counter/drop/vendor/<vendor>/<platform>/packet-processing/state` aggregate into `.../pipeline-counter/drop/state/packet-processing-aggregate`
 
+If these aggregate counters are implemented, the sum of the vendor-specific counters must match the aggregate counters.
+
+If an integrated-circuit has a vendor-specific packet drop counter which cannot differentiate between packet-processing, congestion and adverse drops, then that counter should still be exposed as a vendor-specific packet-processing counter with an appropriate description in the vendor's augmentation. Such a counter is undesirable as it means this hardware cannot meet the goal for identifying adverse packet drops in the ASIC, but it is better not to ruin the fidelity of the adverse-aggregate drop counter with noise of intended packet drops.
 ## Example
 
 This following is a sample augmentation file.
@@ -117,4 +120,3 @@ module: openconfig-platform
                               +--ro acme-ppc:packet-processing-reason-counter-b?    oc-yang:counter64
                               +--ro acme-ppc:packet-processing-reason-counter-c?    oc-yang:counter64
 ```
-
--- a/doc/vendor_counter_guide.md
+++ b/doc/vendor_counter_guide.md
@@ -1,0 +1,120 @@
+# Vendor-Specific Augmentation for Pipeline Counter
+
+**Contributors**: roland@arista.com
+
+This document provides the guidelines for the vendor-specific portions of openconfig-pipeline-counters.yang. As implementations differ from vendor to vendor and platform to platform, a process of adding vendor-specific counters will be defined here.
+
+## Usage: Vendor-specific pipeline drop counter
+
+Each implementor should augment `/components/component/integrated-circuit/pipeline-counter/drop/vendor` with their own vendor and platform containers. Within the platform container, that container may use the utility grouping `oc-ppc:pipeline-vendor-drop-containers` that provides the adverse/congestion/packet-processing specific containers. For each set of adverse/congestion/packet-processing counters augmented into `oc-ppc:pipeline-vendor-drop-containers`, the sum of the counters should be included in the values of the aggregate leaves:
+
+- Counters within  `.../pipeline-counter/drop/vendor/<vendor>/<platform>/adverse/state` aggregate into `.../pipeline-counter/drop/state/adverse-aggregate`
+- Counters within  `.../pipeline-counter/drop/vendor/<vendor>/<platform>/congestion/state`  aggregate into `.../pipeline-counter/drop/state/congestion-aggregate`
+- Counters within  `.../pipeline-counter/drop/vendor/<vendor>/<platform>/packet-processing/state` aggregate into `.../pipeline-counter/drop/state/packet-processing-aggregate`
+
+## Example
+
+This following is a sample augmentation file.
+
+- Vendor: Acme
+- Platform: AsicFamily
+
+### Example YANG Augmentation
+
+release/platform/acme-asicfamily-drop-augments.yang
+
+```yang
+grouping acme-asicfamily-adverse-drop-counters {
+  leaf adverse-reason-counter-a {
+    type oc-yang:counter64;
+  }
+
+  leaf adverse-reason-counter-b {
+    type oc-yang:counter64;
+  }
+
+  leaf adverse-reason-counter-c {
+    type oc-yang:counter64;
+  }
+}
+
+grouping acme-asicfamily-congestion-drop-counters {
+  leaf congestion-reason-counter-a {
+    type oc-yang:counter64;
+  }
+
+  leaf congestion-reason-counter-b {
+    type oc-yang:counter64;
+  }
+
+  leaf congestion-reason-counter-c {
+    type oc-yang:counter64;
+  }
+}
+
+grouping acme-asicfamily-packet-processing-drop-counters {
+  leaf packet-processing-reason-counter-a {
+    type oc-yang:counter64;
+  }
+
+  leaf packet-processing-reason-counter-b {
+    type oc-yang:counter64;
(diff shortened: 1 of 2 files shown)


dplore:
Hi @rolandphung , the PR has some differences from the linked proposal.  Can you clarify where you are intending the vendor specific augments to occur?  Providing an example (here in the PR) would clarify what the intended format is.  I would like to see if we can achieve a template for a uniform structure for augments.

rolandphung:
I had left this in the original message:
```
There are two major differences from the original proposal that was made:

* No vendor type was needed as this could be abstracted in a generic vendor container. This also simplifies the extra changes to add a vendor type which explicitly defines the vendor platform pairs.
* Adverse/Congestion/Packet Processing counter types are not needed, instead it is split between their individual containers.
```
> No vendor type was needed as this could be abstracted in a generic vendor container. This also simplifies the extra changes to add a vendor type which explicitly defines the vendor platform pairs.

I wanted to simplify the changes as much as possible. It made sense when you had suggested having a `vendor` container, instead of the actual vendor name.

> Adverse/Congestion/Packet Processing counter types are not needed, instead it is split between their individual containers.

The goal here was to reduce the number of non-standard use of yang modeling, such as extending a base case type to associate additional severity meta information.

> Providing an example (here in the PR) would clarify what the intended format is

Here's a quick example:
```
grouping sand-platform-adverse-drop-counters {
  leaf drop-voq-in-null-route {
    type oc-yang:counter64;
    description
      "Drops occuring in VOQ where the destination is pointing to a null route.
      This is expected when the traffic hits a route that points to DROP rule
      or when there is no route to destination.";
  }
}

augment "/components/component/integrated-circuit/pipeline-counters/drop/vendor/adverse/state" {
  uses sand-platform-adverse-counters;
}
```

earies:
From the understanding of often vast implementation differences - +1 for this design pattern in general.  This provides controlled anchor points vs. either random inconsistent augments or resorting to native model paths and drawing what can be complex relationships between model sets.  A few initial comments on the structure/pattern...

Looking collectively at the platform domain, we have a few other design patterns that exist within that I think should be discussed on how to handle in parallel and which pattern(s) may be deemed for deprecation if we provide alternate means of representation.  Most notably `properties` and `resources`.  The prior being somewhat of a 1:1 opaque k/v while the latter being a slightly more targeted 1:many pairing.

As far as the doc vs. this PR differences - I just want to call out that if we consolidate down to just a `vendor` container (vs. per implementation named as suggested in the doc), any uniqueness of a path would be lost thus stressing the importance of per implementation handlers/logic (If unique, you could skirt around this in alternate fashions).  For instance, vendor A could augment node `foo` of type `uint32` while vendor B augments node `foo` of type `counter64` - same path in theory to the human reader (considering no namespace/prefix handling and just relying off origin=`openconfig` since augments will roll up to that origin) but _not_ at all the same (representation, behavioral, etc..) - something a consumer has to be aware of in any case.  With that said, the approach in the doc may be the less confusing approach imo.

It is also likely this design pattern could be used throughout the model sets for conveying per implementation specifics - just that there will likely need to be a protocol distinguishing the correct and incorrect way to augment

dplore:
Thanks Roland.  Perhaps the vendor could be a named grouping vs. container.  For example (admit I am cheating here in that I haven't provided the yang, but just illustrate the concept):

### Vendor independent portion of the schema
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:drop
                 +--ro oc-ppc:state
                    +--ro oc-ppc:adverse-aggregate?             oc-yang:counter64
                    +--ro oc-ppc:congestion-aggregate?          oc-yang:counter64
                    +--ro oc-ppc:packet-processing-aggregate?   oc-yang:counter64
```

### Vendor dependent portion of the schema
This serves as a template structure for what openconfig will expect augments to look like
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:drop
                    +--ro oc-ppc:vendor* [name]
                        +--ro oc-ppc:adverse
                        |  +--ro oc-ppc:state
                        +--ro oc-ppc:congestion
                        |  +--ro oc-ppc:state
                        +--ro oc-ppc:packet-processing
                           +--ro oc-ppc:state
```

### Vendor dependent portion of the schema with example leafs

```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:drop
                    +--ro oc-ppc:vendor* [name]
                        +--ro oc-ppc:adverse
                        |  +--ro oc-ppc:state
                        |     +--ro oc-ppc:addrop-leaf1?   oc-yang:counter64 
                        +--ro oc-ppc:congestion
                        |  +--ro oc-ppc:state
                        |     +--ro oc-ppc:codrop-leaf1?   oc-yang:counter64 
                        +--ro oc-ppc:packet-processing
                           +--ro oc-ppc:state
                              +--ro oc-ppc:padrop-leaf1-drop?        oc-yang:counter64 
```

### Flattened example
```
/components/component/integrated-circuit/pipelines-counters/drop/arista/adverse/state/addrop-leaf1-drop
```

rolandphung:
>                     +--ro oc-ppc:vendor* [name]

@dplore  I've been mulling over how to implement this container as a singleton list to allow the vendor's `name` to exist alongside `vendor` container. Under the constraints of the [openconfig_style_guide#list](https://github.com/openconfig/public/blob/master/doc/openconfig_style_guide.md#list), it would look like the following:
```
container vendors {
  list vendor {
    key "name";
    leaf name {
      type leafref {
        path "../state/name";
      }
    }

    container state { 
      leaf name {
        type string;
      }
    }
  }
}

module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:drop
                 +--ro oc-ppc:vendors
                    +--ro oc-ppc:vendor* [name]
                       +--ro oc-ppc:name                 -> ../state/name
                       +--ro oc-ppc:state
                       |  +--ro oc-ppc:name?   string
```
This to me implies multiple vendors and not a singleton which could be misinterpreted. I don't really see any other way without requesting a change in the style guide. Having a long `vendor[name]` in your proposal would be clearer and provide both benefits of querying a vendor agnostic path and identifying the vendor within the path: `/components/component/integrated-circuit/pipeline-counters/drop/vendor/adverse/state`

Alternatively there could be a `drop/vendor/state/name` and avoid the list altogether: 
```
  container vendor {
    container state { 
      leaf name {
        type string;
      }
    }
  }

module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:drop
                 +--ro oc-ppc:vendor
                    +--ro oc-ppc:state
                    |  +--ro oc-ppc:name?   string
```

jsterne:
Discussed in the call today: If an IC detects errors on the Rx side of a link into that IC, those errors would be reported against this IC (not against a separate "link" component). The errors reported against an IC aren't necessarily guaranteed to be caused by that IC. They are just error counters reported *by* that IC (and could be errors attributable to some other component).  It would be good to ensure this is clear in the YANG model.

rolandphung:
>As far as the doc vs. this PR differences - I just want to call out that if we consolidate down to just a vendor container (vs. per implementation named as suggested in the doc), any uniqueness of a path would be lost thus stressing the importance of per implementation handlers/logic (If unique, you could skirt around this in alternate fashions). For instance, vendor A could augment node foo of type uint32 while vendor B augments node foo of type counter64 - same path in theory to the human reader (considering no namespace/prefix handling and just relying off origin=openconfig since augments will roll up to that origin) but not at all the same (representation, behavioral, etc..) - something a consumer has to be aware of in any case. With that said, the approach in the doc may be the less confusing approach imo.

@earies Can you elaborate on how the approach in the doc would be less confusing/better?

rolandphung:
I made an update that matches more closely to what was in the proposal. After some thought about this, I agree with @earies about the uniqueness of path would be better maintained if vendor is the landing point for vendor augments.

Description is updated and more fleshed out with vendor independent/dependent examples.

dplore:
@rolandphung can you add a markdown document to this PR which specifies an example vendor augment?  For example, a vendor name, platform and a leaf inside each of the adverse, congestion and packet-processing containers?    I think we need this as a markdown doc that is part of this PR to formalize this and to provide clarity to the community.

rgwilton:
Just to record the comments that I made verbally on the call:

1) Is it fixed to only two levels (vendor + platform)? The indication is that it was.

2) Is their flexibility in the name of the platform (e.g., to allow platform/linecard specific ASIC counters)?  It would be useful it there is some flexibility, and the documentation indicates that this is allowed (e.g., per platform, per asic family, or platform + asic specific, or platform + linecard specific).

3) What are rules relative to updating the aggregate vs specific counters.
 - If specific counter updated, aggregate MUST be updated.  I.e., to ensure that they are consistent.
 - If aggregate counter is updated, then MUST update the specific counter, if an appropriate counter is available, but allowed to just update the aggregate counter if there is not a suitable more specific counter?

dplore:
> Just to record the comments that I made verbally on the call:
> 
> 1. Is it fixed to only two levels (vendor + platform)? The indication is that it was.

Yes. the model have a vendor container and a platform container.

> 2. Is their flexibility in the name of the platform (e.g., to allow platform/linecard specific ASIC counters)?  It would be useful it there is some flexibility, and the documentation indicates that this is allowed (e.g., per platform, per asic family, or platform + asic specific, or platform + linecard specific).

The names used for the two containers are determined by the vendor yang definition.  Note that this model is specific to integrated circuits.  One must only report drop counters for integrated circuits (asic / switch-chip) in this tree.  How the vendor chooses to name the integrated circuit is up to the vendor.   The operational use case for this data is to support root cause analysis for packet drops in the integrated circuit, so the name should be useful to the vendor in this context.  

> 3. What are rules relative to updating the aggregate vs specific counters.
> 
> * If specific counter updated, aggregate MUST be updated.  I.e., to ensure that they are consistent.
> * If aggregate counter is updated, then MUST update the specific counter, if an appropriate counter is available, but allowed to just update the aggregate counter if there is not a suitable more specific counter?

I would like to see a device always update vendor+platform specific counters AND aggregate.  Previous efforts to track packet loss at the integrated circuit have been challenged by not being able to differentiate between intended (packet processing and congestion) packet drops  and unintended (adverse) packet drops.    

If an integrated circuit has a vendor+platform specific packet drop counter which cannot differentiate between packet-processing, congestion and adverse drops, then that counter should still be exposed as a vendor specific, `packet-processing` counter with an appropriate description in the vendor yang.  In addition, the `packet-processing-aggregate` should be incremented.  Such a counter is undesirable as it means this hardware cannot meet the goal for identifying adverse packet drops (for the use cases where this 'catch-all drop counter' is used)  in the ASIC.  But it is better to not ruin the fidelity of the `adverse-aggregate` drop counter with noise of intended packet drops.

dplore:
@rolandphung could you update the PR to cover the above?  I think the counter update expectations should be covered in leaf descriptions in the yang and optionally may also be covered in the .md.

rolandphung:
> 2. Is their flexibility in the name of the platform (e.g., to allow platform/linecard specific ASIC counters)?  It would be useful it there is some flexibility, and the documentation indicates that this is allowed (e.g., per platform, per asic family, or platform + asic specific, or platform + linecard specific).

The flexibility should go as far as (per platform, per asic family, or platform + asic specific). As Darren has noted, it is up to the vendor name the container on the platform level. For linecard counters, I would expect this require additional model updates to add `/components/component/linecard/drop/vendor` or equivalent.

> could you update the PR to cover the above? I think the counter update expectations should be covered in leaf descriptions in the yang and optionally may also be covered in the .md.

Will do.

rolandphung:
@dplore PTAL

doc/vendor_counter_guide.md
```diff
@@ -0,0 +1,122 @@
...
+
+- Counters within `.../pipeline-counter/drop/vendor/<vendor>/<platform>/adverse/state` aggregate into `.../pipeline-counter/drop/state/adverse-aggregate`
+- Counters within `.../pipeline-counter/drop/vendor/<vendor>/<platform>/congestion/state`  aggregate into `.../pipeline-counter/drop/state/congestion-aggregate`
+- Counters within `.../pipeline-counter/drop/vendor/<vendor>/<platform>/packet-processing/state` aggregate into `.../pipeline-counter/drop/state/packet-processing-aggregate`
+
+If these aggregate counters are implemented, the sum of the vendor-specific counters must match the aggregate counters.
+
+If an integrated-circuit has a vendor-specific packet drop counter which cannot differentiate between packet-processing, congestion and adverse drops, then that counter should still be exposed as a vendor-specific packet-processing counter with an appropriate description in the vendor's augmentation. Such a counter is undesirable as it means this hardware cannot meet the goal for identifying adverse packet drops in the ASIC, but it is better not to ruin the fidelity of the adverse-aggregate drop counter with noise of intended packet drops.
```
dplore, line 17:
```suggestion
If an integrated-circuit has a vendor-specific packet drop counter which cannot differentiate between packet-processing, congestion and adverse drops, then that counter should still be exposed as a vendor-specific packet-processing counter with an appropriate description in the vendor's augmentation.   The packet-processing-aggregate counter should be incremented in this scenario as expected above.  Such a counter is undesirable as it means this hardware cannot meet the goal for identifying adverse packet drops in the ASIC, but it is better not to ruin the fidelity of the adverse-aggregate drop counter with noise of intended packet drops.
```

dplore (approved):
LGTM.  Thanks to all contributors and to Roland for authoring this model

PR #716, opened 2022-09-27
Add base-mac-address

dplore:
### Change Scope
* Add base-mac-address leaf `/components/component/state/base-mac-address`.

### Platform Implementations
* [JunOS - show chassis mac address](https://www.juniper.net/documentation/us/en/software/junos/chassis/topics/ref/command/show-chassis-mac-addresses.html)
* [Arista EOS - show version](https://www.arista.com/en/um-eos/eos-command-line-interface-cli#xx1125082)
* [Nokia SRL - show version](https://infocenter.nokia.com/public/SRLINUX200R6A/index.jsp?topic=%2Fcom.srlinux.interfaceconfig%2Fhtml%2Fcli-interface.html)

### DIFF
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -69,9 +69,10 @@ module openconfig-platform {
 
   revision "2022-09-26" {
     description
-      "Add new state data for component CLEI code.";
+      "Add state data for base-mac-address.";
     reference "0.21.0";
   }
+
   revision "2022-08-31" {
     description
       "Add new state data for component CLEI code.";
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,8 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.20.0";
+  oc-ext:openconfig-version "0.21.0";
 
+  revision "2022-09-26" {
+    description
+      "Add new state data for component CLEI code.";
+    reference "0.21.0";
+  }
   revision "2022-08-31" {
     description
       "Add new state data for component CLEI code.";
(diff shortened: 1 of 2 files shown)


rolandphung (approved):

dplore:
Reviewed on Sep 27, 2022 OpenConfig Operators meeting.  Will merge on Oct 11, 2022 if there are no objections.

release/models/platform/openconfig-platform.yang
```diff
@@ -542,6 +548,17 @@ module openconfig-platform {
...
         report the same value.";
     }
+
+    leaf base-mac-address {
+      type oc-yang:mac-address;
+      description
+        "This is a MAC address representing the root or primary MAC
+        address for a component.  Only components such as CHASSIS and
+        CONTROLLER_CARD are expected to providea base-mac-address.  The
+        base mac-address for CHASSIS and a PRIMARY CONTROLLER_CARD may
```
robshakir, line 558: Is there a reason that only the primary controller card and not the secondary can have the same `base-mac-address`? Does this imply that when there is a switchover of controller cards there would be a change in base MAC address? (Seems a challenging idea!)

robshakir (changes requested):

release/models/platform/openconfig-platform.yang
dplore: Good question.  I am collecting some more information on the use case so I can clarify appropriately.

release/models/platform/openconfig-platform.yang
```diff
@@ -542,6 +548,17 @@ module openconfig-platform {
         device, one as primary and the other as secondary, should both
         report the same value.";
     }
+
+    leaf base-mac-address {
+      type oc-yang:mac-address;
+      description
+        "This is a MAC address representing the root or primary MAC
+        address for a component.  Only components such as CHASSIS and
```
mike-albano, line 556: If we remove the word "only" this seems to increase the re-usability of the model. As an example Access Point radios also utilize a "base" mac address from which all BSSIDs utilize over-the-air.

release/models/platform/openconfig-platform.yang
dplore: Updated description with use case.

robshakir (approved):
Thanks for the fix.

PR #719, opened 2022-09-28
add last configuration timestamp leaf

### LINKED ISSUE #545: openconfig-system/state/config-time ?
leaf boot-time {
        type oc-types:timeticks64;
        units "nanoseconds";
        description
          "This timestamp indicates the time that the system was last
          restarted.  The value is the timestamp in nanoseconds relative
          to the Unix Epoch (Jan 1, 1970 00:00:00 UTC).";
    }

Is there any plan to have a "config-time" leaf which would indicate the timestamp of the last configuration change?
And perhaps another leaf to indicate when the configuration was last saved to persistent/startup configuration?

RonanMacF:
Some users may have a desire to know if the configuration on the box has changed. Subscribing to the entire tree could work, however, all the state leafs would have to be filtered out. This proposes a new leaf `last-configuration-timestamp` which represents the timestamp at which the last configuration was made on the box. E.g. an OpenConfig Set request would update this time stamp or a CLI configuration change on the box would update this timestamp. However, counters data being updated in the YANG tree does not update this timestamp. This provides a mean by which a user can subscribe to the leaf to determine if configuration was made on the box.

*pyang output*

```
module: openconfig-system
  +--rw system
     +--rw config
     |  +--rw hostname?       oc-inet:domain-name
     |  +--rw domain-name?    oc-inet:domain-name
     |  +--rw login-banner?   string
     |  +--rw motd-banner?    string
     +--ro state
     |  +--ro hostname?                       oc-inet:domain-name
     |  +--ro domain-name?                    oc-inet:domain-name
     |  +--ro login-banner?                   string
     |  +--ro motd-banner?                    string
     |  +--ro current-datetime?               oc-yang:date-and-time
     |  +--ro boot-time?                      oc-types:timeticks64
     |  +--ro software-version?               string
     |  +--ro last-configuration-timestamp?   oc-types:timeticks64
```
## References
* [Cisco IOS XR - show configuration running-config](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r3-9/system_management/command/reference/yr39xr12k_chapter5.html#wp1406184717)
* [Juniper JunOS - show system commit](https://www.juniper.net/documentation/us/en/software/junos/cli/topics/ref/command/show-system-commit.html)

Resolves #545

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -290,15 +290,6 @@ module openconfig-system {
         information such as the current system date and time, uptime,
         last login timestamp, etc.";
     }
-
-    leaf last-configuration-timestamp {
-      type oc-types:timeticks64;
-      units "nanoseconds";
-      description
-        "Indicates the timestamp at which the last configuration change
-        was made. This may may be through CLI, OpenConfig or some other
-        mechanism.";
-    }
   }
 
   grouping mount-point-state {
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -374,7 +374,7 @@ module openconfig-system {
       units "nanoseconds";
       description
         "Indicates the timestamp at which the last configuration change
-        was made. This may may be through CLI, OpenConfig or some other
+        was made. This may may be through CLI, gNMI or some other
         mechanism.";
     }
   }
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -373,9 +373,9 @@ module openconfig-system {
       type oc-types:timeticks64;
       units "nanoseconds";
       description
-        "Indicates the timestamp at which the last configuration change
-        was made. This may may be through CLI, gNMI or some other
-        mechanism.";
+        "Indicates the monotonically increasing timestamp at which the
+        last configuration change was made. This may may be through CLI, 
+        gNMI or some other mechanism.";
     }
   }
 
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -377,6 +377,15 @@ module openconfig-system {
           value of the state/software-version leaf in the component
           of type OPERATING_SYSTEM.";
     }
+
+    leaf last-configuration-timestamp {
+      type oc-types:timeticks64;
+      units "nanoseconds";
+      description
+        "Indicates the timestamp at which the last configuration change
+        was made. This may may be through CLI, OpenConfig or some other
+        mechanism.";
+    }
   }
 
   grouping mount-points-top {
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -374,7 +374,7 @@ module openconfig-system {
       units "nanoseconds";
       description
         "Indicates the monotonically increasing timestamp at which the
-        last configuration change was made. This may may be through CLI, 
+        last configuration change was made. This may may be through CLI,
         gNMI or some other mechanism.";
     }
   }
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -46,7 +46,13 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.15.0";
+  oc-ext:openconfig-version "0.16.0";
+
+  revision "2022-09-28" {
+    description
+      "Add last configuration timestamp leaf.";
+    reference "0.16.0";
+  }
 
   revision "2022-07-25" {
     description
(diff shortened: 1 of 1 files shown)


RonanMacF:
yamllint is failing on unrelated code, the system module has passed

release/models/system/openconfig-system.yang
```diff
@@ -284,6 +290,15 @@ module openconfig-system {
         information such as the current system date and time, uptime,
         last login timestamp, etc.";
     }
+
+    leaf last-configuration-timestamp {
```
dplore, line 294: I think this should be under [grouping system-global-state](https://github.com/openconfig/public/pull/719/files#diff-39c8ff1320ad76b4552208d599622a306cc84862f6dd94d378573c980e43e85cR353)

dplore (changes requested):
Suggestion to move this under global state instead of config.  Also, can you provide public links for reference to some vendor documentation that indicate some support for showing the last configuration changed time

release/models/system/openconfig-system.yang
RonanMacF: Right you are, mistakenly put this under config

RonanMacF:
> Suggestion to move this under global state instead of config. Also, can you provide public links for reference to some vendor documentation that indicate some support for showing the last configuration changed time

This leaf will be implemented Arista Networks. This will be used to determine any out of band switch configuration. At the moment none of these documents are public.

dplore:
It's true that this is a new openconfig leaf, but the underlying functionality of storing the last configuration change timestamp is widely supported.  There are CLI commands which expose this data today.  I will add a couple of references for us in the PR description.  I do this because it helps set a standard for pull requests that's needed for quality and to speed reviews of pull requests.

dplore:
Scheduled for OC operator review on Oct 4, 2022

release/models/system/openconfig-system.yang
```diff
@@ -362,6 +368,15 @@ module openconfig-system {
           value of the state/software-version leaf in the component
           of type OPERATING_SYSTEM.";
     }
+
+    leaf last-configuration-timestamp {
+      type oc-types:timeticks64;
+      units "nanoseconds";
+      description
+        "Indicates the timestamp at which the last configuration change
+        was made. This may may be through CLI, OpenConfig or some other
```
robshakir, line 377: `s/OpenConfig/gNMI/` -- configurations are manipulated through gNMI rather than OpenConfig :-)

robshakir (changes requested):
One clarification, otherwise LGTM.

dplore (approved):
LGTM.  The clarification Rob has asked for is a good one to follow up on.

release/models/system/openconfig-system.yang
```diff
@@ -362,6 +368,15 @@ module openconfig-system {
           value of the state/software-version leaf in the component
           of type OPERATING_SYSTEM.";
     }
+
+    leaf last-configuration-timestamp {
+      type oc-types:timeticks64;
+      units "nanoseconds";
+      description
+        "Indicates the timestamp at which the last configuration change
```
dplore, line 376: @rolandphung what will happen if the configuration is not 'saved' or 'stored' to the device and the device is rebooted?  Perhaps we should clarify the description to specify the behavior or allow for differences.

dplore (commented):
Upon discussion in the OC operators meeting, this question came up.

release/models/system/openconfig-system.yang
RonanMacF: agreed, done!

release/models/system/openconfig-system.yang
RonanMacF: The current implementation in mind would have been that the data would have been lost, upon reboot the timestamp would update until the box reaches a steady state. Some metadata could be persisted to the filesystem to help identify if a config change was a real change or a reread of current config already accounted for. I suppose it depends on how strict we want this leaf to be, if its for people who just want to see if an unexpected configuration occurs by subscribing to the leaf then this is probably an ok edge case for them.

dplore: For our operational use case, it is good to see the timestamp update if a reboot triggered a configuration change because the running configuration was not saved at the reboot time.  Can you confirm this is what you're expecting the behavior to be?

dplore: Please update the description to require that the timestamp must only be incremented.    

This still leaves it up to the implementation on exactly how to behave in this scenario where 'unsaved' configuration is lost due to a reboot.  The implementation can leave the timestamp as the last time the config was changed, or update to the reboot time.

dplore (changes requested):

release/models/system/openconfig-system.yang
RonanMacF: Done

dplore (commented):
LGTM, there is a trailing whitespace issue if you can correct that please, then we are ready.

dplore:
This was reviewed in the OC operators meeting on Oct 4 and 18th and comments have been addressed in the PR.  Thanks @RonanMacF for the contribution.

dplore (approved):

PR #723, opened 2022-10-06
Update ACL sequence-id to start from 1

dplore:
### Change Scope

* It is observed a number of network OS implementations do not support starting ACL's at '0', but rather start at '1'.
* This change is not backwards compatible

### Platform Implementations

* [Cisco  IP ACL sequence number range](https://www.cisco.com/en/US/docs/ios-xml/ios/sec_data_acl/configuration/15-2mt/sec-acl-seq-num.html#GUID-C6FA179D-880C-4812-9062-F5A85F6F8D10)

* [Arista EOS IP ACL](https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1149143)

### DIFF
--- a/release/models/acl/openconfig-acl.yang
+++ b/release/models/acl/openconfig-acl.yang
@@ -34,9 +34,15 @@ module openconfig-acl {
     packets should be handled. Entries have a type that indicates
     the type of match criteria, e.g., MAC layer, IPv4, IPv6, etc.";
 
-  oc-ext:openconfig-version "1.2.2";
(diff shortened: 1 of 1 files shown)


dplore:
Reviewed this at the Oct 11, 2022 OpenConfig operators meeting.  There is not consensus that this is an important change to make.  Since it is a breaking change and not particularly important, we should not merge this.  

Operators are advised to start ACL sequence-id from ‘1’ to avoid issues.

PR #725, opened 2022-10-11
Hardware AFT Programming errors module

akalluru1:
Initial proposal to model the hardware programming errors in a network instance. Note the model only defines leaves for IP route programming errors. In future it can be extended to other type of routes as well, for ex: MPLS, ethernet & etc...

### Change Scope

* Adding a new module named `programming-errors`.
$ pyang -p release/models -f tree release/models/network-instance/openconfig-programming-errors.yang
module: openconfig-programming-errors
```
  augment /oc-ni:network-instances/oc-ni:network-instance:
    +--rw programming-errors
       +--rw ip-routes
          +--rw config
          |  +--rw enabled?   boolean
          +--ro state
          |  +--ro total-errors?   oc-yang:counter64
          |  +--ro enabled?        boolean
          +--ro failed-routes
          |  +--ro failed* [prefix]
          |     +--ro prefix    -> ../state/prefix
          |     +--ro state
          |        +--ro prefix?           oc-inet:ip-prefix
          |        +--ro time?             oc-types:timeticks64
          |        +--ro dest-component*   -> /oc-platform:components/component/name
          +--ro stale-routes
          |  +--ro stale* [prefix]
          |     +--ro prefix    -> ../state/prefix
          |     +--ro state
          |        +--ro prefix?           oc-inet:ip-prefix
          |        +--ro time?             oc-types:timeticks64
          |        +--ro dest-component*   -> /oc-platform:components/component/name
          +--ro drop-routes
             +--ro drop* [prefix]
                +--ro prefix    -> ../state/prefix
                +--ro state
                   +--ro prefix?           oc-inet:ip-prefix
                   +--ro time?             oc-types:timeticks64
                   +--ro dest-component*   -> /oc-platform:components/component/name
```

### DIFF
--- a/release/models/network-instance/openconfig-programming-errors.yang
+++ b/release/models/network-instance/openconfig-programming-errors.yang
@@ -34,6 +34,9 @@ module openconfig-programming-errors {
   }
 
   grouping ip-routes-common-config {
+    description
+      "IP routes programming error common configuration parameters";
+
     leaf enabled {
       type boolean;
       description
--- a/release/models/network-instance/openconfig-programming-errors.yang
+++ b/release/models/network-instance/openconfig-programming-errors.yang
@@ -8,6 +8,7 @@ module openconfig-programming-errors {
   import openconfig-network-instance { prefix "oc-ni"; }
   import openconfig-platform { prefix "oc-platform"; }
   import openconfig-types { prefix "oc-types"; }
+  import openconfig-yang-types { prefix "oc-yang"; }
 
   organization
     "OpenConfig working group";
@@ -17,9 +18,10 @@ module openconfig-programming-errors {
     www.openconfig.net";
 
   description
-    "This module provides detailed information about the programming
+    "This module provides detailed information about the hardware programming
     state of various types of routes within a particular network instance.
-    It can be used to track where there are routing programming errors on a device.";
+    It can be used to track where there are routing hardware programming errors
+    on a device.";
 
   oc-ext:regexp-posix;
   oc-ext:catalog-organization "openconfig";
@@ -35,44 +37,47 @@ module openconfig-programming-errors {
 
   grouping ip-routes-common-config {
     description
-      "IP routes programming error common configuration parameters";
+      "IP routes hardware programming error common configuration parameters";
 
     leaf enabled {
       type boolean;
       description
-        "This leaf indicates whether IP routes programming error tracking
+        "This leaf indicates whether IP routes hardware programming error tracking
         is enabled or not";
     }
   }
 
   grouping programming-errors-top {
     description
-      "Top-level grouping for programming errors in OpenConfig.";
+      "Top-level grouping for AFT hardware programming errors in OpenConfig.";
 
     container programming-errors {
       description
-        "Detailed information about the programming state of various types of routes
-        within a particular network instance. It can be used to track where there are
-        routing programming errors on a device.";
+        "Detailed information about the AFT hardware programming state of various types of
+        routes within a particular network instance. It can be used to track where there
+        are AFT hardware programming errors on a device.";
 
       container ip-routes {
         description
-          "The programming status of IP routes within the network-instance. IP Route
-          programming reports are divided into a set of sub-categories:
-           - failed programming -- A prefix which is not present in the AFT is
-             attempted to be added into the AFT but failed.  Traffic destined for
+          "The IP AFT/routes hardware programming status within the network-instance.
+          IP Route programming reports are divided into a set of sub-categories:
+           - failed programming -- A prefix which is not present in the AFT state is
+             attempted to be added into the hardware but failed.  Traffic destined for
              this prefix will not be matched.
            - stale programming -- A prefix is already in the AFT is requested to
              be updated but failed.  Traffic destined for this prefix will be
              forwarded to the old next-hop.  AFT telemetry should continue to
              reflect the old next-hop for the prefix.
            - drop programming - these routes are explicitly programmed into hardware
-             to point to a destination that discards packets, captures both intentional
-             and un-intentional drop routes.";
+             to point to a destination that discards packets.
+
+          Lifecycle of entries/prefixes in the above 3 containers is tied to AFT
+          telemetry data. When the route gets removed from AFT telemetry data
+          corresponding entry/prefix will be deleted from above containers as well";
 
         container config {
           description
-            "IP route programming errors Configuration parameters";
+            "IP route hardware programming errors Configuration parameters";
 
           uses ip-routes-common-config;
         }
@@ -80,7 +85,13 @@ module openconfig-programming-errors {
         container state {
           config false;
           description
-            "IP route programming errors State parameters";
+            "IP route hardware programming errors State parameters";
+
+          leaf total-errors {
+            type oc-yang:counter64;
+            description
+              "Total number of IP route hardware programming errors encountered";
+          }
 
           uses ip-routes-common-config;
         }
@@ -88,7 +99,8 @@ module openconfig-programming-errors {
         container failed-routes {
           config false;
           description
-            "Surrounding container for the list of routes that fail programming.";
+            "Surrounding container for the list of routes that failed hardware
+            programming.";
 
           list failed {
             key "prefix";
@@ -156,8 +168,7 @@ module openconfig-programming-errors {
             description
               "A prefix that is currently installed in hardware but with an explicit
               instruction that it should discard packets that are destined towards
-              it. This list includes both intentional and un-intentional
-              drop routes.";
+              it.";
 
             leaf prefix {
               type leafref {
@@ -178,7 +189,7 @@ module openconfig-programming-errors {
     }
   }
 
- grouping route-state-common {
+  grouping route-state-common {
     description
       "Common parameters that correspond to a particular route type.";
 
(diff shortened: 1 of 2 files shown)


akalluru1:
@sulrich @xw-g @earies @mayukh288 for review.

release/models/network-instance/openconfig-programming-errors.yang
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    leaf time {
+      type oc-types:timeticks64;
+      description
+        "Represents the time the route programming state change was detected by the
+        monitoring subsystem expressed as nanoseconds since the Unix epoch.";
+    }
+
+    leaf dest-component {
```
bstoll, line 198: dest-component should be able to support multiple components (ie, modular chassis).

bstoll (commented):
It seems like it would be useful to have a leaf for the total number of programming errors encountered, such as following (`route-programming-errors`):

```
  augment /oc-ni:network-instances/oc-ni:network-instance:
    +--rw programming-errors
       +--rw ip-routes
          +--rw config
          |  +--rw enabled?   boolean
          +--ro state
          |  +--ro enabled?                    boolean
          |  +--ro route-programming-errors?   openconfig-yang-types:counter64
          +--ro failed-routes
          |  +--ro failed* [prefix]
...
```

This would allow an operator to see the total number of bad operations that have happened on a device, regardless of if they recover in the future.  This leaf should not count intentional drop routes, since they are not programming failures.

release/models/network-instance/openconfig-programming-errors.yang
```diff
@@ -0,0 +1,240 @@
...
+      type leafref {
+        path "/oc-platform:components/oc-platform:component/oc-platform:name";
+      }
+      description
+        "The destination component for the route programming";
+     }
+
+    leaf reason {
+      type enumeration {
```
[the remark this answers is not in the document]
akalluru1: `UNKOWN` reason covers any other programming errors which doesn't fall under `INVALID` or `HW_PROGRAMMING_FAILED`, let me see if I can find an example for such a scenario.

release/models/network-instance/openconfig-programming-errors.yang
akalluru1: Are you suggesting to convert this into a `leaf-list`?

chrisy (changes requested):
We discussed this at the Oct 18th community meeting and that raised a few areas of concern:

* There's no discussion of extant or proposed vendor support for this capability. We typically ask for examples of existing support in at least two vendors native implementations, or seek some evidence that two or more vendors could and would support it.

* Some of the language is ambiguous over whether these are focused on the layer that implements the AFT (whether hardware or software) or the network instance RIB. It ought to be the AFT layer; please consider tightening the language to make this clear everywhere.

* There is no discussion over the lifecycle of entries in these tables; in particular, with no aging mechanism they could conceivably grow to consume large amounts of memory indefinitely.

mayukh288:
> We discussed this at the Oct 18th community meeting and that review raised a few areas of concern:
> 
> * There's no discussion of extant or proposed vendor support for this capability. We typically ask for examples of existing support in at least two vendors native implementations, or seek some evidence that two or more vendors could and would support it.
> * Some of the language is ambiguous over whether these are focused on the layer that implements the AFT (whether hardware or software) or the network instance RIB. It ought to be the AFT layer; please consider tightening the language to make this clear everywhere.

--> This is indeed focused on error notifications from the AFT layer and below. We can fix up the language to emphasize that point. Do you have examples handy where the language is confusing ?

> * There is no discussion over the lifecycle of entries in these tables; in particular, with no aging mechanism they could conceivably grow to consume large amounts of memory indefinitely.
--> The number of entries in these tables will be proportional to the number of routes in the system at any point, so there won't be unbounded growth. If a route gets removed from the AFT/FIB the telemetry data for it will get deleted.

release/models/network-instance/openconfig-programming-errors.yang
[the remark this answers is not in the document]
mayukh288: We discussed this a bit offline, and my proposal is to remove the reason leaf from the various containers. The containers ( failed-routes, stale-routes, drop-routes ) by themselves give us an initial categorization of the error class. The reason here is a sub-category within that, so the reason would be specific to each category per se and should not be common. 

Currently we do not sub-categorize the routes beyond the top level category, so having this reason enum doesn't help. If at a later point we can collectively come up with applicable sub-categories for this we can add it then IMHO.

chrisy:
> > We discussed this at the Oct 18th community meeting and that review raised a few areas of concern:
> > 
> > * There's no discussion of extant or proposed vendor support for this capability. We typically ask for examples of existing support in at least two vendors native implementations, or seek some evidence that two or more vendors could and would support it.
> > * Some of the language is ambiguous over whether these are focused on the layer that implements the AFT (whether hardware or software) or the network instance RIB. It ought to be the AFT layer; please consider tightening the language to make this clear everywhere.
> 
> --> This is indeed focused on error notifications from the AFT layer and below. We can fix up the language to emphasize that point. Do you have examples handy where the language is confusing ?

I think in general this PR uses varying terminology in a few places for the same general thing; "programming" is used throughout and is taken to mean the act of programming and the programming that has already completed. It's often used without the context of what was being programmed or what it was programmed into. "The programming status of IP routes within the network-instance" is what led me to ponder this, because this expression if read only at-a-glance directs me to believe it was the network instance that was programmed, and it's only later that it mentions the AFT in some of the sub-containers. I admit that just a moments thought reveals the context and perhaps I am being nit-picky, but I think clarity goes a long way.

release/models/network-instance/openconfig-programming-errors.yang
```diff
@@ -0,0 +1,240 @@
...
+              }
+              description
+                "Reference to the prefix that keys the drop list.";
+            }
+
+            container state {
+              description
+                "Operational state parameters relating to a drop route.";
+              uses route-state-common;
+            }
```
dplore, line 174: While I understand the operational use case to detect the root cause of packet loss due to routes pointing to null, but successfully programmed drop routes are not route programming errors I think.    Routes with a null nexthop are already represented in the aft model.    Therefore, I think this container may not be necessary and is duplicative of AFT.

Here is the traversal one would need to do to resolve a route which is pointing to a null interface:
```
/openconfig-network-instance:network-instances/network-instance/afts/ipv4-unicast/ipv4-entry/prefix
/openconfig-network-instance:network-instances/network-instance/afts/next-hop-groups/next-hop-group/id
/openconfig-network-instance:network-instances/network-instance/afts/next-hops/next-hop/interface-ref/state/interface
```

I am not sure what an un-intentional drop route is?    @akalluru1 can you elaborate?

mayukh288: un-intentional drops are routes which are programmed as drop routes due to some error. There are cases where we program a drop route if the resolution of the nexthopgroup fails. Another thing to clarify is that the AFT export may not always capture drop routes in the way you indicated. We export drop routes as a prefix pointing to a NHG with no nexthop(s) in it ( an empty nexthopgroup ).

So overall the telemetry data here is trying to capture the case that there is a drop route programmed in the hw, trying to figure whether that is due to an error/at which layer that error is would be better served by querying other parts of the model like the RIB/looking at the config.

akalluru1: I've removed the un-intentional/intentional terminology from the latest diff. Please take a look & let us know if something is not clear.

release/models/network-instance/openconfig-programming-errors.yang
akalluru1: Changed it to `leaf-list` in latest diff.

release/models/network-instance/openconfig-programming-errors.yang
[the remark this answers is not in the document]
akalluru1: As mentioned by Mayukh above `reason` leaf has been removed in latest diff.

akalluru1:
> 

Included new leaf called `total-errors`

akalluru1:
> > > We discussed this at the Oct 18th community meeting and that review raised a few areas of concern:
> > > 
> > > * There's no discussion of extant or proposed vendor support for this capability. We typically ask for examples of existing support in at least two vendors native implementations, or seek some evidence that two or more vendors could and would support it.
> > > * Some of the language is ambiguous over whether these are focused on the layer that implements the AFT (whether hardware or software) or the network instance RIB. It ought to be the AFT layer; please consider tightening the language to make this clear everywhere.
> > 
> > 
> > --> This is indeed focused on error notifications from the AFT layer and below. We can fix up the language to emphasize that point. Do you have examples handy where the language is confusing ?
> 
> I think in general this PR uses varying terminology in a few places for the same general thing; "programming" is used throughout and is taken to mean the act of programming and the programming that has already completed. It's often used without the context of what was being programmed or what it was programmed into. "The programming status of IP routes within the network-instance" is what led me to ponder this, because this expression if read only at-a-glance directs me to believe it was the network instance that was programmed, and it's only later that it mentions the AFT in some of the sub-containers. I admit that just a moments thought reveals the context and perhaps I am being nit-picky, but I think clarity goes a long way.

Updated the language/description to be more clear & precise to provide complete context. Please take a look at latest diff & let us know if something doesn't make sense.

akalluru1:
@bstoll @chrisy sorry for the delay in addressing your comments. Can you please take a look at the latest changes?

chrisy:
> @bstoll @chrisy sorry for the delay in addressing your comments. Can you please take a look at the latest changes?

I am comfortable with the documentation updates.

bstoll:
I also like the latest proposed changes.

akalluru1:
> I also like the latest proposed changes.

would you be able to approve this PR? what are the next steps to merge it up? @bstoll @dplore

dplore:
@earies @sulrich @aaronmillisor @aredmon8551  for visibility

dplore (approved):
Thanks for the contribution @akalluru1

PR #727, opened 2022-10-18
decap-then-lookup OC schema

xw-g:
Propose an OC schema to support/unify the representation of use cases where
- The decapsulation action is controlled by SDN controllers populated network-instance (e.g. via gRIBI). 
- And the lookup on the new header happens in another network instance, which is chosen by fields that are not used for resolution.

This will be useful for chained decap-then-lookup TE scenarios.

```
  +--rw network-instances
        +--rw policy-forwarding
           +--rw policies
              +--rw policy* [policy-id]
                 +--rw rules
                    +--rw rule* [sequence-id]
                       +--rw action
                          +--rw config
                             +--rw discard?                          
                             +--rw decapsulate-gre?                  
                             +--rw decap-network-instance?            <<<<<<<<<<< new introduced leaf.
                             +--rw decap-fallback-network-instance?   <<<<<<<<<<< new introduced leaf.
                             +--rw post-decap-network-instance?       <<<<<<<<<<< new introduced leaf.
                             +--rw network-instance?                 
                             +--rw path-selection-group?              
                             +--rw next-hop?                          
                             +--rw decapsulate-mpls-in-udp?          
                             +--rw decapsulate-gue?                  
```


xw-g:
+ @rolandphung and @aaronmillisor to help review.

nokia1adam:
I feel this modeling is too specific to one use case. I am wondering if it can be generalized.

A packet matching a PBF rule with this new decap-lookup action is not guaranteed to be a packet that can be decapsulated and even if it can be decapsulated, the PBF action itself has no control on whether the decapsulation is done or not -- this outcome is entirely based on whether the referenced network-instance (or its fallback) has a matching route with next-hop action that performs decapsulation.

Given this, it seems to me that you could support the new functionality just by introducing one new PBF action called fallback-network-instance at the same level as the existing actions.

A complete action configuration could be:
action
+--network-instance <foo>
+--fallback-network-instance <bar>

If an IPinP packet matches this rule, there will be a first lookup in <foo> and if there is a matching entry in <foo> it may or may not perform decapsulation; the OC model is agnostic. If the first lookup in <foo> finds no matching route, a second lookup will occur in <bar>. Again, the matching entry in <bar> may or may not perform decapsulation, and the OC model should not care.

My 2 cents.

xw-g:
> I feel this modeling is too specific to one use case. I am wondering if it can be generalized.
> 
> A packet matching a PBF rule with this new decap-lookup action is not guaranteed to be a packet that can be decapsulated and even if it can be decapsulated, the PBF action itself has no control on whether the decapsulation is done or not -- this outcome is entirely based on whether the referenced network-instance (or its fallback) has a matching route with next-hop action that performs decapsulation.
> 
> Given this, it seems to me that you could support the new functionality just by introducing one new PBF action called fallback-network-instance at the same level as the existing actions.
> 
> A complete action configuration could be: action +--network-instance +--fallback-network-instance
> 
> If an IPinP packet matches this rule, there will be a first lookup in and if there is a matching entry in it may or may not perform decapsulation; the OC model is agnostic. If the first lookup in finds no matching route, a second lookup will occur in . Again, the matching entry in may or may not perform decapsulation, and the OC model should not care.
> 
> My 2 cents.

I think your proposal missed one more NI that's to be used for post decap lookup? We can sync offline to chat more.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,6 +270,41 @@ submodule openconfig-pf-forwarding-policies {
         according to the IP destination).";
     }
 
+    container decap-lookup {
```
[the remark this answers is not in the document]
xw-g: Done. Can you help take a look? Thanks!

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,6 +270,46 @@ submodule openconfig-pf-forwarding-policies {
...
+          - The decapsulation decision to be controlled by the
+            controllers.
+          - The new header after decapsulation to be looked up in
+            another NI, and the NI is chosen by fields (e.g. source IP,
+            DSCP, etc) that are not used for route resolution.
+        When this leaf is set, packets matching the match criteria for
+        the forwarding rule should be looked up in the specified NI.
+        It is expected that the NI should only contain routes that have
+        next hop action as decapsualtion.
+          - If there is a match, the packet should be decapsulated and
```
aredmon8551, line 291: What if a match occurs for a route with a NH action other than decapsulation (i.e., a successful LPM result is obtained but the resulting NH does not have a decap action)?  I'm assuming based on https://github.com/openconfig/public/pull/727/files#diff-6e72bad429863fd2ad7d49eb5ef687e6e158ea87b354c7582125144242b4d798R310 that we should fallback

xw-g:
@nokia1adam, @rolandphung @aaronmillisor, Can you help review again? Thanks.

nokia1adam:
Looks OK to me. The description for decap-network-instance states that this network-instance is expected to have entries that perform decapsulation as a next-hop action. What if it has other types of entries? Should those be ignored so they can never match? I think some text should be included about this.

Why are there decapsulate-mpls-in-udp, decapsulate-gue and decapsulate-gre actions but no decapsulate-ip-in-ip action? Including one of these in combination with decap-network-instance could further specify the types of decapsulation entries that are expected in the decap-network-instance.

xw-g:
> Looks OK to me. The description for decap-network-instance states that this network-instance is expected to have entries that perform decapsulation as a next-hop action. What if it has other types of entries? Should those be ignored so they can never match? I think some text should be included about this.
>

I think the current wording, ` It is expected that the NI should only contain routes that have next hop action as decapsualtion` indicates that vendors can implement needed precondition to avoid having such entries? 

> Why are there decapsulate-mpls-in-udp, decapsulate-gue and decapsulate-gre actions but no decapsulate-ip-in-ip action? Including one of these in combination with decap-network-instance could further specify the types of decapsulation entries that are expected in the decap-network-instance.

The usage of those leaves are for the policy rules that does match and then the specific decapsulation action. However, this PR is about rely on NI that is controlled by SDN controllers who push the decapsulation action.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,14 +270,72 @@ submodule openconfig-pf-forwarding-policies {
...
+        populated by SDN controllers (e.g. via gRIBI), where we want:
+          - The decapsulation decision to be controlled by the
+            controllers.
+          - The new header after decapsulation is to be looked up in
+            another NI, and the NI is chosen by fields (e.g. source IP,
+            DSCP, etc) that are not used for route resolution.
+        When this leaf is set, packets matching the match criteria for
+        the forwarding rule should be looked up in the specified NI.
+        It is expected that the NI should only contain routes that have
+        next hop action as decapsualtion.
```
adrawatcisco, line 291: Typo "decapsulation"

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,14 +270,72 @@ submodule openconfig-pf-forwarding-policies {
...
+        // this must be an absolute reference to allow another NI
+        // to be referenced.
+        path "/network-instances/network-instance/config/name";
+      }
+      description
+        "This leaf has to be set when ../decap-network-instance is set.
+        When this leaf is set, the specified NI should be used as a
+        fallback instance for lookup when
+        ../decap-network-instance produces no match for the packet.
+        NI specified by this leaf might conatin different mixed next hop
```
adrawatcisco, line 310: Typo "conatin"

adrawatcisco:
Do you foresee any further updates to the PR?

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
xw-g: done. Thanks.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
xw-g: Done. Thnaks!

xw-g:
> Do you foresee any further updates to the PR?

Not for now.

xw-g:
@adrawatcisco, @nokia1adam, @rolandphung, approve?

rolandphung:
Just a nit on the process, can we get the relevant pyang output in the description?

xw-g:
> Just a nit on the process, can we get the relevant pyang output in the description?

Ah, forgot that. Done. Thanks!

aaronmillisor:
This looks good for cisco.

aaronmillisor (approved):
lgtm.

rolandphung (approved):

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,14 +270,72 @@ submodule openconfig-pf-forwarding-policies {
...
+      }
+      description
+        "This leaf is mutually exclusive with ../network-instance.
+        This leaf is normally used with NIs (network instances) that are
+        populated by SDN controllers (e.g. via gRIBI), where we want:
+          - The decapsulation decision to be controlled by the
+            controllers.
+          - The new header after decapsulation is to be looked up in
+            another NI, and the NI is chosen by fields (e.g. source IP,
+            DSCP, etc) that are not used for route resolution.
```
dplore, line 287: Does this help further clarify the use case?

```suggestion
          - The new header after decapsulation is to be looked up in
            another NI.  The target NI is set by the SDN controller
            possibly using fields that are not use for route resolution
            (e.g. source IP, DSCP, etc).
```

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,14 +270,72 @@ submodule openconfig-pf-forwarding-policies {
...
+        When this leaf is set, packets matching the match criteria for
+        the forwarding rule should be looked up in the specified NI.
+        It is expected that the NI should only contain routes that have
+        next hop action as decapsulation.
+          - If there is a match, the packet should be decapsulated.
+            The route lookup on the new header should happen in the
+            ../post-decap-network-instance.
+          - If there is no match, the packet should be looked up in the
+            NI set by ../decap-fallback-network-instance.";
```
dplore, line 296: How is this leaf different than a leaf network-instance which simply contains match criteria which all have decap action?  In other words, do we need a decap-network-instance leaf at all?  Could we use some network-instance populated exclusively with match criteria with decap action set?  The network-instance name could include the substring 'decap' as a hint to it's purpose.

dplore: Is the difference that  `post-decap-network-instance` is expected to be used together / chained  with `decap-network-instance`?  If so, perhaps it will help to also add this expectation into the description for `decap-network-instance`.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
xw-g: It might not be SDN controllers/pipeline that populate the fields. It might be config generation pipelines. Therefore, I didn't plan to call it out specifically.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
xw-g: Updated the description. Thanks!

Regarding why we don't "reuse" the existing "network-instance", we think explicitly setting a leave is more reliable than string parsing the VRF name. it gives an explicit signal which would make implementation robust.

xw-g: Updated the description. Thanks!

Regarding why we don't "reuse" the existing "network-instance", we think explicitly setting a leave is more reliable than string parsing the VRF name. it gives an explicit signal which would make implementation robust.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: SGTM

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: SGTM

dplore (approved):

PR #732, opened 2022-11-01
Clarify static route index description to use a number

dplore:
### Change Scope
This brings static route index in line with with other openconfig models containing indicies for nexthops such as the aft model.  This change is not backwards compatible.

###References
* [openconfig-aft-common.yang](https://github.com/openconfig/public/blob/00a27a5f0f3c472205ef293ab238e0530b610811/release/models/aft/openconfig-aft-common.yang#L268-L272)

*  https://github.com/openconfig/featureprofiles/pull/692

### DIFF
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -210,12 +210,13 @@ module openconfig-local-routing {
       specified for a static route";
 
     leaf index {
-      type uint64;
+      type string;
       description
         "An user-specified identifier utilised to uniquely reference
         the next-hop entry in the next-hop list. The value of this
         index has no semantic meaning other than for referencing
-        the entry.";
+        the entry.  It is observed that implementations typically 
+        only support a number value for this string. ";
     }
 
     leaf next-hop {
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -204,7 +204,7 @@ module openconfig-local-routing {
       specified for a static route";
 
     leaf index {
-      type string;
+      type uint64;
       description
         "An user-specified identifier utilised to uniquely reference
         the next-hop entry in the next-hop list. The value of this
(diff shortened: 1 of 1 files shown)


robshakir:
What is the motivation for this change please? It's a pretty major backwards incompatible change for a model that is in use, so the bar should be high. One can also use an integer marshalled to a string if needed - so it seems like whilst this might not be ideal, it would be compatible.

brwallis:
The history for this is in the featureprofiles repo (https://github.com/openconfig/featureprofiles/pull/692).

I do think it's acceptable to say that the values must be valid integers from an implementation standpoint, but as was seen in featureprofiles this likely needs to be something that is explicit (hence the PR). In the AFT model this value is a uint64, and since it is named as an index I think most implementations assume it is an integer.

dplore:
Discussion in OC operators meeting prefers to update the description to clarify that this leaf should be a number, but remains string data type for backwards compatibility.

brwallis:
That sounds OK to me, if this didn't pop up in featureprofiles (and the tests all used integers) I don't think it would have been flagged. Is there some cadence in which OC would accept NBC changes like this? It does _seem_ like the right thing to do, but obviously NBC changes can be incredibly disruptive.

xw-g:
cc: @xw-g

dplore:
I have revised this PR to only update the description as a clarification.

dplore:
> That sounds OK to me, if this didn't pop up in featureprofiles (and the tests all used integers) I don't think it would have been flagged. Is there some cadence in which OC would accept NBC changes like this? It does _seem_ like the right thing to do, but obviously NBC changes can be incredibly disruptive.

@brwallis see #720 for proposal for release management to include periodic non-breaking releases followed by less frequent breaking releases.  Comments on the release process are appreciated!

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -209,7 +215,8 @@ module openconfig-local-routing {
         "An user-specified identifier utilised to uniquely reference
         the next-hop entry in the next-hop list. The value of this
         index has no semantic meaning other than for referencing
-        the entry.";
+        the entry.  It is observed that implementations typically
+        only support a number value for this string. ";
```
robshakir, line 219:
```suggestion
        only support a numeric value for this string. ";
```

robshakir (approved):

PR #736, opened 2022-11-07
Normalize and remove single 'unique' statement among model set

earies:
    - Remove network-instance interface list unique statement

### Change Scope

This PR is to initiate discussion on the path forward for the use of the YANG `unique` statement in OpenConfig models.

Per https://www.rfc-editor.org/rfc/rfc6020#section-7.8.3, the `unique` statement is meant to put constraints on non-key list entry child nodes however the use throughout OpenConfig models is limited to a single use: `/network-instances/network-instance/interfaces/interface` and is a valid case of the use of this statement however for consistency purposes, I would like to raise the point that its use should either be expanded to all cases of lists where it is applicable throughout the model set or propose the removal of this statement until further decision and backwards incompatible model updates across the model set (e.g. all or nothing).

Per YANG backwards compatibility rules, this change is backwards incompatible in either case and major revision numbers need to be incremented.

### Platform Implementations

N/A

### DIFF
--- a/release/models/network-instance/openconfig-network-instance-l2.yang
+++ b/release/models/network-instance/openconfig-network-instance-l2.yang
@@ -24,12 +24,12 @@ submodule openconfig-network-instance-l2 {
     Layer 2 network instance configuration and operational state
     parameters.";
 
-  oc-ext:openconfig-version "2.0.0";
+  oc-ext:openconfig-version "1.5.0";
 
-  revision "2022-11-07" {
+  revision "2022-12-14" {
     description
       "Removal of interface list unique statement";
-    reference "2.0.0";
+    reference "1.5.0";
   }
 
   revision "2022-09-15" {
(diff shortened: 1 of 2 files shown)


earies:
@dplore @robshakir PTAL thx

jsterne:
Agree this one use of "unique" feels a bit out of place (not technically incorrect, but inconsistent).

dplore:
My preference is to add unique to lists which we know in implementation are in fact, unique.   In addition, at least some lists are also appropriate to be keyed lists such as  #747  #750  #752

robshakir:
I agree with @earies here -- let's not make this PR a "fix all places where unique COULD be specified through the models". I think generally there are constraints other than uniqueness that we could express, but this is a major undertaking and not clear how it is actually achieved within reasonable use of the YANG language that tooling actually supports.

So, +1 to removing this unique statement.

robshakir:
BTW, given that this makes the model less constrained, I don't think it needs a major revision change.

dplore (approved):

earies:
Thx @dplore @robshakir - I've gone ahead and relaxed the revision change to 1.5.0 - we'll keep the versioning bound to the commits to keep the history vs. combine multiple PRs into a new aggregate PR

Issue #733, opened 2022-11-03
Update KeyChain model to accept hex string for key-id 


m26singhvi:
As part of https://github.com/openconfig/public/issues/606, Macsec model was updated to use global key chain model. However,  this will not work because key-id type in key chain model is U64 : 
```
leaf key-id {
      type uint64;
      description
        "Identifier for the key within the keychain.";
    }
```

whereas prior to the above change, key id in Macsec was hex string as per below : 

```
grouping macsec-mka-key-config {
    description
      "MKA Key config grouping";

    leaf id {
      type oc-yang:hex-string {
        length "1..64";
      }
      description
        "Key identifier is used as the
         Connectivity Association Key name (CKN)";
    }
```

So, currently this is  broken and proposal is to make the `key-id` in keychain to be a union of hex string and U64.

dplore:
Thanks @m26singhvi ,   can you raise a pull request to add this?  Please also reference at least two implementations and/or specifications that show the need for a hex-string as macsec key id.

m26singhvi:
Thanks for taking a look at this @dplore. I have raised a pull request here - https://github.com/openconfig/public/pull/737

As I captured in the issue itself, before #606 was fixed, Macsec was using hex string only as key identifier.  801.X-2020 has details about how a typical CKN ( CAK Name ) looks like.

dplore:
Implemented in #737 which was merged.

PR #737, opened 2022-11-08
Update KeyChain model to accept hex string for key-id #733

m26singhvi:
### Change Scope

* Update KeyChain model to accept hex string for key-id
*  Since  a union is being proposed here, this change is backward compatible.


dplore:
For references this appears compatible with

[Cisco IOS XR keychain key-id](https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-0/security/configuration/guide/sc40crsbook_chapter5.html)

[JunOS key-identifier](https://www.juniper.net/documentation/us/en/software/junos/user-access/topics/ref/statement/key-edit-security-authentication-key-chains.html)

dplore (approved):
LGTM.  I am requesting last call for comments. To be merged December 2nd, 2022

PR #738, opened 2022-11-08
Removal of network-instance enabled-address-families leaf-list

earies:
  * (M) release/models/network-instance/openconfig-network-instance-l3.yang
    - Removal of unnecessary enabled-address-families leaf-list node

### Change Scope

The `enabled-address-families` leaf-list currently exists at the top level of network-instances, lacks implementation guidance and is questionable in operational reality.

* This node appears specifically derived from one implementations native behaviors specific to IP based VRFs - Are there vendor references to this leaf-list?
* The model defines this at an AF level only but the single implementation above does this based on AF + SAF
* The model uses an identityref that does not gate the other possible identities (L2_ETHERNET, MPLS + any future) so it would require implementation detail to gate this outside of the data-model
* Implementation guidance, expectations or vendor support is not documented
* Is this really a sane operational workflow for what are meant to be operator use-cases?  Would anyone ever enable/disable L3 protocols such as IPv4 or IPv6 per instance at runtime?  What if all child nodes (such as bound interfaces or protocols) contain configuration for these protocols?  What is the expected behavior?  Would all such instances cease to function?

This PR is to propose removal of undefined and undocumented behavior for this leaf-list.

Due to the nature of this change and removal of the leaf-list, this change is deemed backwards incompatible thus the major version is incremented.

### Platform Implementations

N/A: Lacks overall purpose and appears to be derived from a single native implementation.

### DIFF
--- a/release/models/network-instance/openconfig-network-instance-l2.yang
+++ b/release/models/network-instance/openconfig-network-instance-l2.yang
@@ -24,7 +24,13 @@ submodule openconfig-network-instance-l2 {
     Layer 2 network instance configuration and operational state
     parameters.";
 
-  oc-ext:openconfig-version "1.4.0";
+  oc-ext:openconfig-version "2.0.0";
+
+  revision "2022-11-08" {
+    description
+      "Removal of top-level enabled-address-families leaf-list";
+    reference "2.0.0";
+  }
 
   revision "2022-09-15" {
     description
--- a/release/models/network-instance/openconfig-network-instance-l3.yang
+++ b/release/models/network-instance/openconfig-network-instance-l3.yang
@@ -23,7 +23,13 @@ module openconfig-network-instance-l3 {
     Layer 3 network instance configuration and operational state
     parameters.";
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "2.0.0";
+
+  revision "2022-11-08" {
+    description
+      "Removal of top-level enabled-address-families leaf-list";
+    reference "2.0.0";
+  }
 
   revision "2022-09-15" {
     description
@@ -125,21 +131,6 @@ module openconfig-network-instance-l3 {
       "Configuration and operational state parameters relevant
       to network instances that include a Layer 3 type";
 
-  }
-
-  grouping l3ni-instance-common-config {
-    description
-      "Configuration parameters that are common to L3 network
-      instances other than the default instance";
-
-    leaf-list enabled-address-families {
-      type identityref {
-        base octypes:ADDRESS_FAMILY;
-      }
-      description
-        "The address families that are to be enabled for this
-        network instance.";
-    }
   }
 
   grouping l3ni-route-limit-structural {
--- a/release/models/network-instance/openconfig-network-instance.yang
+++ b/release/models/network-instance/openconfig-network-instance.yang
@@ -293,7 +293,12 @@ module openconfig-network-instance {
         key "name";
 
         description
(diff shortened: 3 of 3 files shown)


nokia1adam:
I fully support this PR. It is completely unclear what 'enabled' means in this context. Is it meant to imply "forwarding" behaviors or does it also extend to control plane population of routes in said address families? Without guidance on what this means I think the OC model would be better off without this leaf-list at all.

earies:
@dplore @robshakir - PTAL

There are many nodes across the model set that came by way of initial publish with no vendor reference or implementation guidance and the suggestion should be to remove all such nodes in this state or where _all_ implementations have had to deviate as not-supported

earies:
@wenovus - checking back if you have any comments/feedback... There are a number of early/initial model publishes that have either no mapping to vendor implementation or implementation guidance/behavior as such that should be removed.  Thx

robshakir:
The intention here way to map to the configuration structure of a couple of NOSes:

**IOS XE** ([Reference](https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/mp_l3_vpns/configuration/xe-16/mp-l3-vpns-xe-16-book/ipv6-vrf-aware-system-message-logging.html))

```
vrf definition FOO
  address-family ipv4
  address-family ipv6
```

**Arista EOS** ([Reference](https://www.arista.com/en/um-eos/eos-bgpmpls-l3-vpn?searchword=eos%2041%204%20vm%20tracer%20configuration%20commands))

```
ip routing vrf FOO
ipv6 routing vrf FOO
```

AIUI, these are the two implementations that support this configuration, whereas JUNOS and SROS do not support enabling specific address families within a particular VRF. I think that the decision that we need to make here is whether we need to have means to actively enable IPv4+IPv6 within a VRF, or whether both are simply implicitly enabled.

CC: @rolandphung @brwallis @rgwilton

robshakir:
To clarify, the use case here was originally to ensure that as an implementation was translating to CLI it could determine that it needed to populate these commands. For native implementations I do not think this is relevant, but what I would like to replace it with is some base expectation that says something like:

> IPv4 and IPv6 forwarding are enabled by default within a VRF, and subsequently, routes can be populated into the VRF using protocols that enable IPv4 and IPv6 configuration without explicitly enabling these.

We can support this with a feature profiles test that validates this fact.

dplore:
@earies if you want to address the comment from Rob regarding description update, we can mark this approved.  Once approved, you may merge this with #749 if you prefer.  I am ok with multiple version bumps or combining.  Your preference.

earies:
Thx @dplore + @robshakir - I've gone ahead and updated the description containing the above suggestion to the `network-instance` list statement.

robshakir:
Thanks @earies -- I've added a test in the feature profiles repo (https://github.com/openconfig/featureprofiles/pull/876) to validate this. It currently runs against OTG, I'll ping the various teams in that PR to run it against their virtualised images.

earies:
We can keep the revision as-is vs. combining to maintain the true revision/commit history here.  For the other PR addressing these same models, I'll adjust the revision bump depending on order of merge

sachendras (approved):

vcheny (approved):

dplore:
Scheduled for review at Dec 20th OpenConfig operators meeting.

dplore:
Approved by openconfig operators meeting.    Seeing as there is only support, no objection and that this PR have been open since Nov 8th, I will merge now

dplore (approved):

PR #739, opened 2022-11-09
Update openconfig-policy-types.yang

ElodinLaarz:
Add new INSTALL_PROTOCOL_TYPE local to openconfig-policy-types.yang.

### Change Scope

We add a new install protocol type named local to openconfig-policy-types.yang.

This addition causes no backwards compatibility problems, only adding to OC.

### Platform Implementations

 * Cisco Reference and Implementation Output: [link to documentation](https://www.cisco.com/c/en/us/support/docs/ip/ip-routing/116264-technote-ios-00.html)
 * Juniper Reference and Implementation Output: [link to documentation](https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/command/show-route-detail.html)

### DIFF
--- a/release/models/policy/openconfig-policy-types.yang
+++ b/release/models/policy/openconfig-policy-types.yang
@@ -275,7 +275,7 @@ module openconfig-policy-types {
 
       10.244.136.79/31
 
-      the derived direct route is
+      the derived DIRECTLY_CONNECTED route is
 
       10.244.136.78/31
 
--- a/release/models/policy/openconfig-policy-types.yang
+++ b/release/models/policy/openconfig-policy-types.yang
@@ -259,7 +259,7 @@ module openconfig-policy-types {
     reference
       "RFC 5440";
   }
-  
+
   identity LOCAL {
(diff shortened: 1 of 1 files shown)


release/models/policy/openconfig-policy-types.yang
```diff
@@ -253,4 +259,28 @@ module openconfig-policy-types {
...
+      For example, when configuring an interface with the ip
+      address
+
+      10.244.136.79/31
+
+      the derived direct route is
+
+      10.244.136.78/31
+
+      and the derived local route is
```
dplore, line 282:
```suggestion
      and the derived LOCAL route is
```

dplore (commented):
I suggest using referring the to the identity names defined in the model

dplore:
@oscargdd volunteered to add links to additional vendor implementation of LOCAL routes

dplore (approved):

dplore:
Last call for comments on this change.  We will merge in two weeks on Nov 29, 2022

PR #741, opened 2022-11-10
Add support for IPv6 RA features.

robshakir:
```
 * (M) release/models/interfaces/openconfig-if-ip.yang
   - Add support for explicit configuration of the M and O bits.
   - Add support for configuration of prefixes within an RA and the
     flags associated with them.
```

### Change Scope

This change adds support for setting the M and O bits within an RA advertised by a system configured for IPv6.

Additionally, this change adds support for prefixes being advertised within an RA, and specifying the flags that are described by RFC4861 for such prefixes.

The change is backwards compatible (new fields added only).

### Platform Implementations

 * JUNOS [documentation](https://www.juniper.net/documentation/us/en/software/junos/neighbor-discovery/topics/topic-map/ipv6-neighbor-discovery.html)
   - Juniper supports all fields but does not appear to support the ability to disable an advertisement.
 * Arista EOS [documentation](https://www.arista.com/en/um-eos/eos-ipv6).
   - Arista supports all fields in this change.
 * Cisco IOS XR [documentation](https://www.cisco.com/c/en/us/td/docs/ios_xr_sw/iosxr_r3-7/addr_serv/command/reference/ir37ntsk.html)
   - XR supports all fields in this change.
 * Nokia SRLinux [documentation](https://yang.srlinux.dev/v22.3.2/tree.html) (see `/interface/subinterface/ipv6/router-advertisement` and child paths)
   - SRLinux does not support the ability to disable an advertisement

The main discussion point based on these platform implementations is whether the disable advertisement leaf should be included.

### Tree Diff

This diffs current `HEAD` against the change.

```diff
▶ diff -prauwN tree1 tree2
--- tree1	2022-11-09 16:27:45.000000000 -0800
+++ tree2	2022-11-09 16:27:52.000000000 -0800
@@ -134,10 +134,31 @@ module: openconfig-if-ip
        |  |  +--rw interval?   uint32
        |  |  +--rw lifetime?   uint32
        |  |  +--rw suppress?   boolean
+       |  |  +--rw managed?        boolean
+       |  |  +--rw other-config?   boolean
        |  +--ro state
-       |     +--ro interval?   uint32
-       |     +--ro lifetime?   uint32
-       |     +--ro suppress?   boolean
+       |  |  +--ro interval?       uint32
+       |  |  +--ro lifetime?       uint32
+       |  |  +--ro suppress?       boolean
+       |  |  +--ro managed?        boolean
+       |  |  +--ro other-config?   boolean
+       |  +--rw prefixes
+       |     +--rw prefix* [cidr]
+       |        +--rw cidr      -> ../config/cidr
+       |        +--rw config
+       |        |  +--rw cidr?                        oc-inet:ipv4-prefix
+       |        |  +--rw valid-lifetime?              uint32
+       |        |  +--rw preferred-lifetime?          uint32
+       |        |  +--rw disable-advertisement?       boolean
+       |        |  +--rw disable-autoconfiguration?   boolean
+       |        |  +--rw enable-onlink?               boolean
+       |        +--ro state
+       |           +--ro cidr?                        oc-inet:ipv4-prefix
+       |           +--ro valid-lifetime?              uint32
+       |           +--ro preferred-lifetime?          uint32
+       |           +--ro disable-advertisement?       boolean
+       |           +--ro disable-autoconfiguration?   boolean
+       |           +--ro enable-onlink?               boolean
        +--rw neighbors
        |  +--rw neighbor* [ip]
        |     +--rw ip        -> ../config/ip
@@ -319,10 +340,31 @@ module: openconfig-if-ip
        |  |  +--rw interval?   uint32
        |  |  +--rw lifetime?   uint32
        |  |  +--rw suppress?   boolean
+       |  |  +--rw managed?        boolean
+       |  |  +--rw other-config?   boolean
        |  +--ro state
-       |     +--ro interval?   uint32
-       |     +--ro lifetime?   uint32
-       |     +--ro suppress?   boolean
+       |  |  +--ro interval?       uint32
+       |  |  +--ro lifetime?       uint32
+       |  |  +--ro suppress?       boolean
+       |  |  +--ro managed?        boolean
+       |  |  +--ro other-config?   boolean
+       |  +--rw prefixes
+       |     +--rw prefix* [cidr]
+       |        +--rw cidr      -> ../config/cidr
+       |        +--rw config
+       |        |  +--rw cidr?                        oc-inet:ipv4-prefix
+       |        |  +--rw valid-lifetime?              uint32
+       |        |  +--rw preferred-lifetime?          uint32
+       |        |  +--rw disable-advertisement?       boolean
+       |        |  +--rw disable-autoconfiguration?   boolean
+       |        |  +--rw enable-onlink?               boolean
+       |        +--ro state
+       |           +--ro cidr?                        oc-inet:ipv4-prefix
+       |           +--ro valid-lifetime?              uint32
+       |           +--ro preferred-lifetime?          uint32
+       |           +--ro disable-advertisement?       boolean
+       |           +--ro disable-autoconfiguration?   boolean
+       |           +--ro enable-onlink?               boolean
        +--rw neighbors
        |  +--rw neighbor* [ip]
        |     +--rw ip        -> ../config/ip
```

### DIFF
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -941,7 +941,7 @@ module openconfig-if-ip {
       router advertisement.";
 
     leaf prefix {
-      type oc-inet:ipv4-prefix;
+      type oc-inet:ipv6-prefix;
       description
         "IPv6 prefix to be advertised within the router advertisement
         message.";
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -44,7 +44,13 @@ module openconfig-if-ip {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "3.0.0";
+  oc-ext:openconfig-version "3.1.0";
+  
+  revision "2022-11-09" {
+    description
+      "Add additional IPv6 router-advertisement features.";
+    reference "3.1.0";
+  }
 
   revision "2019-01-08" {
     description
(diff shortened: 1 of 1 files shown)


release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -907,6 +913,84 @@ module openconfig-if-ip {
...
+      default false;
+      description
+        "When set to true, the other configuration (O) flag is set in the
+        advertised router advertisement. The O flag indicates that there is
+        other configuration available via DHCPv6 (e.g., DNS servers).";
+      reference "RFC4861: Neighbor Discovery for IPv6, section 4.2";
+    }
+  }
+
+  grouping ipv6-ra-prefix-config {
```
rolandphung, line 938: It might be worth considering to have some of these leaves be defaulted to what is defined in https://datatracker.ietf.org/doc/html/rfc4861#section-6.2.1

robshakir:
Credit to kiakhontov@google.com for much of the research here.

release/models/interfaces/openconfig-if-ip.yang
dplore: Some of the leafs above for the configuratoin of ipv6 router advertisements per interface do have defaults as specified in the rfc.  But the options for prefixes ([rfc4861 section-4.6.2](https://datatracker.ietf.org/doc/html/rfc4861#section-4.6.2)) don't seem to have defaults?

dplore (approved):
The model LGTM.   Thanks for the thorough investigation of the implementation support.

dplore:
Last call for comments.   Will merge on Jan 17th.

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -907,6 +913,84 @@ module openconfig-if-ip {
...
+    description
+      "Configuration parameters for an individual prefix within an IPv6
+      router advertisement.";
+
+    leaf prefix {
+      type oc-inet:ipv4-prefix;
```
dplore, line 944:
```suggestion
      type oc-inet:ipv6-prefix;
```

dplore (changes requested):
found a typo

dplore:
Fixed typo and merging

Issue #726, opened 2022-10-18
Should network-instance type be mandatory?


dplore:
network-instance type is currently optional and has no default.  When configuring a network instance, it is currently ambiguous as to which type a network operating system (NOS) will use.  

It seems network-instance type should be mandatory in order to get deterministic behavior.  Modifying the current leaf would be a breaking change.

Model Reference:
https://github.com/openconfig/public/blob/f3f05d5f6e38ca7c46671ea5a4c11b62d264c10b/release/models/network-instance/openconfig-network-instance.yang#L1193-L1208

featureprofiles test which exposed this issue:
https://github.com/openconfig/featureprofiles/pull/624

robshakir:
We have generally not used mandatory markings in the models. I believe
there is one instance of it - since there was discussion as to whether this
could be consistently done across implementations, and further discussions
around the lack of presence containers in OpenConfig.

Before adding this specifically can we make sure we think about cross model
consistency please?

r.

earies:
There are currently 17 instances of `mandatory=true` across the OC model set

The looseness of OC models by not implementing stricter checks and relationships is a double-edged sword.  The looser the model, the more precise specifications will need to be in order to give implementation guidance and consistency.  Without this you can expect implementation differences across the board.

For this case, since `type` is optional, we have undefined behavior as a result (There are many other cases of this as well).  This is currently being triggered by way of ONDATRA tests that assume some level of pre-configuration but is not limited to such.

An implementation can always produce a deviation, make this mandatory on the server (or handle purely internal) but this should be covered in a spec at minimum.  Having it directly encoded in the YANG model means tooling can understand this whereas all side channel specification guidance needs to be manually codified with a constant understanding of spec iteration.

What do we think the harm is in making this leaf mandatory since in theory it already is?

PR #749, opened 2022-11-19
Network-instance type/enabled and YANG/IETF type cleanup

earies:
    - Enforce network-instance type as mandatory for all instances
    - Removal of top-level network-instance enabled leaf
    - Refactor IETF types to OpenConfig types

### Change Scope

Similar to https://github.com/openconfig/public/pull/738 (and can get consolidated if agreed upon as to not have back to back major revision bumps).

This change-set encompasses 3 changes and is backwards incompatible

1. Enforce the network-instance `type` to be mandatory per YANG model.  This is being discussed in https://github.com/openconfig/public/issues/726
2. Removal of the top-level network-instance `enabled` leaf.  There is no implementation reference or guidance for various instance types including `DEFAULT_INSTANCE` and question it's usage in operations
3. Refactoring legacy IETF type references to OpenConfig types

### Platform Implementations

N/A: Mostly cleanup effort and removing nodes not in use without guidance or implementation reference otherwise you can expect deviations, implementation differences or undefined behavior.

### DIFF
--- a/release/models/network-instance/openconfig-network-instance-l2.yang
+++ b/release/models/network-instance/openconfig-network-instance-l2.yang
@@ -7,8 +7,8 @@ submodule openconfig-network-instance-l2 {
   // import some basic types
   import openconfig-extensions { prefix "oc-ext"; }
   import openconfig-interfaces { prefix "oc-if"; }
-  import ietf-yang-types { prefix "yang"; }
-  import ietf-inet-types { prefix "inet"; }
+  import openconfig-yang-types { prefix "oc-yang"; }
+  import openconfig-inet-types { prefix "oc-inet"; }
   import openconfig-evpn-types { prefix oc-evpn-types; }
   import openconfig-evpn { prefix "oc-evpn"; }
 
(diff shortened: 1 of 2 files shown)


robshakir (approved):
This looks good to me. I think the `type` was intended to be mandatory anyway - and since there is no default type, so a NI without a type seems problematic in general.

I'm good with combining this with other changes too.

dplore:
Last call for comments.  To be merged on Dec 14, 2023.

dplore:
@rgwilton @rolandphung

rgwilton:
I'm checking with engineering.  The change looks reasonable to me.

robshakir:
> Last call for comments. To be merged on Dec 14, 2023.

I presume this is to be merged on Dec 14, **2022** right? :-)

dplore (approved):
LGTM, thanks Ebben.  I will go ahead and merge this now as it is overdue.  #738 can merge separately, pending feedback from the operator community meeting on Dec 20, 2022.

PR #750, opened 2022-11-21
Convert `.../bgp/rib/attr-sets/attr-set/as-path/as-segment` and `as4-segment` to keyed lists

wenovus:
### Change Scope

* Use an `uint32` index value as the key.
* Non-backward compatible.

### Rationale behind Change
Some OpenConfig tooling (e.g. gNMI, ygot, Ondatra) currently don't provide good support for unkeyed lists. The root problem being gNMI as a path-based configuration cannot indexed into the unkeyed list, forcing a non-leaf representation.

While unkeyed lists can sometimes be a natural representation of some device state ([ref](https://github.com/openconfig/public/pull/750#discussion_r1051282375)), using a keyed list would circumvent this issue. The downside is that there is extra work on the device side to stream the indices of the previous unkeyed list, but this is perceived to be a small price to pay to simplify the tooling.

### DIFF
--- a/release/models/rib/openconfig-rib-bgp-attributes.yang
+++ b/release/models/rib/openconfig-rib-bgp-attributes.yang
@@ -24,7 +24,13 @@ submodule openconfig-rib-bgp-attributes {
     attributes for use in BGP RIB tables.";
 
 
-  oc-ext:openconfig-version "0.8.1";
+  oc-ext:openconfig-version "0.9.0";
+
+  revision "2022-11-21" {
+    description
+      "Convert as-segment to a keyed list.";
+    reference "0.9.0";
+  }
 
   revision "2022-06-06" {
     description
@@ -81,6 +87,12 @@ submodule openconfig-rib-bgp-attributes {
     description
       "Data for representing BGP AS-PATH attribute";
 
+    leaf index {
+      type uint64;
+      description
+        "A unique index identifying the AS-PATH segment.";
+    }
+
     leaf type {
       type oc-bgpt:as-path-segment-type;
       description
@@ -117,8 +129,18 @@ submodule openconfig-rib-bgp-attributes {
         RFC 5065 - Autonomous System Confederations for BGP";
 
       list as-segment {
+        key "index";
+
         description
-          "Unkeyed list of AS PATH segments";
+          "List of AS-PATH segments";
+
+        leaf index {
+          type leafref {
+            path "../state/index";
+          }
+          description
+            "A unique index identifying the AS-PATH segment.";
+        }
 
         container state {
           config false;
--- a/release/models/rib/openconfig-rib-bgp-attributes.yang
+++ b/release/models/rib/openconfig-rib-bgp-attributes.yang
@@ -26,9 +26,9 @@ submodule openconfig-rib-bgp-attributes {
 
   oc-ext:openconfig-version "0.9.0";
 
-  revision "2022-11-21" {
+  revision "2022-12-20" {
     description
-      "Convert as-segment to a keyed list.";
+      "Convert as-segment and as4-segment to keyed lists.";
     reference "0.9.0";
   }
 
@@ -88,9 +88,13 @@ submodule openconfig-rib-bgp-attributes {
       "Data for representing BGP AS-PATH attribute";
 
     leaf index {
-      type uint64;
+      type uint32;
       description
-        "A unique index identifying the AS-PATH segment.";
+        "A unique ordering index starting from 0 identifying the position of
+        the AS-PATH segment in the list of segments.
+
+        The index MUST start from 0 and end at (length-1), where length is the
+        number of segments in the list of AS-PATH segments.";
     }
 
     leaf type {
@@ -129,6 +133,7 @@ submodule openconfig-rib-bgp-attributes {
         RFC 5065 - Autonomous System Confederations for BGP";
 
       list as-segment {
+        oc-ext:telemetry-atomic;
         key "index";
 
         description
@@ -139,7 +144,8 @@ submodule openconfig-rib-bgp-attributes {
             path "../state/index";
           }
           description
-            "A unique index identifying the AS-PATH segment.";
+            "Reference to the unique ordering index starting from 0 identifying
+            the position of the AS-PATH segment in the list of segments.";
         }
 
         container state {
@@ -170,13 +176,25 @@ submodule openconfig-rib-bgp-attributes {
           "RFC 6793 - BGP Support for Four-octet AS Number Space";
 
       list as4-segment {
+        oc-ext:telemetry-atomic;
+        key "index";
+
         description
-          "Unkeyed list of AS PATH segments";
+          "List of AS4-PATH segments";
+
+        leaf index {
+          type leafref {
+            path "../state/index";
+          }
+          description
+            "Reference to the unique ordering index starting from 0 identifying
+            the position of the AS4-PATH segment in the list of segments.";
+        }
 
         container state {
           config false;
           description
-            "Opstate data for AS-PATH segments";
+            "Opstate data for AS4-PATH segments";
 
           uses bgp-as-path-attr-state;
         }
--- a/release/models/rib/openconfig-rib-bgp-attributes.yang
+++ b/release/models/rib/openconfig-rib-bgp-attributes.yang
@@ -115,6 +115,7 @@ submodule openconfig-rib-bgp-attributes {
       "Top-level grouping for AS-PATH attribute data";
 
     container as-path {
+      oc-ext:telemetry-atomic;
       description
         "Enclosing container for the list of AS path segments.
 
@@ -133,7 +134,6 @@ submodule openconfig-rib-bgp-attributes {
         RFC 5065 - Autonomous System Confederations for BGP";
 
       list as-segment {
-        oc-ext:telemetry-atomic;
         key "index";
 
         description
@@ -164,6 +164,7 @@ submodule openconfig-rib-bgp-attributes {
       "Top-level grouping for AS4-PATH attribute data";
 
     container as4-path {
+      oc-ext:telemetry-atomic;
       description
         "This is the path encoded with 4-octet
         AS numbers in the optional transitive AS4_PATH attribute.
@@ -176,7 +177,6 @@ submodule openconfig-rib-bgp-attributes {
           "RFC 6793 - BGP Support for Four-octet AS Number Space";
 
       list as4-segment {
-        oc-ext:telemetry-atomic;
         key "index";
 
         description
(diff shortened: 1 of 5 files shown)


dplore:
Last call for comments, to be merged on Dec 15 ,2022

dplore:
@earies @rgwilton @rolandphung @hobbydos for comment on this change

release/models/rib/openconfig-rib-bgp-attributes.yang
```diff
@@ -81,6 +87,12 @@ submodule openconfig-rib-bgp-attributes {
     description
       "Data for representing BGP AS-PATH attribute";
 
+    leaf index {
+      type uint64;
```
jhaas-pfrc, line 91: uint32 is plenty here.  Very likely uint16 is sufficient for standard RFC 4271 4k max pdu packets, but with extended 65k packets with extension uint16 may not be sufficient for pedantic implementation.  In practice, segment list size will be small.

release/models/rib/openconfig-rib-bgp-attributes.yang
```diff
@@ -81,6 +87,12 @@ submodule openconfig-rib-bgp-attributes {
...
+      description
+        "A unique index identifying the AS-PATH segment.";
+    }
+
```
jhaas-pfrc, line 95: The procedure here is insufficiently specific.  BGP normally discusses its AS_PATH in terms of prepending each AS to the left.  Segments are limited in the protocol to 255 AS numbers in a given segment.

The members of the segment are also under-specified.  Is it the natural vector of the BGP prepend as usually displayed in implementations?  I.e. the first element of the leaf-list is the "left-most" member AS in that segment? (I believe that's the intent and general implementation).

The indices are necessary to provide order to the segments.  Is index 0 the right-most segment or the left-most?

wenovus: The type for AS-PATH segments is defined as [this leaf-list](https://github.com/openconfig/public/blob/0bdeaa7e1bb8c7833ab379a7acf6aa4a424ea8e9/release/models/rib/openconfig-rib-bgp-attributes.yang#L102-L106), whereas the the `as-segment` list in question (currently unkeyed) can contain multiple individual AS-PATH segments, so I believe the prepend operation you're referring to is still available for each segment.

Do you know if the ordering of these segments is important? In either case I think we should update the description to make this clear.

release/models/rib/openconfig-rib-bgp-attributes.yang
wenovus: Does this comment still hold given the other reply?

release/models/rib/openconfig-rib-bgp-attributes.yang
jhaas-pfrc: Yes, the ordering of the segments is important.

The leaf list is the element that I was referring to as also under-specified.

To give an easy example, consider an as_path using Juniper notation:
https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/command/show-route-detail.html

(10000) ([1000 2000]) 200 100 [ 10 20 ]

Note that the above example is contrived, and you won't typically find such a thing in the wild.  However, it is permitted to be valid by BGP aggregation rules.

The above as_path is constituted from four different as-segments.
The members of each of these segments is ordered.

What is the normative relationship for the elements in the YANG model to the above segments?

Minimally, I believe it's the case that the elements of the leaf-list for members of a given segment correspond to the natural left-to-right order.  Thus, the first element in the leaf-list for the as-set is "10", the second is "20".

The segments have a similarly natural order, however because of the constraints of gNMI which violates IETF YANG expectations, there's a desire to force them to be indexed.  However, the indexing property is less import than the ordering property.  The confed segment precedes the confed-set segment precedes the as-sequence segment precedes the as-set segment.  

The minute you supply a numeric index to elements, there's a push for implementations to want to also re-order them.  This is because if you don't, you have to maintain not only the index for key purposes, but add a secondary ordering index.  Ideally, the two properties are covered by the same thing.

If the simplest behavior were done, the unsigned integer index for the segments would still be left-to-right starting at zero and increasing by one.  Thus, the confed-seq would be index 0, confed-set would be index 1, as-seq would be index 2, as-set would be index 3.  I.e.:

(10000) ([1000 2000]) 200 100 [ 10 20 ]
segment[0] = {
type = confed-seq
members = «1000»
}
segment[1] = {
type = confed-set
members = «1000 2000»
}
segment[2] = {
type = as-seq
members = «200 100»
}
segment[3] = {
type = as-set
members = «10 20»
}

Unfortunately, YANG simply cares about the index values to make the contents unique.  An anti-social but compliant implementation would put random but unique nonsense in the index and force the implementation to have to store it while also storing the position.

wenovus: Thanks for the analysis. I think the problem is "how do we model config-false lists whose element ordering matters?" in an appropriate way for the implementation and client.

I want to clarify here that this list is `config false`, such that when using an indexed list with an additional ordering leaf, the implementation would be responsible for generating both values, I don't know if this makes the implementation easier.

The problem with unkeyed lists is that if a single leaf or element within the entire list is modified, then the entire thing needs to be sent due to the inability to index into any particular element. This incurs at least `O(n)` cost on both the server and client side. I don't know how many segments are expected to be in this list, but if we ever hit a case where there are many, then an unkeyed list may not scale.

One way of making this easier on the telemetry client is to add a second keyed list that maps {`order-index`: `index`}, where `order-index` starts from 0 and increments by one. Both lists would be under a container that has the [telemetry-atomic](https://github.com/openconfig/public/blob/master/release/models/openconfig-extensions.yang#L154) extension to always keep a consistent view. This allows list ordering to be easily accessible to the client while minimizing telemetry updates.

Do you think this solution works for the implementation? If so I will discuss this further and change the PR to incorporate this.

jhaas-pfrc: I'm not familiar with order-index, but based on how it is written up, you're trying to place another wrapper on top of stuff that already is a set of lists and that just unnecessarily complicates things

Fundamentally, it is insane to try to take some operational state that is being represented for modeling purposes as sets of containers and lists that may or may not have natural keying and pretending that you can get on-change behaviors for those contents at a leaf level.  These objects are typically embedded in other similarly modeled collections and once you work your way far enough up the stack are often not even directly embedded but are referenced by named index.  Compare vs. the general  attr-set grouping: While it's possible that you may want to update contents for that entire set of things for all referring entities simultaneously, it's just as likely that the change for contents are only applicable to a set of referring entities.

The only sane answer is to treat updates to such things atomically, and to similarly do so in a recursive fashion.  You don't update one member AS in the list and expect that you've got good referential behavior for all routes referring to this as-path.  You'd issue a new as-path, stream the set of routes that have switched to it, and once references to the prior subtree have gone to zero you'd issue a delete for the containers.

wenovus: Let me try to understand your points:
A. Let's not add another list to supplement an existing list.
B. A client cannot expect to usefully do gNMI subscribe STREAM on a leaf within a list that's "config false".
C. We update things in a make-before-break fashion, and in most cases the implementation will treat AS-PATH with its possibly multiple segments as one unit during modification -- either adding the whole thing or deleting the whole thing.

My goal is to find the right model for both the client and the implementation. Right now the candidates I see are,
1. Drop this PR, keep the same unkeyed list. Implementation streams the entire unkeyed list **if** anything within changes.
2. Keep this PR as-is, with the index determining the order of the segments. Here we have the option of enforcing that,
2a. the indices increment by one each time, or
2b. have "anti-social but compliant implementation" you described above, where the implementation uses indices that can skip numbers.
3. Add a secondary `order-index` leaf on top of this PR. This means that the index no longer indicates the order to the client. The client would have to sort based on the `order-index`.
4. Add a sibling list keyed by `order-index` that maps to `index` -- this spares clients from having to run a sorting algorithm.

From the constraints,
* A rules out 4.
* B sounds like it rules out 2-4 because it almost sounds like leaves should not be accessible and all lists should be unkeyed.
* C sounds like it is in favour of 1 since we're not expecting the unkeyed list to be streaming a lot of redundant telemetry values in this case.

Did I understand your argument correctly?

jhaas-pfrc: You've understood enough of the relevant points.  Another thing that will make you want to avoid the "order-index" ideas is that you're simply adding another set of elements that need to stream atomically for them to be useful.

In general, there will exist protocol elements that have been modeled in a way to ease management use of the state that aren't readily amenable to partial or incremental updates. The element as a whole, e.g. bgp as-paths, community lists, path attributes, largely will have to be treated as read-only once it has been streamed.  Partial updates will lead to state referencing these items as broken during operations that touch them.  The lack of streaming protocol markers for atomic updates simply make this an issue.

The issue in general, a flavor of referential integrity, really means that streaming updates of leaf nodes isn't a helpful thing.  The parties working on the streaming protocols should focus on updating clusters of object sets atomically.

With regard to this specific PR, you may feel free to do something like 2 and make sure that it's clear what the behaviors are.  However, if you're only doing this to theoretically help with streaming individual leaves, you hopefully can see that's not helpful here.  If that's the case, dropping the change is likely appropriate.

release/models/rib/openconfig-rib-bgp-attributes.yang
wenovus: Changed to uint32

release/models/rib/openconfig-rib-bgp-attributes.yang
wenovus: I've selected 2a, the reason is to avoid having to stream JSON for gNMI's PROTO encoding. The tradeoff being an indexed list is less natural than an unkeyed list.

wenovus: Thanks for the input, I consider the feedback to be incorporated.

jhaas-pfrc: The one final bit of feedback is that it's less important for individually contained segments to be atomic, it's the entire container as-path.  I'm not familiar with the constraints that directive implies, but consider whether it's appropriate to move the atomic keyword up to the enclosing container for a single path.

wenovus: Thanks you're right, done.

dplore:
Added to Jan 10, 2023 OpenConfig Operators agenda

dplore:
Reviewed in Jan 12 OC operators meeting without objection.  @wenovus Please summarize the motivation for this change in the first comment of the PR, then we are good to go.

wenovus:
> Reviewed in Jan 12 OC operators meeting without objection. @wenovus Please summarize the motivation for this change in the first comment of the PR, then we are good to go.

@dplore Update description, PTAL

dplore (approved):
LGTM, will merge on Jan 26, 2023

wenovus:
@dplore ping on merging this

PR #752, opened 2022-11-21
Convert `vlans/vlan/members/member` to a keyed list.

wenovus:
### Change Scope

* Use existing `interface` leaf as the key. This is a leafref that points to /interfaces/interface/name, which is already a key.
* Non-backward compatible.

### Platform Implementations

Plan to discuss this with vendors in the next OpenConfig community meeting.


earies:
+1 - see https://github.com/openconfig/public/pull/747

wenovus:
Looks like a dup.

PR #753, opened 2022-11-23
Removal of global per network-instance MTU

earies:
    - Removal of global per network-instance MTU (config/state)

### Change Scope

Initial publishes of various models brought in nodes that either do not map to existing implementations or have no defined/expected behavior.

This patch set removes `/network-instances/network-instance/config/mtu` (as well as the corresponding state node)

Until there is implementation guidance or proof of >1 existing implementations that a per-instance MTU is supported, the suggestion is to remove such nodes from the model set as it produces issues with folks mandating full compliance to OpenConfig model-sets and/or a set of deviations a server and client must implement and understand.

Per YANG backwards compatibility rules, this change is removing a node and thus is backwards incompatible though highly unlikely to affect clients since every implementation has likely produced a deviation for this node.

### Platform Implementations

Noted Deviations:

* Cisco IOS-XR: https://github.com/YangModels/yang/blob/main/vendor/cisco/xr/772/cisco-xr-openconfig-network-instance-deviations.yang#L76-L78
* Nokia SROS: https://github.com/nokia/7x50_YangModels/blob/master/latest_sros_22.7/openconfig/nokia-sr-openconfig-network-instance-deviations.yang#L80-L82
* Arista EOS: https://github.com/aristanetworks/yang/blob/master/EOS-4.29.0.2F/release/openconfig/models/not-supported/arista-network-instance-notsupported-deviations.yang#L2354-L2358
* Juniper JUNOS: https://github.com/Juniper/telemetry/blob/master/22.2/22.2R1/yangs/openconfig/deviation/jnx-openconfig-dev.yang#L1023-L1025

### DIFF
--- a/release/models/network-instance/openconfig-network-instance-l2.yang
+++ b/release/models/network-instance/openconfig-network-instance-l2.yang
@@ -24,7 +24,13 @@ submodule openconfig-network-instance-l2 {
     Layer 2 network instance configuration and operational state
     parameters.";
 
-  oc-ext:openconfig-version "1.4.0";
+  oc-ext:openconfig-version "2.0.0";
+
+  revision "2022-11-22" {
+    description
+      "Removal of global per network-instance MTU";
+    reference "2.0.0";
+  }
 
   revision "2022-09-15" {
     description
@@ -244,19 +250,6 @@ submodule openconfig-network-instance-l2 {
        uses oc-evpn:evpn-arp-proxy-top;
        uses oc-evpn:evpn-nd-proxy-top;
        uses l2ni-l2rib-top;
-    }
-  }
-
-  grouping l2ni-instance-common-config {
-    description
-      "Common configuration options which are specific to Layer 2
-      network instances";
-
-    leaf mtu {
-      type uint16;
-      description
-        "The maximum frame size which should be supported for this
-        instance for Layer 2 frames";
     }
   }
 
--- a/release/models/network-instance/openconfig-network-instance.yang
+++ b/release/models/network-instance/openconfig-network-instance.yang
@@ -48,7 +48,13 @@ module openconfig-network-instance {
     virtual switch instance (VSI). Mixed Layer 2 and Layer 3
     instances are also supported.";
 
-  oc-ext:openconfig-version "1.4.0";
+  oc-ext:openconfig-version "2.0.0";
+
+  revision "2022-11-22" {
+    description
+      "Removal of global per network-instance MTU";
+    reference "2.0.0";
+  }
 
   revision "2022-09-15" {
     description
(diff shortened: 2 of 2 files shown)


dplore:
This sounds good to me.  The deviations referenced indicate that this is not supported and I am not aware of a strong operational use case that requires MTU to be set at a network instance level.  It seems the common way to set MTU is per interface (Ethernet MTU) and per subinterface (IP MTU)

dplore:
Last call for comments on this PR.  We will merge this in 2 weeks on December 13, 2022 if there are no material objections.

mikewiebe:
@dplore I just checked and it does appear that this property is supported on Cisco NX platforms.  I need to do a bit more investigation into how it's mapped and follow up with the team that did the mapping.

earies:
My opinion is that for all nodes that were published initially before any proof of implementation, mapping or use/usefulness (and fully described intent), we now have a variety of cases where a node may map to 0 or 1 implementations - 0 is _not_ ok and gives a perception to those not following the development of OpenConfig that there is implementation support and anything less is not desirable.  Any such cases should be removed and ideally this does not occur again.

For the case of 1 implementation (and I'm not saying that is exactly the case here but so far seems to be), that can be fine but this is not implementation neutrality.  It's just a common model that just so happened to get a native node inserted into it.

One can always deviate but when the majority deviates, this is questionable practice vs. having wider support with fewer implementations deviating.  In addition, there is a perception going around that deviations are a bad thing (and they can be in other circumstances) and that 100% compliance of a given model is mandated.

The proper way to then handle this type of scenario may very well likely be that any global MTU is set via native schema rather (since this is likely necessary for many other things anyways)

dplore:
The [contributions guide](https://github.com/openconfig/public/blob/2a54814329edb18a2607d8e1ed430b3e9c4890b1/doc/contributions-guide.md?plain=1#L25-L28) now requires a minimum of two implementations be referenced and that if there are not at least two, that vendor native models should be used instead.  

This guidance came after the addition of this leaf.  We don't have a policy for what to do in this scenario.  We'll make it a operator topic for Dec 6th 2022.

earies:
> The [contributions guide](https://github.com/openconfig/public/blob/2a54814329edb18a2607d8e1ed430b3e9c4890b1/doc/contributions-guide.md?plain=1#L25-L28) now requires a minimum of two implementations be referenced and that if there are not at least two, that vendor native models should be used instead.
> 
> This guidance came after the addition of this leaf. We don't have a policy for what to do in this scenario. We'll make it a operator topic for Dec 6th 2022.

Thx @dplore - as you can probably tell by a few other current PRs, that this scenarios exists in various fashions across the model set and not specific to this one leaf.  I've found most cases usually point to a day 1 introduction for that given domain.

earies:
@dplore - any updates on this one? thx

dplore:
The operator community did not have any objection to this change.

dplore:
needs rebase and version bump based on #736 getting merged ahead of this one

earies:
thx @dplore - updated and currently running checks

dplore (approved):
LGTM

PR #754, opened 2022-11-28
Hardware Resource Thresholds

rolandphung:
### Change Scope

* Adding threshold configuration and threshold-exceeded state for platform resource utilization components.
pyang output:
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
           +--rw utilization
              +--rw resources
                 +--rw resource* [name]
                    +--rw config
                    |  +--rw name?                         string
                    |  +--rw used-threshold-upper?         oc-types:percentage
                    |  +--rw used-threshold-upper-clear?   oc-types:percentage
                    +--ro state
                       +--ro name?                            string
                       +--ro used-threshold-upper?            oc-types:percentage
                       +--ro used-threshold-upper-clear?      oc-types:percentage
                       +--ro used?                            uint64
                       +--ro committed?                       uint64
                       +--ro free?                            uint64
                       +--ro max-limit?                       uint64
                       +--ro high-watermark?                  uint64
                       +--ro last-high-watermark?             oc-types:timeticks64
                       +--ro used-threshold-upper-exceeded?   boolean

```

### Platform Implementations

 * Arista `hardware capacity alert table <table> [feature <feature>] threshold <threshold>`: [link to documentation](https://www.arista.com/en/um-eos/eos-logging-of-event-notifications?searchword=threshold%20capacity)

### DIFF
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -138,7 +138,7 @@ submodule openconfig-platform-common {
     leaf used-threshold-upper {
       type oc-types:percentage;
       description
-        "The used percentage value (used / (used + free) * 100) that 
+        "The used percentage value (used / (used + free) * 100) that
         when crossed will set utilization-threshold-exceeded to 'true'.";
     }
 
@@ -198,10 +198,10 @@ submodule openconfig-platform-common {
     leaf used-threshold-upper-exceeded {
       type boolean;
       description
-        "This value is set to true when the used percentage value (used / (used + free) * 100)  
-        has crossed the used-threshold-upper for this resource and false when the
-        used percentage value has crossed the configured used-threshold-upper-clear
-        value  for this resource.";
+        "This value is set to true when the used percentage value
+        (used / (used + free) * 100) has crossed the used-threshold-upper for this
+        resource and false when the used percentage value has crossed the configured
+        used-threshold-upper-clear value for this resource.";
     }
   }
 
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -195,13 +195,13 @@ submodule openconfig-platform-common {
         (Jan 1, 1970 00:00:00 UTC).";
     }
 
-    leaf threshold-exceeded {
+    leaf used-threshold-upper-exceeded {
       type boolean;
       description
-        "This value is set to true when the system usage has crossed the
-        configured threshold-upper for this resource and false when the
-        system usage has crossed the configured threshold-lower for this
-        resource.";
+        "This value is set to true when the used percentage value (used / (used + free) * 100)  
+        has crossed the used-threshold-upper for this resource and false when the
+        used percentage value has crossed the configured used-threshold-upper-clear
+        value  for this resource.";
     }
   }
 
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -146,7 +146,7 @@ submodule openconfig-platform-common {
     leaf threshold-lower {
       type oc-types:percentage;
       description
-        "The percentage value that will send an alert or system alarm when
+        "The percentage value that will reset an alert or system alarm when
         the used percentage (used / (used + free) * 100) is lower than
         this value";
     }
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -135,20 +135,18 @@ submodule openconfig-platform-common {
         "Resource name within the component.";
     }
 
-    leaf threshold-upper {
+    leaf used-threshold-upper {
       type oc-types:percentage;
       description
-        "The percentage value that will send an alert or system alarm when
-        the used percentage (used / (used + free) * 100) is higher than
-        this value";
+        "The used percentage value (used / (used + free) * 100) that 
+        when crossed will set utilization-threshold-exceeded to 'true'.";
     }
 
-    leaf threshold-lower {
+    leaf used-threshold-upper-clear {
       type oc-types:percentage;
       description
-        "The percentage value that will reset an alert or system alarm when
-        the used percentage (used / (used + free) * 100) is lower than
-        this value";
+        "The used percentage value (used / (used + free) * 100) that when
+        crossed will set utilization-threshold-exceeded to 'false'.";
     }
   }
 
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -20,7 +20,13 @@ submodule openconfig-platform-common {
     "This modules contains common groupings that are used in multiple
     components within the platform module.";
 
-  oc-ext:openconfig-version "0.21.1";
+  oc-ext:openconfig-version "0.22.0";
+
+  revision "2022-12-20" {
+    description
+      "Add threshold and threshold-exceeded for resource usage.";
+    reference "0.22.0";
+  }
 
   revision "2022-12-19" {
     description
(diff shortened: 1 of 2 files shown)


aredmon8551:
What is the expected system behavior when one of these thresholds is exceeded?

rolandphung:
> What is/are the expected system behavior(s) when one of these thresholds is exceeded? Is there a shipping example of something similar? I _think_ this would be largely open to vendors to interpret (e.g., "major" alarm for upper-threshold, "minor" for lower) but would like to confirm I'm not missing a more prescriptive behavior requirement somewhere else.

Oh the documentation is incorrect, I will fix that up. `threshold-lower` is meant for *reseting* the alarm or `threshold-exceeded` when the `usage` value goes below `threshold-lower`. This is trying to prevent noise from the system when the `usage` value is hovering around one `threshold` value, preventing a hysteresis behavior.

cfyo:
Please add a second platform vendor implementation reference.

dplore:
Implementation references attached.  I based this investigation on @robshakir's from #506 and added more links and changes appropriate to this proposal.

### Cisco
  * IOS XE:
    * Example taken: ([TCAM threshold alarm](https://www.cisco.com/c/en/us/td/docs/routers/asr903/software/guide/tcam-threshold-config-xe-3s-asr903-book/nt-tcam.html#task_777E99665BE842AE8ECA212CFA2D5F82))
    * `platform tcam-threshold enable all <%>` allows all TCAMs across the system to alert at `<%>` percent utilization. This would correspond to a related alarm within this `/components/component/integrated-circuit` hierarchy. 
    * IOS XR:
    * `show prm server tcam summary ...` displays TCAM utilization across different components.   When these resources are exceeded, alarms are raised of the form `pfilter_ea[292]: %PKT_INFRA-FEA_DLL-3-TCAM_ERR : TCAM create region error: 'prm_server' detected the 'resource not available' condition 'TCAM resource exhausted.'`
    * The PM threshold monitoring allows for [thresholds to be configured](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/system-monitoring/73x/b-system-monitoring-cg-cisco8k-73x/implementing_performance_management.html#concept_o5z_hdg_p3b) for the component utilizations.
    * I do not see a hysteresis parameter similar to the proposed `threshold-lower` leaf in this PR

### Juniper JUNOS

JUNOS supports some thresholding, e.g., `edit system services resource monitor`. 
The [high-threshold](https://www.juniper.net/documentation/us/en/software/junos/network-mgmt/topics/ref/statement/high-threshold-edit-system-services-resource-monitor.html
) item roughly maps to  leaf `threshold-upper` of this proposal.  I do not see a hysteresis parameter similar to the proposed `threshold-lower` leaf in this PR.

### Nokia SROS

Nokia SROS supports [defining arbitrary thresholds](https://documentation.nokia.com/cgi-bin/dbaccessfilename.cgi/9300700701_V1_7750%20SR%20OS%20Basic%20System%20Configuration%20Guide%208.0r1.pdf) through `config>system>threshold` for example:

```
config>system>thresholds# memory-use-alarm rising-threshold
50000000 falling-threshold 45999999 interval 500 both startup-alarm
either
```
The `rising-threshold` parameter maps to the proposed leaf `threshold-upper` in this PR.
The `falling-threshold`  parameter performs a hysteresis function like the proposed `threshold-lower` leaf in this PR.

dplore (commented):
A couple of clarifying suggestions for your consideration @rolandphung.   

I do think this is a bit of a 'single purpose' solution versus growing a more reusable alarming and thresholds model.  However, there is an urgent, high business impact need need for this specific operational use case (observing integrated circuit utilization crossing a threshold).  Enhancing the openconfig alarms and thresholds model to scale to a larger number of use cases may be visited in a future PR.

dplore:
Last call for comments on this update, this will merge on Jan 31, 2023.  @rgwilton @hobbydos @earies please comment.

Hobbydos:
@dplore Someone from my org will provide feedback shortly.

robshakir:
This looks a reasonable addition to the model to me.

It is worth re-iterating some of the history of `/system/alarms` based on the discussion that @dplore mentioned above. Originally, there was no such container, and we rather said that where there were specific events that resulted in an alarm, it was best to have these modelled in the section of the tree where they were surrounded by their context, e.g., where some threshold on a transceiver or NPU was exceeded, this should be co-located in the model with the other data relating to that component. This decision allowed implementations like:

```
let sub = gNMI.Subscribe(/components/component/transceiver/state/magic-threshold-exceeded)

when (update is received on sub) then
   let upd be the received update
   gNMI.Subscribe(trim(upd.Path, "state/magic-threshold-exceeded"), mode=ONCE)
```

where the `trim` action would give you the telemetry that relates entire component that had triggered the alert.

The downside of this is that for data to be transitioned to streaming over gRPC, every leaf that indicates such an event would need to be explicitly modelled. In surveying vendors, we found that this wasn't the case, and would mean that transitioning to gNMI-only telemetry would take longer. Thus, @ejbrever made a proposal to accommodate "generic" alarms in the model. Quoting from his design document:

> The proposal to model alarms contains two items that will create a path to achieve the 
objectives outlined above. In addition, this will allow us to model alarms in the short term without losing any alarm data. The two items would be:
> * Support a generic alarm list.
> * Begin modeling our most important alarms and continue to drive towards to our long term vision by adding more and more alarm variables into the model.
> This will allow us to model all alarms in the short term through a generic alarm list. Any OpenConfig users who prefer this approach would also have this available to them. We can then work with vendors to model specific alarms through boolean variables. This will take some time to work through the long tail of alarms, but provides a path to achieve the core objective.

This strategy was agreed and reviewed by the operators group in the autumn of 2017. We can re-examine this approach, but this essentially reflects the wider strategy for the models.

Given this strategy -- how does this PR fit in? This seems (to me) to be exactly the "modelling of our most important alarms" in an area of the model that complies with the original approach. There are other examples in the model - for example `prefix-limit-exceeded` in BGP is a similar "alarm-esque" leaf.

Thus, I believe we should pursue this approach, and this PR LGTM.

dplore (approved):
Still targeted for Jan 31, 2023 merge

Hobbydos:
cc @nokia1adam

nokia1adam:
This is a useful PR. The implementation I represent can align to it without too much difficulty. However, if I can inject two comments:

1. It seems overkill to me that the thresholds are configured per IC. Do most implementations actually support this, or are the thresholds usually set at the system level?

2. would it be useful to break up the list of resources by subsystem? In other words you would have a **list subsystem** right above the **list resource**, with the subsystem names being vendor-specific, but you could imagine common categories such as "qos", "counters", "acl", "fib", etc. If not, then these categories will be incorporated into the resource names, which is not the end of the world, but worth considering if that's really the best organization.

hellt:
@dplore in https://github.com/openconfig/public/pull/754#issuecomment-1381152745 you referred to SR OS docs from ancient times.

This link should lead readers to the recent one
https://documentation.nokia.com/sr/22-10-2/books/basic-system-configuration/system-management.html#ai9emdyopc

rolandphung:
> It seems overkill to me that the thresholds are configured per IC. Do most implementations actually support this, or are the thresholds usually set at the system level?

This is a very good point that I had originally missed. It might be better for the model to support system wide, but there isn't really a simple place to move the configuration to (maybe `/system/utilization/...`?). Even that move would probably require additional modeling to associate the one resource to many components. @dplore  Thoughts or comments on changing this model to be system wide?

> would it be useful to break up the list of resources by subsystem? In other words you would have a list subsystem right above the list resource, with the subsystem names being vendor-specific, but you could imagine common categories such as "qos", "counters", "acl", "fib", etc. If not, then these categories will be incorporated into the resource names, which is not the end of the world, but worth considering if that's really the best organization.

@nokia1adam what do you mean by subsystem? We already categorize this by component and resource.

dplore:
> > It seems overkill to me that the thresholds are configured per IC. Do most implementations actually support this, or are the thresholds usually set at the system level?
> 
> This is a very good point that I had originally missed. It might be better for the model to support system wide, but there isn't really a simple place to move the configuration to (maybe `/system/utilization/...`?). Even that move would probably require additional modeling to associate the one resource to many components. @dplore Thoughts or comments on changing this model to be system wide?

A system wide model for thresholds (and alarms which are closely related) is an issue that has been proposed and debated 
 without consensus to date.  To make progress on this we will merge this PR.  As far configuring different thresholds per component, you can refer to the above links.  At least Cisco XR has some ability to configure different thresholds for different components (but the example I could find was CPU, not INTEGRATED_CIRCUIT).   To work with the model presented here, an implementation could chose to  offer partial support by placing a constraint that it only supports configuring the same threshold for all  INTEGRATED_CIRCUIT's present.

PR #758, opened 2022-12-02
Creating mutually exclusive tcp details

sourcequench:
Adding match-any / match-all necessary for a useful flags list Adding common keywords such as established and fragments, which cannot be used at the same time as explicit flags matching.

### Change Scope

* It is very common for ACLs to be written using keywords such as "established" across a range of vendors.
* Having a tcp-flags leaf list is nice, but incomplete, as you'd need to specify whether you wish to match-any or match-all.
* This change is **not** backward compatible.

### Platform Implementations

 * Cisco: [acl guide](https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/sec_data_acl/configuration/15-sy/sec-data-acl-15-sy-book/sec-refine-ip-al.html) 
 * Juniper: [stateless match conditions](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/firewall-filter-stateless-match-conditions-bit-field-values.html) and/or
   implementation output.

### DIFF
--- a/release/models/acl/openconfig-packet-match.yang
+++ b/release/models/acl/openconfig-packet-match.yang
@@ -394,12 +394,32 @@ module openconfig-packet-match {
         "Destination port or range";
     }
 
-    leaf-list tcp-flags {
-      type identityref {
-        base oc-pkt-match-types:TCP_FLAGS;
+    container transport-details {
+      choice explicit-or-builtin {
+        case explicit-flags {
+          leaf-list tcp-flags {
+            type identityref {
+              base oc-pkt-match-types:TCP_FLAGS;
+            }
+            description
+              "List of TCP flags to match";
+          }
+          leaf match {
+            type enumeration {
+              enum ANY;
+              enum ALL;
+          }
+        }
+        case builtin {
+          leaf keyword {
+            type enumeration {
+              enum TCP-ESTABLISHED;
+              enum TCP-INITIAL;
+              enum FRAGMENT;
+            }
+          }
+        }
       }
-      description
-        "List of TCP flags to match";
     }
   }
 
--- a/release/models/acl/openconfig-acl.yang
+++ b/release/models/acl/openconfig-acl.yang
@@ -388,7 +388,7 @@ module openconfig-acl {
 
   grouping access-list-entries-top {
     description
-      "Access list entries to level container";
+      "Access list entries top level container";
 
     container acl-entries {
       description


release/models/acl/openconfig-packet-match.yang
```diff
@@ -394,12 +394,32 @@ module openconfig-packet-match {
...
+            type identityref {
+              base oc-pkt-match-types:TCP_FLAGS;
+            }
+            description
+              "List of TCP flags to match";
+          }
+          leaf match {
+            type enumeration {
+              enum ANY;
+              enum ALL;
```
robshakir, line 410: Please add a description for each of these options, and the `match` leaf itself.

release/models/acl/openconfig-packet-match.yang
```diff
@@ -394,12 +394,32 @@ module openconfig-packet-match {
...
+          }
+        }
+        case builtin {
+          leaf keyword {
+            type enumeration {
+              enum TCP-ESTABLISHED;
```
robshakir, line 416: The format here should be `TCP_ESTABLISHED`,

release/models/acl/openconfig-packet-match.yang
```diff
@@ -394,12 +394,32 @@ module openconfig-packet-match {
         "Destination port or range";
     }
 
-    leaf-list tcp-flags {
-      type identityref {
-        base oc-pkt-match-types:TCP_FLAGS;
+    container transport-details {
+      choice explicit-or-builtin {
```
robshakir, line 398: We explicitly do not use the `choice`/`case` pattern in OpenConfig models.

We would rather do something like:

```yang

container transport-details {
   container config {
      description "...";

      leaf mode {
         type enumeration {
             enum BUILTIN { description "..."; };
             enum EXPLICIT { description "..."; };
         }
         description "...";
      }

      leaf-list explicit {
        type identityref { base oc-pkt-match-types:TCP_FLAGS; 
        when "../mode = 'EXPLICIT';
      }

      leaf explicit-match-option {
        type enumeration {
           enum ANY { description "..."; }
           enum ALL { description "..."; }
        }
        description "...";
      }

      leaf builtin {
        type enumeration {
          enum TCP_INITIAL { description "..."; }
          enum FRAGMENT { description "..."; }
          enum TCP_FRAGMENT { description "..."; }
        }
        description "...";
      }
   }
}
```

robshakir (changes requested):
Do we think that the `explicit` option is useful here, or is the `builtin` actually what would be supported by platforms? I'd be OK to remove the `explicit` option if we felt that this was more compatible.

I added some comments re: YANG style too. I am happy to help with getting this together in the right way.

robshakir:
Made the suggested changes to the YANG in https://github.com/openconfig/public/pull/796 -- we'll take that PR for this functionality. Thanks for the contribution @sourcequench :-)

PR #759, opened 2022-12-02
Adjust description statement for lacp-timeout-transitions

earies:
* (M) release/models/lacp/openconfig-lacp.yang
    - Clarify wording around lacp-timeout-transitions

### Change Scope

Clarify wording surrounding `lacp-timeout-transitions` to specify the number of times the LACP state has transitioned with a timeout since

* The time the device restarted
* The interface was added in the LACP bundle

Since this change only accounts for a description statement update, the patch version is incremented and is backwards compatible in nature however implementations could very well need to adjust as the previous wording was ambiguous.

### Platform Implementations

N/A: Description statement update for implementation guidance

### DIFF
--- a/release/models/lacp/openconfig-lacp.yang
+++ b/release/models/lacp/openconfig-lacp.yang
@@ -26,7 +26,13 @@ module openconfig-lacp {
     managing aggregate interfaces.   It works in conjunction with
     the OpenConfig interfaces and aggregate interfaces models.";
 
-  oc-ext:openconfig-version "1.2.0";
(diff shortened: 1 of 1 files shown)


earies:
@dplore - simple description update - could we get a reviewer assigned - thx

dplore:
Adding to Jan 10th openconfig operators meeting for review

cfyo:
During the operators meeting can we make sure we have at least two vendors references that this update does not affect any current implementations.

dplore:
@earies please provide references for at least 2 implementations

PR #582, opened 2022-02-10
Proposal for control-plane ACL and QoS model.

robshakir:
```
commit 0ef7e4ab4fb8f0c0dd4594de6fbbac290a44f5ef
Author: Rob Shakir <robjs@google.com>
Date:   Thu Feb 10 07:31:47 2022 -0500

    Add control-plane QoS and ACL model.
    
      * (M) release/models/system/.spec.yml
      * (A) release/models/system/openconfig-system-controlplane.yang
        - Initial proposal for a model that covers control-plane policing
          (QoS policy) and ACL for control-plane protection. The proposal is
          to provide a central point of attachment under the system model.
```

PR #762, opened 2022-12-03
Control plane access control list and QoS

dplore:
### Change Scope

Originally authored by @robshakir  at #582.  Refactored to reference [openconfig-defined-sets](https://github.com/openconfig/public/blob/master/release/models/defined-sets/openconfig-defined-sets.yang) model.

* Creates a new system container `control-plane-traffic` to associate ACLs with representing the control plane of a network device.
* This change only adds new leafs and is therefore backwards compatible.
### Platform Implementations

 * [Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/security/70x/b-system-security-cg-cisco8000-70x/implementing-management-plane-protection.html#concept_f14_pxg_s3b) defines management plane protection
 * [JunOS](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/bgp/topics/example/firewall-filter-stateless-example-prefix-list.html#configuration1029__d3762e75)  uses firewall filters applied to interface lo0  

```
module: openconfig-system-controlplane

  augment /oc-sys:system:
    +--rw control-plane-traffic
       +--rw ingress
       |  +--rw acl
       |  |  +--rw acl-set* [set-name type]
       |  |     +--rw set-name       -> ../config/set-name
       |  |     +--rw type           -> ../config/type
       |  |     +--rw config
       |  |     |  +--rw set-name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
       |  |     |  +--rw type?       -> /oc-acl:acl/acl-sets/acl-set[oc-acl:name=current()/../set-name]/config/type
       |  |     +--ro acl-entries
       |  |        +--ro acl-entry* [sequence-id]
       |  |           +--ro sequence-id    -> ../state/sequence-id
       |  |           +--ro state
       |  |              +--ro sequence-id?       -> /oc-acl:acl/acl-sets/acl-set[oc-acl:name=current()/../../../../set-name][oc-acl:type=current()/../../../../type]/oc-acl:acl-entries/acl-entry/sequence-id
       |  |              +--ro matched-packets?   oc-yang:counter64
       |  |              +--ro matched-octets?    oc-yang:counter64
       |  +--rw qos
       |     +--rw classifier
       |     |  +--rw config
       |     |  |  +--rw name?   -> /oc-qos:qos/classifiers/classifier/config/name
       |     |  +--ro state
       |     |  |  +--ro name?   -> /oc-qos:qos/classifiers/classifier/config/name
       |     |  +--ro terms
       |     |     +--ro term* [id]
       |     |        +--ro id       -> ../state/id
       |     |        +--ro state
       |     |           +--ro id?                -> /oc-qos:qos/classifiers/classifier[oc-qos:name=current()/../../../../config/name]/terms/term/config/id
       |     |           +--ro matched-packets?   oc-yang:counter64
       |     |           +--ro matched-octets?    oc-yang:counter64
       |     +--rw scheduler-policy
       |        +--rw config
       |        |  +--rw name?   -> /oc-qos:qos/scheduler-policies/scheduler-policy/config/name
       |        +--ro state
       |        |  +--ro name?   -> /oc-qos:qos/scheduler-policies/scheduler-policy/config/name
       |        +--ro scheduler-statistics
       |           +--ro scheduler* [sequence]
       |              +--ro sequence    -> ../state/sequence
       |              +--ro state
       |                 +--ro sequence?            -> /oc-qos:qos/scheduler-policies/scheduler-policy[oc-qos:name=current()/../../../../config/name]/schedulers/scheduler/config/sequence
       |                 +--ro conforming-pkts?     oc-yang:counter64
       |                 +--ro conforming-octets?   oc-yang:counter64
       |                 +--ro exceeding-pkts?      oc-yang:counter64
       |                 +--ro exceeding-octets?    oc-yang:counter64
       |                 +--ro violating-pkts?      oc-yang:counter64
       |                 +--ro violating-octets?    oc-yang:counter64
       +--rw egress
          +--rw acl
             +--rw acl-set* [set-name type]
                +--rw set-name       -> ../config/set-name
                +--rw type           -> ../config/type
                +--rw config
                |  +--rw set-name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
                |  +--rw type?       -> /oc-acl:acl/acl-sets/acl-set[oc-acl:name=current()/../set-name]/config/type
                +--ro acl-entries
                   +--ro acl-entry* [sequence-id]
                      +--ro sequence-id    -> ../state/sequence-id
                      +--ro state
                         +--ro sequence-id?       -> /oc-acl:acl/acl-sets/acl-set[oc-acl:name=current()/../../../../set-name][oc-acl:type=current()/../../../../type]/oc-acl:acl-entries/acl-entry/sequence-id
                         +--ro matched-packets?   oc-yang:counter64
                         +--ro matched-octets?    oc-yang:counter64
```


dplore:
Feedback from operator meeting is to update this to reference the new access control set defined in #649

release/models/system/openconfig-system-controlplane.yang
```diff
@@ -0,0 +1,181 @@
...
+      }
+
+      container acls {
+        description
+          "Access control lists (ACLs) that are to be applied on traffic
+          destined for the control-plane. An ACL per traffic type
+          can be applied.";
+
+        list acl {
+          key "type";
```
rolandphung, line 82: Do we want to consider a per network-instance control-plane ACL configuration?

See https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1302661

release/models/system/openconfig-system-controlplane.yang
```diff
@@ -0,0 +1,181 @@
...
+
+      container state {
+        config false;
+        description
+          "Operational state parameters relating to control-plane traffic.";
+
+        uses system-controlplane-config;
+      }
+
+      container acls {
```
rolandphung, line 75: Should this be made to be future proof by supporting both `ingress` and `egress` control plane ACL configurations?

dplore:
Added to openconfig operator meeting agenda for Jan 10, 2023

release/models/system/openconfig-system-controlplane.yang
```diff
@@ -0,0 +1,141 @@
...
+            }
+            description
+              "Reference to a scheduler policy that determines rate limits, or
+               shaping of packets towards the control-plane.";
+        }
+
+        leaf acl-v4 {
+            type leafref {
+                path "/oc-sets:defined-sets/oc-sets:ipv4-prefix-sets/" +
+                "oc-sets:ipv4-prefix-set/oc-sets:config/oc-sets:name";
```
rolandphung, line 116: Should this be the leafref of an ACL from the `acl-set`? How does a prefix-set associate itself with an ACL here?

release/models/system/openconfig-system-controlplane.yang
```diff
@@ -0,0 +1,141 @@
...
+            type leafref {
+                path "/oc-qos:qos/oc-qos:scheduler-policies/oc-qos:scheduler-policy/" +
+                "oc-qos:config/oc-qos:name";
+            }
+            description
+              "Reference to a scheduler policy that determines rate limits, or
+               shaping of packets towards the control-plane.";
+        }
+
+        leaf acl-v4 {
```
robshakir, line 113: For consistency, let's call this `-ipv4` and similarly for IPv6.

release/models/system/openconfig-system-controlplane.yang
robshakir: Agreed. This should reference `/acl/acl-sets/acl-set/name`.

release/models/system/openconfig-system-controlplane.yang
rolandphung: @dplore The more I think about it the more we likely need to make this a per network-instance collection of `acl-set` leafrefs. Within the `ipv4` and `ipv6` containers, there is no indication of which network-instance to refer to. It also doesn't make sense for `acl-sets` to be NI aware as it can be associated per interface which is NI inherited. Therefore this control-plane configuration would have to be NI aware.

One side note is if the intention is to apply the ACL across all VRFs (with an `acl-set` that doesn't have any IP addresses, usually default ACLs) then that configuration should be an replacement of whatever default ACL that comes on a switch and there should be a separate per VRF collection which would override this default.

jsterne:
From the back-and-forth above it is difficult to know what the proposed model is. Maybe someone can update the proposed PR with the latest thoughts?

It doesn't seem correct to me to reference any prefix sets here (wouldn't the ACL itself reference whatever prefix sets it needs?). 

Having an ctl plane ACL per VRF seems expensive. Maybe we need an NI leafref in the ACL match criteria itself (but that is only used for ctl plane ACLs)?

We may also want to separate out the ACL aspects of this PR and the qos/classifier/scheduling. Those are different functions with likely different discussions around them.

dplore:
I will make changes to update this PR to reference ACL instead of prefix list.  

Regarding network-instance, we are clarifying our operational use case when it comes to ACL with IP prefixes and network-instances.  When this use case is well defined we will share in this PR and update the model accordingly.

release/models/system/openconfig-system-controlplane.yang
dplore: today we only require ingress.  We should be explicit about that here so egress could be added later as a nonbreaking change.

release/models/system/openconfig-system-controlplane.yang
robshakir: I'm not sure that this is something that is generally compatible.

* Cisco IOS XR LPTS does not match on a [VRF](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/ip-addresses/75x/b-ip-addresses-cg-8k-75x/implementing_lpts.html#Cisco_Concept.dita_906ea9e6-b511-4c74-a3b4-50819baf860d).
* JUNOS allows `fxp0` to be assigned to one routing-instance, and firewall filters don't seem to support VRF matches [doc](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/firewall-filter-match-conditions-for-ipv4-traffic.html)
* Nokia SROS - see [here](https://infocenter.nokia.com/public/7750SR150R4A/index.jsp?topic=/com.sr.system.mgmt/html/getstarted.html) does not have a binding of `cpm-filter` to a particular VRF.

To this end, I don't propose to extend the model to cover VRFs. If this is something that we want to add in the future, we could add a VRF list that gives us ACL-per-VRF and leave a single pair (ingress/egress) of references here for "all VRFs". The new list would be overrides and/or mutually exclusive with this leaf (if we decide to add it based on device compatibility).

release/models/system/openconfig-system-controlplane.yang
robshakir: I've added an egress ACL configuration, but not the QoS policy. 

JUNOS supports this `set interfaces lo0 unit 0 family inet filter output`, I think EOS supports this with `system control-plane\n ip access-group FOO out`.

robshakir:
@jsterne @rolandphung @dplore - PTAL at this updated model here which reflects the discussion that we've had in this thread.

I'll be updating #763 to include the QoS counters that are discussed in that thread momentarily.

robshakir:
Sorry, since #762 belongs to @rolandphung - I can't add to it there. I'm going to create a new PR that specifically addresses that.

robshakir:
Adding the QoS counters and ACL counters required some restructuring of the initial proposal. This is now ready for review.

Ping @dplore, @rolandphung, @jsterne, @aaronmillisor, @aredmon8551. 

Thanks.

robshakir:
Here is the final proposed model tree view:

```
module: openconfig-system-controlplane
  augment /oc-sys:system:
    +--rw control-plane-traffic
       +--rw ingress
       |  +--rw acl
       |  |  +--rw ipv4
       |  |  |  +--rw config
       |  |  |  |  +--rw name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
       |  |  |  +--ro state
       |  |  |  |  +--ro name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
       |  |  |  +--ro acl-entries
       |  |  |     +--ro acl-entry* [sequence-id]
       |  |  |        +--ro sequence-id    -> ../state/sequence-id
       |  |  |        +--ro state
       |  |  |           +--ro sequence-id?       -> /acl/acl-sets/acl-set[name=current()/../../../../set-name][type=current()/../../../../type]/acl-entries/acl-entry/sequence-id
       |  |  |           +--ro matched-packets?   oc-yang:counter64
       |  |  |           +--ro matched-octets?    oc-yang:counter64
       |  |  +--rw ipv6
       |  |     +--rw config
       |  |     |  +--rw name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
       |  |     +--ro state
       |  |     |  +--ro name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
       |  |     +--ro acl-entries
       |  |        +--ro acl-entry* [sequence-id]
       |  |           +--ro sequence-id    -> ../state/sequence-id
       |  |           +--ro state
       |  |              +--ro sequence-id?       -> /acl/acl-sets/acl-set[name=current()/../../../../set-name][type=current()/../../../../type]/acl-entries/acl-entry/sequence-id
       |  |              +--ro matched-packets?   oc-yang:counter64
       |  |              +--ro matched-octets?    oc-yang:counter64
       |  +--rw qos
       |     +--rw classifier
       |     |  +--rw config
       |     |  |  +--rw name?   -> /oc-qos:qos/classifiers/classifier/config/name
       |     |  +--ro state
       |     |  |  +--ro name?   -> /oc-qos:qos/classifiers/classifier/config/name
       |     |  +--ro terms
       |     |     +--ro term* [id]
       |     |        +--ro id       -> ../state/id
       |     |        +--ro state
       |     |           +--ro id?                -> /oc-qos:qos/classifiers/classifier[oc-qos:name=current()/../../../../config/name]/terms/term/config/id
       |     |           +--ro matched-packets?   oc-yang:counter64
       |     |           +--ro matched-octets?    oc-yang:counter64
       |     +--rw scheduler-policy
       |        +--rw config
       |        |  +--rw name?   -> /oc-qos:qos/scheduler-policies/scheduler-policy/config/name
       |        +--ro state
       |        |  +--ro name?   -> /oc-qos:qos/scheduler-policies/scheduler-policy/config/name
       |        +--ro scheduler-statistics
       |           +--ro scheduler* [sequence]
       |              +--ro sequence    -> ../state/sequence
       |              +--ro state
       |                 +--ro sequence?            -> /oc-qos:qos/scheduler-policies/scheduler-policy[oc-qos:name=current()/../../../../config/name]/schedulers/scheduler/config/sequence
       |                 +--ro conforming-pkts?     oc-yang:counter64
       |                 +--ro conforming-octets?   oc-yang:counter64
       |                 +--ro exceeding-pkts?      oc-yang:counter64
       |                 +--ro exceeding-octets?    oc-yang:counter64
       |                 +--ro violating-pkts?      oc-yang:counter64
       |                 +--ro violating-octets?    oc-yang:counter64
       +--rw egress
          +--rw acl
             +--rw ipv4
             |  +--rw config
             |  |  +--rw name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
             |  +--ro state
             |  |  +--ro name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
             |  +--ro acl-entries
             |     +--ro acl-entry* [sequence-id]
             |        +--ro sequence-id    -> ../state/sequence-id
             |        +--ro state
             |           +--ro sequence-id?       -> /acl/acl-sets/acl-set[name=current()/../../../../set-name][type=current()/../../../../type]/acl-entries/acl-entry/sequence-id
             |           +--ro matched-packets?   oc-yang:counter64
             |           +--ro matched-octets?    oc-yang:counter64
             +--rw ipv6
                +--rw config
                |  +--rw name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
                +--ro state
                |  +--ro name?   -> /oc-acl:acl/acl-sets/acl-set/config/name
                +--ro acl-entries
                   +--ro acl-entry* [sequence-id]
                      +--ro sequence-id    -> ../state/sequence-id
                      +--ro state
                         +--ro sequence-id?       -> /acl/acl-sets/acl-set[name=current()/../../../../set-name][type=current()/../../../../type]/acl-entries/acl-entry/sequence-id
                         +--ro matched-packets?   oc-yang:counter64
                         +--ro matched-octets?    oc-yang:counter64
```

jsterne:
Thx for the updated tree view. 

Curious why we'd put acl-entry counters in the tree here under system. Since we reference an acl-set, don't the counters over in the ACL model apply?

That brings up another point: is there any specific expectation one way or the other that the acl-set referred to under system is used exclusively as the control plane ACL? (and not also used by other interfaces?).

Note that in SR OS, the control plane ACL is configured separately than the data plane ACLs.  (i.e. the control plane ACL isn't just another ACL in the same unified list of ACLs as data plane ACLs).

The model doesn't explicitly say how many ACLs are allowed in the list of control plane ACLs. I suspect the assumption is max one per 'type'? It would get tricky to define the precedence/behavior if multiple of the same type were allowed (better to specify that within the one list of that type IMO). I think maybe the association of ACL to interfaces has the same issue but perhaps the intent is to leave this flexible and potentially implementation-specific.

Do many implementations support an egress ACL on the control plane? I would think that other knobs (protocol specific, etc) are used to control the *output* rather than post-filtering on "my own output" as a router.

release/models/system/openconfig-system-controlplane.yang
jsterne: Nokia SR OS has a VRF as a match criteria inside the cpm-filter:
https://documentation.nokia.com/sr/22-10-2/books/system-management/security-system-management-1.html#ai9exj5yak__ai9exj5yaq

robshakir:
The reason that we discussed modelling it this way is that we don't necessarily have an entry in `/interfaces/interface` for the CPU bound traffic -- if we did, then we could say that this interface would be where this is bound, but we also need to have something that is not a vendor-dependent name there. The modelling choice was thus made to have this be something that is configured under `/system` and states in an opinionated way that this should be matched to the underlying interface(s) that map there.

Because then there isn't an interface in `/interfaces/interface` we need to replicate the counters that are for the ACL and QoS policies into this context to meet the current constraints.

Generally, you're asking about implementation specific constraints on how this is used -- operationally, I think an ACL that is used only for the control-plane would match all the deployments that I have seen. This seems a reasonable constraint for these use cases. We'd expect that tests for compliance enumerate (e.g., `featureprofiles` cover cases like this that are not fully constrained by the model).

I can add a note into the model to say that just one `IPV4` and one `IPV6` ACL is allowed. This is in common with ACL to interfaces, as you say.

I found that both JUNOS and IOS support the concept of an egress ACL, hence including it here.

robshakir:
Updating to reflect that Adam/Jason and I discussed offline to clarify the choice of not using a generic interface here, and agreed this route was OK.

marcushines (approved):

PR #763, opened 2022-12-05
Vendor-specific control-plane traffic counters

rolandphung:
### Change Scope

* (M) doc/vendor_counter_guide.md
  - Updated guide on how to augment /components/component/integrated-circuit/control-plane/vendor
* (M) release/models/platform/openconfig-platform-pipeline-counters.yang
  - Add support with a vendor container for queued/dropped packets/bytes
  - Add support for aggregate counters of each queued/dropped total packets/bytes
* These changes are backwards compatible

#### pyang Output
Model changes:
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:control-plane-traffic
                 +--ro oc-ppc:state
                 |  +--ro oc-ppc:total-queued?          oc-yang:counter64
                 |  +--ro oc-ppc:total-queued-bytes?    oc-yang:counter64
                 |  +--ro oc-ppc:total-dropped?         oc-yang:counter64
                 |  +--ro oc-ppc:total-dropped-bytes?   oc-yang:counter64
                 +--ro oc-ppc:vendor
```
Vendor-specific model example:
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw integrated-circuit
              +--ro oc-ppc:control-plane-traffic
                 +--ro oc-ppc:vendor
                    +--ro acme-ppc:acme
                       +--ro acme-ppc:asic-family
                          +--ro acme-ppc:state
                             +--ro acme-ppc:queue-counter-a
                             +--ro acme-ppc:queue-counter-b
                             +--ro acme-ppc:queue-counter-c
```

### Platform Implementations

 * Arista Sand: [link to documentation](https://aristanetworks.force.com/AristaCommunity/s/article/troubleshooting-based-on-control-plane-policing-copp-for-sand-platform)

### DIFF
--- a/release/models/platform/openconfig-platform-pipeline-counters.yang
+++ b/release/models/platform/openconfig-platform-pipeline-counters.yang
@@ -69,7 +69,7 @@ module openconfig-platform-pipeline-counters {
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
-  revision "2022-11-29" {
+  revision "2022-02-03" {
     description
       "Add vendor-specific control-plane traffic queue counters";
     reference "0.4.0";
@@ -1142,7 +1142,8 @@ module openconfig-platform-pipeline-counters {
           }
         }";
 
-      reference "doc/vendor_counter_guide.md";
+      reference
+        "https://github.com/openconfig/public/tree/master/doc/vendor_counter_guide.md";
     }
   }
 
@@ -1155,7 +1156,8 @@ module openconfig-platform-pipeline-counters {
       packet-processing should still be exposed as a vendor-specific,
       packet-processing counter.";
 
-    reference "doc/vendor_counter_guide.md";
+    reference
+      "https://github.com/openconfig/public/tree/master/doc/vendor_counter_guide.md";
 
     container adverse {
       description
@@ -1248,7 +1250,7 @@ module openconfig-platform-pipeline-counters {
         "This counter counts the number of packets enqueued.
 
         This counter should contribute to the total aggregate of
-        .../pipeline-counters/control-plane-traffic/state/total-queued.";
+        .../pipeline-counters/control-plane-traffic/state/queued-aggregate.";
     }
 
     leaf queued-bytes {
@@ -1257,7 +1259,7 @@ module openconfig-platform-pipeline-counters {
         "This counter counts the number of bytes enqueued.
 
         This counter should contribute to the total aggregate of
-        .../pipeline-counters/control-plane-traffic/state/total-queued-bytes.";
+        .../pipeline-counters/control-plane-traffic/state/queued-bytes-aggregate.";
     }
 
     leaf dropped {
@@ -1266,7 +1268,7 @@ module openconfig-platform-pipeline-counters {
         "This counter counts the number of packets dropped.
 
         This counter should contribute to the total aggregate of
-        .../pipeline-counters/control-plane-traffic/state/total-dropped.";
+        .../pipeline-counters/control-plane-traffic/state/dropped-aggregate.";
     }
 
     leaf dropped-bytes {
@@ -1275,7 +1277,7 @@ module openconfig-platform-pipeline-counters {
         "This counter counts the number of bytes dropped.
 
         This counter should contribute to the total aggregate of
-        .../pipeline-counters/control-plane-traffic/state/total-dropped-bytes.";
+        .../pipeline-counters/control-plane-traffic/state/dropped-bytes-aggregate.";
     }
   }
 
@@ -1323,7 +1325,8 @@ module openconfig-platform-pipeline-counters {
             }
           }";
 
-        reference "doc/vendor_counter_guide.md";
+        reference
+          "https://github.com/openconfig/public/tree/master/doc/vendor_counter_guide.md";
       }
     }
   }
--- a/release/models/platform/openconfig-platform-pipeline-counters.yang
+++ b/release/models/platform/openconfig-platform-pipeline-counters.yang
@@ -69,7 +69,7 @@ module openconfig-platform-pipeline-counters {
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
-  revision "2022-02-03" {
+  revision "2023-02-03" {
     description
       "Add vendor-specific control-plane traffic queue counters";
     reference "0.4.0";
--- a/release/models/platform/openconfig-platform-pipeline-counters.yang
+++ b/release/models/platform/openconfig-platform-pipeline-counters.yang
@@ -65,14 +65,14 @@ module openconfig-platform-pipeline-counters {
     5 blocks, is to have the abililty to receive all drop counters from
     all 5 blocks, for example, with one request.";
 
-  oc-ext:openconfig-version "0.3.1";
+  oc-ext:openconfig-version "0.4.0";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
   revision "2022-11-29" {
     description
       "Add vendor-specific control-plane traffic queue counters";
-    reference "0.3.1";
+    reference "0.4.0";
   }
 
   revision "2022-11-09" {
--- a/release/models/platform/openconfig-platform-pipeline-counters.yang
+++ b/release/models/platform/openconfig-platform-pipeline-counters.yang
@@ -65,9 +65,15 @@ module openconfig-platform-pipeline-counters {
     5 blocks, is to have the abililty to receive all drop counters from
     all 5 blocks, for example, with one request.";
 
-  oc-ext:openconfig-version "0.3.0";
+  oc-ext:openconfig-version "0.3.1";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
+
+  revision "2022-11-29" {
+    description
+      "Add vendor-specific control-plane traffic queue counters";
+    reference "0.3.1";
+  }
 
   revision "2022-11-09" {
     description
@@ -410,6 +416,8 @@ module openconfig-platform-pipeline-counters {
           }
         }
       }
+
+      uses pipeline-control-plane-top;
     }
   }
 
@@ -444,7 +452,6 @@ module openconfig-platform-pipeline-counters {
         "Outgoing bytes towards the line interfaces or fabric from the
         integrated-circuit interface subsystem block.";
     }
-
   }
 
   grouping pipeline-counters-common-high-low-packets {
@@ -1106,7 +1113,7 @@ module openconfig-platform-pipeline-counters {
         parts where packets may be dropped at any point in time. Providing
         specific hardware counters provides better visibility into traffic drop.
 
-        The recommended useage of this container is to create an augment at
+        The recommended usage of this container is to create an augment at
         .../pipeline-counter/drop/vendor that contains additional vendor/platform
         specific containers.
 
@@ -1177,6 +1184,130 @@ module openconfig-platform-pipeline-counters {
       container state {
         description
           "State container for vendor specific packet processing counters.";
+      }
+    }
+  }
+
+  grouping control-plane-traffic-counters-state {
+    description
+      "Control plane traffic counter state grouping.";
+
+    leaf total-queued {
+      type oc-yang:counter64;
+      description
+        "This captures the aggregation of all counters where the switch has enqueued
+        traffic related to the control-plane.";
+    }
+
+    leaf total-queued-bytes {
+      type oc-yang:counter64;
+      description
+        "This captures the aggregation of all counters in bytes where the switch has
+        enqueued traffic related to the control-plane.";
+    }
+
+    leaf total-dropped {
+      type oc-yang:counter64;
+      description
+        "This captures the aggregation of all counters where the switch has dropped
+        traffic related to the control-plane.";
+    }
+
+    leaf total-dropped-bytes {
+      type oc-yang:counter64;
+      description
+        "This captures the aggregation of all counters in bytes where the switch has
+        dropped traffic related to the control-plane.";
+    }
+  }
+
+  grouping control-plane-traffic-vendor-counters {
+    description
+      "A utility grouping for vendors to use when augmenting the vendor-specific
+      control-plane traffic container.";
+
+    leaf queued {
+      type oc-yang:counter64;
+      description
+        "This counter counts the number of packets enqueued.
+
+        This counter should contribute to the total aggregate of
+        .../pipeline-counters/control-plane-traffic/state/total-queued.";
+    }
+
+    leaf queued-bytes {
+      type oc-yang:counter64;
+      description
+        "This counter counts the number of bytes enqueued.
+
+        This counter should contribute to the total aggregate of
+        .../pipeline-counters/control-plane-traffic/state/total-queued-bytes.";
+    }
+
+    leaf dropped {
+      type oc-yang:counter64;
+      description
+        "This counter counts the number of packets dropped.
+
+        This counter should contribute to the total aggregate of
+        .../pipeline-counters/control-plane-traffic/state/total-dropped.";
+    }
+
+    leaf dropped-bytes {
+      type oc-yang:counter64;
+      description
+        "This counter counts the number of bytes dropped.
+
+        This counter should contribute to the total aggregate of
+        .../pipeline-counters/control-plane-traffic/state/total-dropped-bytes.";
+    }
+  }
+
+  grouping pipeline-control-plane-top {
+    description
+      "Top-level structural grouping for control-plane traffic counters.";
+
+    container control-plane-traffic {
+      description
+        "Counters that are related to traffic destined to the control-plane.";
+
+      container state {
+        config false;
+        description
+          "State container for control-plane traffic counters.";
+
+        uses control-plane-traffic-counters-state;
+      }
+
+      container vendor {
+        description
+          "Counters within these containers are defined and augmented by vendors.
+          As each ASIC and vendor has different implementation and internal
+          parts where packets may be dropped at any point in time. Providing
+          vendor-specific counters provides better visibility into control-plane traffic.
+
+          The recommended usage of this container is to create an augment at
+          .../pipeline-counter/control-plane-traffic/vendor that contains additional
+          vendor/platform specific containers.
+
+          e.g.
+          augment /components/component/integrated-circuit/pipeline-counter/control-plane-traffic/vendor {
+            container <vendor name> {
+              container <platform name> {
+                container state {
+                  leaf counter-a {
+                    uses control-plane-traffic-vendor-counters;
+                  }
+
+                  leaf counter-b {
+                    uses control-plane-traffic-vendor-counters;
+                  }
+                }
+              }
+            }
+          }";
+
+        reference "doc/vendor_counter_guide.md";
       }
     }
   }
--- a/release/models/platform/openconfig-platform-pipeline-counters.yang
+++ b/release/models/platform/openconfig-platform-pipeline-counters.yang
@@ -1192,28 +1192,28 @@ module openconfig-platform-pipeline-counters {
     description
       "Control plane traffic counter state grouping.";
 
-    leaf total-queued {
+    leaf queued-aggregate {
       type oc-yang:counter64;
       description
         "This captures the aggregation of all counters where the switch has enqueued
         traffic related to the control-plane.";
     }
 
-    leaf total-queued-bytes {
+    leaf queued-bytes-aggregate {
       type oc-yang:counter64;
       description
         "This captures the aggregation of all counters in bytes where the switch has
         enqueued traffic related to the control-plane.";
     }
 
-    leaf total-dropped {
+    leaf dropped-aggregate {
       type oc-yang:counter64;
       description
         "This captures the aggregation of all counters where the switch has dropped
         traffic related to the control-plane.";
     }
 
-    leaf total-dropped-bytes {
+    leaf dropped-bytes-aggregate {
       type oc-yang:counter64;
       description
         "This captures the aggregation of all counters in bytes where the switch has
--- a/doc/vendor_counter_guide.md
+++ b/doc/vendor_counter_guide.md
@@ -2,7 +2,7 @@
 
 **Contributors**: roland@arista.com
 
-This document provides the guidelines for the vendor-specific portions of openconfig-pipeline-counters.yang. As implementations differ from vendor to vendor and platform to platform, a process of adding vendor-specific counters will be defined here.
+This document provides the guidelines for the vendor-specific portions of openconfig-platform. As implementations differ from vendor to vendor and platform to platform, a process of adding vendor-specific counters will be defined here.
 
 ## Usage: Vendor-specific pipeline drop counter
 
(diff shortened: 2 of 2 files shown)


release/models/platform/openconfig-platform-pipeline-counters.yang
```diff
@@ -65,10 +65,16 @@ module openconfig-platform-pipeline-counters {
     5 blocks, is to have the abililty to receive all drop counters from
     all 5 blocks, for example, with one request.";
 
-  oc-ext:openconfig-version "0.3.0";
+  oc-ext:openconfig-version "0.3.1";
```
robshakir, line 68: This should be a minor revision since it adds leaves.

release/models/platform/openconfig-platform-pipeline-counters.yang
```diff
@@ -1181,6 +1188,130 @@ module openconfig-platform-pipeline-counters {
     }
   }
 
+  grouping control-plane-traffic-counters-state {
```
robshakir, line 1191: Do we believe that this is the maximum commonality of these counters that we can get? It seems to me like there is an opportunity for us to have breakdowns of counters based on the configured control-plane protection/classifiers that are there.... would this be handled in the proposed way to add such policers? If not, is there a proposal as to how to handle this?

robshakir (changes requested):

release/models/platform/openconfig-platform-pipeline-counters.yang
dplore: I can speculate that JunOS may have flexibility report counters based on the configured policy as they appear to use the same firewall filters used on regular interfaces.  But the other implementations appear to have configuration and counters that are unique to control plane traffic and not shared with QoS and ACL counters.  

@rolandphung @aredmon8551 @rgwilton @hobbydos for comment on their platform's ability to to report control plane protection counters using defined QoS and ACL rules vs. system defined/fixed counters specific to control plane protection.

I think we should keep front of mind the operational use cases for these counters which are the same as the pipeline counters defined in #715 to which these are an expansion of the same pattern.  The use cases being:
1.  coarse identification of the cause or source of device packet loss as either intended (due to policy) or unintended (due to device health/error) to make a decision on isolating/draining the device to mitigate loss due if the cause is device errors/un-healthy device. 
2. After mitigation, assist deep root cause analysis by making time series debug data available to the device designer/owner which includes device hardware specific counters.

dplore:
adding to OpenConfig Operators meeting on Jan 10th,2023 for review

release/models/platform/openconfig-platform-pipeline-counters.yang
```diff
@@ -1181,6 +1188,130 @@ module openconfig-platform-pipeline-counters {
...
+      type oc-yang:counter64;
+      description
+        "This captures the aggregation of all counters in bytes where the switch has
+        enqueued traffic related to the control-plane.";
+    }
+
+    leaf total-dropped {
+      type oc-yang:counter64;
+      description
+        "This captures the aggregation of all counters where the switch has dropped
```
aredmon8551, line 1212: Would this apply to both queue drops and filter/ACL drops?

dplore: Yes, this should capture all packet drops to the control plane whether they are due to policer, queue tail drop, ACL drop or a drop due to an error.

release/models/platform/openconfig-platform-pipeline-counters.yang
dplore: In #762 (Control Plane Protection model) the referenced QoS and [defined sets ACL](https://github.com/openconfig/public/blob/master/release/models/defined-sets/openconfig-defined-sets.yang) models have counters.  (ref: [OC QoS interface counters](https://github.com/openconfig/public/blob/f04dff4e782f9795da6e06323353c44000853a07/release/models/qos/openconfig-qos-interfaces.yang#L325-L335)). Using the proposed #762 model for protection policy and the related OC QoS interface counters would be the most vendor neutral and represent a consistent counter model.  

I'd like to hear from @rolandphung @aredmon8551 @rgwilton @hobbydos if their platforms implementation of control  plane protection can map to [OC QoS interface counters](https://github.com/openconfig/public/blob/f04dff4e782f9795da6e06323353c44000853a07/release/models/qos/openconfig-qos-interfaces.yang#L325-L335),

dplore:
@rolandphung  is this also intended to cover counters for interfaces/ports such as an Ethernet interface which is part of a CONTROLLER_CARD component?

rolandphung:
> @rolandphung is this also intended to cover counters for interfaces/ports such as an Ethernet interface which is part of a CONTROLLER_CARD component?

Sorry, I'm a bit confused about the interface counters requirements, @dplore can you help clarify a few things:

- Are you thinking about any specific counters?
- Are those counters defined in `/interfaces/interface`?
- Also which interface/ports would be a part of the CONTROLLER_CARD component?

dplore:
> > @rolandphung is this also intended to cover counters for interfaces/ports such as an Ethernet interface which is part of a CONTROLLER_CARD component?
> 
> Sorry, I'm a bit confused about the interface counters requirements, @dplore can you help clarify a few things:
> 
> * Are you thinking about any specific counters?
> * Are those counters defined in `/interfaces/interface`?

These are the counters I am thinking of:
```
/openconfig-qos:qos/interfaces/interface/input/queues/queue/state/avg-queue-len
/openconfig-qos:qos/interfaces/interface/input/queues/queue/state/max-queue-len
/openconfig-qos:qos/interfaces/interface/input/queues/queue/state/transmit-octets
/openconfig-qos:qos/interfaces/interface/input/queues/queue/state/transmit-pkts
/openconfig-qos:qos/interfaces/interface/input/queues/queue/state/dropped-octets
/openconfig-qos:qos/interfaces/interface/input/queues/queue/state/dropped-pkts

/openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/dropped-octets
/openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/dropped-pkts
/openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/max-queue-len
/openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/name
/openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/transmit-octets
/openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/transmit-pkts

/openconfig-qos:qos/interfaces/interface/output/queues/queue/state/avg-queue-len
/openconfig-qos:qos/interfaces/interface/output/queues/queue/state/dropped-octets
/openconfig-qos:qos/interfaces/interface/output/queues/queue/state/dropped-pkts
/openconfig-qos:qos/interfaces/interface/output/queues/queue/state/max-queue-len
/openconfig-qos:qos/interfaces/interface/output/queues/queue/state/transmit-octets
/openconfig-qos:qos/interfaces/interface/output/queues/queue/state/transmit-pkts
```

> * Also which interface/ports would be a part of the CONTROLLER_CARD component?

For example, a typical Arista supervisor module (CONTROLL_CARD) has 2 Ethernet ports named `management 1` and `management 2`.

dplore:
This was reviewed at the Jan 17th OC operators meeting with the following comment:
- update the vendor neutral leaves to use the same name pattern as previous leaves (*-aggregate)
- Note: #762 is intended to also be supported to configure CoPP and the related QoS counters should also be supported.  

@rgwilton @hobbydos @aredmon8551 please provide feedback.

Hobbydos:
@dplore someone from my org will provide feedback shortly.

release/models/platform/openconfig-platform-pipeline-counters.yang
robshakir: I agree that this is a good pattern to form.

Note, I don't think the references to the "defined sets ACL" model is useful here. This is essentially an optimisation to the existing ACL model to have multiple prefixes that are in reusable sets be referenced. This is _not_ a requirement to implement to have these counters - as long as the input configuration is one-prefix-per-entry then the existing model will continue to operate.

The QoS model being how we get counters - and this applying to the aggregate CPU bound traffic - seems a reasonable way to get the same granularity as is implemented there. This model would be how counters that are more specific than that, and potentially NPU or implementation specific would be exposed.

robshakir:
> > > @rolandphung is this also intended to cover counters for interfaces/ports such as an Ethernet interface which is part of a CONTROLLER_CARD component?
> > 
> > 
> > Sorry, I'm a bit confused about the interface counters requirements, @dplore can you help clarify a few things:
> > 
> > * Are you thinking about any specific counters?
> > * Are those counters defined in `/interfaces/interface`?
> 
> These are the counters I am thinking of:
> 
> ```
> /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/avg-queue-len
> /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/max-queue-len
> /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/transmit-octets
> /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/transmit-pkts
> /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/dropped-octets
> /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/dropped-pkts
> 
> /openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/dropped-octets
> /openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/dropped-pkts
> /openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/max-queue-len
> /openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/name
> /openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/transmit-octets
> /openconfig-qos:qos/interfaces/interface/input/virtual-output-queues/voq-interface/queues/queue/state/transmit-pkts
> 
> /openconfig-qos:qos/interfaces/interface/output/queues/queue/state/avg-queue-len
> /openconfig-qos:qos/interfaces/interface/output/queues/queue/state/dropped-octets
> /openconfig-qos:qos/interfaces/interface/output/queues/queue/state/dropped-pkts
> /openconfig-qos:qos/interfaces/interface/output/queues/queue/state/max-queue-len
> /openconfig-qos:qos/interfaces/interface/output/queues/queue/state/transmit-octets
> /openconfig-qos:qos/interfaces/interface/output/queues/queue/state/transmit-pkts
> ```
> 
> > * Also which interface/ports would be a part of the CONTROLLER_CARD component?
> 
> For example, a typical Arista supervisor module (CONTROLL_CARD) has 2 Ethernet ports named `management 1` and `management 2`.

@dplore -- I'm not sure whether this is actually the right mapping  -- can you explain a bit further? We have these counters but these are not where the control-plane facing QoS would be instantiated in the proposals that we have open today.

In #762, we have `/system/control-plane-traffic` defined. I think that we would really need to have `/system/control-plane-traffic/input/queues/queue/state` and `/system/control-plane-traffic/output/queues/queues/state` defined such that we have a way to get the same counters but in a way that doesn't imply a new  "interface" (which would need to be in `/interfaces/interface`) as the way that traffic is sent to the control-plane.

I can add to this PR that we have this if we agree that this is the right way forward.

release/models/platform/openconfig-platform-pipeline-counters.yang
dplore: I am ok with having both vendor neutral counters in #762, the vendor neutral integrated circuit aggregate counters here and the vendor specific integrated circuit counters for control plane here.   Regarding the ACL, we can resolve that in #762.

dplore:
> > > > @rolandphung is this also intended to cover counters for interfaces/ports such as an Ethernet interface which is part of a CONTROLLER_CARD component?
> > > 
> > > 
> > > Sorry, I'm a bit confused about the interface counters requirements, @dplore can you help clarify a few things:
> > > 
> > > * Are you thinking about any specific counters?
> > > * Are those counters defined in `/interfaces/interface`?
> > 
> > 
> > These are the counters I am thinking of:
> > ```
> > /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/avg-queue-len
> > /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/max-queue-len
> > /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/transmit-octets
> > /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/transmit-pkts
> > /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/dropped-octets
> > /openconfig-qos:qos/interfaces/interface/input/queues/queue/state/dropped-pkts
> > ```
> 
> @dplore -- I'm not sure whether this is actually the right mapping -- can you explain a bit further? We have these counters but these are not where the control-plane facing QoS would be instantiated in the proposals that we have open today.
> 
> In #762, we have `/system/control-plane-traffic` defined. I think that we would really need to have `/system/control-plane-traffic/input/queues/queue/state` and `/system/control-plane-traffic/output/queues/queues/state` defined such that we have a way to get the same counters but in a way that doesn't imply a new "interface" (which would need to be in `/interfaces/interface`) as the way that traffic is sent to the control-plane.
> 
> I can add to this PR that we have this if we agree that this is the right way forward.

You are correct, I intended the vendor neutral control plane QoS counters to be aligned in concept with what is proposed in #762 and that the counters be the same queue state counters already defined for traditional interfaces.  Agreed that #762 does not yet have a container with those counters, but the pattern might look something like the above where `/system/control-plane-traffic` is substituted for `/interfaces/interface`.  Thanks for volunteering to add that to #762 (or a fresh PR if you prefer).

I am also ok with also having vendor specific control plane counters in addition to vendor neutral counters to be defined in #762.

release/models/platform/openconfig-platform-pipeline-counters.yang
dplore: @robshakir after the above discussion, shall we mark this comment resolved?   This comment is requesting changes.  Are there changes or open questions you would like answered to see to this PR currently?

Hobbydos: > In #762 (Control Plane Protection model) the referenced QoS and [defined sets ACL](https://github.com/openconfig/public/blob/master/release/models/defined-sets/openconfig-defined-sets.yang) models have counters. (ref: [OC QoS interface counters](https://github.com/openconfig/public/blob/f04dff4e782f9795da6e06323353c44000853a07/release/models/qos/openconfig-qos-interfaces.yang#L325-L335)). Using the proposed #762 model for protection policy and the related OC QoS interface counters would be the most vendor neutral and represent a consistent counter model.
> 
> I'd like to hear from @rolandphung @aredmon8551 @rgwilton @Hobbydos if their platforms implementation of control plane protection can map to [OC QoS interface counters](https://github.com/openconfig/public/blob/f04dff4e782f9795da6e06323353c44000853a07/release/models/qos/openconfig-qos-interfaces.yang#L325-L335),

@nokia1adam, Could you PTAL?

nokia1adam: As a standalone PR I'm not sure what this accomplishes. We already have pipeline-counters/host-interface-block, which to me is indistinguishable in scope from the pipeline-counters/control-plane-traffic addition proposed by this PR.

My general issue with the current host-interface-block state model is that it only splits the counters up into "high" and "low" priority buckets, which assumes a scheduling model in which there are two classes (or two queues) and one is strict priority over the other. Real implementations likely have 8, or 16, or 32+ queues from each IC, with the scheduling of those queues possibly WRR instead of strict priority. These queues could also have PIR/CIR rate limits. If you imagined a state-only qos scheduler-policy applied to each IC you would be much closer to reality.

Finally, I don't think this topic should be mixed up with configured CoPP policies (PR 762 for example). The rules that a user configures for protection of their control plane are not 1:1 aligned with the datapath pipeline rules that determine which flows should be extracted to the control plane engine. The former should be presented as CoPP policy statistics while the latter should be presented as pipeline counters.

dplore:
Last call for comments, this will be merged on Feb 2, 2023

aredmon8551:
Challenge I see: Because this is nested under integrated-circuit it would seem to scope this to only control-plane traffic received via the ASIC/front-panel ports.  Many systems today have management ports that are connected directly to the same board that the control-plane CPU resides on.  Bottom line is you can have two places where these counters live: in the ASIC/hardware and in the kernel of the control-plane.

aredmon8551:
> Challenge I see: Because this is nested under integrated-circuit it would seem to scope this to only control-plane traffic received via the ASIC/front-panel ports. Many systems today have management ports that are connected directly to the same board that the control-plane CPU resides on. Bottom line is you can have two places where these counters live: in the ASIC/hardware and in the kernel of the control-plane.

Forgot to include that backpressure within the control-plane (i.e., CPU time contention) could result in non-zero dropped/queued counters for the kernel's software interface.  So I guess I'm making the point that the OS would have its own set of counters and it may be worth modeling those separately.

dplore:
> > Challenge I see: Because this is nested under integrated-circuit it would seem to scope this to only control-plane traffic received via the ASIC/front-panel ports. Many systems today have management ports that are connected directly to the same board that the control-plane CPU resides on. Bottom line is you can have two places where these counters live: in the ASIC/hardware and in the kernel of the control-plane.
> 
> Forgot to include that backpressure within the control-plane (i.e., CPU time contention) could result in non-zero dropped/queued counters for the kernel's software interface. So I guess I'm making the point that the OS would have its own set of counters and it may be worth modeling those separately.

This model is intended to only cover drops in INTEGRATED_CIRCUIT components.  Drops in control plane packets that occur outside of an implementation's INTEGRATED_CIRCUIT components should be counted in in the /system/control-plane-traffic model (#762).   

The most important operational goal of this PR and the counter portion of #762 is to not miss counting of packet drops in the control plane pipeline with models that allows implementations to express those drops.

aredmon8551:
> > > Challenge I see: Because this is nested under integrated-circuit it would seem to scope this to only control-plane traffic received via the ASIC/front-panel ports. Many systems today have management ports that are connected directly to the same board that the control-plane CPU resides on. Bottom line is you can have two places where these counters live: in the ASIC/hardware and in the kernel of the control-plane.
> > 
> > 
> > Forgot to include that backpressure within the control-plane (i.e., CPU time contention) could result in non-zero dropped/queued counters for the kernel's software interface. So I guess I'm making the point that the OS would have its own set of counters and it may be worth modeling those separately.
> 
> This model is intended to only cover drops in INTEGRATED_CIRCUIT components. Drops in control plane packets that occur outside of an implementation's INTEGRATED_CIRCUIT components should be counted in in the /system/control-plane-traffic model (#762).
> 
> The most important operational goal of this PR and the counter portion of #762 is to not miss counting of packet drops in the control plane pipeline with models that allows implementations to express those drops.

Ack, thanks for the clarification.  I think this is fine as specified and believe Junos has the appropriate native counters to support it as defined.

I'll comment on #762 regarding kernel counters.

dplore:
@Hobbydos @nokia1adam do you have further comments given the recent explanations above?

nokia1adam:
@dplore, thanks. We are fine with this PR, however I would still like some clarity on two points:
1. what is the role/purpose of pipeline-counters/host-interface-block after this change? seems redundant to me.
2. queued-aggregate is a bit ambiguous. It would have been good if the description had stated whether this count is pre-discard (entry to the queues) or post-discard (exit from the queues).

dplore:
> @dplore, thanks. We are fine with this PR, however I would still like some clarity on two points:
> 
> 1. what is the role/purpose of pipeline-counters/host-interface-block after this change? seems redundant to me.

Agreed, it is redundant.  At this moment we are not planning to create featureprofiles tests for pipeline-counters/host-interface-block.  The focus for featureprofiles test coverage of pipeline-counters is #715 and #763 (this PR).  

> 2. queued-aggregate is a bit ambiguous. It would have been good if the description had stated whether this count is pre-discard (entry to the queues) or post-discard (exit from the queues).

@rolandphung can you clarify queued-aggregate intention?

release/models/platform/openconfig-platform-pipeline-counters.yang
```diff
@@ -65,10 +65,16 @@ module openconfig-platform-pipeline-counters {
     5 blocks, is to have the abililty to receive all drop counters from
     all 5 blocks, for example, with one request.";
 
-  oc-ext:openconfig-version "0.3.1";
+  oc-ext:openconfig-version "0.4.0";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
+  revision "2022-11-29" {
```
dplore, line 72: @rolandphung please update the revision date to fix the pyangbind check.

```messages:{path:"/workspace/release/models/platform/openconfig-platform-pipeline-counters.yang" line:78 code:"REVISION_ORDER" type:"warning" level:4 message:'the revision statements are not given in reverse chronological order'}
```

robshakir, line 72:
```suggestion
  revision "2023-02-02" {
```

release/models/platform/openconfig-platform-pipeline-counters.yang
```diff
@@ -1197,6 +1204,130 @@ module openconfig-platform-pipeline-counters {
...
+
+                  leaf counter-b {
+                    uses control-plane-traffic-vendor-counters;
+                  }
+                }
+              }
+            }
+          }";
+
+        reference "doc/vendor_counter_guide.md";
```
robshakir, line 1326:
```suggestion
        reference
          "https://github.com/openconfig/public/tree/master/doc/vendor_counter_guide.md";
```

kamiraux:
> > 2. queued-aggregate is a bit ambiguous. It would have been good if the description had stated whether this count is pre-discard (entry to the queues) or post-discard (exit from the queues).
> 
> @rolandphung can you clarify queued-aggregate intention?

I think this is simply meant to be the sum of the counters defined by vendor-specific augments (the ones in control-plane-traffic-vendor-counters), so probably vague on purpose with the intention to have vendor clarify their specific behavior in the augment.

Regarding the pre/post, most queue counters are entry counts, that is "queued" means the packet entered the queue, and "dropped" means it didn't enter the queue, but it doesn't say whether the packet exited the queue. Packets being dropped at the exit of the queue is possible but generally unexpected AFAIK (though, my view could be biased by the chips I worked on).

release/models/platform/openconfig-platform-pipeline-counters.yang
```diff
@@ -1197,6 +1204,130 @@ module openconfig-platform-pipeline-counters {
...
+      "A utility grouping for vendors to use when augmenting the vendor-specific
+      control-plane traffic container.";
+
+    leaf queued {
+      type oc-yang:counter64;
+      description
+        "This counter counts the number of packets enqueued.
+
+        This counter should contribute to the total aggregate of
+        .../pipeline-counters/control-plane-traffic/state/total-queued.";
```
kamiraux, line 1251: Probably need to change total-queued to queued-aggregate, following the rename in [2614b5a429f352175e4ffbf6bd0c0ec119d18771](https://github.com/openconfig/public/pull/763/commits/2614b5a429f352175e4ffbf6bd0c0ec119d18771).

Same on the similar attributes below.

robshakir (approved):

dplore:
> > > 2. queued-aggregate is a bit ambiguous. It would have been good if the description had stated whether this count is pre-discard (entry to the queues) or post-discard (exit from the queues).
> > 
> > 
> > @rolandphung can you clarify queued-aggregate intention?
> 
> I think this is simply meant to be the sum of the counters defined by vendor-specific augments (the ones in control-plane-traffic-vendor-counters), so probably vague on purpose with the intention to have vendor clarify their specific behavior in the augment.
> 
> Regarding the pre/post, most queue counters are entry counts, that is "queued" means the packet entered the queue, and "dropped" means it didn't enter the queue, but it doesn't say whether the packet exited the queue. Packets being dropped at the exit of the queue is possible but generally unexpected AFAIK (though, my view could be biased by the chips I worked on).

I would expect any packet/byte making it into a queue is counted as queued.    Any packet dropped (pre or post or tail dropped) would be counted as `dropped-aggregate`.   In the end, it is up to the vendor to expose more detailed counters for pre/post queue drop or any other details that might be available to count packet drops in the hardware.

`pipeline-counters/control-plane-traffic/
                 +--ro oc-ppc:vendor
                    +--ro acme-ppc:acme
                       +--ro acme-ppc:asic-family

dplore (approved):

PR #764, opened 2022-12-05
Cleaning up trailing whitespace

rolandphung:
This is to get the trailing whitespace checker to pass for other PRs
* (M) release/models/optical-transport/optical-transport/openconfig-transport-types.yang


dplore (approved):

dplore:
Can you also update openconfig-transport-types.yang version?  I will approve and merge for you right away

PR #767, opened 2022-12-06
Added NTP configuration and state to network-instance model.

lgomez9:
### Change Scope
Added NTP to the network-instance model.

```
     +--rw ntp
     |  +--rw config
     |  |  +--rw enabled?           boolean
     |  |  +--rw enable-ntp-auth?   boolean
     |  +--ro state
     |  |  +--ro enabled?           boolean
     |  |  +--ro enable-ntp-auth?   boolean
     |  |  +--ro auth-mismatch?     oc-yang:counter64
     |  +--rw ntp-keys
     |  |  +--rw ntp-key* [key-id]
     |  |     +--rw key-id    -> ../config/key-id
     |  |     +--rw config
     |  |     |  +--rw key-id?      uint16
     |  |     |  +--rw key-type?    identityref
     |  |     |  +--rw key-value?   string
     |  |     +--ro state
     |  |        +--ro key-id?      uint16
     |  |        +--ro key-type?    identityref
     |  |        +--ro key-value?   string
     |  +--rw servers
     |     +--rw server* [address]
     |        +--rw config
     |        |  +--rw address?              oc-inet:host
     |        |  +--rw port?                 oc-inet:port-number
     |        |  +--rw version?              uint8
     |        |  +--rw association-type?     enumeration
     |        |  +--rw iburst?               boolean
     |        |  +--rw prefer?               boolean
     |        |  +--rw network-instance?     oc-ni:network-instance-ref
     |        |  +--rw ntp-source-address?   oc-inet:ip-address
     |        +--ro state
     |           +--ro address?              oc-inet:host
     |           +--ro port?                 oc-inet:port-number
     |           +--ro version?              uint8
     |           +--ro association-type?     enumeration
     |           +--ro iburst?               boolean
     |           +--ro prefer?               boolean
     |           +--ro network-instance?     oc-ni:network-instance-ref
     |           +--ro ntp-source-address?   oc-inet:ip-address
     |           +--ro stratum?              uint8
     |           +--ro root-delay?           uint32
     |           +--ro root-dispersion?      uint64
     |           +--ro offset?               uint64
     |           +--ro poll-interval?        uint32
```

##References
[NTP client on Arista EOS](https://www.arista.com/en/um-eos/eos-system-clock-and-time-protocols#xx1106323) supports specifying a vrf per ntp peer/server.

[NTP client on Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/system_management/command/reference/yr40xr12k_chapter9.html#wp1282222071) supports specifying a vrf per ntp peer/server

[NTP on JunOS](https://www.juniper.net/documentation/us/en/software/junos/time-mgmt/topics/topic-map/network-time-protocol.html#id_fqr_1dg_grb) supports a management vrf

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -47,12 +47,13 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.16.2";
+  oc-ext:openconfig-version "0.17.0";
   
   revision "2022-12-20" {
     description
-      "Added network instance leaf to ntp client.";
-    reference "0.16.2";
+      "Added network instance leaf to ntp client, and dropped
+      ntp-source-address from global to individual config.";
+    reference "0.17.0";
   }
 
   revision "2022-12-19" {
@@ -714,6 +715,12 @@ module openconfig-system {
       description
         "The network instance used to find this server.";
     }
+
+    leaf ntp-source-address {
+      type oc-inet:ip-address;
+      description
+        "Source address to use on outgoing NTP packets";
+    }
   }
 
   grouping system-ntp-server-state {
@@ -912,12 +919,6 @@ module openconfig-system {
         "Enables and disables the NTP protocol and indicates that the system should
         attempt to synchronize the system clock with an NTP server
         from the servers defined in the 'ntp/server' list.";
-    }
-
-    leaf ntp-source-address {
-      type oc-inet:ip-address;
-      description
-        "Source address to use on outgoing NTP packets";
     }
 
     leaf enable-ntp-auth {
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -51,7 +51,11 @@ module openconfig-system {
 
   revision "2022-12-14" {
     description
+<<<<<<< HEAD
       "Added network instance leaf to ntp client.";
+=======
+      "Added network instance leaf to ntp server.";
+>>>>>>> f8d063f103bc8b7d56e1e640642dd7a7ad9e948e
     reference "0.16.1";
   }
 
@@ -702,10 +706,17 @@ module openconfig-system {
         or not.";
     }
 
+<<<<<<< HEAD
     leaf network-instance {
       type oc-ni:network-instance-ref;
       description
         "The network instance used to find this server.";
+=======
+    leaf-list network-instances {
+      type oc-ni:network-instance-ref;
+      description
+        "The list of network instances to which the NTP server is listening.";
+>>>>>>> f8d063f103bc8b7d56e1e640642dd7a7ad9e948e
     }
   }
 
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -49,13 +49,9 @@ module openconfig-system {
 
   oc-ext:openconfig-version "0.16.1";
 
-  revision "2022-12-14" {
+  revision "2022-12-19" {
     description
-<<<<<<< HEAD
       "Added network instance leaf to ntp client.";
-=======
-      "Added network instance leaf to ntp server.";
->>>>>>> f8d063f103bc8b7d56e1e640642dd7a7ad9e948e
     reference "0.16.1";
   }
 
@@ -706,17 +702,10 @@ module openconfig-system {
         or not.";
     }
 
-<<<<<<< HEAD
     leaf network-instance {
       type oc-ni:network-instance-ref;
       description
         "The network instance used to find this server.";
-=======
-    leaf-list network-instances {
-      type oc-ni:network-instance-ref;
-      description
-        "The list of network instances to which the NTP server is listening.";
->>>>>>> f8d063f103bc8b7d56e1e640642dd7a7ad9e948e
     }
   }
 
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -49,9 +49,9 @@ module openconfig-system {
 
   oc-ext:openconfig-version "0.16.1";
 
-  revision "2022-12-06" {
+  revision "2022-12-14" {
     description
-      "Added NTP to the network instance schema.";
+      "Added network instance leaf to ntp client.";
     reference "0.16.1";
   }
 
@@ -701,6 +701,12 @@ module openconfig-system {
         "Indicates whether this server should be preferred
         or not.";
     }
+
+    leaf network-instance {
+      type oc-ni:network-instance-ref;
+      description
+        "The network instance used to find this server.";
+    }
   }
 
   grouping system-ntp-server-state {
@@ -1256,13 +1262,6 @@ module openconfig-system {
       uses oc-license:license-top;
       uses system-macaddr-top;
     }
-  }
-
-  augment "/oc-ni:network-instances/oc-ni:network-instance" {
-    description
-      "Augment the NTP configuration and state into network instance.";
-
-    uses system-ntp-top;
   }
 
   // data definition statements
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -20,6 +20,7 @@ module openconfig-system {
   import openconfig-alarms { prefix oc-alarms; }
   import openconfig-messages { prefix oc-messages; }
   import openconfig-license { prefix oc-license; }
+  import openconfig-network-instance { prefix oc-ni; }
 
   // meta
   organization "OpenConfig working group";
@@ -1249,6 +1250,13 @@ module openconfig-system {
       uses oc-license:license-top;
       uses system-macaddr-top;
     }
+  }
+
+  augment "/oc-ni:network-instances/oc-ni:network-instance" {
+    description
+      "Augment the NTP configuration and state into network instance.";
+
+    uses system-ntp-top;
   }
 
   // data definition statements
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -47,11 +47,17 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.16.1";
+  oc-ext:openconfig-version "0.16.2";
+  
+  revision "2022-12-20" {
+    description
+      "Added network instance leaf to ntp client.";
+    reference "0.16.2";
+  }
 
   revision "2022-12-19" {
     description
-      "Added network instance leaf to ntp client.";
+      "Update last configuration timestamp documentation.";
     reference "0.16.1";
   }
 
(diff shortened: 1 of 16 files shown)


dplore:
Adding to openconfig operators agenda for Dec 13, 2022

earies:
I feel this pattern is incorrect as proposed - you are attempting to just reuse an existing `/system/ntp` construct also underneath `/network-instances/network-instance[name=XYZ]/ntp` so this has now created 2 methods or a possible migration path but now throws a domain at the NI root.  You now have a conflict between the `DEFAULT_INSTANCE` and the global `/system/ntp` method.  Organization and hierarchy should be carefully considered otherwise we are likely to see scattered constructs over time.

In addition, due to how network-instance models import a variety of other models, augmenting back into the network-instance domain needs to be cautious otherwise there is the possible risk of circular dependencies - in this case, it is safe from this issue and augments should be a preferable design pattern vs. wide imports or submodules imo.

This issue is larger however onto network-instance "aware" services in general.  To date, there have been 2 design patterns

1. Root service domains under network-instances
2. Reference a network-instance at a list member construct in a global service domain

For the 1st method, this can be desirable, expansible and will require unmodeled server-side restrictions per implementation.  This means *all* services should be reconsidered for this approach - not just NTP and a proper hierarchy be introduced under NI for items such as "system" services

For the 2nd, some constructs could merely add a leaf that is a leafref to an NI while others may require redesign to now pair up any source related information (e.g. source-address per list member since a global definition is no longer valid if list members can belong to different NIs)

dplore:
We could consider this a migration and mark [/system/ntp](https://github.com/openconfig/public/pull/767/files#diff-39c8ff1320ad76b4552208d599622a306cc84862f6dd94d378573c980e43e85cR1246)  for removal at a future date.  Thoughts?

robshakir:
This looks the wrong pattern to me too. 

What is the change that we're proposing here in support of? Do we want to have an NTP server that is looked up in a specific NI? If so, the pattern we have elsewhere is that we have a leaf underneath a specific server specifying what network instance to look this up in, e.g., `/system/grpc-servers/grpc-server/config/network-instance`.

dplore:
@lgomez9 , the recommendation  from Dec 13, 2022 OC Operators meeting is to add a leafref defining the NI where the server is
Ie: system/ntp/servers/server/config/network-instance  

This is the same pattern as grpc for example:
/system/grpc-servers/grpc-server/state/network-instance

This patterns is suitable for other portions of the models such as tacacs.

@lgomez9  will you update your pull request please?

lgomez9:
Will update later this afternoon. Thank you everyone!

dplore:
The relevant gRPC model for `/system/grpc-servers/grpc-server/config/network-instance` is at 
https://github.com/openconfig/public/blob/69f571c3b2e2db93dd88d66c424abed45281e527/release/models/system/openconfig-system-grpc.yang#L195-L200

nkitchen:
The pattern used by gRPC can't be applied to some other services, such as NTP,  with the same expressiveness.

The config values for the `grpc-server` entries are unconstrained with respect to network instance -- any two entries in different NIs can have different ports, **or the same ones,** and likewise for `listen-addresses`, etc....because the list is keyed by name, and the operator can choose any name they want.

In contrast, `/system/ntp/servers/server` is keyed by address, so it's not possible to use the same server address in different network instances (for example, `pool.ntp.org`).

Similarly, two entries in `/sampling/sflow/collectors/collector` cannot use the same address and port in different network instances.

To get the same flexibility as we have with the gRPC model, lists such as the NTP servers and the sFlow collectors would need to be keyed by arbitrary identifiers such as names, instead of addresses. That's a bigger change to the current model structure.

dplore:
The descriptions for ntp in this module may be bit confusing.  The intent of the NTP related config/state model in /system/ntp pertains to an NTP client on a network device.  The   The `system-ntp-server-config` group is for configuring a list of NTP servers which the NTP client on the device will attempt to open connections and form ntp associations.   This PR should add the config needed for the NTP client on the device to contact NTP servers which may exist within different  network-instances.  There should be zero or one network-instance per ntp server entry.

As @nkitchen points out, the IP address of an NTP server might be the same within different network-instances.  While the operational use case we have in mind does not require this, it seems good to have the model accommodate that possibility.  Therefore the list key for /system/ntp/servers should be updated to include an index or name and the key should include or be changed to this index or name.  

for example, the tree should look something like:
```
     +--rw ntp
     |  +--rw servers
     |     +--rw server* [name]
     |        +--rw name                   string
     |        +--rw address                -> ../config/address
     |        +--rw config
     |        |  +--rw port?               oc-inet:port-number
     |        |  +--rw version?            uint8
     |        |  +--rw association-type?   enumeration
     |        |  +--rw iburst?             boolean
     |        |  +--rw prefer?             boolean
```

earies:
@dplore @nkitchen - not only do we have these design pattern inconsistencies for these that are pointed out across domains (noticed the sFlow related one shortly back as well) but flipping keys to arbitrary user-defined names not only poses the `unique` stmt issue (See the comments in https://github.com/openconfig/public/pull/736) where an implementation needs to implement its own guardrails but often will require feature overhauls in implementations since these user-defined keys do not map to existing implementations across various domains (essentially non-existant keys) - We've breached the boundaries of simple mapping exercises to attempting to restructure long lasting data structures and organization in the native implementation.

Case in point is defining 2 list members `S1` and `S2` where either the underlying attributes are identical or where you may have something like identical IP addresses but ports differ and the underlying implementation cannot support such.  The `unique` statement can help us on a direct descendant list here to ensure that not only are list keys unique but certain child nodes are unique across the list entries as well - This however cannot solve all cases in OC modeling as pointed out in the PR mentioned above

Merely adding a `network-instance` leaf is not enough.  We currently have a global `/system/ntp/config/ntp-source-address` (not sure why ntp is reiterated in the node name) that if each list entry for a server can be part of a separate network-instance, then you need to drop the source-address down to this level as well

Putting services like this under `/network-instances` may not be a bad idea after all.  We get some of what we are discussing here for free but safeguards still need to be put in place by the implementation as there is not always an expectation that all services are NI aware (along w/ the NI types driving what is/isn't supported)

In either case - we need to drive some design consistency as this problem statement applies to handfuls of subtrees atm

dplore:
Another way to simplify this is based on use case.  Our use case is to have an NTP client use only one 'management' network-instance.  We do not have a use case for multiple network instances at this time.  This would lead us to a tree looking like:

```
     +--rw ntp
     |  +--rw config
     |  |  +--rw enabled?              boolean
     |  |  +--rw ntp-source-address?   oc-inet:ip-address
     |  |  +--rw enable-ntp-auth?      boolean
     |  |  +--rw network-instance?     oc-ni:network-instance-ref
     |  +--rw servers
     |     +--rw server*
     |        +--rw address                -> ../config/address
     |        +--rw config
     |        |  +--rw port?               oc-inet:port-number
     |        |  +--rw version?            uint8
     |        |  +--rw association-type?   enumeration
     |        |  +--rw iburst?             boolean
     |        |  +--rw prefer?             boolean

```

robshakir:
IMHO, this is a worse solution than just adding the `network-instance` leaf under the server. I do see the challenge of the inconsistency here -- but can I ask whether anyone actually has a NTP server that has the same IP in a different network instance? I would strongly recommend that this is a bad network design since the amount of confusion caused by re-using infrastructure addresses across multiple VRFs would be extremely large.

It is much more typical to see that there might be NTP servers in a single VRF with unique addresses, I think this is true for most cases. I think it seems a good enough solution -- and if we want to come back and support this use case later, because some operator raises a requirement, then we can do so with a backwards incompatible change.

lgomez9:
Just to make sure I understand - @robshakir, you're saying to leave it as a single network-instance leaf, rather than the list that I changed it to this afternoon. Thank you all for the help ironing this out!

robshakir:
What @dplore said in this PR and what the leaf description says are different things. I think this is where the confusion is. @lgomez9 - what is the intention here? Is the device an NTP client or an NTP server? AIUI it's an NTP client.

Given this - I think we need to:

 * make this a singular `leaf` rather than a `leaf-list`
 * specify that this is the network instance used to resolve the reachability of the specified server in the description

If this is really the device acting as an NTP server, then we need to add a whole new subtree for this, since this is not supported in the model today.

dplore:
> What @dplore said in this PR and what the leaf description says are different things. I think this is where the confusion is. @lgomez9 - what is the intention here? Is the device an NTP client or an NTP server? AIUI it's an NTP client.
> 
> Given this - I think we need to:
> 
> * make this a singular `leaf` rather than a `leaf-list`
> * specify that this is the network instance used to resolve the reachability of the specified server in the description
> 
> If this is really the device acting as an NTP server, then we need to add a whole new subtree for this, since this is not supported in the model today.

Leo's intent is to configure an NTP client with IP addresses pointing to multiple NTP servers in a single network-instance.  Agreed, the PR as it stands does not do this.  :)

lgomez9:
I understand and apologize for the semantic confusion!

release/models/system/openconfig-system.yang
```diff
@@ -46,7 +47,13 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.16.1";
+  oc-ext:openconfig-version "0.16.2";
+  
+  revision "2022-12-20" {
+    description
+      "Added network instance leaf to ntp client.";
+    reference "0.16.2";
```
earies, line 55: New leaf additions have historically incremented the minor version - patch versions mostly just for doc changes/clarifications that are not adding/removing nodes to the tree

release/models/system/openconfig-system.yang
```diff
@@ -701,6 +708,12 @@ module openconfig-system {
         "Indicates whether this server should be preferred
         or not.";
     }
+
+    leaf network-instance {
```
earies, line 712: I mentioned this in previous comments but the addition of this needs to take into account `/system/ntp/config/ntp-source-address` now.  If a server is reachable via an NI then any source-addressing needs to be addressed per server definition

lgomez9: Is simply moving the source address down a level to the individual server config enough, or would more changes be necessary? Making that change, but can change it again!

release/models/system/openconfig-system.yang
dplore: I see Leo has updated the minor version

release/models/system/openconfig-system.yang
dplore: Leo has added source-address as a leaf in addition to network-instance as a leafref

dplore (commented):
Last call for comments, this PR will merge in two weeks on Jan 11, 2023

dplore:
@lgomez9 can you fix the whitespace issue for us please?

lgomez9:
Done!

nkitchen:
I think Ebben had a good point, and I didn't see any response to it:

> not sure why ntp is reiterated in the node name

None of other node names in the server list have `ntp-` as a prefix. Shouldn't we also drop it from `ntp-source-address`?

release/models/system/openconfig-system.yang
```diff
@@ -701,6 +709,18 @@ module openconfig-system {
...
+      type oc-ni:network-instance-ref;
+      description
+        "The network instance used to find this server.";
+    }
+
+    leaf ntp-source-address {
```
dplore, line 719:
```suggestion
    leaf source-address {
```

dplore (approved):
Thanks for the contribution Leo.

PR #768, opened 2022-12-06
Move ethernet segments to top-level container (EVPN)

oscargdd:
### Change Scope
In the first version of the Openconfig EVPN model ethernet-segments were under network instance container.
However, Ethernet Segment can be shared by a number of network-instances on the same physical interface, so the ES model shouldn’t really be defined at the MAC-VRF level.
Hence, the proposal is to move ethernet-segements from network-instance/evpn to a top level container.
The change is  NOT backwards compatible. However, there are yet very few Openconfig EVPN implementations. 
### Platform Implementations

Up to the best of my knowledge all implementations are modelling ethernet segments outside the network instance. Will update with proper references here


jsnyder81:
This makes sense to me, considering situations where you might have multiple network-instances on a single ethernet segment.  While some may choose to limit an ethernet-segment to a single network-instance, this is a relatively common scenario in enterprise networking.

release/models/network-instance/openconfig-network-instance-l2.yang
aappachi, line 1: what is changing in this file? Does it need a new revision?

aappachi: I see the oc version change. never mind.

aappachi (approved):
Changes look good to me.

dplore (approved):
LGTM.  Will merge on 2023-08-01

PR #770, opened 2022-12-19
Fix timestamp documentation

rolandphung:
### Change Scope

* Update documentation to be more specific to which time epoch to use
  * /components/component/<component-type>/utilization/resources/resource/state/last-high-watermark
  * /system/state/last-configuration-timestamp


dplore (approved):
LGTM

PR #771, opened 2022-12-20
Remove unused import from openconfig-acl

earies:
* (M) release/models/acl/openconfig-acl.yang
    - Remove unused openconfig-inet-types import

### Change Scope

Previous modifications to openconfig-acl introduced an unused import `openconfig-inet-types`.  This PR removes the import and increments the patch version.

### Platform Implementations

N/A: Backwards compatible cleanup


earies:
@dplore - simple patch fix for unused import previously introduced which results in unnecessary compiler warnings

Is likely warnings should be treated as errors to catch these in CI checks against various compilers

dplore (approved):

PR #772, opened 2022-12-22
    Add support for defining a queue identifier under queues.

aredmon8551:
* (M) qos/openconfig-qos-interfaces.yang
* (M) qos/openconfig-qos-elements.yang
* (M) qos/openconfig-qos.yang

- Add support for defining a queue identifier per queue.

This additional leaf allows the mapping of hardware or software queues to traffic classes using the system's queue identifiers (often 0-7).

<pre>
Tree view:
module: openconfig-qos
+--rw qos
    +--rw queues
        +--rw queue* [name]
            +--rw config
            |  +--rw name?       string
            |  +--rw queue-id?   uint8
            +--ro state
               +--ro name?       string
               +--ro queue-id?   uint8
</pre>
Some systems allow the mapping of traffic classes to specific queues, some just apply a default mapping.

EOS:
https://www.arista.com/en/um-eos/eos-quality-of-service#xx1156350

Juniper:
https://www.juniper.net/documentation/us/en/software/junos/cos/topics/task/cos-configuring-forwarding-classes.html

### DIFF
--- a/release/models/qos/openconfig-qos-elements.yang
+++ b/release/models/qos/openconfig-qos-elements.yang
@@ -43,8 +43,6 @@ submodule openconfig-qos-elements {
     reference "0.8.0";
   }
 
-  oc-ext:openconfig-version "0.7.0";
-
   revision "2023-02-08" {
     description
       "Remove incorrect output placement of interface-ref";
--- a/release/models/qos/openconfig-qos-elements.yang
+++ b/release/models/qos/openconfig-qos-elements.yang
@@ -35,7 +35,13 @@ submodule openconfig-qos-elements {
       packets for transmission, including policer and shaper
       functions";
 
-  oc-ext:openconfig-version "0.6.0";
+  oc-ext:openconfig-version "0.7.0";
+
+  revision "2022-12-20" {
+    description
+      "Add queue identifier.";
+    reference "0.7.0";
+  }
 
   revision "2022-09-13" {
     description
@@ -559,6 +565,11 @@ submodule openconfig-qos-elements {
       type string;
       description
         "User-defined name of the queue";
+    }
+    leaf queue-id {
+      type uint8;
+      description
+        "Identifier of the queue";
     }
   }
 
--- a/release/models/qos/openconfig-qos-elements.yang
+++ b/release/models/qos/openconfig-qos-elements.yang
@@ -569,7 +569,8 @@ submodule openconfig-qos-elements {
     leaf queue-id {
       type uint8;
       description
-        "Identifier of the queue";
+        "An optional identifier which may be required by some hardware to map 
+        the named queue to a hardware queue";
     }
   }
 
--- a/release/models/qos/openconfig-qos-elements.yang
+++ b/release/models/qos/openconfig-qos-elements.yang
@@ -569,7 +569,7 @@ submodule openconfig-qos-elements {
     leaf queue-id {
       type uint8;
       description
-        "An optional identifier which may be required by some hardware to map 
+        "An optional identifier which may be required by some hardware to map
         the named queue to a hardware queue";
     }
   }
--- a/release/models/qos/openconfig-qos-interfaces.yang
+++ b/release/models/qos/openconfig-qos-interfaces.yang
@@ -25,7 +25,13 @@ submodule openconfig-qos-interfaces {
     configuration and operational state associated with
     interfaces.";
 
-  oc-ext:openconfig-version "0.6.0";
+  oc-ext:openconfig-version "0.7.0";
+
+  revision "2022-12-20" {
+    description
+      "Add queue identifier.";
+    reference "0.7.0";
+  }
 
   revision "2022-09-13" {
     description
--- a/release/models/qos/openconfig-qos-mem-mgmt.yang
+++ b/release/models/qos/openconfig-qos-mem-mgmt.yang
@@ -29,7 +29,13 @@ submodule openconfig-qos-mem-mgmt {
       per-queue basis, and determine how packets are marked/dropped within
       the queue instantiation.";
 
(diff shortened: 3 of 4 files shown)


release/models/qos/openconfig-qos-elements.yang
```diff
@@ -560,6 +566,11 @@ submodule openconfig-qos-elements {
       description
         "User-defined name of the queue";
     }
+    leaf queue-id {
+      type uint8;
+      description
+        "Identifier of the queue";
```
dplore, line 572: Here is an attempt to elaborate on the description of `queue-id` to clarify what the meaning of the queue-id is and when to use it.  

```suggestion
        "An optional identifier which may be required by some hardware to map 
        the named queue to a hardware queue";
```

dplore (commented):
Great suggestion.    I have a comment to elaborate on the description.

release/models/qos/openconfig-qos-elements.yang
aredmon8551: Thanks, good feedback - adopted as the above as the description.

dplore:
@sulrich @amilisor @hobbydos for comment.

dplore:
Added to Jan 10, 2023 OpenConfig operator review

sulrich:
@panduranganrs for viz.

this seems reasonable to me.  we also need to do something similar to provide forwarding group mappings.

dplore:
@rgwilton @hobbydos please provide feedback

Hobbydos:
@dplore someone from my org will provide feedback shortly. Thanks.

nokia1adam:
> @dplore someone from my org will provide feedback shortly. Thanks.

this is a good change. +1 for this PR.

aredmon8551:
> @panduranganrs for viz.
> 
> this seems reasonable to me. we also need to do something similar to provide forwarding group mappings.

Hey Steve,

Maybe I'm missing something/confused - that should be possible with the combination of this PR and `/qos/forwarding-groups/forwarding-group/config/output-queue`, no?

sulrich:
@aredmon8551 - 

> Maybe I'm missing something/confused - that should be possible with the combination of this PR and `/qos/forwarding-groups/forwarding-group/config/output-queue`, no?

sorry, to clarify, some systems have fixed names for the moral equivalent of forwarding-groups.  i believe cisco has a `qos-group`[^1] construct, within arista eos these are traffic-classes [^2]. 

these need to be mapped from the freeform string to their associated platform-specific. class/group IDs.  it's the same thing your queue naming PR addresses, just for forwarding-groups. 

[^1]: [set qos-group](https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5500/qos/b-ncs5500-qos-cli-reference/b-ncs5500-qos-cli-reference_chapter_01.html#wp2627207313)
[^2]: [traffic classes](https://www.arista.com/en/um-eos/eos-quality-of-service#xx1436242)

dplore (approved):
LGTM, last call for comments, this will merge on Feb 16, 2023

m-hesketh:
LGTM - it covers a gap where implementations currently need to infer/derive a queue-id from unrelated fields in the config.

A couple of points that would be good to clarify/check. 

1. The presence of queue-id on the global queue config (rather than the interface input/output profiles) suggest it is intended to be used for both input and output queues. This seems to create some cases where there is redundancy/ambiguity for implementations where there is an expectation that the queue-id is consistent between input-queue and output-queue (which I believe is true for most vendors).
 - Where the input classifier forwards directly to an output queue (via fwd-group), should the queue-id for the input path be inferred from the queue-id for the output queue? (An alternative would be to enforce that there is an input-queue with queue-id present to make the value explicit rather than inferred).
 - If the implementation requires a queue-id for an input queue and there is an input queue present but no queue-id associated with it should: 
   - the queue-id be inferred from the output-queue (via forwarding-group output-queue)
   - the input-queue be rejected (non-backwards compatible).
   - An implementation-dependent default queue-id assumed.
 - If an implementation requires a consistent queue-id value but the input and output queues (linked by forwarding-group output queue) have inconsistent queue-ids, is there an obligation for the implementation to reject the configuration.

2. There is a potential gap where on the input side there is an implicit input queue for a classifier default (i.e. any packets that don't match any of the classifier terms) and on the output side there is a need to perform actions for that default- i.e. the config must configure an output queue-id -  How does the operator know which queue-id that output queue should use (This could be covered by vendor specific documentation, but opens up the possibility of inconsistency between vendors)

If needed, I'm happy to try putting together some configs to give more concrete examples for these scenarios

m-hesketh (approved):

dplore:
> LGTM - it covers a gap where implementations currently need to infer/derive a queue-id from unrelated fields in the config.
> 
> A couple of points that would be good to clarify/check.
> 
> 1. The presence of queue-id on the global queue config (rather than the interface input/output profiles) suggest it is intended to be used for both input and output queues. This seems to create some cases where there is redundancy/ambiguity for implementations where there is an expectation that the queue-id is consistent between input-queue and output-queue (which I believe is true for most vendors).
> 

Yes, in the scenario we are addressing here, there is no need to specify a hardware queue  id (or more abstractly, a hardware traffic class) that is different for input/output.

> 2. There is a potential gap where on the input side there is an implicit input queue for a classifier default (i.e. any packets that don't match any of the classifier terms) and on the output side there is a need to perform actions for that default- i.e. the config must configure an output queue-id -  How does the operator know which queue-id that output queue should use (This could be covered by vendor specific documentation, but opens up the possibility of inconsistency between vendors)
> 
> If needed, I'm happy to try putting together some configs to give more concrete examples for these scenarios


Thanks for the thoughtful comments.  Please consider opening a new pull request to describe the input queue configuration scenarios, especially where there might be a gap.  Referencing 2+ implementations will help  everyone understand a gap in more detail.

PR #774, opened 2022-12-24
Remove the default of "enabled" of BGP AFI-SAFI to support BGP inheritance model

yaonanliang:
Most BGP features can be configured at multiple levels in the BGP configuration level hierarchy, i.e. global, AFI-SAFI, peer-group, neighbor and neighbor/AFI-SAFI. The inheritance model supported by most vendors allows the more specific configuration (e.g. neighbor) to inherit from or override the less specific configuration (e.g. global).

Currently `enabled` leaf of BGP AFI-SAFI is available at global, peer-group and neighbor levels. The inheritance model works as follow: if `enabled` leaf is present at neighbor level, it overrides `enabled` leafs present at peer-group and global level; if `enabled` leaf is not present at neighbor level, the effective `enabled` inherits from `enabled` leaf at peer-group level or global level whichever is present.

The inheritance model requires a `not present` (not set/configured) state of leaf nodes to derive the inheritance relationship, and therefore leaf nodes must not have a default value due to the treatment of default values per [Section 7.6.1 of RFC6020](https://www.rfc-editor.org/rfc/rfc6020#section-7.6.1):
```
The default value of a leaf is the value that the server uses if the leaf does not exist in the data tree.
And..
When the default value is in use, the server MUST operationally behave as if the leaf was present in the data tree with the default value as its value.

```

This change defines the inheritance model for BGP and removes the default (`false`) of `enabled` leaf node to support this model. Similarly the default should be removed for other BGP leafs as well, but they are not included in the scope of this PR. The change is not backward compatible. Users must explicitly initialize `enabled` after the default is removed.

Most platforms support some (if not all) levels of BGP inheritance hierarchy. For example, the most common inheritance relationship is between peer-group and neighbor. A neighbor in a peer-group can either inherit from or override the peer-group configuration. 

Implementation example of this inheritance model:

Arista: [Configuring BGP Neighbors](https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1115325)

### DIFF
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -193,8 +193,11 @@ submodule openconfig-bgp-common-multiprotocol {
       type boolean;
       default false;
       description
-        "This leaf indicates whether the AFI-SAFI is
-        enabled for all neighbors or groups";
+        "This leaf indicates whether the AFI-SAFI is enabled for all
+        neighbors or groups.  For historical reasons AF_IPV4 must be 
+        enabled by default and all other AF are disabled by default.  This
+        supports implementations of BGP RFC4271 which only defined
+        IPv4 and is always enabled when BGP is configured.";
     }
   }
 
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -195,7 +195,6 @@ submodule openconfig-bgp-common-multiprotocol {
       description
         "This leaf indicates whether the AFI-SAFI is enabled for all
         neighbors or groups.";
-        supports implementations of BGP RFC4271 which only defined
         IPv4 and is always enabled when BGP is configured.";
     }
   }
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -195,7 +195,6 @@ submodule openconfig-bgp-common-multiprotocol {
       description
         "This leaf indicates whether the AFI-SAFI is enabled for all
         neighbors or groups.";
-        enabled by default and all other AF are disabled by default.  This
         supports implementations of BGP RFC4271 which only defined
         IPv4 and is always enabled when BGP is configured.";
     }
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -178,9 +178,30 @@ submodule openconfig-bgp-common-multiprotocol {
     }
   }
 
+  grouping bgp-common-mp-global-afi-safi-config {
+    description
+      "Configuration parameters used for all BGP global AFI-SAFIs";
+
+    leaf afi-safi-name {
+      type identityref {
+        base oc-bgp-types:AFI_SAFI_TYPE;
+      }
+      description "AFI,SAFI";
+    }
+
+    leaf enabled {
+      type boolean;
+      default false;
+      description
+        "This leaf indicates whether the AFI-SAFI is
+        enabled for all neighbors or groups";
+    }
+  }
+
   grouping bgp-common-mp-afi-safi-config {
     description
-      "Configuration parameters used for all BGP AFI-SAFIs";
+      "Configuration parameters used for all BGP neighbors
+      or groups AFI-SAFIs";
 
     leaf afi-safi-name {
       type identityref {
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -194,7 +194,7 @@ submodule openconfig-bgp-common-multiprotocol {
       default false;
       description
         "This leaf indicates whether the AFI-SAFI is enabled for all
-        neighbors or groups.  For historical reasons AF_IPV4 must be 
+        neighbors or groups.";
         enabled by default and all other AF are disabled by default.  This
         supports implementations of BGP RFC4271 which only defined
         IPv4 and is always enabled when BGP is configured.";
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -29,7 +29,7 @@ submodule openconfig-bgp-common-multiprotocol {
   revision "2022-12-12" {
     description
       "Removed the default of enabled leaf of AFI-SAFI
-      to support BGP inheritance model.";
+      at neighbor/peer-group to support BGP inheritance model.";
     reference "9.2.0";
   }
 
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -195,7 +195,6 @@ submodule openconfig-bgp-common-multiprotocol {
       description
         "This leaf indicates whether the AFI-SAFI is enabled for all
         neighbors or groups.";
-        IPv4 and is always enabled when BGP is configured.";
     }
   }
 
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -24,7 +24,14 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "9.1.0";
+  oc-ext:openconfig-version "9.2.0";
+
+  revision "2022-12-12" {
+    description
+      "Removed the default of enabled leaf of AFI-SAFI
+      to support BGP inheritance model.";
+    reference "9.2.0";
+  }
 
   revision "2022-05-21" {
     description
@@ -184,7 +191,6 @@ submodule openconfig-bgp-common-multiprotocol {
 
     leaf enabled {
       type boolean;
-      default false;
       description
         "This leaf indicates whether the AFI-SAFI is
         enabled for the neighbor or group";
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -26,7 +26,7 @@ submodule openconfig-bgp-common-structure {
   revision "2022-12-12" {
     description
       "Removed the default of enabled leaf of AFI-SAFI
-      to support BGP inheritance model."
+      to support BGP inheritance model.";
     reference "9.2.0";
   }
 
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -26,7 +26,7 @@ submodule openconfig-bgp-common-structure {
   revision "2022-12-12" {
     description
       "Removed the default of enabled leaf of AFI-SAFI
-      to support BGP inheritance model.";
+      at neighbor/peer-group to support BGP inheritance model.";
     reference "9.2.0";
   }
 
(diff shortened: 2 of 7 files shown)


dplore:
Thanks for this contribution.  Here are two additional references: 

[JunOS Reference](https://www.juniper.net/documentation/en_US/junose15.1/information-products/topic-collections/swconfig-bgp-mpls/index.html?id-16052.html)

[Cisco IOS XR Reference](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r3-9/routing/configuration/guide/xr12krc39_chapter1.html#con_1046664)

dplore:
Please fix the associated lint errors.

dplore:
I wanted to write down my interpretation for future reference as this took some time to derive and the description of the [yang default](https://www.rfc-editor.org/rfc/rfc6020#section-7.5.1) behavior is rather confusing.  

```
The usage of the default value depends on the leaf's closest ancestor 
node in the schema tree that is not a non-presence container:

   o  If no such ancestor exists in the schema tree, the default value
      MUST be used.

   o  Otherwise, if this ancestor is a case node, the default value MUST
      be used if any node from the case exists in the data tree, or if
      the case node is the choice's default case, and no nodes from any
      other case exist in the data tree.

   o  Otherwise, the default value MUST be used if the ancestor node
      exists in the data tree.
```

In the case of the bgp afi-safi enabled leaf, the closest ancestor which is not a non-presence container is the [afi-safis list](https://github.com/openconfig/models/blob/b0e74f26222bc570c9cbd40e0abed83859c4e9cd/yang/bgp/openconfig-bgp-global.yang#L296-L339).  So if I interpret the RFC correctly, the `default false` MUST be applied because any afi-safi entity is part of this list.  

Given this interpretation, I believe this change is needed to achieve the inheritance behavior where the configuration of a parent in the hierarchy is applied to a child unless explicitly set by the child.  For example, if an BGP peer group afi-safi  is enabled then the individual neighbors should also be enabled, without explicit configuration.  With today's `enabled` leaf with a default of `false` , a neighbor will always be disabled unless explicitly enabled, even if the parent peer-group is explicitly `enabled`.  

Added to Jan 17, 2023 OpenConfig Operators agenda

yaonanliang:
Thanks Darren for the additional references. I completely agree with your interpretation. I have fixed the lint errors.

yaonanliang:
> OC Operators reviewed. The group recommends retaining a default false value for the global level and removing the default value for peer group and neighbor. You could add a grouping just for global in order to achieve this.

Thanks Darren. Sorry I didn't make it clear as the existing OC BGP configuration hierarchy only describes the levels within the same VRF (network-instance). Many BGP features (including this in EOS) also maintain an inheritance relationship between default VRF and non-default VRF. A global default value would still forbid the non-default VRF to inherit global "enabled" from that of the default VRF.

dplore:
> Thanks Darren. Sorry I didn't make it clear as the existing OC BGP configuration hierarchy only describes the levels within the same VRF (network-instance). Many BGP features (including this in EOS) also maintain an inheritance relationship between default VRF and non-default VRF. A global default value would still forbid the non-default VRF to inherit global "enabled" from that of the default VRF.

This is a level of inheritance that I hadn't considered.  It seems strange to me to inherit address families between network instances.  Rather it does seem required to configure an address family per network instance.    If the address family isn't configured for a network instance, then the address family isn't enabled. (ie: default = disable)

I don't see reference to inheritance between network-instances in [JunOS SE](https://www.juniper.net/documentation/en_US/junose15.1/topics/reference/command-summary/address-family.html) or in [Cisco XR OS](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r3-9/routing/configuration/guide/xr12krc39_chapter1.html#con_1046664)

So I am not seeing a compelling operational or implementation reason to require inheritance of bgp configuration between network instances.

yaonanliang:
> > Thanks Darren. Sorry I didn't make it clear as the existing OC BGP configuration hierarchy only describes the levels within the same VRF (network-instance). Many BGP features (including this in EOS) also maintain an inheritance relationship between default VRF and non-default VRF. A global default value would still forbid the non-default VRF to inherit global "enabled" from that of the default VRF.
> 
> This is a level of inheritance that I hadn't considered. It seems strange to me to inherit address families between network instances. Rather it does seem required to configure an address family per network instance. If the address family isn't configured for a network instance, then the address family isn't enabled. (ie: default = disable)
> 
> I don't see reference to inheritance between network-instances in [JunOS SE](https://www.juniper.net/documentation/en_US/junose15.1/topics/reference/command-summary/address-family.html) or in [Cisco XR OS](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r3-9/routing/configuration/guide/xr12krc39_chapter1.html#con_1046664)
> 
> So I am not seeing a compelling operational or implementation reason to require inheritance of bgp configuration between network instances.

Fair enough. I will update the PR as recommended.

dplore:
@yaonanliang can you make this update today?

robshakir:
Thanks for the change here -- this looks good to me, essentially I think this fixes a long-standing issue with the inheritance model that wasn't clear before, so thanks for the observation.

One clarification that we should probably make, when the `afi-safis` list is not specified, then this means that the nearest non-presence container is not specified, so the default does not apply. However, the behavioural semantics are that the unspecified set of `afi-safis` means that they are not enabled or disabled, I think we need to clarify that if there are no entries in the `afi-safi` list then all address-families are disabled.

yaonanliang:
> Thanks for the change here -- this looks good to me, essentially I think this fixes a long-standing issue with the inheritance model that wasn't clear before, so thanks for the observation.
> 
> One clarification that we should probably make, when the `afi-safis` list is not specified, then this means that the nearest non-presence container is not specified, so the default does not apply. However, the behavioural semantics are that the unspecified set of `afi-safis` means that they are not enabled or disabled, I think we need to clarify that if there are no entries in the `afi-safi` list then all address-families are disabled.

Thanks Rob. I think the behavior is defined by the RFC if I interpret it right: if afi-safis list is not specified (has no entries), the closest ancestor that is not a non-presence container will go up to /network-instance/protocols/protocl (list). In this case the default value (false) should be used for all afi-safis (the last otherwise clause). The wording of RFC is a little confusing. It seems that the default value MUST be used in most cases except  the "case" node case. 
 
https://www.rfc-editor.org/rfc/rfc6020.html#section-7.6.1

dplore:
Last call for comments, this will merge on Feb 3, 2023.

Note, a follow up PR is planned to clarify the behavior of the BGP AF global default enabled leaf.

dplore (approved):

aredmon8551:
Feedback from a Junos/EVO perspective:
1. The Junos implementation uses positive configuration today, so the most specific instance of configuration has precedence (e.g., a neighbor level configuration overrides a group level configuration).  This aligns to the inheritance model I believe is being specified here.
2. Disabling AFI/SAFIs: Bear in mind that AFI/SAFIs are a MP-BGP concept and were introduced after the initial BGP RFCs to allow support for protocols beyond IPv4 unicast.  In short: You cannot disable IPv4 unicast as it's an inextricable component of the protocol itself.  That's why most implementations support it today.  There is no mechanism for disabling it in Junos.  If the intention of the model is to disable it by default then we'll need to work out how to achieve that through either import/export policies or disabling the protocol entirely (which I assume would conflict with enable/disable leaves).

yaonanliang:
Thanks for the feedback.

> Feedback from a Junos/EVO perspective:
> 
>     1. The Junos implementation uses positive configuration today, so the most specific instance of configuration has precedence (e.g., a neighbor level configuration overrides a group level configuration).  This aligns to the inheritance model I believe is being specified here.
> 
Yes. I believe this inheritance model will work for your model.

>     2. Disabling AFI/SAFIs: Bear in mind that AFI/SAFIs are a MP-BGP concept and were introduced after the initial BGP RFCs to allow support for protocols beyond IPv4 unicast.  In short: You cannot disable IPv4 unicast as it's an inextricable component of the protocol itself.  That's why most implementations support it today as a default.  There is no mechanism for disabling it in Junos.  If the intention of the model is to disable it by default then we'll need to work out how to achieve that through either import/export policies or disabling the protocol entirely (which I assume would conflict with enable/disable leaves).

Arista/EOS has the similar difficulty with the current default for the reason you raised for ipv4-unicast. The global default (esp. for ipv4-unicast) will need further discussion and clarification with the community.

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -185,6 +192,26 @@ submodule openconfig-bgp-common-multiprotocol {
     leaf enabled {
       type boolean;
       default false;
+      description
+        "This leaf indicates whether the AFI-SAFI is
+        enabled for all neighbors or groups";
```
dplore, line 197:
```suggestion
        "This leaf indicates whether the AFI-SAFI is enabled for all
        neighbors or groups.  For historical reasons AF_IPV4 must be 
        enabled by default and all other AF are disabled by default.  This
        supports implementations of BGP RFC4271 which only defined
        IPv4 and is always enabled when BGP is configured.";
```

aredmon8551: LGTM

dplore (changes requested):
@yaonanliang another change requested to clarify the global AF enable defaults.  There is not a reasonable way to indicate this condition in our model using a `when` statement, so I proposed this wording in the description for the global AF enable leaf.

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
robshakir: I don't agree with adding this. Please can we leave this clarification for a subsequent CL.

marcushines: changing defaults to be inconsistent based on out of touch and out of date reasons is counter to the goal of openconfig

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -185,6 +192,29 @@ submodule openconfig-bgp-common-multiprotocol {
     leaf enabled {
       type boolean;
       default false;
+      description
+        "This leaf indicates whether the AFI-SAFI is enabled for all
+        neighbors or groups.  For historical reasons AF_IPV4 must be
```
robshakir, line 197:
```suggestion
        neighbors or groups.";
```

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -185,6 +192,29 @@ submodule openconfig-bgp-common-multiprotocol {
     leaf enabled {
       type boolean;
       default false;
+      description
+        "This leaf indicates whether the AFI-SAFI is enabled for all
+        neighbors or groups.  For historical reasons AF_IPV4 must be 
+        enabled by default and all other AF are disabled by default.  This
```
robshakir, line 198:
```suggestion
```

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -185,6 +192,29 @@ submodule openconfig-bgp-common-multiprotocol {
...
+        supports implementations of BGP RFC4271 which only defined
```
robshakir, line 199:
```suggestion
```

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -185,6 +192,29 @@ submodule openconfig-bgp-common-multiprotocol {
...
+        IPv4 and is always enabled when BGP is configured.";
```
robshakir, line 200:
```suggestion
```

robshakir (changes requested):

nokia1adam:
Nokia SROS/SRL inheritance model is aligned with these changes. We are planning to throw a validation error if, as a result of the total configuration, including inheritance, a session ends up with no enabled AFI/SAFI. It makes no sense to allow this, since it is an undefined behavior with respect to RFC 4271. This will allow the OC default for ipv4-unicast enabled to stay false.

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -24,7 +24,14 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "9.1.0";
+  oc-ext:openconfig-version "9.2.0";
+
+  revision "2022-12-12" {
+    description
+      "Removed the default of enabled leaf of AFI-SAFI
```
marcushines, line 31: this is not true

robshakir (approved):

dplore (approved):

PR #775, opened 2022-12-27
Add enabled-address-families back for backward compatibility. 

vcheny:
Even though the leaf is deprecated some existing implementation still uses it. It is better to keep it in the model for sometime for backward compatibility.

### DIFF
--- a/release/models/network-instance/openconfig-network-instance-l2.yang
+++ b/release/models/network-instance/openconfig-network-instance-l2.yang
@@ -24,7 +24,14 @@ submodule openconfig-network-instance-l2 {
     Layer 2 network instance configuration and operational state
     parameters.";
 
-  oc-ext:openconfig-version "4.0.0";
+  oc-ext:openconfig-version "4.1.0";
+
+  revision "2022-12-22" {
+    description
+      "Add enabled-address-families leaf-list back for backward
+      compatibility";
+    reference "4.1.0";
+  }
 
   revision "2022-12-21" {
     description
--- a/release/models/network-instance/openconfig-network-instance-l3.yang
+++ b/release/models/network-instance/openconfig-network-instance-l3.yang
@@ -152,7 +152,7 @@ module openconfig-network-instance-l3 {
       description
         "The address families that are to be enabled for this
         network instance. This leaf has been deprecated. It is
-        kept in the model for back compatibility and will be
+        kept in the model for backward compatibility and will be
         removed.";
     }
   }
--- a/release/models/network-instance/openconfig-network-instance-l3.yang
+++ b/release/models/network-instance/openconfig-network-instance-l3.yang
@@ -23,7 +23,14 @@ module openconfig-network-instance-l3 {
     Layer 3 network instance configuration and operational state
     parameters.";
 
-  oc-ext:openconfig-version "2.0.0";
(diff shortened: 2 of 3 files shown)


vcheny:
@dplore @robshakir

dplore:
I have added to the Jan 10th openconfig operators agenda.

earies:
These types of edits have been the case since inception of OpenConfig so what is new here?

Nodes are being removed if ambiguous or irrelevant especially in the context of the use-cases that drive OC modeling.  Many items slipped in in the early models that are now being caught during implementation attempts.  This was already discussed in https://github.com/openconfig/public/pull/738 w/ major version increments so your proposal here is to "revert" but increment again.

There has been little to no use of the YANG `status` statement to date in which slow deprecation could be introduced but reintroducing it with description updates in this fashion imo solves nothing.  The server side implementation can choose not to update to this model version.

There are bigger issues and questions at play here wrt. backwards compatibility and this is not specific to this set of model nodes but what is the precise reasoning behind trying to reintroduce these nodes?  Is there an implementation making this mandatory internally?

earies:
I agree there needs to be additional consideration of transition mechanisms, deprecation and upgrades.  This has been a long standing issue w/ this modeling work.  To date, backwards incompatible changes have been introduced with only a signal in the version - there today, gone tomorrow...  The end result means when an implementation supports a future version, it is an abrupt transition for both the network-element and any clients communicating w/ the network-element.  TLDR: This is a global problem w/ iterative modeling work and even more so when models are prematurely published w/o proper review and detail.

But I'm afraid the reasoning for this PR is not justified.  You appear to have created a hard dependency on the main/master branch of these models and are assuming that network-elements support _every_ latest version which is incorrect.  The fact that this "broke the implementation" is not true.  The implementation you are referring to supports a past version of the model is all - not what is published to this branch which is meant to be "top of tree" and looking out for the best interest of the model set on removing undefined behavior.

In addition, there is no mandate by way of the data-model that these nodes are mandatory for operation or management via this model-set so 1 implementations mandate does not correspond to anothers.

With the lofty goal being common modeling across implementations, any loose definition or undefined behaviors are going to create divergence and these either get course corrected or there is going to be pain to live with.

But if there is code generation and dependencies being made (e.g. ONDATRA featureprofiles) on all latest publishes, I'm afraid that this assumption is incorrect as all implementations will be on varying degrees of support of data-models at any given point in time.  The client side really needs to understand this aspect vs. making that assumption.

dplore:
After some offline discussion we have decided to drop this request.  Thanks for the comments.

oscargdd:
Hi 

I disagree on having this non-backwards compatible change fully removed without having a solution to cover the implementations that need to support explicity IPv4 or IPv6, such as IOS XR or Huawei.

dplore:
Some recommended solutions for this issue include:

* The implementation may support the current model with a deviation that adds with the enabled property back in
* The implementation may assert support for a previous version of the model which contains the leaf with the enabled property
* The implementation can add the support for this model in a future release

oscargdd:
Thanks @dplore I guess between supporting a deprecated leaf and a proprietary augmentation there is not much difference… I hope implementations can support not having to explicitly tell you IPv6 or ipv4.

PR #789, opened 2023-01-18
Fix `.spec.yml` where name doesn't match root module name to fix online documentation.

wenovus:
### Summary of Change
This PR fixes up the `.spec.yml` files such that for each path, there is a `.spec.yml` entry whose name matches the root module of the path so that almost all paths now point to valid links.

Example of broken link: all "openconfig-messages" paths.

### Background
The way the path links at
https://openconfig.net/projects/models/paths/index.html works is that the root module of each path is assumed to be the name of the HTML file containing the documentation of the path.

For example, the root module of `/interfaces/interface/state/name` is `openconfig-interfaces`, so the [link](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-interfaces.html#interfaces-interface-state-name) refers to
`openconfig-interfaces.html`.

This means that for this to work, there must be
an entry in *some* `spec.yml` file whose name matches the root module for the path.

For example, the above path points to a valid link because there is a spec file with the following:
```yaml
- name: openconfig-interfaces
  docs:
      - yang/interfaces/openconfig-interfaces.yang
```

This means that if we changed the `name` field in the above YAML to `openconfig-intf`, then the path no longer points to a valid link.

Ideally, for each path we know the YAML entry which compiled it, but this is hard to do since we would have to analyze every YAML spec entry. In general, it is not guaranteed that the file defining the leaf will appear in the relevant spec entry due to module includes and submodules.

### Testing
Done locally for openconfig.net

### DIFF
--- a/release/models/ate/.spec.yml
+++ b/release/models/ate/.spec.yml
@@ -1,8 +1,6 @@
-- name: openconfig-ate
+- name: openconfig-ate-flow
   docs:
     - yang/ate/openconfig-ate-flow.yang
-    - yang/ate/openconfig-ate-intf.yang
   build:
     - yang/ate/openconfig-ate-flow.yang
-    - yang/ate/openconfig-ate-intf.yang
   run-ci: true
--- a/release/models/catalog/.spec.yml
+++ b/release/models/catalog/.spec.yml
@@ -1,4 +1,4 @@
-- name: openconfig-catalog
+- name: openconfig-module-catalog
   build:
     - yang/catalog/openconfig-module-catalog.yang
   docs:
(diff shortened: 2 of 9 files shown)


robshakir:
This LGTM, approving.

One question is how we keep this being true -- is there something that we should add to a check somewhere that says "for each module that defines an element (i.e., doesn't solely augment) then there must be such a rule?

wenovus:
> This LGTM, approving.
> 
> One question is how we keep this being true -- is there something that we should add to a check somewhere that says "for each module that defines an element (i.e., doesn't solely augment) then there must be such a rule?

I'll add a task for this. I want to do some refactoring of the CI code first to make such rules easier to add.

robshakir (approved):
SGTM.

PR #790, opened 2023-01-19
Add remote-ports and deprecate remote-port in /system/logging

LeonGWang:
### Change Scope

* Add `remote-ports` leaf-list and deprecate the `remote-port` leaf in /system/logging.
* This is backwards compatible.

### Reference Implementations

* rsyslog: https://www.rsyslog.com/doc/v8-stable/configuration/modules/imudp.html#port
* syslog-ng: https://www.syslog-ng.com/technical-documents/doc/syslog-ng-open-source-edition/3.37/administration-guide/46#TOPIC-1829084
Note that while syslog-ng's [port] option is a single number, we can still configure multiple ports for a single destination like so:
```
destination d_tcp1 { network("10.1.2.3" port(1999)); };
destination d_tcp2 { network("10.1.2.3" port(2000)); };

log {
     source(s_net_tcp);
     destination (d_tcp1);
     destination (d_tcp2);
};
```

### DIFF
--- a/release/models/system/openconfig-system-logging.yang
+++ b/release/models/system/openconfig-system-logging.yang
@@ -23,7 +23,13 @@ module openconfig-system-logging {
     "This module defines configuration and operational state data
     for common logging facilities on network systems.";
 
-  oc-ext:openconfig-version "0.3.1";
+  oc-ext:openconfig-version "0.3.2";
+
+  revision "2023-01-18" {
+    description
+      "Add remote-ports and deprecate remote-port.";
+    reference "0.3.2";
+  }
 
   revision "2018-11-21" {
     description
@@ -427,10 +433,18 @@ module openconfig-system-logging {
     }
 
     leaf remote-port {
+      status deprecated;
       type oc-inet:port-number;
       default 514;
       description
         "Sets the destination port number for syslog UDP messages to
+        the server.  The default for syslog is 514.";
+    }
+
+    leaf-list remote-ports {
+      type oc-inet:port-number;
+      description
+        "Sets the destination port numbers for syslog UDP messages to
         the server.  The default for syslog is 514.";
     }
   }


dplore:
Does it make sense to have a single server with multiple remote ports?  I think the use case is more likely to be using different ports on different servers.

jsterne:
Are there references to multiple implementations that support this concept of sending a syslog stream of events from a single router using multiple configurable destination UDP ports (as the same dest server context)?

LeonGWang:
@dplore It does make sense to configure multiple ports on a single remote-server. For example, one such use case is illustrated by syslog-ng to improve multi-threading on the destination remote-server (see Fig. 40): https://www.syslog-ng.com/technical-documents/doc/syslog-ng-premium-edition/7.0.10/administration-guide/80

@jsterne I've added implementation reference to rsyslog and syslog-ng.

jsterne:
I'm not sure that syslog-ng approach is multiple ports from the destination context. That's more along the lines of what Darren was saying (different destination contexts, each with its own port).

But in any case, aren't references to router data models more relevant here than rsyslog or syslog-ng?

robshakir:
* Cisco IOS XR [docs](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/system_monitoring/command/reference/b-sysmon-cr-asr9k/b-sysmon-cr-asr9k_chapter_0100.html) - `logging ... port <x>`. This does not seem to support >1 port.
* Arista EOS [docs](https://www.arista.com/assets/data/docs/Manuals/EOS-4.17.0F-Manual.pdf) - `logging [VRF_INSTANCE] host syslog_host [PORT] [PROT_TYPE]`. This doesn't seem to support >1 port.
* Nokia [docs](https://yang.srlinux.dev/v22.3.2?pathType=Config&hidePrefix=false&path=%2Fsrl_nokia-system%3Asystem%2Fsrl_nokia-logging%3Alogging%2Fremote-server%5Bhost%3D*%5D%2Fhost) - SRLinux supports a leaf `remote-port`.
* Juniper [docs](https://www.juniper.net/documentation/us/en/software/junos/junos-overview/topics/ref/statement/port-edit-system.html) - `edit system syslog host FOO port BAR`, supports a single port.

I'm not clear that given that this survey that this should be a `leaf-list`.

robshakir:
Ah, one thing I missed here -- with some of these commands you may be able to specify `logging 192.0.2.1 port 514` and `logging 192.0.2.1 port 500`, which today the model doesn't accept because the key is IP address.

This would be a motivation to move to a `leaf-list` to make this change in a way that doesn't impact how this list is keyed.

LeonGWang:
@robshakir 
Arista doesn't support specifying additional ports to the same remote-host in consecutive CLI commands, i.e.
```
logging 192.0.2.1 514
logging 192.0.2.1 500
```
The second CLI command in this example sequence will override the first CLI command, with the final configuration being the same as just running `logging 192.0.2.1 500` by itself.

However, Arista does support specifying multiple ports to the same remote-host in one single CLI command, i.e.
```
logging host 127.0.0.1 500 501 513 514 515
```
`show logging` will then shows:
```
...
    Logging to '127.0.0.1' port 500 in VRF default via udp
    Logging to '127.0.0.1' port 501 in VRF default via udp
    Logging to '127.0.0.1' port 513 in VRF default via udp
    Logging to '127.0.0.1' port 514 in VRF default via udp
    Logging to '127.0.0.1' port 515 in VRF default via udp
...
```

Unfortunately, I couldn't find any publicly available Arista TOI that illustrates this capability.

robshakir:
@LeonGWang - thanks for the clarification. I'd missed in the docs that this was a possibility, but your confirmation/assertion here is sufficient for one platform to be covered.

I'm confirming whether this is supported elsewhere (or planned to be supported).

robshakir:
Since rsyslog is used under the hood of SONIC, this is also supported on that platform.

aredmon8551:
@robshakir correct, Junos/EVO do not support one-to-many mappings between syslog destinations and destination ports.  Today it's a 1:1.

LimeHat:
> and SRLinux do not support this
Correct, not supported in SRLinux as of today

Similar configuration can be done in SROS with proprietary yang models, but this will be configured/modeled as two different destinations (we use named syslog destinations like in the IETF model, thus the key is independent from dst. IP).

@LeonGWang could you elaborate on the expected syslog client behavior in this case? Is a router supposed to replicate & send each entry multiple times (via each socket)?

LeonGWang:
@LimeHat Yes, if multiple ports are configured on the same `remote-server`, the client should replicate the syslog and send a copy to each configured port.

jsterne:
I thought the multi-port thing (at least in syslog-ng) was a performance optimization (to have multiple threads handing the event stream). But if multiple ports means sending every event to every port, that doesn't seem like a performance improvement.

dplore:
Reviewed in OC operators meeting on Jan 31, 2023.   Will merge on Feb 3, 2023.

LimeHat:
> I thought the multi-port thing (at least in syslog-ng) was a performance optimization (to have multiple threads handing the event stream). But if multiple ports means sending every event to every port, that doesn't seem like a performance improvement.

it is a performance optimization on a server side. And indeed, if we assume that we model a syslog server configuration - this makes a lot of sense.

@LeonGWang is there a good use case for this type of config on a client side?
I'm struggling to see it in this scenario, and just wondering what am I missing. E.g. even if you use some sort of proxy in front of your servers that listens on multiple sockets, it will be easier to do replication of the proxy. 
Otherwise, a modeling with different selectors per IP/port pair seems much more useful for a client config (but this configuration will not achieve that).

It feels more like a quirk that is based on a simple fact that you can do this easily with rsyslog that is used in your backend (just by replicating a filter multiple times in rsyslog.conf).

LeonGWang:
@LimeHat The notable utility here is that only a single destination IP address needs to be exposed/anycasted to allow  syslog collection on multiple destination ports, which different processes, containers, etc may manage.

> Otherwise, a modeling with different selectors per IP/port pair seems much more useful for a client config (but this configuration will not achieve that).

I do agree that a model where selectors are configured per IP/port pair can achieve the same functionality, as this is how `syslog-ng` models outgoing syslogs. However, that would be a major backward-incompatible change to the model which this PR is trying to avoid.

LimeHat:
Thanks, understood. A corner case I'd say (a middle layer that can replicate messages makes more sense to me personally), but still valid.

dplore:
Additional review of the use case for this revealed that we do in fact want to filter by severity/facility for each selected port.  I will submit a new PR to reflect this.

dplore:
This PR is in last-call, to be merged Feb 3, 2023.  There is a follow up PR which is compatible with this PR and can be merged at a later date.

marcushines (approved):

dplore:
Based on comments and further review, it's decided to close (obsolete) this PR in favor of #799 which also includes filtering.

PR #791, opened 2023-01-19
Add a clarification and examples for port breakout mode.

robshakir:
```
 * (M) release/models/platform/openconfig-platform-port.yang
  - Add clarification that a physical channel is defined to be an
    electrical channel.
  - Add examples for the existing cases described in the model, as
    we as a simple 400G->100G breakout case.
```

### Change Scope

 * Clarification of the existing port breakout model in order to aid implementation understanding.
 * This change moves the revision to 1.0.0 based on the fact that this model is implemented.

### Platform Implementations

 * No new modelling that requires vendor feature mapping.

### DIFF
--- a/release/models/platform/openconfig-platform-port.yang
+++ b/release/models/platform/openconfig-platform-port.yang
@@ -24,7 +24,14 @@ module openconfig-platform-port {
     "This module defines data related to PORT components in the
     openconfig-platform model";
 
-  oc-ext:openconfig-version "0.4.2";
+  oc-ext:openconfig-version "1.0.0";
+
+  revision "2023-01-19" {
+    description
+      "Add clarification of the definition of a physical channel, and
+      example configurations.";
+    reference "1.0.0";
+  }
 
   revision "2021-10-01" {
     description
@@ -129,7 +136,10 @@ module openconfig-platform-port {
         to the interfaces in this breakout group. This leaf need
         not be set if there is only one breakout group where all
         the interfaces are of equal speed and have equal number
-        of physical channels";
+        of physical channels.
+
+        The physical channels referred to by this leaf electrical channels
+        towards the transceiver.";
     }
   }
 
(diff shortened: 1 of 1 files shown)


marcushines (approved):

oscargdd:
Hi @robshakir Wouldn't it be better to include the examples in a separated MD file?  Or, complementary to this one, open a specific document on Breakout port modeling.
Also, I see that the PR is labeled with breaking changes.. I guess it is not the case as only description is changed.

release/models/platform/openconfig-platform-port.yang
```diff
@@ -129,7 +136,10 @@ module openconfig-platform-port {
         to the interfaces in this breakout group. This leaf need
         not be set if there is only one breakout group where all
         the interfaces are of equal speed and have equal number
-        of physical channels";
+        of physical channels.
+
+        The physical channels referred to by this leaf electrical channels
+        towards the transceiver.";
```
jsterne, line 142: minor editorial but should it say "...by this leaf are electrical..."?

robshakir:
@oscargdd -- The breaking change here is that I cut the 1.0.0 version of the model. It seems like this is overdue. WDYT?

I'm fine to put these examples in another MD file and then link it here, I just thought that they were compact enough to include here. Again - happy to take feedback.

release/models/platform/openconfig-platform-port.yang
robshakir: Fixed, thanks.

dplore:
> @oscargdd -- The breaking change here is that I cut the 1.0.0 version of the model. It seems like this is overdue. WDYT?
> 
> I'm fine to put these examples in another MD file and then link it here, I just thought that they were compact enough to include here. Again - happy to take feedback.

I am ok either way regarding documentation.  

Should moving from 0.x.x to 1.x.x be considered a breaking change?

dplore (approved):
Presented at Jan 24 OpenConfig operators meeting without objection.  Last call for comments, this will merge on Feb 7th 2023

Issue #97, opened 2017-10-05
openconfig-isis: redundant "enabled" leaf under ISIS levels?


kavitharp:
Hello, 

What is the use for the enabled leaf under protocols/protocol/isis/levels/level/config/enabled.
There is a leaf called level-capability under : protocols/protocol/isis/global/config/. Level-capability says what level/s the given ISIS instance should operate at. I think the enabled flag under the levels container and the level-capability maybe stepping on each others toes here, IIUC.
Could you please clarify what is the purpose of level-capability and enabled leaf under level/config for ISIS?

Thanks, 
Kavitha

robshakir:
Thanks -- I agree that this can be determined from the `level-capability` leaf. I'll remove this leaf in the next revision of the model and make the clarification that the `level-capability` should be used to determine whether a particular level is enabled.

kavitharp:
Thanks!

xw-g:
@robshakir do you still plan to do it or anything changed after last comment? cc: @dplore 

cc: @aredmon8551,  @nokia1adam, @rolandphung,  @aaronmillisor

robshakir:
Created #794 to implement this.

PR #731, opened 2022-10-26
modify /interfaces/interface/loopback-mode to allow specification of loopback-mode-type

sulrich:
### Change Scope

* this changes the noted leaf from a `boolean` to `loopback-mode-type`.  
* this is a backward **incompatible** change. 

the leaf, as currently defined (boolean), indicates whether the interface is looped onto itself. more broadly, interfaces can be looped in either direction.  the `loopback-mode` should allow for the specification of the direction of the loopback as defined in the optical-transport models.

### pyang output
```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw name             -> ../config/name
        +--rw config
        |  +--rw name?            string
        |  +--rw type             identityref
        |  +--rw mtu?             uint16
        |  +--rw loopback-mode?   oc-opt-types:loopback-mode-type << leaf of interest
        |  +--rw description?     string
        |  +--rw enabled?         boolean
        +--ro state
        |  +--ro name?            string
        |  +--ro type             identityref
        |  +--ro mtu?             uint16
        |  +--ro loopback-mode?   oc-opt-types:loopback-mode-type << leaf of interest
        |  +--ro description?     string
        |  +--ro enabled?         boolean
        |  +--ro ifindex?         uint32
        |  +--ro admin-status     enumeration
        |  +--ro oper-status      enumeration
        |  +--ro last-change?     oc-types:timeticks64
        |  +--ro logical?         boolean
        |  +--ro management?      boolean
        |  +--ro cpu?             boolean
        |  +--ro counters

 { ... snipped ... }

        +--rw hold-time
        |  +--rw config
        |  |  +--rw up?     uint32
        |  |  +--rw down?   uint32
        |  +--ro state
        |     +--ro up?     uint32
        |     +--ro down?   uint32
        +--rw subinterfaces

{ ... snipped ... }
```

PR #794, opened 2023-01-25
Deprecate global per-level 'enabled' leaf for IS-IS.

robshakir:
```
   - Remove the per-level IS-IS enabled leaf within the global
     configuration, since it is duplicated by the level-capability
     leaf. See #97.
 * (M) release/models/isis/openconfig-isis-{routing,lsp}.yang
   - Bump model version to match master module.
```

### Change Scope

* Remove the `enabled` leaf per-level within global IS-IS configuration. Enabling levels is supported across implementations, but this leaf duplicated the `level-capability` leaf.
* See #97.

### DIFF
--- a/release/models/isis/openconfig-isis.yang
+++ b/release/models/isis/openconfig-isis.yang
@@ -54,13 +54,13 @@ module openconfig-isis {
             +-> { levels config }
             +-> { level adjacencies }";
 
-  oc-ext:openconfig-version "2.0.0";
+  oc-ext:openconfig-version "1.2.0";
 
   revision "2023-01-25" {
     description
       "Per-level global enabled configuration removed, since it duplicates
       the level-capability leaf.";
-    reference "2.0.0";
+    reference "1.2.0";
   }
 
   revision "2022-09-20" {
--- a/release/models/isis/openconfig-isis.yang
+++ b/release/models/isis/openconfig-isis.yang
@@ -54,9 +54,16 @@ module openconfig-isis {
             +-> { levels config }
             +-> { level adjacencies }";
 
-  oc-ext:openconfig-version "1.1.0";
+  oc-ext:openconfig-version "2.0.0";
 
-    revision "2022-09-20" {
+  revision "2023-01-25" {
+    description
+      "Per-level global enabled configuration removed, since it duplicates
+      the level-capability leaf.";
+    reference "2.0.0";
+  }
+
+  revision "2022-09-20" {
     description
       "Add CSNP enable to IS-IS global configuration.";
     reference "1.1.0";
@@ -1328,7 +1335,6 @@ module openconfig-isis {
       description
         "This container defines ISIS level based configuration.";
 
-      uses admin-config;
       uses isis-base-level-config;
       uses isis-metric-style-config;
       uses isis-authentication-check-config;
--- a/release/models/isis/openconfig-isis.yang
+++ b/release/models/isis/openconfig-isis.yang
@@ -69,13 +69,6 @@ module openconfig-isis {
     reference "1.2.0";
   }
 
-  revision "2023-01-25" {
-    description
-      "Per-level global enabled configuration removed, since it duplicates
-      the level-capability leaf.";
-    reference "1.2.0";
-  }
-
   revision "2022-09-20" {
     description
       "Add CSNP enable to IS-IS global configuration.";
--- a/release/models/isis/openconfig-isis.yang
+++ b/release/models/isis/openconfig-isis.yang
@@ -299,7 +299,7 @@ module openconfig-isis {
   grouping admin-config-deprecated {
     description
       "Re-usable grouping for the enabled leaf in contexts where it is deprecated.";
-    
+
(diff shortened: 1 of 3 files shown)


dplore:
should we [deprecate](https://www.rfc-editor.org/rfc/rfc6020#section-7.19.2) vs remove?

robshakir:
Deprecating in this case is rather ugly, since the leaf is not actually individually defined. Is the concern here that removing this leaf needs to wait for a major release? i.e., can you explain what the benefits of your suggestion are :-)

dplore:
> Deprecating in this case is rather ugly, since the leaf is not actually individually defined. Is the concern here that removing this leaf needs to wait for a major release? i.e., can you explain what the benefits of your suggestion are :-)

I am attempting to exercise bias for backwards compatibility.  If this PR can wait until the next major release (date TBD) then perhaps no action is needed.

If the change is needed to happen much sooner and be available in the next minor release (date also TBD, but sooner than next major release) then the enabled leaf in the grouping could be left alone and a new grouping like `admin-deprecated` could be added with a deprecated enable leaf.  This would keep the tree backwards compatible and be clean?  Just a thought.

robshakir:
thank you for the endorsement @OpenConfigBot :-)

robshakir:
@dplore - this change is now backwards compatible, using the approach you mention.

release/models/isis/openconfig-isis.yang
```diff
@@ -54,9 +54,16 @@ module openconfig-isis {
             +-> { levels config }
             +-> { level adjacencies }";
 
-  oc-ext:openconfig-version "1.1.0";
+  oc-ext:openconfig-version "2.0.0";
```
dplore, line 57:
```suggestion
  oc-ext:openconfig-version "1.2.0";
```

release/models/isis/openconfig-isis.yang
```diff
@@ -54,9 +54,16 @@ module openconfig-isis {
...
 
-    revision "2022-09-20" {
+  revision "2023-01-25" {
+    description
+      "Per-level global enabled configuration removed, since it duplicates
+      the level-capability leaf.";
+    reference "2.0.0";
```
dplore, line 63:
```suggestion
    reference "1.2.0";
```

dplore (commented):
I think the version should show a minor increment?  @wenovus can you help with the failing CI checks?

release/models/isis/openconfig-isis.yang
```diff
@@ -289,6 +296,19 @@ module openconfig-isis {
...
+  grouping admin-config-deprecated {
+    description
+      "Re-usable grouping for the enabled leaf in contexts where it is deprecated.";
+
+    leaf enabled {
+      type boolean;
+      status deprecated;
+      description
+        "Formerly this leaf was used to enable or disable the functionality
+        within the context specified. This leaf is DEPRECATED."
```
wenovus, line 308: Missing `;` at Line 308

wenovus (changes requested):

release/models/isis/openconfig-isis.yang
robshakir: Done, thanks.

release/models/isis/openconfig-isis.yang
robshakir: Done, thanks.

dplore:
This will be reviewed at March 1st OC Operators meeting

rgwilton:
No objections.

dplore (approved):

dplore:
This will merge on March 2, 2023

release/models/isis/openconfig-isis.yang
```diff
@@ -295,6 +302,19 @@ module openconfig-isis {
...
+    description
+      "Re-usable grouping for the enabled leaf in contexts where it is deprecated.";
+
+    leaf enabled {
+      type boolean;
+      status deprecated;
+      description
+        "Formerly this leaf was used to enable or disable the functionality
+        within the context specified. This leaf is DEPRECATED.";
+    }
```
wenovus, line 315: Was the removal of `default false` on purpose here? Would it make this PR backwards-incompatible since the default statement changed?

@dplore There has already been other backwards-incompatible changes since last release (e.g. https://github.com/openconfig/public/pull/731), so perhaps the comment https://github.com/openconfig/public/pull/794#issuecomment-1420228979 no longer applies? In that case, should we simply remove this grouping altogether as was originally proposed in the PR? Alternatively I think we should add `default false` to match the previous leaf's definition to keep backwards compatibility.

dplore: Regarding breaking changes: it's true that there are other breaking changes since release 1.0 which have been merged.  I would like to get into the habit of making non-breaking changes when possible.  Although we've had breaking changes merged which requires the next release to be a major release, I would still like to minimize the quantity of breaking changes.    

Unless there is some justification for why this should be a breaking change, I'm biased to make it non-breaking (by submitting a new PR to add `default false`.

PR #798, opened 2023-01-30
[Forked from #609] Add ICMP type/code and packet length matches.

robshakir:
- Add ICMPv[46] type and code defininitions, update ICMPv[46] match.
 - See #609 for the background of this PR.

### Change Scope

* Add ICMPv[46] matches to ACL model.
* Add packet length matches to ACL model.

#### ICMPv[46]

 * IOS: [doc](https://www.cisco.com/c/en/us/support/docs/security/ios-firewall/23602-confaccesslists.html) - `deny icmp any any <type> <code>`
 * IOS XR: [doc](https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5000/ip-addresses/64x/b-ip-addresses-configuration-guide-ncs5000-64x/m-implementing-access-lists-prefix-lists-ncs5000.html) - `30 icmp permit any any <type> <code>`.
 * Arista EOS: [doc](https://www.arista.com/en/um-eos/eos-acls-and-route-maps) - `deny icmp any any <type> <code>`
 * JUNOS: [doc](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/firewall-filter-match-conditions-for-ipv4-traffic.html) - see `icmp-type` and `icmp-code`.
 * Nokia SROS: [doc](https://documentation.nokia.com/html/0_add-h-f/93-0073-10-01/7750_SR_OS_Router_Configuration_Guide/Filter-CLI.html) - see `icmp-type` and `icmp-code`.
 * Huawei VRP: [Doc](https://support.huawei.com/enterprise/en/doc/EDOC1100069339/79d97573/configuring-and-applying-an-advanced-acl6) - `rule 10 permit icmpv6 any any icmp6-type <type> | <type>  <code> `

#### Packet Length

 * JUNOS: [doc](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/firewall-filter-match-conditions-for-ipv4-traffic.html) see `packet-length`.
 * Cisco IOS XR: [doc](https://xrdocs.io/ncs5500/tutorials/acl-packet-length-matching-ncs55xx-and-ncs5xx/) - see `permit ipv4 any any packet-length XXX`.


release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    base TYPE;
+    description
+      "ICMP Node Information Query.";
+    reference
+      "RFC4620: IPv6 Node Information Queries";
+  }
+  identity NDDE_INFORMATION_RESPONSE {
+    base TYPE;
+    description
+      "ICMP Node Information Response.";
```
oscargdd, line 174:
```suggestion
      "Type 140: ICMP Node Information Response.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    description
+      "Home Agent Address Dsicovery Request Message.";
+    reference "RFC6275: Mobility Support in IPv6";
+
+  }
+
+  identity HOME_AGENT_ADDRESS_DISCOVERY_REPLY {
+    base TYPE;
+    description
+      "Home Agent Address Discovery Reply Message.";
```
oscargdd, line 214:
```suggestion
      "Type 145: Home Agent Address Discovery Reply Message.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    description
+      "Fast handover mode for IPv6.";
+    reference
+      "RFC5568: Mobile IPv6 Fast Handovers";
+  }
+
+  identity RPL_CONTROL {
+    base TYPE;
+    description
+      "RPL Control Message.";
```
oscargdd, line 279:
```suggestion
      "Type 155: RPL Control Message.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference
+      "RFC6550: RPL: IPv6 Routing Protocol for Low-Power and Lossy Networks";
+  }
+
+  identity ILNPV4 {
+    base TYPE;
+    description
+      "ILNPv6 Locator Update Message.";
```
oscargdd, line 287:
```suggestion
      "Type 156: ILNPv6 Locator Update Message.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+      "Duplicate address request.";
+    reference
+      "RFC6775: Neighbor Discovery Optimization for IPv6 over Low-Power Wireless
+                Personal Area Networks (6LoWPANs)";
+  }
+
+  identity DUPLICATE_ADDRESS_CONFIRMATION {
+    base TYPE;
+    description
+      "Duplicate address confirmation.";
```
oscargdd, line 305:
```suggestion
      "Type 158: Duplicate address confirmation.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    description
+      "MPL Control Message.";
+    reference
+      "RFC7731: Multicast Protocol for Low-Power and Lossy Networks (MPL)";
+  }
+
+  identity EXT_ECHO_REQUEST {
+    base TYPE;
+    description
+      "Extended echo request.";
```
oscargdd, line 322:
```suggestion
      "Type 160: Extended echo request.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference
+      "RFC8335: PROBE: A Utility for Probing Interfaces";
+  }
+
+  identity EXT_ECHO_REPLY {
+    base TYPE;
+    description
+      "Extended echo reply.";
```
oscargdd, line 330:
```suggestion
      "Type 161: Extended echo reply.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+
+  identity DST_UNREACHABLE_CODE {
+    base CODE;
+    description
+      "ICMPv6 destination unreachable subcodes.";
+  }
+  identity DST_UNREACHABLE_NO_ROUTE_TO_DST {
+    base DST_UNREACHABLE_CODE;
+    description
+      "No route to destination.";
```
oscargdd, line 343:
```suggestion
      "Code 0: No route to destination.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference "RFC4443: Internet Control Message Protocol (ICMPv6)
+          for the Internet Protocol Version 6 (IPv6) Specification";
+  }
+  identity DST_UNREACHABLE_DST_ADMIN_PROHIBITED {
+    base DST_UNREACHABLE_CODE;
+    description
+      "Communication with destination adminstratively prohibited.";
```
oscargdd, line 350:
```suggestion
      "Code 1: Communication with destination adminstratively prohibited.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference "RFC4443: Internet Control Message Protocol (ICMPv6)
+              for the Internet Protocol Version 6 (IPv6) Specification";
+  }
+
+  identity DST_UNREACHABLE_BEYOND_SCOPE_OF_SRC {
+    base DST_UNREACHABLE_CODE;
+    description
+      "Beyond scope of source address.";
```
oscargdd, line 358:
```suggestion
      "Code 2: Beyond scope of source address.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference "RFC4443: Internet Control Message Protocol (ICMPv6)
+              for the Internet Protocol Version 6 (IPv6) Specification";
+  }
+
+  identity DST_UNREACHABLE_ADDR {
+    base DST_UNREACHABLE_CODE;
+    description
+      "Address unreachable.";
```
oscargdd, line 366:
```suggestion
      "Code 3: Address unreachable.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference "RFC4443: Internet Control Message Protocol (ICMPv6)
+              for the Internet Protocol Version 6 (IPv6) Specification";
+  }
+
+  identity DST_UNREACHABLE_PORT {
+    base DST_UNREACHABLE_CODE;
+    description
+      "Port unreachable.";
```
oscargdd, line 374:
```suggestion
      "Code 4: Port unreachable.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference "RFC4443: Internet Control Message Protocol (ICMPv6)
+              for the Internet Protocol Version 6 (IPv6) Specification";
+  }
+
+  identity DST_UNREACHABLE_SRC_ADDR_FAILED_POLICY {
+    base DST_UNREACHABLE_CODE;
+    description
+      "Source address failed ingress/egress policy.";
```
oscargdd, line 382:
```suggestion
      "Code 5: Source address failed ingress/egress policy.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference "RFC4443: Internet Control Message Protocol (ICMPv6)
+              for the Internet Protocol Version 6 (IPv6) Specification";
+  }
+
+  identity DST_UNREACHABLE_REJECT_ROUTE_TO_DST {
+    base DST_UNREACHABLE_CODE;
+    description
+      "Reject route to destination.";
```
oscargdd, line 390:
```suggestion
      "Code 6: Reject route to destination.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference "RFC4443: Internet Control Message Protocol (ICMPv6)
+              for the Internet Protocol Version 6 (IPv6) Specification";
+  }
+
+  identity DST_UNREACHABLE_ERR_IN_SRC_ROUTING_HDR {
+    base DST_UNREACHABLE_CODE;
+    description
+      "Error in Source Routing Header.";
```
oscargdd, line 398:
```suggestion
      "Code 7: Error in Source Routing Header.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference "RFC8554: An IPv6 Routing Header for Source Routes with
+              the Routing Protocol for Low-Power and Lossy Networks (RPL)";
+  }
+
+  identity DST_UNREACHABLE_HDRS_TOO_LONG {
+    base DST_UNREACHABLE_CODE;
+    description
+      "Headers too long";
```
oscargdd, line 406:
```suggestion
      "Type 8: Headers too long";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+  identity TIME_EXCEEDED_CODE {
+    base CODE;
+    description
+      "Subcodes for the Time Exceeded ICMPv6 type.";
+  }
+
+  identity TIME_EXCEEDED_HOP_LIMIT {
+    base TIME_EXCEEDED_CODE;
+    description
+      "Hop limit exceeded in transit.";
```
oscargdd, line 434:
```suggestion
      "Code 0: Hop limit exceeded in transit.";
```

release/models/acl/openconfig-icmpv6-types.yang
```diff
@@ -0,0 +1,1004 @@
...
+    reference "RFC4443: Internet Control Message Protocol (ICMPv6)
+              for the Internet Protocol Version 6 (IPv6) Specification";
+  }
+  identity TIME_EXCEEDED_FRAGMENT_REASSEMBLY {
+    base TIME_EXCEEDED_CODE;
+    description
+      "Fragment reassembly time exceeded.";
```
oscargdd, line 441:
```suggestion
      "Code 1: Fragment reassembly time exceeded.";
```

release/models/acl/openconfig-icmpv4-types.yang
```diff
@@ -0,0 +1,530 @@
...
+    reference "RFC792: INTERNET CONTROL MESSAGE PROTOCOL";
+  }
+
+  identity TIMESTAMP_REPLY {
+    description
+      "ICMP timestamp reply, value 14.";
+    base TYPE;
+    reference "RFC792: INTERNET CONTROL MESSAGE PROTOCOL";
+  }
+
```
oscargdd, line 107:
```suggestion
  identity TRACEROUTE{
    description
      "Traceroute (deprecated), value 30.";
    base TYPE;
    reference "RFC1393: Traceroute Using an IP Option";
  }
```

oscargdd, line 107: I have added TRACEROUTE, even if it is deprecated, it seems it is still used a bit. I highlighted the deprecated in the description.

oscargdd (changes requested):
I have reviewed the ICMPv4 and ICMPv6 types. I have fixed some typos and a couple of descriptions swaped. I submitted the fixes as suggestions. I also  added the type and code numbers in the description, I think they are helpful.

oscargdd:
@robshakir For the documentation you can also add Huawei:
Huawei VRP: [Doc](https://support.huawei.com/enterprise/en/doc/EDOC1100069339/79d97573/configuring-and-applying-an-advanced-acl6) - `rule 10 permit icmpv6 any any icmp6-type <type> | <type>  <code> `

release/models/acl/openconfig-icmpv4-types.yang
robshakir: Thanks, accepted the suggestion :-)

release/models/acl/openconfig-icmpv4-types.yang
```diff
@@ -0,0 +1,530 @@
...
+  identity DST_UNREACHABLE_HOST_PRECEDENCE_VIOLATION {
+    description
+      "ICMPv4 destination host unreachable.";
+    base DST_UNREACHABLE_CODE;
+  }
+
+  identity DST_UNREACHABLE_PRECEDENCE_CUTOFF {
+    description
+      "ICMPv4 destination is unreachable as the first-hop router has determined
+      that the destination cannot be reached for the specified source/
```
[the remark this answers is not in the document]
robshakir: Thanks. Accepted suggestion.

robshakir (commented):
Thanks for the detailed review Oscar!

release/models/acl/openconfig-packet-match.yang
```diff
@@ -500,4 +524,110 @@ module openconfig-packet-match {
...
+    }
+  }
+
+  grouping ip-icmpv4-header-fields-config {
+    description
+      "Collection of ICMPv4 header fields that can be
+       used to set up a match filter.";
+
+    leaf type {
+      type identityref {
```
oscargdd, line 583: Would it be better to make it an union as in the case of the protocol match? either a number or the identify?

release/models/acl/openconfig-packet-match.yang
```diff
@@ -500,4 +524,110 @@ module openconfig-packet-match {
...
+      }
+      description
+        "ICMPv4 type to be matched.";
+      reference
+        "RFC 792: Internet Control Message Protocol";
+    }
+
+    leaf code {
+      type identityref {
+        base oc-icmpv4-types:CODE;
```
oscargdd, line 594: Same comment as with the type.. Code could be an union of the identity with the number.

release/models/acl/openconfig-packet-match.yang
robshakir: I pondered doing so, the main reason I decided against this was that this just gave two ways to express the same thing. The only benefit of doing so is that you can express something that we didn't add as an identity here, but given that we've been somewhat exhaustive, it's not clear to me that we need both.

Let's discuss tomorrow.

oscargdd: I get your point. I was reviewing our ACLs requirements and found gaps in the protocol match identities, but as a number is also allowed, there was no urgent issue… So it immediately reminded me of the icmp type/ code matches…

I doubt that there is going to appear in the near future a value that is not in the exhaustive list and someone needs to match… 

On the other hand, the identities need maintenance sooner or later (in IETF there is an ongoing debate on IANA maintained modules that keep track of identities in parallel to number assignments)

By allowing the number too… you can take out the “urgency” to assign new identities to new type/codes when they appear.

robshakir: Agreed -- but equally, I think these things are relatively slow to be added -- an implementation can always define an identity with the same `base` to be able to extend these to different values. My preference is to keep this as a simple identity.

dplore: I'm ok with identity only.

release/models/acl/openconfig-packet-match.yang
dplore: For now let's stick with identity only as we have extensive identities defined as discussed above.

dplore:
Last call for comments.  This will merge on Mar 14.

rszarecki:
The ICMP is L4 transport protocol.
This Leafs shluld be under ../transport branch and without "icmp" container.
Similar to ports for TCP/UDP

robshakir:
Ahh, the age old discussion of what layer ICMP is at.  My understanding here is that since ICMP is considered to be a "key" part of the Internet suite, then its considered at L3. This is the consensus that [wikipedia](https://en.wikipedia.org/wiki/Internet_Control_Message_Protocol) has too.

rszarecki:
This PR adds support for ICMP_TUPE and ICMP_CODE. This are fields of ICMP header which is L4 protocol. It is illogical to define ICMP under IPv4 and IPv6 branches, while other IP protocols (L4) protocols header field are under /transport.

IMHO, it shall be under /acl/acl-sets/acl-set/acl-entries/acl-entry/transport/config/{icmp-type, icmp-type}
I also think this shall be lists of enums not just INT16. It is quite common to match on "any of code in set: (foo, bar)"

dplore:
Thank you for the comments @rszarecki .    The open comments regarding L3 or L4 placement while perhaps poke at some lack of consensus in the community, they do not materially affect the functionality of this match criteria.  In the interest of obtaining OpenConfig coverage for these commonly used ACL match criteria I am going to go ahead and merge this PR.

dplore (approved):

PR #799, opened 2023-02-02
Add port list and selector-set to system logging

dplore:
### Change Scope
The objective of this change is to allow two features:
* definition of more than one port per logging server 
* The ability to select the logging severities and facilities mapped to each port.
This allows an operator to send different logging messages to different ports.  For example, to send debug messages to one port and errors to another port.  This is a practical aid in filtering and scaling the large quantities logs on remote logging servers.

### Change summary
* builds off of #790 by adding 
* A port-selector list which associates a selector-set with a reference to remote-ports leaflist
* The selector-set contains leaf lists for logging severity and facility

###Tree output
<pre>
module: openconfig-system
  +--rw system
     +--rw logging
     |  +--rw remote-servers
     |  |  +--rw remote-server* [host]
     |  |     +--rw host         -> ../config/host
     |  |     +--rw config
     |  |     |  +--rw host?               oc-inet:host
     |  |     |  +--rw source-address?     oc-inet:ip-address
     |  |     |  +--rw network-instance?   oc-ni:network-instance-ref
<b>     |  |     |  x--rw remote-port?        oc-inet:port-number
     |  |     |  +--rw remote-ports*        oc-inet:port-number
     |  |     |  +--rw port-selector* [remote-port selector-set-name]
     |  |     |     +--rw remote-port    -> /system/logging/remote-servers/remote-server/config/remote-ports
     |  |     |     +--rw selector-set-name    -> /system/logging/selector-sets/selector-set/config/name </b>
     |  |     +--ro state
     |  |     |  +--ro host?               oc-inet:host
     |  |     |  +--ro source-address?     oc-inet:ip-address
     |  |     |  +--ro network-instance?   oc-ni:network-instance-ref
<b>     |  |     |  x--ro remote-port?        oc-inet:port-number   <i><--- deprecated</i>
     |  |     |  +--ro port-selector* [remote-port selector-set-name]
     |  |     |     +--ro remote-port     -> /system/logging/remote-servers/remote-server/config/remote-ports
     |  |     |     +--ro selector-set-name    -> /system/logging/selector-sets/selector-set/config/name </b>
     |  |     +--rw selectors
     |  |        +--rw selector* [facility severity]
     |  |           +--rw facility    -> ../config/facility
     |  |           +--rw severity    -> ../config/severity
     |  |           +--rw config
     |  |           |  +--rw name?       string
     |  |           |  +--rw facility*   identityref
     |  |           |  +--rw severity*   syslog-severity
     |  |           +--ro state
     |  |              +--ro name?       string
     |  |              +--ro facility*   identityref
     |  |              +--ro severity*   syslog-severity
<b>     |  +--rw selector-sets
     |     +--rw selector-set* [name]
     |        +--rw name      -> ../config/name
     |        +--rw config
     |        |  +--rw name?       string
     |        |  +--rw facility*   identityref
     |        |  +--rw severity*   syslog-severity
     |        +--ro state
     |           +--ro name?       string
     |           +--ro facility*   identityref
     |           +--ro severity*   syslog-severity </b>

</pre>

### Reference Implementations

* Cisco IOS XR [docs](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/system_monitoring/command/reference/b-sysmon-cr-asr9k/b-sysmon-cr-asr9k_chapter_0100.html) - `logging ... port <x>`. 

* Arista EOS [docs](https://www.arista.com/assets/data/docs/Manuals/EOS-4.17.0F-Manual.pdf) - `logging [VRF_INSTANCE] host syslog_host [PORT] [PROT_TYPE]`. 
Arista does support specifying multiple ports to the same remote-host in one single CLI command, i.e.
```
logging host 127.0.0.1 500 501 513 514 515
```
`show logging` will then shows:
```
...
    Logging to '127.0.0.1' port 500 in VRF default via udp
    Logging to '127.0.0.1' port 501 in VRF default via udp
    Logging to '127.0.0.1' port 513 in VRF default via udp
    Logging to '127.0.0.1' port 514 in VRF default via udp
    Logging to '127.0.0.1' port 515 in VRF default via udp
...
```

* Nokia [docs](https://yang.srlinux.dev/v22.3.2?pathType=Config&hidePrefix=false&path=%2Fsrl_nokia-system%3Asystem%2Fsrl_nokia-logging%3Alogging%2Fremote-server%5Bhost%3D*%5D%2Fhost) - SRLinux supports a leaf `remote-port`

* Juniper [docs](https://www.juniper.net/documentation/us/en/software/junos/junos-overview/topics/ref/statement/port-edit-system.html) - `edit system syslog host FOO port BAR`

### DIFF
--- a/release/models/system/openconfig-system-logging.yang
+++ b/release/models/system/openconfig-system-logging.yang
@@ -306,7 +306,7 @@ module openconfig-system-logging {
   // grouping statements
 
   grouping logging-selector-set-state {
-    description 
+    description
       "Operational state data for logging-selector-set-state";
   }
 
@@ -515,7 +515,7 @@ module openconfig-system-logging {
 
     list port-selector {
       key "remote-port selector-set-name";
-      description 
+      description
         "This list defines ports associated with selector-sets.  If no
         port-selector is configured, then all logs are sent.";
 
@@ -523,7 +523,7 @@ module openconfig-system-logging {
         type leafref {
           path "/system/logging/remote-servers/remote-server/config/remote-ports";
         }
-        description 
+        description
           "Associate a selector-set to a remote-port.";
       }
 
@@ -531,7 +531,7 @@ module openconfig-system-logging {
         type leafref {
           path "/system/logging/selector-sets/selector-set/config/name";
         }
-        description 
+        description
           "Name of the selector set used to filter logs sent to this
           port";
       }
--- a/release/models/system/openconfig-system-logging.yang
+++ b/release/models/system/openconfig-system-logging.yang
@@ -506,16 +506,25 @@ module openconfig-system-logging {
         port-selector list";
     }
 
-    list port-selector {
-      key "destination-port selector-set-name";
-      description 
-        "A list of ports associated selector.";
+    leaf-list remote-ports {
+      type oc-inet:port-number;
+      description
+        "Sets the destination port numbers for syslog UDP messages to
+        the server.  The default for syslog is 514.";
+    }
 
-      leaf destination-port {
-        type oc-inet:port-number;
+    list port-selector {
+      key "remote-port selector-set-name";
+      description 
+        "This list defines ports associated with selector-sets.  If no
+        port-selector is configured, then all logs are sent.";
+
+      leaf remote-port {
+        type leafref {
+          path "/system/logging/remote-servers/remote-server/config/remote-ports";
+        }
         description 
-          "Sets the destination port numbers for syslog messages to
-          the server.";
+          "Associate a selector-set to a remote-port.";
       }
 
       leaf selector-set-name {
--- a/release/models/system/openconfig-system-logging.yang
+++ b/release/models/system/openconfig-system-logging.yang
@@ -23,7 +23,13 @@ module openconfig-system-logging {
     "This module defines configuration and operational state data
     for common logging facilities on network systems.";
 
-  oc-ext:openconfig-version "0.4.1";
+  oc-ext:openconfig-version "0.5.0";
+
+  revision "2023-02-01" {
+    description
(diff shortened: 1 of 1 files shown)


release/models/system/openconfig-system-logging.yang
```diff
@@ -300,19 +305,71 @@ module openconfig-system-logging {
...
+        key "name";
+        description
+          "List of selectors for log messages";
+
+        leaf name {
+          type leafref {
+            path "../config/name";
+          }
+          description
+            "Reference to prefix name list key";
```
jsterne, line 331: wrong description?

jsterne:
The tree above seems to be missing the remote-ports leaf-list from the state container. It also doesn't highlight the fact that the facility and severity leafs under the normal selectors suddenly become leaf-lists (instead of leafs) - that feels odd.

I'm not a huge fan of adding all this complexity to this model for a configuration that most major routers don't support. It also puts these new "selector-sets" at the top level of logging, which gives the impression they are generically applicable, but then they are only used in the multiple-ports case (not for more typical single-port remote-server or console cases).

The selector-sets are also just independant lists of severities and of facilities. In the normal selectors those are individual specific tuples <facility, severity>.

LeonGWang:
> I'm not a huge fan of adding all this complexity to this model for a configuration that most major routers don't support. It also puts these new "selector-sets" at the top level of logging, which gives the impression they are generically applicable, but then they are only used in the multiple-ports case (not for more typical single-port remote-server or console cases).

I agree with @jsterne on this, especially regarding the additional complexity introduced by this PR.

Would it be better if `selector-sets` were also used under `remote-servers` and `console`, with the `selectors` list deprecated?
That way, at least it makes sense that `selector-sets` is defined at the top level of logging.

Issue #642, opened 2022-05-06
Incorrect when statement in vlan module


eugeneyudanov:
setup versions:
libyang - 2.0.112
sysrepo - 2.0.53

I have some problems with openconfig-vlan module. Leafs access-vlan, native-vlan, trunk-vlans have 'when' condition like `when "../interface-mode = 'oc-vlan-types:TRUNK'"` and `when "../interface-mode = 'oc-vlan-types:ACCESS'"`. Leaf interface-mode has the type `type oc-vlan-types:vlan-mode-type;`. vlan-mode-type is enumeration with values 'ACCESS' and 'TRUNK'. 

Problem: When I am trying to work with leafs access-vlan, native-vlan, trunk-vlans i got error because 'when' condition cannot be true in any cases. I can set to the leaf vlan-mode-type only values 'ACCESS' and 'TRUNK'. If I remove prefix in 'when' condition everything works fine.

I checked modules openconfig-vlan and openconfig-vlan-types through online validator Yanglint. It return warnings like
> libyang warn: Invalid value "oc-vlan-types:TRUNK" which does not fit the type (Invalid enumeration value "oc-vlan-types:TRUNK".).
> libyang warn: Previous warning generated by XPath subexpression[3] "interface-mode = 'oc-vlan-types:TRUNK'" with context node "/openconfig-interfaces:interfaces/interface/openconfig-if-aggregate:aggregation/openconfig-vlan:switched-vlan/state/trunk-vlans".
> libyang warn: Invalid value "oc-vlan-types:ACCESS" which does not fit the type (Invalid enumeration value "oc-vlan-types:ACCESS".).
> libyang warn: Previous warning generated by XPath subexpression[3] "interface-mode = 'oc-vlan-types:ACCESS'" with context node "/openconfig-interfaces:interfaces/interface/openconfig-if-aggregate:aggregation/openconfig-vlan:switched-vlan/state/access-vlan".
> ...

Prefixes was added by [commit](https://github.com/openconfig/public/commit/b3e96c91a86fc26af12b79bb7fa9b7b646a4a4c4).

It looks like it is the module problem. As I know we should use prefixes for identityref. For enumeration shouldn't. 
If I am wrong in something or do incorrect, please tell me.
Thank you!

robshakir:
Thanks for catching this -- this does indeed sound like an issue in the module. I think this is one of the few cases where we have a `when` statement that references an `enumeration` value rather than an `identity`.

[...]

PR #800, opened 2023-02-02
vlan - remove prefixes from enum values in condition clauses

donaldh:
### Change Scope

* Bugfix in the openconfig-vlan module

When trying to work with leafs access-vlan, native-vlan, trunk-vlans you get an error because 'when' condition cannot be true in any cases. It's only valid to set the leaf vlan-mode-type to values 'ACCESS' and 'TRUNK'. By removing the prefix in the 'when' conditions everything works fine.

Fixes issue #642

### DIFF
--- a/release/models/vlan/openconfig-vlan.yang
+++ b/release/models/vlan/openconfig-vlan.yang
@@ -26,7 +26,13 @@ module openconfig-vlan {
     "This module defines configuration and state variables for VLANs,
     in addition to VLAN parameters associated with interfaces";
 
-  oc-ext:openconfig-version "3.2.1";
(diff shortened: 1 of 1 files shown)


lllyfeng (approved):

robshakir:
Thanks for this change, looks like this was a collateral damage error when we prefixed all identities.

Please can you update the `openconfig-version` statement in this module and any submodules of the module, and add the relevant `revision` statements per the automated check? 

Many thanks!

dplore:
Reviewed in OC operators meeting today without objection.

dplore:
Last call, this will merge on Feb 21, 2023

dplore (approved):

PR #801, opened 2023-02-05
fixed typo CSFP to CSPF in openconfig-mpls-te.yang

bensons:
### Change Scope

* Fixed a typo in the description of cspf-tiebreaker
* removed some extra whitespace in front of the path-control leaf to align with the rest of the model

### Platform Implementations

 * NA, but useful for anybody that might wish to grep for *SFP optical module references...


robshakir:
Thanks for the change! Please can you update the `openconfig-version` statement in the module (and any submodules), along with adding `revision` statement per the automated check?

Thanks!

PR #804, opened 2023-02-07
network-instance - add prefixes to identity values in when statements

donaldh:
### Change Scope

* Bugfix in openconfig-network-instance.yang

Validated with yanglint from libyang
Tested with sysrepo


dplore (approved):
LGTM, thanks for the bug fix.

PR #807, opened 2023-02-07
Patch 1

sachendras:
Suggested wording for the interface-ref applicability for vrf-selection-policy


dplore:
Thanks for this @sachendras 

@robshakir any thoughts on updating your PR #802  to include the change here in #807?

robshakir:
@dplore Yes, this is a PR against my branch. I'll review and then we can merge into there and then merge that into `master`.

PR #808, opened 2023-02-08
openconfig-transport-types.yang add ETH_100GBASE_DR PMD type

liulk:
100G-DR (range 500m) is interoperable with 100G-FR (2km). This addition is backwards compatible.

Co-authored-by: Likai Liu <liulk@google.com>

### DIFF
--- a/release/models/optical-transport/openconfig-transport-types.yang
+++ b/release/models/optical-transport/openconfig-transport-types.yang
@@ -1051,6 +1051,11 @@ module openconfig-transport-types {
     description "Ethernet compliance code: 100GBASE_FR";
   }
 
+  identity ETH_100GBASE_DR {
+    base ETHERNET_PMD_TYPE;
+    description "Ethernet compliance code: 100GBASE_DR";
+  }
+
   identity ETH_400GBASE_ZR {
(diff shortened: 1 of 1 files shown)


release/models/optical-transport/openconfig-transport-types.yang
```diff
@@ -22,7 +22,13 @@ module openconfig-transport-types {
     "This module contains general type definitions and identities
     for optical transport models.";
 
-  oc-ext:openconfig-version "0.17.1";
+  oc-ext:openconfig-version "0.17.2";
```
dplore, line 25: I think formally in [semver.org](https://semver.org/) this change is considered a minor rev vs. patch as we are adding data vs. bug fixing existing data.

dplore (approved):
LGTM, I only have one non-blocking suggestion to consider incrementing minor version vs. patch version.

dplore:
Plan to merge after OC operators review on Feb 14,2023

release/models/optical-transport/openconfig-transport-types.yang
liulk: Done.

dplore:
Reviewed without comment from OC operators

PR #811, opened 2023-02-11
Refactor resource utilization

rolandphung:
### Change Scope
* (M) release/models/platform/openconfig-platform-common.yang
* (A) release/models/platform/openconfig-platform-utilization.yang
  * Move resource utilization into it's own yang file 
  * Update the grouping names and documentation to be more consistent
    * From `utilization (resource)` to `resource utilization`
  * Use `augment` instead of importing directly into the respective `/components/component/<component-type>` containers

* These changes are backwards compatible


robshakir:
Note, I think this is actually *not* backwards compatible because the RFC7951 representation will change the namespace that is used. Since this is telemetry generally then this is less of a concern, but I think we should flag this breaking.

dplore:
To be reviewed on Feb 14 OC Operators meeting

PR #813, opened 2023-02-17
Revert "Add support for AND/OR for flags; built-in flag match criteria."

robshakir:
Reverts openconfig/public#796 - upstream breakage. To be resubmitted.


greg-dennis:
This can be abandoned. Just need a PR to remove the newly non-existent paths from https://github.com/openconfig/featureprofiles/blob/main/feature/experimental/policy/policy_base/feature.textproto

PR #815, opened 2023-02-22
Clarify the default of global AF enabled leaf for ipv4-unicast

yaonanliang:
This change clarifies that BGP global AF `enabled` leaf for `ipv4-unicast` afi-safi is enabled by default. This is the followup PR to https://github.com/openconfig/public/pull/774 More context can be found over there.

`ipv4-unicast` is the only afi-safi supported in BGP protocl ([RFC4271](https://www.rfc-editor.org/rfc/rfc4271)). Other afi-safis are supported later in the context of MP-BGP ([RFC4760](https://www.rfc-editor.org/rfc/rfc4760)). Since `ipv4-unicast` is an inextricable component of the BGP protocol itself, it's enabled by default in most implementations today. 

There is not a reasonable way to indicate this condition in our model using a `when` statement, so this change proposes to clarify that `ipv4-unicast` is enabled by default in the the description of the global AF `enable` leaf.

### DIFF
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -193,8 +193,11 @@ submodule openconfig-bgp-common-multiprotocol {
       type boolean;
       default false;
       description
-        "This leaf indicates whether the AFI-SAFI is
-        enabled for all neighbors or groups";
+        "This leaf indicates whether the AFI-SAFI is enabled for all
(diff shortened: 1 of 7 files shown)


dplore:
This will be reviewed at March 1 OC Operators meeting.  @yaonanliang if you are affiliated with a network operator, please consider joining.

dplore:
Discussed in Feb 28th OC operators meeting.  There is some preference to update the model to have ipv4 and ipv6 enabled leafs and set the defaults appropriately vs. a description change that says to ignore the default in the case of ipv4.

PR #816, opened 2023-02-22
Clarify IS-IS instance identifiers per #710.

robshakir:
```
   - Remove the `instance` leaf that was ambiguous as to whether
     this is the name of the instance, or the instance identifier.
   - Add a new uint16 instance-identifier leaf.
```

### Change Scope

* removes the `instance` leaf (deprecated) since it was of type `string`
  and causes ambiguity as to whether this is the name of the instance
  or the instance ID.
* adds a new `instance-id` leaf that explicitly specifies the instance
  identifier TLV value.

### Platform Implementations

* see #710 for discussion of support. confirmed supported in ios xr and sros.

### DIFF
--- a/release/models/isis/openconfig-isis-lsp.yang
+++ b/release/models/isis/openconfig-isis-lsp.yang
@@ -34,7 +34,15 @@ submodule openconfig-isis-lsp {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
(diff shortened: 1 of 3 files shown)


dplore (approved):
LGTM, essentially a cleanup/bugfix to align with https://www.rfc-editor.org/rfc/rfc6822#section-2.1

dplore:
Reviewed in Feb 28 OC operators call without objection.  Will merge on March 14, 2023

PR #817, opened 2023-02-23
Add support for utilization thresholds config per resource

rolandphung:
### Change Scope

* Adding system level configuration for resource utilization thresholds. The motivation for this change is that there needs to be a way to configure per resource, rather than per component per resource. It should make sense that for whichever component is added into the switch dynamically, should already be configured with the corresponding threshold.
  * This is a follow up to #754 

pyang output:
```pyang
module: openconfig-system
  +--rw system
     +--rw oc-sys-util:utilization
        +--rw oc-sys-util:resources
           +--rw oc-sys-util:resource* [name]
              +--rw oc-sys-util:name      -> ../config/name
              +--rw oc-sys-util:config
              |  +--rw oc-sys-util:name?                         string
              |  +--rw oc-sys-util:used-threshold-upper?         oc-types:percentage
              |  +--rw oc-sys-util:used-threshold-upper-clear?   oc-types:percentage
              +--ro oc-sys-util:state
                 +--ro oc-sys-util:name?                         string
                 +--ro oc-sys-util:used-threshold-upper?         oc-types:percentage
                 +--ro oc-sys-util:used-threshold-upper-clear?   oc-types:percentage
                 +--ro oc-sys-util:active-component-list*        -> /oc-platform:components/component/config/name
```

### Platform Implementations

 * Arista `hardware capacity alert table <table> [feature <feature>] threshold <threshold>`: [link to documentation](https://www.arista.com/en/um-eos/eos-logging-of-event-notifications)

### DIFF
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -20,7 +20,14 @@ submodule openconfig-platform-common {
     "This modules contains common groupings that are used in multiple
     components within the platform module.";
 
-  oc-ext:openconfig-version "0.22.0";
+  oc-ext:openconfig-version "0.23.0";
+
+  revision "2023-02-13" {
+    description
+      "Refactor resource utilization threshold config into a separate grouping.
+      Update 'utilization resource' to 'resource utilization'.";
+    reference "0.23.0";
+  }
 
   revision "2022-12-20" {
     description
@@ -80,13 +87,13 @@ submodule openconfig-platform-common {
 
   // grouping statements
 
-  grouping platform-utilization-top {
+  grouping platform-resource-utilization-top {
     description
-      "Top level of utilization.";
+      "Top level grouping of platform resource utilization.";
 
     container utilization {
       description
-        "Utilization of the component.";
+        "Resource utilization of the component.";
 
       container resources {
         description
@@ -109,7 +116,7 @@ submodule openconfig-platform-common {
             description
               "Configuration data for each resource.";
 
-            uses platform-utilization-resource-config;
+            uses platform-resource-utilization-config;
           }
 
           container state {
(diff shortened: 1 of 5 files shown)


rolandphung:
> It seems overkill to me that the thresholds are configured per IC. Do most implementations actually support this, or are the thresholds usually set at the system level?

@nokia1adam This PR is in response to your comment from #754. I do agree that there should be a configuration across all ICs. PTAL

lllyfeng:
should add interval leaf to indicate the interval time of threshold monitoring?

rolandphung:
> should add interval leaf to indicate the interval time of threshold monitoring?

At what granularity? How do you propose this should look?

dplore:
Is the platform tree changed?  Please  show a diffs for both the platform and system trees.    (Looks like system tree is indicated in the introductory comment as just an add which seems to match the yang)

rolandphung:
The platform tree has not changed. The yang code was refactored to pull `used-threshold-upper` and `used-threshold-clear` into `resource-utilization-threshold-common` which is reused in `/system/utilization/resources/resource/config`

dplore:
Last call, to be merged on May 4th, 2023

dplore:
Cisco XROS has an implementation of [utilization and threshold configuration ](https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5500/system-monitoring/b-ncs5500-system-monitoring-cli-reference/oor-handling-commands.html#wp8797494580)

JunOS has utilization and threshold capability via  it's [resource-monitor](https://www.juniper.net/documentation/us/en/software/junos/network-mgmt/topics/topic-map/troubleshooting-system-performance-with-resource-monitoring-methodology.html) feature

dplore (approved):
LGTM, thank you for your contribution.  Note, #870 (a copy of this PR rebased to with updates to CI checks) shows these models pass all CI checks.

PR #820, opened 2023-02-27
fixing type of set-metric-type field

awebsters:
### Change Scope

Changing the type of the set-metric-type leaf in IS-IS (previously incorrect).
Change is not backwards compatible, however, the field changed would be unuseable in previous versions

### Platform Implementations

Relevant to any implementation of IS-IS


dplore (approved):
LGTM.  I will label as a breaking change.  It will be merged shortly before releasing OpenConfig 2.0.  It is a goal to release OpenConfig 2.0 by the end of March 2023.

dplore:
I have proposed a non-breaking form of this PR at #841

PR #825, opened 2023-03-04
add control plane queue config and state

dplore:
### Change Scope

* Add queue config and state to control plane traffic

The intent of these queues is expose their existence and statistics.  In the referenced implementations the traffic classes and queues (if any) for control plane traffic offer no configuration that I could find.  These queues are 'system defined' queues which have system defined classifiers, queues and schedulers.  This is compliant with the QoS model as it explicitly permits systems to expose default queue names and their associated statistics.  Future PR's could add configurability of these queues if there is an operational need and implementation.
 
(Note: Control plane policers are typically configurable and are intended to be covered by the traffic scheduler paths which are already defined for control-plane-traffic).    

### Platform Implementations

 *  [Cisco XR OS Control Plane Queue](https://content.cisco.com/chapter.sjs?uri=/searchable/chapter/content/en/us/td/docs/ios-xml/ios/qos_plcshp/configuration/15-mt/qos-plcshp-15-mt-book/qos-plcshp-cpp.html.xml#GUID-F05A63BE-DC5E-4D0B-9B61-FA8B78F5ECAB) thresholding feature defines queues as part of it's control plane traffic features
 * [Arista EOS CoPP policy](https://www.arista.com/en/um-eos/eos-traffic-management#xx1152216) defines queues depending on the hardware platform supported.  
 * [JunOS](https://www.juniper.net/documentation/us/en/software/junos/security-services/topics/topic-map/configuring-ddos-protection.html) supports control plane traffic policing which likely maps to the policers already defined.  It is not clear if JunOS also has queues for control plane traffic.

### Tree view

```
     +--rw oc-sys-copp:control-plane-traffic
     |  +--rw oc-sys-copp:ingress
        |     +--rw oc-sys-copp:queues
        |        +--ro oc-sys-copp:queue* [name]
        |           +--ro oc-sys-copp:name     -> ../state/name
        |           +--ro oc-sys-copp:state
        |              +--ro oc-sys-copp:name?              -> /oc-qos:qos/queues/queue/state/name
        |              +--ro oc-sys-copp:max-queue-len?     oc-yang:counter64
        |              +--ro oc-sys-copp:avg-queue-len?     oc-yang:counter64
        |              +--ro oc-sys-copp:transmit-pkts?     oc-yang:counter64
        |              +--ro oc-sys-copp:transmit-octets?   oc-yang:counter64
        |              +--ro oc-sys-copp:dropped-pkts?      oc-yang:counter64
        |              +--ro oc-sys-copp:dropped-octets?    oc-yang:counter64
```

### DIFF
--- a/release/models/system/openconfig-system-controlplane.yang
+++ b/release/models/system/openconfig-system-controlplane.yang
@@ -204,28 +204,44 @@ module openconfig-system-controlplane {
               }
             }
 
-            container queue {
+            container queues {
               description
                 "Configuration and operational state relating to the QoS
-                queues that are used for control-plane traffic.  The
+                queues that are used for control plane traffic.  The
                 control-plane container is treated like an interface.";
-
-              container config {
+              
+              list queue {
+                key "name";
+                
                 description
-                  "Configuration parameters relating to the queues used
-                  for control-plane traffic.";
+                "Top-level container for the queue associated with 
+                control plane traffic";
 
-                uses system-control-plane-queue-config;
-              }
+                leaf name {
+                  type leafref {
+                    path "../state/name";
+                  }
+                  description
+                    "Reference to the name of the queue
+                    for control plane traffic.";
+                }
 
-              container state {
-                config false;
-                description
-                  "Operational state parameters relating to the queues
-                  used for control-plane traffic.";
+                container state {
+                  config false;
+                  description
+                    "Operational state parameters relating to the queues
+                    used for control-plane traffic.";
 
-                uses system-control-plane-queue-config;
-                uses system-control-plane-queue-state;
+                  leaf name {
+                    type string;
+                    description
+                      "Reference to the queue associated with control plane 
+                      traffic.  These queues are expected to be implicitly
+                      created by the system.";
+                  }
+
+                  uses oc-qos:qos-interface-queue-state;
+                }
               }
             }
           }
@@ -242,26 +258,20 @@ module openconfig-system-controlplane {
       }
     }
 
-    grouping system-control-plane-queue-config {
-      description
-        "Configuration data for queues associated with the
-        interface, this is re-used across input/output queues.";
-
-      leaf name {
-        type string;
-        description
-          "Reference to the queue associated with the control plane.
-          A queue may be explicitly configured, or implicitly created
-          by the system based on default queues that are instantiated
-          by a hardware component, or are assumed to be default on
-          the system.";
-      }
-    }
-
   grouping system-control-plane-queue-state {
     description
       "Operational state data for the queue associated with the
       control plane";
+
+    leaf name {
+      type string;
+      description
+        "Reference to the queue associated with the control plane.
+        A queue may be explicitly configured, or implicitly created
+        by the system based on default queues that are instantiated
+        by a hardware component, or are assumed to be default on
+        the system.";
+    }
 
     leaf max-queue-len {
       type oc-yang:counter64;
--- a/release/models/system/openconfig-system-controlplane.yang
+++ b/release/models/system/openconfig-system-controlplane.yang
@@ -4,18 +4,11 @@ module openconfig-system-controlplane {
     namespace "http://openconfig.net/yang/system-controlplane";
     prefix "oc-sys-copp";
 
-    import openconfig-extensions {
-        prefix oc-ext;
-    }
-    import openconfig-system {
-        prefix oc-sys;
-    }
-    import openconfig-acl {
-        prefix oc-acl;
-    }
-    import openconfig-qos {
-        prefix oc-qos;
-    }
+    import openconfig-extensions { prefix oc-ext; }
+    import openconfig-system { prefix oc-sys; }
+    import openconfig-acl { prefix oc-acl; }
+    import openconfig-qos { prefix oc-qos; }
+    import openconfig-yang-types { prefix oc-yang; }
 
     organization
       "OpenConfig working group";
@@ -31,9 +24,13 @@ module openconfig-system-controlplane {
           a QoS classifier.
         - schedule traffic that has been forwarded towards the control-plane,
           to allow for policies such as rate limits to be applied.
+        - queue traffic ingress or egress from the control plane, allowing
+          packets to be queued separately based on a QoS classifier.
+
        The configured policies apply generically to all control-planes that
        exist within the system, and should be mapped to the internal interfaces
        via which packets are forwarded to control-plane modules.
+
        When a packet is received at an input interface - it is classified into a
        forwarding group which drains to a specific queue. If this input mapping
        is sufficient, the CPU-facing interface uses the specified scheduler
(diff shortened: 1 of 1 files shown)


release/models/system/openconfig-system-controlplane.yang
```diff
@@ -198,6 +203,31 @@ module openconfig-system-controlplane {
...
               }
             }
+
+            container queue {
+              description
+                "Configuration and operational state relating to the QoS
+                queues that are used for control-plane traffic.  The
+                control-plane container is treated like an interface.";
+
+              container config {
```
rszarecki, line 213: I'm not sure if this is necessary to configure queue name. But it doesn't not harm either.

rszarecki (approved):

release/models/system/openconfig-system-controlplane.yang
dplore: This leaf is added so one can stream a list of queues, keyed by name.  I expect systems will only expose these as state and not allow configuration as I understand these names are typically fixed based on the hardware that is used.

dplore: I see your point though, if we expect the control traffic queue names are statically defined by the system, we could make the name state only.   My guess is like you said, leave name as configurable as it is not harmful, even if the referenced platforms do not support it currently.  @robshakir any thoughts?

robshakir:
Couple of questions:

1. This only seems (from the tree view) to do one queue -- don't we need this to be N queues?
2. Can we re-use the queue leaf definitions from the QoS model?  This seems to be a case of just re-using (`qos-interface-queue-state`)[https://github.com/openconfig/public/blob/master/release/models/qos/openconfig-qos-interfaces.yang#L318]

release/models/system/openconfig-system-controlplane.yang
robshakir: In the QoS model we have this a read-write leaf because there are other parameters that are being specified under this queue (per-interface per-queue queue-management approach for example). If we don't foresee a need here that this will ever be configurable in this context then a ro leaf is possible.

I don't think this is the same as the discussion above though - the classifier that is applied here can specify a group to queue mapping, which may be more than just the set of queues that are available in hardware. For example, CoPP policies can have concepts such as the "configured BGP peers" in a queue with a specific rate limit etc. This however, would be done through the classifier and the queue counters would still be here with that static name.

release/models/system/openconfig-system-controlplane.yang
```diff
@@ -212,6 +242,66 @@ module openconfig-system-controlplane {
       }
     }
 
+    grouping system-control-plane-queue-config {
+      description
+        "Configuration data for queues associated with the
+        interface, this is re-used across input/output queues.";
+
+      leaf name {
```
robshakir, line 250: Should this be a leafref to the queue that is defined in the QoS model per the other approaches that we have in this model?

release/models/system/openconfig-system-controlplane.yang
dplore: For a moment I would like to focus on what is the intended way to count policer drops for control plane traffic, since that is the operational goal.   Is my interpretation below correct?

`/qos/scheduler-policies/scheduler-policy/schedulers/scheduler/one-rate-two-color/config/queuing-behavior` is set to POLICE. (see [openconfig type-queue-behavior](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-qos.html#type-queue-behavior))

`/qos/scheduler-policies/scheduler-policy/schedulers/scheduler/one-rate-two-color/violate-action/config/drop` is set to TRUE (ie: violate = packet drop)

That scheduler is then applied to an interface.  Packets that violate the policer are counted via `openconfig/system/controlplane/control-plane-traffic/ingress/qos/scheduler-policy/scheduler-statistics/scheduler[sequence]/state/violating-pkts`
Since the violate action is drop, this counter will count dropped packets, yes?

Reading the [OC qos docs](https://www.openconfig.net/docs/models/qos/), it seems a queue is meant to be created implicitly by the device, and this same queue is implicitly associated with the interface where the scheduler is applied.    When the scheduler (policier) drops packets, should the queue drop packet counter also increment?  Or should the queue counter only count TAIL,RED,WRED dropped packets?  It seems most useful for the queue counter to only count  TAIL,RED,WRED dropped packets?

release/models/system/openconfig-system-controlplane.yang
dplore: > Couple of questions:
> 
> 1. This only seems (from the tree view) to do one queue -- don't we need this to be N queues?
> 2. Can we re-use the queue leaf definitions from the QoS model?  This seems to be a case of just re-using (`qos-interface-queue-state`)[https://github.com/openconfig/public/blob/master/release/models/qos/openconfig-qos-interfaces.yang#L318]

Done.

Also used leafref for the queue name.

dplore:
> Couple of questions:
> 
> 1. This only seems (from the tree view) to do one queue -- don't we need this to be N queues?
> 2. Can we re-use the queue leaf definitions from the QoS model?  This seems to be a case of just re-using (`qos-interface-queue-state`)[https://github.com/openconfig/public/blob/master/release/models/qos/openconfig-qos-interfaces.yang#L318]

Updated to use a list and leafref to the qos queue name

dplore:
Added tp agenda for April 18, 2023 OC Operators review.

dplore:
Reviewed without objection from OC Operators.  Setting last call, will merge on May 2nd.

@sulrich @earies @rgwilton @hellt  for any additional comments.

release/models/system/openconfig-system-controlplane.yang
```diff
@@ -198,6 +202,53 @@ module openconfig-system-controlplane {
...
+                container state {
+                  config false;
+                  description
+                    "Operational state parameters relating to the queues
+                    used for control-plane traffic.";
+
+                    leaf name {
+                      config false;
+                      type leafref {
+                        path "/oc-qos:qos/oc-qos:queues/oc-qos:queue/" +
```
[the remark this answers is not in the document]
dplore: Is there a functional difference between leaf-ref and  cutting a pasting the queue configuration?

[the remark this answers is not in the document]
aashaikh: Referencing the QoS model implies that the COPP queues are part of the same set of queues -- is that the case on most platforms?   If they are a separate structure,  we could just instantiate the queue definitions here with `uses` (i.e., just reuse the groupings, no  copy/paste).

cosmos2k17:
control plane queues and physical interface (user facing port) queues will be different in most platforms, hope we should have them in different space.
Both of these queues should be created by system and we should refer to those queue.

rszarecki:
> * [Cisco XR OS Control Plane Queue](https://content.cisco.com/chapter.sjs?uri=/searchable/chapter/content/en/us/td/docs/ios-xml/ios/qos_plcshp/configuration/15-mt/qos-plcshp-15-mt-book/qos-plcshp-cpp.html.xml#GUID-F05A63BE-DC5E-4D0B-9B61-FA8B78F5ECAB) thresholding feature defines queues as part of it's control plane traffic features

This is link to legacy IOS. Not IOS-XR.
AFAIK IOS XR do not expose control plame queues.

JUNIPER has some control-plane queues which are hardcoded and not exposed.

PR #827, opened 2023-03-07
Add config leafs for bgp dynamic-neighbor-prefixes

xchu-arista:
Adding extra configuration parameters to /network-instances/network-instance/protocols/protocol/bgp/global/dynamic-neighbor-prefixes/dynamic-neighbor-prefix/[config|state]

### Change Scope

* Adding leaf include-router-id to help uniquely identify dynamic BGP peers.
* Adding leaf remote-as to specify BGP neighbor's Autonomous System number.

### DIFF
--- a/release/models/bgp/openconfig-bgp-global.yang
+++ b/release/models/bgp/openconfig-bgp-global.yang
@@ -26,7 +26,14 @@ submodule openconfig-bgp-global {
     "This sub-module contains groupings that are specific to the
     global context of the OpenConfig BGP module";
 
-  oc-ext:openconfig-version "9.2.0";
+  oc-ext:openconfig-version "9.3.0";
+
+  revision "2023-03-07" {
+    description
+      "Added extra configuration parameters relating to an individual prefix
+      from which dynamic neighbors are accepted";
+    reference "9.3.0";
+  }
 
   revision "2022-12-12" {
     description
@@ -297,6 +304,24 @@ submodule openconfig-bgp-global {
         configured.  The configuration parameters used for the dynamic
         neighbor are those specified within the referenced peer
         group.";
+    }
+
+    leaf include-router-id {
+      type boolean;
+      default false;
+      description
+        "This value controls whether the router ID will be used to
+        identify BGP peers. A BGP peer by default is identified by
+        the network instance and peer address.
+        When this is set to true, the router ID will be included in
+        the identification of the BGP peer which will help to uniquely
+        identify dynamic peers.";
+    }
+
+    leaf remote-as {
+      type oc-inet:as-number;
+      description
+        "Neighbor's Autonomous System number."
     }
   }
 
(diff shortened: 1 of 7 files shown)


dplore:
@s19nal for review

release/models/bgp/openconfig-bgp-global.yang
```diff
@@ -298,6 +305,24 @@ submodule openconfig-bgp-global {
...
+        When this is set to true, the router ID will be included in
+        the identification of the BGP peer which will help to uniquely
+        identify dynamic peers.";
+    }
+
+    leaf remote-as {
+      type oc-inet:as-number;
+      description
+        "Neighbor's Autonomous System number.";
+    }
```
s19nal, line 325: I think this may be in the wrong place in the model?   From the main description in openconfig-bgp.yang
```
      BGP
        |
        +-> [ global BGP configuration ]
          +-> AFI / SAFI global
        +-> peer group
          +-> [ peer group config ]
          +-> AFI / SAFI [ per-AFI overrides ]
        +-> neighbor
          +-> [ neighbor config ]
          +-> [ optional pointer to peer-group ]
          +-> AFI / SAFI [ per-AFI overrides ]
```
we already have a leaf peer-as under the peer-group config rather than remote-as, so I think only the boolean (which seems like a reasonable addition to me) needs including, just moved to under the peer-group config grouping.
```
     +--rw peer-groups
     |  +--rw peer-group* [peer-group-name]
     |     +--rw peer-group-name       -> ../config/peer-group-name
     |     +--rw config
     |     |  +--rw peer-group-name?      string
     |     |  +--rw peer-as?              oc-inet:as-number
     |     |  +--rw local-as?             oc-inet:as-number
     |     |  +--rw peer-type?            oc-bgp-types:peer-type
     |     |  +--rw auth-password?        oc-types:routing-password
     |     |  +--rw remove-private-as?    oc-bgp-types:remove-private-as-option
     |     |  +--rw route-flap-damping?   boolean
     |     |  +--rw send-community?       oc-bgp-types:community-type
     |     |  +--rw description?          string
```
cheers
Simon

xchu-arista: Hi Simon,

Thank you for the review!

Grouping these newly proposed leafs under dynamic-neighbor-prefixes instead of peer-groups is intended to give dynamic neighbors under the same BGP peer-group the option to have different configurations.

For example, we might have prefix 11.11.0.1/24 and prefix 12.12.0.1/24 under the same peer-group but using different remote AS numbers and only one of them intends to include router ID in BGP peering considerations.

"Static" BGP neighbors already have such an option. leaf peer-as is defined in a grouping commonly referred to by both BGP peer-groups config and BGP neighbors config. In other words, a BGP neighbor may override the peer-as defined for the peer-group that it belongs to.

Therefore I still there are benefits in keeping these two configuration parameters where they are.

PR #828, opened 2023-03-08
Fix other spec.yml issues for online documentation.

wenovus:
/system/control-plane-traffic/...

and other paths

-------------
tested manually


dplore (approved):
Thanks @wenovus .  @cfyo this will fix the yang doc generation on openconfig.net.

@wenovus we also created a bug to enhance our lint checker to make sure these .spec references aren't missing

PR #829, opened 2023-03-10
Support setting priority for every VRRP track interface

lllyfeng:
### Change Scope

* Currently YANG module can only support set one priority for all VRRP tracked interfaces. It's inconvient 
   if one tracked interface wants to have a different priority. So there should be a list for VRRP track interfaces
  with two numbers, interface and priority-decrement.
OLD:
```
+--rw interface-tracking
    +--rw config
     |  +--rw track-interface* leafref
     |  +--rw priority-decrement? uint8
```
NEW:
```
+--rw interface-tracking
    +--rw track-interface* [interface]
        +--rw interface -> ../config/interface
        +--rw config
        |  +--rw interface? leafref
        |  +--rw priority-decrement? uint8
         +--ro state
            +--ro interface? leafref
            +--ro priority-decrement? uint8
```
   
* non-backwards compatible.
### Platform Implementations

 * H3C: [https://www.h3c.com/en/Support/Resource_Center/HK/Switches/H3C_S7500E/S7500E/Technical_Documents/Configure___Deploy/Configuration_Guides/S7500E_OM(6300_V1.03)/08-System_Volume/200912/658916_294551_0.htm](https://www.h3c.com/en/Support/Resource_Center/HK/Switches/H3C_S7500E/S7500E/Technical_Documents/Configure___Deploy/Configuration_Guides/S7500E_OM(6300_V1.03)/08-System_Volume/200912/658916_294551_0.htm)
 * Junos: 
 [https://www.juniper.net/documentation/us/en/software/junos/high-availability/topics/ref/statement/track-edit-interfaces-vrrp.html](https://www.juniper.net/documentation/us/en/software/junos/high-availability/topics/ref/statement/track-edit-interfaces-vrrp.html)
* Cisco:
[https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/6-x/unicast/configuration/guide/l3_cli_nxos/l3_vrrp.html](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/6-x/unicast/configuration/guide/l3_cli_nxos/l3_vrrp.html)
search 'track interface'
* Huawei:
[https://github.com/Huawei/yang/blob/master/network-router/8.22.0/ne40e-x8x16/huawei-vrrp.yang#L1420](https://github.com/Huawei/yang/blob/master/network-router/8.22.0/ne40e-x8x16/huawei-vrrp.yang#L1420)

### DIFF
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -725,6 +725,7 @@ module openconfig-if-ip {
         type leafref {
           path "/oc-if:interfaces/oc-if:interface/oc-if:name";
         }
+        description "the track interface";
       }
       leaf priority-decrement {
         type uint8 {
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -706,11 +706,38 @@ module openconfig-if-ip {
 
   grouping ip-vrrp-tracking-config {
     description
-      "Configuration data for tracking interfaces
+      "Configuration data for tracking interface
       in a VRRP group";
-    container track-interfaces {
+
+    leaf interface {
+      type leafref {
+        path "/oc-if:interfaces/oc-if:interface/oc-if:name";
+      }
+      description "the track interface";
+    }
+    leaf priority-decrement {
+      type uint8 {
+        range 0..254;
+      }
+      default 0;
+      description "Set the value to subtract from priority when
+                  the tracked interface goes down";
+    }
+  }
+
+  grouping ip-vrrp-tracking-state {
+    description
+      "Operational state data for tracking interfaces in a VRRP
+      group";
+  }
+
+  grouping ip-vrrp-tracking-top {
+    description
+      "Top-level grouping for VRRP interface tracking";
+
+    container interface-tracking {
       description
-        "The configuation data for tracking interfaces";
+        "Top-level container for VRRP interface tracking";
       list track-interface {
         key "interface";
         // TODO: we may need to add some restriction to ethernet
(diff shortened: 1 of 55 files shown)


release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -743,23 +738,43 @@ module openconfig-if-ip {
     container interface-tracking {
```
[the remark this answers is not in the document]
lllyfeng: I have updated it. Thanks.

oscargdd:
I have reviewed the proposal and I have one big concern. Current model DO support setting the priority for every VRRP track interface. The VRRP track interface configuration is applied PER interface, within the interfaces container.

rolandphung:
Now that I think about it, we are making a breaking change and it might be worth considering adding a possible second action that would shutdown the VRRP instance instead. Perhaps this could be represented as an enum/identity leaf called `action` and the possible values would be `PRIORITY_DECREMENT` and `DISABLE`/`SHUTDOWN`.

Ref: https://www.arista.com/en/um-eos/eos-varp#xx1153490

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -700,33 +706,22 @@ module openconfig-if-ip {
...
-      description
-        "Sets a list of one or more interfaces that should
-        be tracked for up/down events to dynamically change the
-        priority state of the VRRP group, and potentially
-        change the mastership if the tracked interface going
-        down lowers the priority sufficiently.  Any of the tracked
-        interfaces going down will cause the priority to be lowered.
-        Some implementations may only support a single
-        tracked interface.";
+      description "the track interface";
```
dplore, line 716:
```suggestion
      description "The tracked interface.";
```

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -700,33 +706,22 @@ module openconfig-if-ip {
...
     }
-
     leaf priority-decrement {
       type uint8 {
         range 0..254;
       }
       default 0;
       description "Set the value to subtract from priority when
-      the tracked interface goes down";
+                  the tracked interface goes down";
```
dplore, line 724: Multi line description should be formatted like:

```
description
  "Set the value to subtract from priority when
  the tracked interface goes down.";
```

dplore (changes requested):
This was reviewed at the Aug 1, 2023 OC operators meeting.    I have a few small comments that should be addressed.

@oscargdd @s19nal will investigate use cases in their own networks for other options that may be needed.  For our network there are scenarios where track interface is used with a priority decrement.  In the examples I found, we we only tracking one interface.

It does seem legitimate to use a different priority for different interfaces, if for example the interfaces being tracked are of different capacity or importance due to the path or cost of the interfaces being tracked.

dplore:
Note, does this need to be a breaking change?   Since a new container with a different name than the leaf-list is being  introduced, the leaf-list could instead be deprecated and the container added.  This would make this a non-breaking change and expidite it's merging into the OpenConfig model set.

dplore:
@lllyfeng waiting for your response on this PR.  Thanks

dplore:
@lllyfeng please address comments if you would like to proceed.  I think the only comment that really should be addressed is the one about deprecation vs. making a breaking change.  (https://github.com/openconfig/public/pull/829#issuecomment-1663060498)

PR #831, opened 2023-03-13
Add isis total-lsp count

dplore:
### Change Scope

* Add  counter for ISIS lsps at the ISIS level

### Platform Implementations

 * Cisco XROS shows total LSP via CLI [show isis checkpoint lsp](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/routing/command/reference/rr40xr12kbook_chapter3.html)
 * Arista EOS shows total LSP via CLI [show isis summary](https://www.arista.com/en/um-eos/eos-is-is)
 * Juniper JunOS shows total LSP via CLI  [show ISIS database](https://www.juniper.net/documentation/us/en/software/junos/is-is/topics/ref/command/show-isis-database.html)

<pre>
  +--rw network-instances
        +--rw protocols
              +--rw isis
                 +--rw levels
                 |  +--rw level* [level-number]
                 |     +--rw level-number             -> ../config/level-number
                 |     +--rw config
                 |     |  x--rw enabled?                boolean
                 |     |  +--rw level-number?           oc-isis-types:level-number
                 |     |  +--rw metric-style?           oc-isis-types:metric-style
                 |     |  +--rw authentication-check?   boolean
                 |     +--ro state
                 |     |  +--ro enabled?                boolean
                 |     |  +--ro level-number?           oc-isis-types:level-number
                 |     |  +--ro metric-style?           oc-isis-types:metric-style
                 |     |  +--ro authentication-check?   boolean
                 |     +--rw system-level-counters
                 |     |  +--ro state
            <b>     |     |     +--ro total-lsps?                       oc-yang:counter32</b>
                 |     |     +--ro corrupted-lsps?                   oc-yang:counter32
                 |     |     +--ro database-overloads?               oc-yang:counter32
                 |     |     +--ro manual-address-drop-from-areas?   oc-yang:counter32
<... snip ...>
</pre>


dplore:
@comment from @cfyo if intent is global lsp count, maybe should create a container in global instead of in levels/level.

cfyo:
I would suggest "total-lsps" be put in a counters container under state for consistency; then when other global counters are added in future this won't have to be done at a later time.

dplore:
> I would suggest "total-lsps" be put in a counters container under state for consistency; then when other global counters are added in future this won't have to be done at a later time.

Done

release/models/isis/openconfig-isis.yang
```diff
@@ -218,6 +224,32 @@ module openconfig-isis {
...
+
+  grouping global-counters-state {
+    description
+      "This container defines global counters";
+
+    leaf total-lsps {
+      type oc-yang:counter32;
+      description
+        "A count of all LSPs in ISIS link state database across all instances
+        and VRFs.";
```
chrisy, line 248: This leaf is located within a specific IS-IS instance; does this really sum the count across _all_ instances? Or do you mean something other than routing protocol instance here?

Corollary, we have the misnomic convention of "global config" container, and repeated here as "global counters", but which are also within specific instances of the protocol. I assume this is a leftover of when we changed from a literal global IS-IS container to one under `network-instances`.

chrisy (changes requested):
One query about how global is global...

release/models/isis/openconfig-isis.yang
dplore: I can see from docs that Juniper and Cisco support a per network instance counter of LSP's based on the links above.  They do not mention support for a 'global' / all instances LSP counter.  

@sulrich it is not clear what LSP counter is supported in EOS from the documentation.  Can you clarify for us?   

I am clarifying with our team what the operational use cases are for these.

sulrich: @dplore a per instance count of LSPs is available via `show isis <instance_name>
summary` command. (`LSDB Size: X`).  i'm not aware of a global multi-instance LSP counter.

additional docs: [IS-IS multiple Instances - Arista](https://www.arista.com/en/support/toi/eos-4-26-1f/14782-is-is-multiple-instances)

s19nal:
@dplore   this LGTM

dplore:
@chrisy I updated this PR to add a leaf for LSP's at the isis level instead of some type of global.  This will meet our use case and avoids ambiguity of what is global in the context of network-instances.

s19nal (approved):

aashaikh:
@dplore my understanding of the examples of support from other vendors is for a _per-instance_ counter -- do those implementations also support a _per-level_ counter?

dplore:
> @dplore my understanding of the examples of support from other vendors is for a _per-instance_ counter -- do those implementations also support a _per-level_ counter?

The examples show sum of lsp's per level per instance.

One example cut/paste here:
```
  RP/0/0/CPU0:router#  show isis checkpoint lsp
  
  Level  LSPID              Chkpt ID 
  1      router-gsr6.00-00  80011f9c 
  1      router-gsr6.01-00  80011f88 
  1      router-gsr8.00-00  80011f74 
  1      router-gsr9.00-00  80011f60 
  2      router-gsr6.00-00  80011f4c 
  2      router-gsr6.01-00  80011f38 
  2      router-gsr8.00-00  80011f24 
  2      router-gsr9.00-00  80011f10 
  Total LSP count: 8 (L1: 4, L2 4, local L1: 2, local L2 2) 
```

aashaikh (approved):
LGTM -- thanks for the clarifications

dplore:
> One query about how global is global...

Hi @chrisy, I have made changes so the global issue I think is no longer in question.  Please approve if you agree (currently blocked merging based on 'request changes' setting in your review.

PR #833, opened 2023-03-15
Clarify routing table deletion

dplore:
### Change Scope

* Adding clarification that network operating systems must delete routing tables for an address family if the address family is removed.   This clarification is introduced based on this featureprofiles test https://github.com/openconfig/featureprofiles/pull/1253/files#r1136470816

### DIFF
--- a/release/models/network-instance/openconfig-network-instance.yang
+++ b/release/models/network-instance/openconfig-network-instance.yang
@@ -574,9 +574,9 @@ module openconfig-network-instance {
               this list when the relevant protocol context is enabled.
               i.e., when a BGP instance is created with IPv4 and IPv6
               address families enabled, the protocol=BGP,
-              address-family=IPv4 table is created by the system.  If
-	      all BGP instances for an address family are deleted, the
-	      associated table must also be deleted.";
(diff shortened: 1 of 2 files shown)


release/models/network-instance/openconfig-network-instance.yang
```diff
@@ -565,7 +574,9 @@ module openconfig-network-instance {
               this list when the relevant protocol context is enabled.
               i.e., when a BGP instance is created with IPv4 and IPv6
               address families enabled, the protocol=BGP,
-              address-family=IPv4 table is created by the system.";
+              address-family=IPv4 table is created by the system.  If
+	      all BGP instances for an address family are deleted, the
+	      associated table must also be deleted.";
```
xw-g, line 579: how about `the removal of the table should not require additional or explicit configurations`?

dplore: Done

xw-g (approved):

dplore:
This was reviewed in the openconfig operators meeting on March 21 without objection.

marcushines (approved):

PR #835, opened 2023-03-21
Route dest policy

awebsters:

Add policy match conditions internal, external, and both to IS-IS

### Change Scope

* Additional feature to IS-IS policy allowing filtering by a routes destination type
* Change is backward compatible

### DIFF
--- a/release/models/isis/openconfig-isis-policy.yang
+++ b/release/models/isis/openconfig-isis-policy.yang
@@ -149,7 +149,7 @@ module openconfig-isis-policy {
       type  isis-types:destination-type;
       description
         "Set the type of routes to redistribute:
-        INTERNAL, EXTERNAL OR BOTH"
+        INTERNAL, EXTERNAL OR BOTH";
     }
   }
 
--- a/release/models/isis/openconfig-isis-policy.yang
+++ b/release/models/isis/openconfig-isis-policy.yang
@@ -30,11 +30,7 @@ module openconfig-isis-policy {
 
   revision "2020-02-04" {
     description
-<<<<<<< HEAD
-      "Adding route destination field";
-=======
       "Adding metric type leaf";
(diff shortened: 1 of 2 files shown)


dplore:
Hi, please sign the CLA and provide at least 2 references to implementations of this feature.  This is needed before the change will be reviewed.  Thanks

awebsters:
> Hi, please sign the CLA and provide at least 2 references to implementations of this feature. This is needed before the change will be reviewed. Thanks

Given the simplicity of the change, is 2 references actually necessary? We could provide a sample code implementation to further cement the need for the change.

dplore:
> > Hi, please sign the CLA and provide at least 2 references to implementations of this feature. This is needed before the change will be reviewed. Thanks
> 
> Given the simplicity of the change, is 2 references actually necessary? We could provide a sample code implementation to further cement the need for the change.

2 references are the minimum and more are preferred.  This helps validate vendor neutrality of the configuration method and feature behavior.   Is there an issue with providing references?      

Occasionally new features are defined in OpenConfig ahead of implementations, but it is discouraged. See the [contribution guide](https://github.com/openconfig/public/blob/master/doc/contributions-guide.md).

chrisy:
> Given the simplicity of the change, is 2 references actually necessary? We could provide a sample code implementation to further cement the need for the change.

Adding on to @dplore's response, by "references" we simply mean links to documentation or examples of the implementation. It's really a very low bar! What we're looking to avoid is replicating vendor-specific structures in OpenConfig since that adds little value and can cause modeling issues later if another vendor does pop up, so our approach is to ask for these pointers to how a mix of vendors implement a feature -- even if it seems obvious, because documentation is always good.

awebsters:
Closed in favour of new PR https://github.com/openconfig/public/pull/863

PR #839, opened 2023-03-27
Fix bgp-set-med-type

dplore:
Clarify the intended use of the types for `bgp-set-med-type`

This change is backwards compatible as no functionality is changed.


swfrench (approved):
Thank you, Darren!

aashaikh (approved):
thanks for the update, LGTM.

PR #841, opened 2023-03-28
Fix ISIS set-metric-type

dplore:
### Change Scope

Deprecate the ISIS set-metric-type as it is using a type which does not meet the intent of the leaf.  Add a new leaf `set-metric-style-type` which references the intended ISIS metric type.

Replaces #820.

### Platform Implementations

 * Cisco XR OS implements [metric-style](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/routing/configuration/guide/rc40xr12k_chapter3.html)
 * JunOS implements a [wide metric opton](https://www.juniper.net/documentation/us/en/software/junos/is-is/topics/ref/statement/metric-edit-protocols-isis.html)


dplore:
@awebsters please consider this replacement which deprecates the leaf with the wrong type assignment.

awebsters:
> @awebsters please consider this replacement which deprecates the leaf with the wrong type of assignment.

This works for us!

release/models/isis/openconfig-isis-policy.yang
```diff
@@ -26,7 +26,13 @@ module openconfig-isis-policy {
     It augments the base routing-policy module with BGP-specific
     options for conditions and actions.";
 
-  oc-ext:openconfig-version "0.5.0";
+  oc-ext:openconfig-version "0.5.1";
```
wenovus, line 29: I think the minor version should be incremented since this is an API change:
http://103.17.196.104.bc.googleusercontent.com/docs/semver/

```
Semver guidelines should be followed while the model is at major number 0, i.e., API or feature changes should increment the minor number, while minor fixes should increment the patch number.
```

I'll create a PR to merge this doc into our release doc for Monday.

release/models/isis/openconfig-isis-policy.yang
```diff
@@ -178,6 +184,15 @@ module openconfig-isis-policy {
 
     leaf set-metric-type {
       type isis-types:level-number;
+      status deprecated;
+      description
+        "Set the type of metric that is to be specified when the
+        set metric leaf is specified.  This leaf is deprecated
+        due to being assigned the wrong type.";
```
wenovus, line 191:
```suggestion
        "Formerly this leaf sets the type of metric that is to be specified
        when the set-metric leaf is specified.  This leaf is DEPRECATED
        due to being assigned the wrong type. Use set-metric-style-type
        instead.";
```

wenovus (commented):
Couple minor comments, otherwise LGTM.

wenovus (approved):

Issue #809, opened 2023-02-08
BGP large communities and send-community leaf


missaesasaya:
Hi all,

I have some questions/suggestions regarding the following leaf:

     /network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/config/send-community

Currently, it has 4 possible values: STANDARD, EXTENDED, BOTH and NONE. But, [RFC 8092](https://www.rfc-editor.org/rfc/rfc8092.html) defines a new type of community: LARGE. From what I could understand, Cisco has redefined BOTH to include all community types, but this prevents the customer to send only large communities or only extended and large communities. Why can't we have send-community changed to a leaf-list or a bits type leaf where we can set any combination of STANDARD, EXTENDED and LARGE values?

Another issue is: the implementation I am working with allows the configuration of send-community per address-family in a peer/peer-group. Is it possible to have this option also? We would keep the current config in peer/peer-group level and add a new send-community leaf inside afi-safi list.

Regards

Missae

robshakir:
hi Missae!

Couple of comments.

1. For adding large communities support, absolutely we can look to make a change here. Generally, we do not use the bits type, but a leaf-list would be a potential option. Here, I'd suggest opening a PR with a proposal - linking to >= 2 different vendor implementations so that we can compare how this maps across different implementations.
2. The answer here is similar -- if this is something that is widely supported then this is a change we can make. The best way to kick off that discussion is to open a PR!

Thanks!
r.

sundaramtirunel:
```
Another issue is: the implementation I am working with allows the configuration of send-community per address-family in a peer/peer-group. Is it possible to have this option also? We would keep the current config in peer/peer-group level and add a new send-community leaf inside afi-safi list.
```
You have mentioned in the description that you want to add a send-community leaf under neighbor and peer-group afi-safi list. I would welcome that move and help us adopt this leaf very easily.
But here, you have added the leaf under global afi-safi list.
Can you move that to neighbor afi-safi list? The grouping name is `bgp-common-mp-afi-safi-config` right under this grouping.

dplore:
@sundaramtirunel note #852 completed the issue regarding LARGE communities.  I recommend starting a new issue regarding "configuration of send-community per address-family in a peer/peer-group. "  if you are interested in pursuing that further?

sundaramtirunel:
Thanks @dplore @missaesasaya for suggestions.
We can open new issue on this.

missaesasaya:
I will close this, as this is being handled in #976

PR #843, opened 2023-03-31
Configuring send-community per AFI-SAFI at neighbor/peer-group (#809)

missaesasaya:
### Change Scope

* Allowing send-community to be configured not only per neighbor/peer-group, but also per AFI-SAFI inside neighbor/peerg-group making the configuration more flexible.

### Platform Implementations

 * Huawei [https://support.huawei.com/enterprise/en/doc/EDOC1100008283/bbae5056/peer-advertise-community]: the peer configuration is done inside the address-family, so the peer-advertise-community configuration is done per peer+afi/safi
 * Juniper [https://www.juniper.net/documentation/en_US/junose15.1/topics/reference/command-summary/neighbor-send-community.html]: Configuration mode indicates the config can be done per address-family
* FRR [https://docs.frrouting.org/en/latest/bgp.html]: Then configuration example show send-community parameter is also configured in address-family mode

### DIFF
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -24,7 +24,14 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "9.2.0";
+  oc-ext:openconfig-version "9.3.0";
+
+  revision "2023-03-31" {
+    description
+      "Allow configuring send-community per AFI-SAFI at
+       neighbor/peer-group.";
+    reference "9.3.0";
+  }
 
   revision "2022-12-12" {
     description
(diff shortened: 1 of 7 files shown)


release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -196,6 +203,16 @@ submodule openconfig-bgp-common-multiprotocol {
         "This leaf indicates whether the AFI-SAFI is enabled for all
         neighbors or groups.";
     }
+
+    leaf send-community {
```
dplore, line 207: This should be `send-community-type`.  Also, similar to #852 ,  should this be a list of types?  Or is only a single type supported?

missaesasaya: I have made this two separate PRs because I figured they were two different issues: one thing is we wanted to make this configuration available also per address-family and this would be a non-breaking change as it is only a new config. While  #852 would change an existing configuration. But, maybe I can merge both changes in one PR.

missaesasaya: @dplore Can I close this PR and leave only #852? The changes are included there anyway.

dplore: Yes you can do this.   I will go ahead and close this one for you for clarity.

dplore:
Closed in favor of #852

PR #856, opened 2023-04-13
Configure BFD min interval, multipleir when activated in protocols

oscargdd:
### Change Scope
Min-tx-interval, required-minimum-receive and multiplier added for the case when BFD is activated at a protocol

### Platform Implementations

 * Cisco: [DOC](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-2/interfaces/command/reference/b_interfaces_cr42xr12k_chapter_010.html)
 * Juniper [DOC](https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/topic-map/bfd-for-bgp-session.html)
 * Huawei: [DOC](https://support.huawei.com/enterprise/en/doc/EDOC1000178171/14038c9c/example-for-configuring-bfd-for-bgp )
 
### Examples  
Cisco
```
neighbor 10.10.10.2
     remote-as 65000
     bfd fast-detect
     bfd multiplier 3
     bfd minimum-interval 100
```
Juniper
```
neighbor 12.12.12.12 {
    description DESCRIPTION;
    import RP_BGP_IMPORT;
    authentication-key KEY;
    export RP_BGP_EXPORT;
    peer-as 65028;
    bfd-liveness-detection {
        minimum-interval 200
        multiplier 4
    }
}
```
Huawei
```
peer 10.10.10.10 as-number 65000
   peer 10.10.10.10 description DESCRIPTION
   peer 10.10.10.10 password PASSWORD
   peer 10.10.10.10 bfd min-tx-interval 200 min-rx-interval 200
   peer 10.10.10.10 bfd enable
```

### DIFF
--- a/release/models/bfd/openconfig-bfd.yang
+++ b/release/models/bfd/openconfig-bfd.yang
@@ -215,43 +215,7 @@ module openconfig-bfd {
         interface.";
     }
 
-    leaf desired-minimum-tx-interval {
-      type uint32;
-      units microseconds;
-      description
-        "The minimum interval between transmission of BFD control
-        packets that the operator desires. This value is advertised to
-        the peer, however the actual interval used is specified by
-        taking the maximum of desired-minimum-tx-interval and the
-        value of the remote required-minimum-receive interval value.
-
-        This value is specified as an integer number of microseconds.";
-    }
-
-    leaf required-minimum-receive {
-      type uint32;
-      units microseconds;
-      description
-        "The minimum interval between received BFD control packets that
-        this system should support. This value is advertised to the
-        remote peer to indicate the maximum frequency (i.e., minimum
-        inter-packet interval) between BFD control packets that is
-        acceptable to the local system.";
-    }
-
-    // rjs: Could have required-minimum-echo-receive here, but this is
-    // generally not configurable.
-
-    leaf detection-multiplier {
-      type uint8 {
-        range "1..max";
-      }
-      description
-        "The number of packets that must be missed to declare this
-        session as down. The detection interval for the BFD session
-        is calculated by multiplying the value of the negotiated
-        transmission interval by this value.";
-    }
+    uses bfd-configuration;
 
     leaf enable-per-member-link {
       type boolean;
@@ -701,8 +665,9 @@ module openconfig-bfd {
         the peer, however the actual interval used is specified by
         taking the maximum of desired-minimum-tx-interval and the
         value of the remote required-minimum-receive interval value.
-
-        This value is specified as an integer number of microseconds.";
+        This value is specified as an integer number of microseconds.
+        The value 0 is reserved and cannot be used.";
+      reference "section 4.1 of RFC 5880";
     }
 
     leaf required-minimum-receive {
@@ -714,6 +679,7 @@ module openconfig-bfd {
         remote peer to indicate the maximum frequency (i.e., minimum
         inter-packet interval) between BFD control packets that is
         acceptable to the local system.";
+      reference "section 4.1 of RFC 5880";
     }
 
     leaf detection-multiplier {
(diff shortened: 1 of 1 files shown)


release/models/bfd/openconfig-bfd.yang
```diff
@@ -682,6 +689,46 @@ module openconfig-bfd {
...
+      units microseconds;
+      description
+        "The minimum interval between transmission of BFD control
+        packets that the operator desires. This value is advertised to
+        the peer, however the actual interval used is specified by
+        taking the maximum of desired-minimum-tx-interval and the
+        value of the remote required-minimum-receive interval value.
+
+        This value is specified as an integer number of microseconds.";
+    }
```
dplore, line 706: Are these leaves defined in RFC? Can you add a reference to the BFD rfc and section?

release/models/bfd/openconfig-bfd.yang
```diff
@@ -682,6 +689,46 @@ module openconfig-bfd {
...
+    leaf required-minimum-receive {
+      type uint32;
+      units microseconds;
+      description
+        "The minimum interval between received BFD control packets that
+        this system should support. This value is advertised to the
+        remote peer to indicate the maximum frequency (i.e., minimum
+        inter-packet interval) between BFD control packets that is
+        acceptable to the local system.";
+    }
```
dplore, line 717: I didn't see this leaf used in the example configs.  Is it supported widely?

dplore (commented):
Couple of questions for you

dplore:
@s19nal to review

release/models/bfd/openconfig-bfd.yang
s19nal: just one quick example from cisco, when bfd configured under an interface:
 bfd interval 999 min_rx 999 multiplier 4

release/models/bfd/openconfig-bfd.yang
oscargdd: Hi @s19nal . You are right, I can do the change you propose and reuse the grouping also from the original place.  My first intention was just that these leaves do make sense when BFD is configured at protocol level beyond current enable. And then, if these are ok, make the necessary optimizations. 

So, @dplore @s19nal , let me add the references and reuse the grouping

oscargdd:
New version uploaded following @dplore and @s19nal comments

release/models/bfd/openconfig-bfd.yang
```diff
@@ -208,43 +215,7 @@ module openconfig-bfd {
...
-      units microseconds;
-      description
-        "The minimum interval between received BFD control packets that
-        this system should support. This value is advertised to the
-        remote peer to indicate the maximum frequency (i.e., minimum
-        inter-packet interval) between BFD control packets that is
-        acceptable to the local system.";
-    }
-
-    // rjs: Could have required-minimum-echo-receive here, but this is
```
jhaas-pfrc, line 235: Echo state is present in RFC 9314.  Useful even if configuration state for it is not done.  Clearly that'd need to be split from the config grouping.

s19nal: Ideally this could be picked up as a new PR?   Getting the config stuff merged would be a good first step :)

s19nal:
@oscargdd this all LGTM now sorry it has taken an age to get round to reviewing!

release/models/bfd/openconfig-bfd.yang
dplore: It seems required-minimum-echo-receive  is rarely if ever supported .  Seems ok for a separate PR to address this.

This change removes what is essentially a 'what-if' or TODO comment.    This seems ok as well, as it's not a critical operational issue (given lack of implementation?)

dplore:
@oscargdd please include a pyang rendering of the tree, and resolve merge conflicts.

s19nal (approved):

PR #517, opened 2021-10-29
Add atomic annotations to containers in openconfig-aft.

drautio:
On branch aft-atomic-508
 Changes to be committed:
	modified:   openconfig-aft-common.yang
	modified:   openconfig-aft-ethernet.yang
	modified:   openconfig-aft-ipv4.yang
	modified:   openconfig-aft-ipv6.yang
	modified:   openconfig-aft-mpls.yang
	modified:   openconfig-aft-pf.yang

Add the atomic attribute to the above aft yang containers.

PR #860, opened 2023-04-19
Add atomic annotations to containers in openconfig-aft.

michaelore:
Refreshing #517 as the author is no longer available to complete the request.

#Operational use case 
The expectation is that for any AFT entity is only complete when it includes  IPv4/IPv6/MPLS/Ethernet, Conditional Next Hop Group (CNHG), Next Hop Group (NHG), or Next Hop (NH).  AFT entities must always be updated as a whole with all attributes of the entity in a single message.  This is required to preserve the integrity of the AFT table stored on a client by preventing fragmented or partial AFT entities during streaming of updates.

This change adds the atomic attribute to AFT lists for IPv4/IPv6/MPLS/Ethernet/PolicyBasedForwarding, Conditional Next Hop Group (CNHG), Next Hop Group (NHG) and Next Hop (NH).

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -424,6 +424,7 @@ submodule openconfig-aft-common {
       list next-hop-group {
         key "id";
 
+        oc-ext:telemetry-atomic;
         description
           "An individual set of next-hops grouped into a common group.
           Each entry within an abstract forwarding table points to a


michaelore:
@dplore This should be ready to review. FYI the CLA check is tripping because I included Dan's prior commits and the check doesn't recognize him anymore. From internal resources sounds like there are ways to override it, but waiting until the PR is otherwise ready to merge.

dplore:
Hi @michaelore , please amend this PR or submit a fresh one so the CLA is intact.

Issue #836, opened 2023-03-22
Wifi-phy RF neighbors RSSI does not indicate that it is negative number representation


jsnyder81:
The description of RSSI in the wifi-phy neighbor-list-state grouping does not include the same language as other leafs.  Specifically that RSSI should be represented as a negative number.

neighbor-list-state RSSI:
https://github.com/openconfig/public/blob/master/release/models/wifi/openconfig-wifi-phy.yang#L387

Example: client-rf-state
https://github.com/openconfig/public/blob/master/release/models/wifi/openconfig-wifi-mac.yang#L1334

I will be submitting a PR for this change.

PR #862, opened 2023-04-25
#836 fixing description of neighbor RSSI

jsnyder81:
### Change Scope

* Updating description of the wifi-phy neighbor RSSI state leaf to be more descriptive and match other references to RSSI to include "represented as a negative number."
* This update is backwards compatible, clarifying description update only.


mike-albano (approved):
LGTM

PR #866, opened 2023-05-02
Fix openconfig-icmpv6-types module prefix

missaesasaya:
### Change Scope

* Fix openconfig-icmpv6-types module prefix

### Platform Implementations

 * n/a


dplore (approved):
Thank you for this bug fix!

PR #868, opened 2023-05-03
Allow 4-byte value of transceiver vendor-rev

nkitchen:
SFF-8472, the spec for the management interface of SFP+ transceivers, specifies that the vendor revision number is a 4-byte field (Section 7.4), not a 2-byte field as in the SFF-8436 spec for QSFP+ and the CMIS spec.

openconfig#347

### Change Scope

* This change expands `vendor-rev`'s length from `1..2` to `1..2 | 4`.
* This change is backwards-compatible.
### Platform Implementations

 * Arista EOS shows the vendor revision via CLI `show inventory`.


nkitchen:
I have never seen a transceiver spec that specified 3 bytes for the revision field. But I haven't checked all the specs that are out there, either.

Maybe the vendor always used 3 bytes of the 4 available, with a single byte of padding.

Anyway, I have no objection to using the length `1..4` instead. It's simpler and still includes all the meaningful possibilities.

ejbrever:
Thanks, I'm thinking 1..4 sounds simplest to me unless there are any objections.

ejbrever (approved):
LGTM

dplore (approved):

PR #869, opened 2023-05-04
removing unused LOG_DESTINATION_TYPE identities

bensons:
### Change Scope

 * removing LOG_DESTINATION_TYPE identities, specifically:
    * DEST_CONSOLE, DEST_BUFFER, DEST_FILE, DEST_REMOTE
 * because they are unused in any known models

### Platform Implementations

 * None. That's the point. :)

### DIFF
--- a/release/models/system/openconfig-system-logging.yang
+++ b/release/models/system/openconfig-system-logging.yang
@@ -25,6 +25,12 @@ module openconfig-system-logging {
 
   oc-ext:openconfig-version "0.5.0";
 
+  revision "2023-05-04" {
+    description
+      "removing LOG_DESTINATION_TYPE identities";
+    reference "0.5.0";
+  }
+
   revision "2022-12-29" {
     description
       "Add network-instance for remote logging servers";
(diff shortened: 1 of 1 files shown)


bensons:
@aashaikh PTAL

aashaikh:
Thanks, this looks consistent with our offline discussion.  Pls also check w/ @dplore about how we are currently handling changes like this to pre v1.0 models.

dplore:
> ### Change Scope
> * removing LOG_DESTINATION_TYPE identities, specifically:
>   
>   * DEST_CONSOLE, DEST_BUFFER, DEST_FILE, DEST_REMOTE
> * because they are unused in any known models
> 
> ### Platform Implementations
> * None. That's the point. :)

Minor nit: implementations is intended to refer to network operating systems, which of course do support console and remote logging.   Agreed however that these identities are not used and also not needed because one can configure the destination of syslog using the following paths.

* `/openconfig-access-points:access-points/access-point/system/logging/console`
* `/openconfig-access-points:access-points/access-point/system/logging/remote-servers/remote-server`

There are open items for syslog modeling, but it is not clear they need such identities either.

dplore:
> Thanks, this looks consistent with our offline discussion. Pls also check w/ @dplore about how we are currently handling changes like this to pre v1.0 models.

We follow semver.org guidelines which says pre v1.0 may be revised with breaking changes and not increment the major version number.

PR #878, opened 2023-05-24
Update openconfig-network-instance-l3.yang to change alarm-threshold value

shasingh27:
Update openconfig-network-instance-l3.yang to change alarm-threshold value from absolute to percentage

I have taken a look at JUNOS implementation and it is percentage there.

I also checked other implementations
 Took a look at Arista EOS and they support the following...
 
 a1(config-router-bgp)#neighbor 1.1.1.1 maximum-routes 10 warning-limit ?
 <0-4294967294 Maximum number of routes after which a warning is issued (0 means never warn)
 <1-100 Percentage of maximum number of routes at which to warn ( 1-100 )

———————————————
I think Cisco’s implementation also takes threshold as percentage. I could find following document for bgp maximum-prefix feature.
 
https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5500/bgp/b-ncs5500-bgp-cli-reference/b-ncs5500-bgp-cli-reference_chapter_01.html#wp6592192440
 
And for nokia sros/srlinux
https://documentation.nokia.com/html/0_add-h-f/93-0074-HTML/7750_SR_OS_Routing_Protols_Guide/BGP-CLI.html#1124424
 
As threshold is relative to the configured max, I think it makes sense to be a % value rather than an absolute value.

https://github.com/openconfig/public/blob/master/doc/external-contributions-guide.md]

### Change Scope

Update openconfig-network-instance-l3.yang to change alarm-threshold value from absolute to percentage

It is not backward compatible
### Platform Implementations

   implementation output.
   implementation output.

### DIFF
--- a/release/models/network-instance/openconfig-network-instance-l3.yang
+++ b/release/models/network-instance/openconfig-network-instance-l3.yang
@@ -23,7 +23,13 @@ module openconfig-network-instance-l3 {
     Layer 3 network instance configuration and operational state
     parameters.";
 
-  oc-ext:openconfig-version "2.0.0";
+  oc-ext:openconfig-version "3.0.0";
+  
+  revision "2023-05-17" {
+    description
+      "Changing alarm-threshold type from absolute to percentage";
+    reference "2.0.0";
+  }
 
   revision "2022-11-08" {
     description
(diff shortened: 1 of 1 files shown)


dplore:
Hi @shasingh27 would you like to continue to pursue this PR?

The change looks reasonable in that there are implementations out there that support using a percentage.  Is there an operational use case for why percentage is more useful than an integer?  On the surface it seems the two are equivalent functionally.   An implementation can trivially map the OC configured integer to a percentage.

For some other areas I sometimes I have seen there is an operational ease to using a percentage, for example where queue depth / buffer size is involved on different speed interfaces.  But this isn't that same kind of scenario where the 'maximum' is highly variable and hardware defined.

PR #879, opened 2023-05-25
Add `gre` and `mpls` containers under next-hops aft entry state. Add `ttl` and `tos` under gre, mpls and ip-in-ip aft entry state for telemetry.

romeyod:
* (M) aft/openconfig-aft-ethernet.yang
* (M) aft/openconfig-aft-ipv4.yang
* (M) aft/openconfig-aft-ipv6.yang
* (M) aft/openconfig-aft-mpls.yang
* (M) aft/openconfig-aft-pf.yang
* (M) aft/openconfig-aft-state-synced.yang
* (M) aft/openconfig-aft.yang

Change Scope
  - Add gre and mpls containers under next-hops aft entry state.
  - Add ttl and tos under gre, mpls and ip-in-ip aft entry state for telemetry.

Platform Implementations
  - Arista [documentation](https://www.arista.com/en/um-eos/eos-nexthop-groups#xx1147395)

Relevant pyang output:
```
  +--rw network-instances
        +--ro afts
        |  +--ro next-hops
        |     +--ro next-hop* [index]
        |        +--ro index            -> ../state/index
        |        +--ro state
        |        |  +--ro index?                       uint64
        |        |  +--ro programmed-index?            uint64
        |        |  +--ro ip-address?                  oc-inet:ip-address
        |        |  +--ro mac-address?                 oc-yang:mac-address
        |        |  +--ro pop-top-label?               boolean
        |        |  +--ro pushed-mpls-label-stack*     oc-mplst:mpls-label
        |        |  +--ro encapsulate-header?          oc-aftt:encapsulation-header-type
        |        |  +--ro decapsulate-header?          oc-aftt:encapsulation-header-type
        |        |  +--ro origin-protocol?             identityref
        |        |  +--ro lsp-name?                    string
        |        |  +--ro counters
        |        |  |  +--ro packets-forwarded?   oc-yang:counter64
        |        |  |  +--ro octets-forwarded?    oc-yang:counter64
        |        |  +--ro vni-label?                   oc-evpn-types:evi-id
        |        |  +--ro tunnel-src-ip-address?       oc-inet:ip-address
        |        |  +--ro oc-aftni:network-instance?   oc-ni:network-instance-ref
        |        +--ro ip-in-ip
        |        |  +--ro state
        |        |     +--ro src-ip?   oc-inet:ip-address
        |        |     +--ro dst-ip?   oc-inet:ip-address
        |        |     +--ro ttl?      uint32
        |        |     +--ro tos?      uint16
        |        +--ro gre
        |        |  +--ro state
        |        |     +--ro ttl?   uint32
        |        |     +--ro tos?   uint16
        |        +--ro mpls
        |        |  +--ro state
        |        |     +--ro ttl?   uint32
        |        |     +--ro tos?   uint16
        |        +--ro interface-ref
        |           +--ro state
        |              +--ro interface?      -> /oc-if:interfaces/interface/name
        |              +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index 
```

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -419,7 +419,7 @@ submodule openconfig-aft-common {
       "Outer header fields of encapsulation applied on a next-hop";
 
     leaf ttl {
-      type uint32;
+      type uint8;
       description
         "This leaf reflects the configured TTL value that is used in the outer
          header during packet encapsulation. When this leaf is not set, the TTL
@@ -428,7 +428,7 @@ submodule openconfig-aft-common {
     }
 
     leaf tos {
-      type uint16;
+      type uint8;
       description
         "This leaf reflects the configured TOS value that is used in the outer
          header during packet encapsulation. When this leaf is not set, the TOS
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,14 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "2.3.0";
+  oc-ext:openconfig-version "2.4.0";
+
+  revision "2023-05-25" {
+    description
+      "Add next-hop-group-name in NHG AFT entry state and
+       Add ttl and tos in NH AFT entry state";
+    reference "2.4.0";
+  }
 
   revision "2023-04-19" {
     description
@@ -359,6 +366,18 @@ submodule openconfig-aft-common {
          specified network-instance.";
     }
 
+    leaf ttl {
+      type uint32;
+      description
+        "TTL used when forwarding a packet to the specified next-hop.";
+    }
+
+    leaf tos {
+      type uint16;
+      description
+        "TOS used when forwarding a packet to the specified next-hop.";
+    }
+
     uses aft-common-install-protocol;
   }
 
@@ -577,6 +596,15 @@ submodule openconfig-aft-common {
         identifier. If the identifier of the next-hop-group
         is changed, all AFT entries that reference it must
         also be updated.";
+    }
+
+    leaf next-hop-group-name {
+      type string;
+      description
+       "Where applicable, this leaf is a unique identifier string for the
+        next-hop-group. This leaf's type is able to accommodate more
+        permutations of identifiers that vendors may have and want to
+        expose.";
     }
 
     leaf programmed-id {
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -421,19 +421,19 @@ submodule openconfig-aft-common {
     leaf ttl {
       type uint8;
       description
-        "This leaf reflects the configured TTL value that is used in the outer
-         header during packet encapsulation. When this leaf is not set, the TTL
-         value of the inner packet is copied over as the outer packet's TTL
-         value during encapsulation.";
+        "This leaf reflects the configured/default TTL value that is used in the
+         outer header during packet encapsulation. When this leaf is not set,
+         the TTL value of the inner packet is copied over as the outer packet's
+         TTL value during encapsulation.";
     }
 
     leaf tos {
       type uint8;
       description
-        "This leaf reflects the configured TOS value that is used in the outer
-         header during packet encapsulation. When this leaf is not set, the TOS
-         value of the inner packet is copied over as the outer packet's TOS
-         value during encapsulation.";
+        "This leaf reflects the configured/default TOS value that is used in the
+         outer header during packet encapsulation. When this leaf is not set,
+         the TOS value of the inner packet is copied over as the outer packet's
+         TOS value during encapsulation.";
     }
   }
 
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -27,8 +27,9 @@ submodule openconfig-aft-common {
 
   revision "2023-05-25" {
     description
-      "Add next-hop-group-name and auto-size in NHG AFT entry state
-       and Add ttl and tos in NH AFT entry state for telemetry";
+      "Add gre and mpls containers under next-hops aft entry state.
+       Add ttl and tos under gre, mpls and ip-in-ip aft entry state
+       for telemetry.";
     reference "2.4.0";
   }
 
@@ -211,6 +212,33 @@ submodule openconfig-aft-common {
             description
               "State parameters relating to IP-in-IP encapsulation.";
             uses aft-common-entry-nexthop-ipip-state;
+            uses aft-common-entry-nexthop-encap-outer-header-state;
+          }
+        }
+
+        container gre {
+          description
+            "When specified, the packet has an GRE header applied to it before
+             forwarding to the specified next-hop.";
+
+          container state {
+            config false;
+            description
+              "State parameters relating to GRE encapsulation.";
+            uses aft-common-entry-nexthop-encap-outer-header-state;
+          }
+        }
+
+        container mpls {
+          description
+            "When specified, the packet has an MPLS header applied to it before
+             forwarding to the specified next-hop.";
+
+          container state {
+            config false;
+            description
+              "State parameters relating to MPLS encapsulation.";
+            uses aft-common-entry-nexthop-encap-outer-header-state;
           }
         }
 
@@ -366,20 +394,6 @@ submodule openconfig-aft-common {
          specified network-instance.";
     }
 
-    leaf ttl {
-      type uint32;
-      description
-        "The TTL value used by the hardware pipeline while forwarding
-         a packet to the specified next-hop.";
-    }
-
-    leaf tos {
-      type uint16;
-      description
-        "The TOS value used by the hardware pipeline while forwarding
-         a packet to the specified next-hop.";
-    }
-
     uses aft-common-install-protocol;
   }
 
@@ -397,6 +411,29 @@ submodule openconfig-aft-common {
       type oc-inet:ip-address;
       description
         "Destination IP address to use for the encapsulated packet.";
+    }
+  }
+
+  grouping aft-common-entry-nexthop-encap-outer-header-state {
+    description
+      "Outer header fields of encapsulation applied on a next-hop";
+
+    leaf ttl {
+      type uint32;
+      description
+        "This leaf reflects the configured TTL value that is used in the outer
+         header during packet encapsulation. When this leaf is not set, the TTL
+         value of the inner packet is copied over as the outer packet's TTL
+         value during encapsulation.";
+    }
+
+    leaf tos {
+      type uint16;
+      description
+        "This leaf reflects the configured TOS value that is used in the outer
+         header during packet encapsulation. When this leaf is not set, the TOS
+         value of the inner packet is copied over as the outer packet's TOS
+         value during encapsulation.";
     }
   }
 
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -28,7 +28,7 @@ submodule openconfig-aft-common {
   revision "2023-05-25" {
     description
       "Add next-hop-group-name in NHG AFT entry state and
-       Add ttl and tos in NH AFT entry state";
+       Add ttl and tos in NH AFT entry state for telemetry";
     reference "2.4.0";
   }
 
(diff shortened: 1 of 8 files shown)


release/models/aft/openconfig-aft-common.yang
```diff
@@ -579,6 +598,15 @@ submodule openconfig-aft-common {
         also be updated.";
     }
 
+    leaf next-hop-group-name {
+      type string;
+      description
+       "Where applicable, this leaf is a unique identifier string for the
+        next-hop-group. This leaf's type is able to accommodate more
+        permutations of identifiers that vendors may have and want to
+        expose.";
```
robshakir, line 607: I'm not quite sure how to parse this -- I understand:

* this is an arbitrary name for the group that is implementation specific

but, it's not clear what "permutation of identifiers" means. What kind of different types of identifiers are we trying to expose?

If it's something that has an enumerated set of values, rather than modelling like this, should we rather model this as an `identity` where each vendor can extend the base identity with enumerated values that have clear definitions?

robshakir (commented):
are these new fields intended for programming or solely telemetry? It'd be good to understand this.

added some questions in-line to clarify some of the new fields.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -359,6 +366,18 @@ submodule openconfig-aft-common {
...
+    leaf ttl {
+      type uint32;
+      description
+        "TTL used when forwarding a packet to the specified next-hop.";
+    }
+
+    leaf tos {
+      type uint16;
+      description
+        "TOS used when forwarding a packet to the specified next-hop.";
```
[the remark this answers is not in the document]
romeyod: My understanding is TOS [Type of Service] is one of the ways for QoS implementation. Can you please clarify/expand your comment further?

romeyod:
> 
> are these new fields intended for programming or solely telemetry? It'd be good to understand this.
> 
> added some questions in-line to clarify some of the new fields.

Solely for a Telemetry based use case.

release/models/aft/openconfig-aft-common.yang
romeyod: My bad. "permutation of identifiers" made sense in my head.

Its like you said an arbitrary name for the group, that is supported by vendors

release/models/aft/openconfig-aft-common.yang
```diff
@@ -359,6 +366,18 @@ submodule openconfig-aft-common {
          specified network-instance.";
     }
 
+    leaf ttl {
+      type uint32;
+      description
+        "TTL used when forwarding a packet to the specified next-hop.";
```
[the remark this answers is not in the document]
romeyod: Yes, its the value that is associated with the Next-Hop. I will update the description

release/models/aft/openconfig-aft-common.yang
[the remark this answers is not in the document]
robshakir: My concern here is that the TOS might not be static for a specific forwarding entry? For example, let's say I have a route 1.1.1.1/32 to some downstream 2.2.2.2 destination, this route is going to be used for any TOS/DSCP marking. In this case, what does this leaf say? If the routing is TOS-specific, then this should be part of the policy-forwarding AFT.

Let's imagine that it's for all TOS input values -- but then I configure a policer that says above 1pps, every packet should be marked with TOS = 8, now, what does this leaf say? It seems like it should either be "the original TOS value OR 8", which isn't something that we can represent here.

release/models/aft/openconfig-aft-common.yang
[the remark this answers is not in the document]
robshakir: I'm not sure that this makes sense to me either -- if the input TTL is 42, then this value would be 41, but if it's 64, then it's 63 if we're decrementing TTL I would assume. In what scenario is TTL statically set (I assume its where we're rewriting or pushing a new encapsulation to the packet).

release/models/aft/openconfig-aft-common.yang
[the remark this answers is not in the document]
dplore: I think more description of the motivation for this leaf (and the others in this PR) is needed.  You can for example refer to specific features that require these leaves.  Is there some feature that requires all traffic to a next hop to be marked with a given TOS value?  If yes, why isn't this QoS policy?  

I browsed the reference documentation quickly.  I see there is some feature for configuring a TOS value for tunneled traffic.  Is that the motivation for this leaf?   Tunnel TOS marking should be implemented as QoS configuration I think.  As an operator if I want to see what TOS would be set for the tunnel, I'd need to inspect the qos policy.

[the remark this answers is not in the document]
romeyod: > I browsed the reference documentation quickly. I see there is some feature for configuring a TOS value for tunneled traffic. Is that the motivation for this leaf? Tunnel TOS marking should be implemented as QoS configuration I think. As an operator if I want to see what TOS would be set for the tunnel, I'd need to inspect the qos policy.
@dplore can you point me to the ` feature for configuring a TOS value for tunneled traffic`

For QoS policy model, I am looking at https://github.com/openconfig/public/tree/master/release/models/qos
Let me know if there is some other documentation that you were looking at.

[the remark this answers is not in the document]
dplore: > 
> @dplore can you point me to the ` feature for configuring a TOS value for tunneled traffic`

https://www.arista.com/en/um-eos/eos-nexthop-groups#xx1147395


> For QoS policy model, I am looking at https://github.com/openconfig/public/tree/master/release/models/qos Let me know if there is some other documentation that you were looking at.

That is the model.  The [OpenConfig QoS guide](https://www.openconfig.net/docs/models/qos/) is probably helpful.  I would expect to have a tunnel exposed as an OpenConfig subinterface and then apply a qos policy which sets ToS.

release/models/aft/openconfig-aft-common.yang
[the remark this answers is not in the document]
dplore: One can configure and set ttl on a tunnel here:
`https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-interfaces.html#interfaces-interface-tunnel-config-ttl`

It seems like if this is configured, one could expose it in aft where the nexthop is the relevant tunnel interface

release/models/aft/openconfig-aft-common.yang
dplore: Looks like this is no longer present in the PR, marking resolved.

dplore (changes requested):
@romeyod could you clarify which feature you are supporting?  

Please include how one would configure this feature (using existing or new OC as required, or if not supported in OC, the relevant vendor CLI ) to demonstrate when these AFT fields would be populated.

romeyod:
> 

@dplore We want to support next-hop-group state and next-hop content streaming for telemetry. For full view of the encapsulation header, streaming TTL and TOS field along with the already existing AFT state.

I did not see any existing OC configs for this. Example: https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-aft.html#mod-openconfig-aft

Arista [documentation](https://www.arista.com/en/um-eos/eos-nexthop-groups#xx1147395) for the CLI

Example CLI config snippet:
```
switch(config)# nexthop-group NH-1
switch(config-nexthop-group-NH-1)# size 4
switch(config-nexthop-group-NH-1)# entry 0 tunnel-destination 10.13.4.4
switch(config-nexthop-group-NH-1)# entry 1 tunnel-destination 10.15.4.22
switch(config-nexthop-group-NH-1)# entry 2 tunnel-destination 10.15.5.37
switch(config-nexthop-group-NH-1)# show active
 nexthop-group NH-1
   size 4
   ttl 64
   entry 0 tunnel-destination 10.13.4.4
   entry 1 tunnel-destination 10.15.4.22
   entry 2 tunnel-destination 10.15.5.37
switch(config-nexthop-group-NH-1)#
```

Let me know if I didnt understand your query correctly. Thanks

release/models/aft/openconfig-aft-common.yang
```diff
@@ -379,6 +414,29 @@ submodule openconfig-aft-common {
     }
   }
 
+  grouping aft-common-entry-nexthop-encap-outer-header-state {
+    description
+      "Outer header fields of encapsulation applied on a next-hop";
+
+    leaf ttl {
+      type uint32;
```
krishna-juniper, line 422: Why ttl is using 32 bits?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -379,6 +414,29 @@ submodule openconfig-aft-common {
...
+         header during packet encapsulation. When this leaf is not set, the TTL
+         value of the inner packet is copied over as the outer packet's TTL
+         value during encapsulation.";
+    }
+
+    leaf tos {
+      type uint16;
+      description
+        "This leaf reflects the configured TOS value that is used in the outer
+         header during packet encapsulation. When this leaf is not set, the TOS
```
krishna-juniper, line 434: Depending on the implementation there can be 3 ways:
1. Setting the value of tos by configuration.
2. Copy from inner payload.
3. Setting default tos.
The proposed changes are catering to 1 and 2 of the above ways, please accommodate all possible ways.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -204,6 +212,33 @@ submodule openconfig-aft-common {
...
+            config false;
+            description
+              "State parameters relating to GRE encapsulation.";
+            uses aft-common-entry-nexthop-encap-outer-header-state;
+          }
+        }
+
+        container mpls {
+          description
+            "When specified, the packet has an MPLS header applied to it before
```
krishna-juniper, line 234: If multiple labels are getting added and each label has its own tos/ttl values then what should be the values?
Should these mpls specific leaves tos/ttl kept under "pushed-mpls-label-stack"?

release/models/aft/openconfig-aft-common.yang
romeyod: IIUC your question: This path is for telemetry only. So we are not setting any values, its the value that is going to be used. So if we had a default value configured [or cause nothing is configured], then this will be a non-zero value that will be streamed, so in that case, 1 and 3 are the same here from a telemetry prespective.

Let me know if I misunderstood.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -379,6 +414,29 @@ submodule openconfig-aft-common {
...
 
+  grouping aft-common-entry-nexthop-encap-outer-header-state {
+    description
+      "Outer header fields of encapsulation applied on a next-hop";
+
+    leaf ttl {
+      type uint32;
+      description
+        "This leaf reflects the configured TTL value that is used in the outer
+         header during packet encapsulation. When this leaf is not set, the TTL
```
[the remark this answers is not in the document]
krishna-juniper: To be more precise please update the description as below:
 "This leaf reflects the configured TTL value" as "This leaf reflects the configured/default TTL value"

release/models/aft/openconfig-aft-common.yang
krishna-juniper: To be more precise please update the description as below:
 "This leaf reflects the configured TOS value" as "This leaf reflects the configured/default TOS value"

also if default TOS value is zero, zero will be streamed out.
is this the expectation?

release/models/aft/openconfig-aft-common.yang
krishna-juniper: @romeyod any comments on this suggestion?

release/models/aft/openconfig-aft-common.yang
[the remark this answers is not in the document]
romeyod: Updated documentation.

release/models/aft/openconfig-aft-common.yang
romeyod: Updated documentation. 

> also if default TOS value is zero, zero will be streamed out.
is this the expectation?
Yes.

release/models/aft/openconfig-aft-common.yang
romeyod: @krishna-juniper Thanks for the suggestion

`pushed-mpls-label-stack` is of type `oc-mplst:mpls-label` which is also used in other models. Example: `popped-mpls-label-stack`, `start-label-value` [more occurences](https://github.com/search?q=repo%3Aopenconfig%2Fpublic%20mpls-label&type=code)

Instead of adding ttl/tos to `oc-mplst:mpls-label` and affecting unrelated models, we can do the following:
define a `typedef encap-outer-header-state` with ttl and tos 
*and* 
in place of mpls container, under `next-hop/state` at the same level as `pushed-mpls-label-stack` we define 
```
leaf-list pushed-mpls-label-header-state {
    type encap-outer-header-state;
}
```
This new `leaf-list pushed-mpls-label-header-state` will have ordered entries with tos/ttl corresponding to `pushed-mpls-label-stack` 

Let me know what you think about this proposal.

krishna-juniper: The approach assumes implicit correlation between labels and ttl/tos. (labels coming from pushed-mpls-label-stack, ttl/tos coming from pushed-mpls-label-header-state).
By adding label stack as part of new container(container for encap-outer-header-state + pushed labels) we can avoid this implicit correlation at the cost of redundant information.
what are your thoughts on this approach?

robshakir: Is this approach supported on any system? Please can we provide _multi-vendor_ examples if we are going to make such a change?

Note that changes around how the MPLS label stack is being pushed have implications in gRIBI and are backwards compatible.  I'd like to make sure we're supporting a use case that is actually supported here, since it also has overhead for those that don't care about this.

I also don't really like this typedef being a single leaf -- i think we rather need to think about how this is actually user friendly to parse.

PR #881, opened 2023-05-25
Update WiFi models field descriptions

xavier-contreras:
### Change Scope
Some of the fields for the WiFi models are fairly ambiguous which has led to incorrect implementations or follow up questions before implementation. This fixes that. No structural changes to the models happening as part of this PR.

### Platform Implementations

N/A : No structural changes to the models happening as part of this PR


mike-albano (approved):
LGTM, thanks.

PR #883, opened 2023-06-06
Refactor BGP community sets

dplore:
### Change Scope

* Deprecate regex string based community and extended community types
* Add containers for standard, 2 byte community type and strongly typed leafs
* Add containers for each extended community type and strongly typed leafs for each

### Platform Implementations

 * Cisco IOS XR implements a command like [set extcommunity bandwidth (2906:45750000)](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/asr9k-r6-4/routing/configuration/guide/b-routing-cg-asr9000-64x/b-routing-cg-asr9000-64x_chapter_01111.html#concept_xsj_2q1_pdb
) with a user specified bandwidth in a route policy similar to this PR.

 * JunOS implements link bandwidth via configuration of the [bgp community with a user specified bandwidth](https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/topic-map/load-balancing-bgp-session.html#d29e113__d29e120) in a route policy list similar to this PR

* [Arista EOS](https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1418621) enables link bandwidth using a per neighbor command.   Bandwidths can optionally be set using a [route-map](https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1153691) and a percent format.  This is similar to this PR, but uses % bandwidth instead of a bandwidth in bits/second.

* [FRR](https://docs.frrouting.org/en/latest/bgp.html#weighted-ecmp-using-bgp-link-bandwidth) implements link bandwidth via configuration applied to a policy.  Bandwidth is automatically calculated and not configurable.

### Tree view
```diff
module: openconfig-routing-policy
  +--rw routing-policy
     +--rw defined-sets
     |  +--rw oc-bgp-pol:bgp-defined-sets
     |     +--rw oc-bgp-pol:community-sets        (changes for standard community sets)

9323c9323
<      |     |     |  +--rw oc-bgp-pol:community-member*     union
---
>      |     |     |  x--rw oc-bgp-pol:community-member*     union     (<-- now deprecated)
9326,9328c9326,9347
<      |     |        +--ro oc-bgp-pol:community-set-name    string
<      |     |        +--ro oc-bgp-pol:community-member*     union
<      |     |        x--ro oc-bgp-pol:match-set-options?    oc-pol-types:match-set-options-type
---
>      |     |     |  +--ro oc-bgp-pol:community-set-name    string
>      |     |     |  x--ro oc-bgp-pol:community-member*     union
>      |     |     |  x--ro oc-bgp-pol:match-set-options?    oc-pol-types:match-set-options-type
>      |     |     +--rw oc-bgp-pol:members
>      |     |        +--rw oc-bgp-pol:member* [index]
>      |     |           +--rw oc-bgp-pol:index             -> ../config/index
>      |     |           +--rw oc-bgp-pol:config
>      |     |           |  +--rw oc-bgp-pol:index?   uint16
>      |     |           +--ro oc-bgp-pol:state
>      |     |           |  +--ro oc-bgp-pol:index?   uint16
>      |     |           +--rw oc-bgp-pol:asn-type
>      |     |           |  +--rw oc-bgp-pol:config
>      |     |           |  |  +--rw oc-bgp-pol:asn?     oc-inet:as-number
>      |     |           |  |  +--rw oc-bgp-pol:value?   uint32
>      |     |           |  +--ro oc-bgp-pol:state
>      |     |           |     +--ro oc-bgp-pol:asn?     oc-inet:as-number
>      |     |           |     +--ro oc-bgp-pol:value?   uint32
>      |     |           +--rw oc-bgp-pol:asn-regex-type
>      |     |              +--rw oc-bgp-pol:config
>      |     |              |  +--rw oc-bgp-pol:regex?   oc-types:posix-eregexp
>      |     |              +--ro oc-bgp-pol:state
>      |     |                 +--ro oc-bgp-pol:regex?   oc-types:posix-eregexp
9334c9353
<      |     |     |  +--rw oc-bgp-pol:ext-community-member*     union
---
>      |     |     |  x--rw oc-bgp-pol:ext-community-member*     union
9337,9339c9356,9428
<      |     |        +--ro oc-bgp-pol:ext-community-set-name?   string
<      |     |        +--ro oc-bgp-pol:ext-community-member*     union
<      |     |        x--ro oc-bgp-pol:match-set-options?        oc-pol-types:match-set-options-type
---
>      |     |     |  +--ro oc-bgp-pol:ext-community-set-name?   string
>      |     |     |  x--ro oc-bgp-pol:ext-community-member*     union
>      |     |     |  x--ro oc-bgp-pol:match-set-options?        oc-pol-types:match-set-options-type
>      |     |     +--rw oc-bgp-pol:members
>      |     |        +--rw oc-bgp-pol:member* [index]
>      |     |           +--rw oc-bgp-pol:index              -> ../config/index
>      |     |           +--rw oc-bgp-pol:config
>      |     |           |  +--rw oc-bgp-pol:index?   uint16
>      |     |           +--ro oc-bgp-pol:state
>      |     |           |  +--ro oc-bgp-pol:index?   uint16
>      |     |           +--rw oc-bgp-pol:asn4-type
>      |     |           |  +--rw oc-bgp-pol:config
>      |     |           |  |  +--rw oc-bgp-pol:asn?     oc-inet:as-number
>      |     |           |  |  +--rw oc-bgp-pol:value?   uint16
>      |     |           |  +--ro oc-bgp-pol:state
>      |     |           |     +--ro oc-bgp-pol:asn?     oc-inet:as-number
>      |     |           |     +--ro oc-bgp-pol:value?   uint16
>      |     |           +--rw oc-bgp-pol:asn4-regex-type
>      |     |           |  +--rw oc-bgp-pol:config
>      |     |           |  |  +--rw oc-bgp-pol:regex?   oc-types:posix-eregexp
>      |     |           |  +--ro oc-bgp-pol:state
>      |     |           |     +--ro oc-bgp-pol:regex?   oc-types:posix-eregexp
>      |     |           +--rw oc-bgp-pol:ip-type
>      |     |           |  +--rw oc-bgp-pol:config
>      |     |           |  |  +--rw oc-bgp-pol:ipv4-address?   oc-inet:ipv4-address
>      |     |           |  |  +--rw oc-bgp-pol:value?          uint16
>      |     |           |  +--ro oc-bgp-pol:state
>      |     |           |     +--ro oc-bgp-pol:ipv4-address?   oc-inet:ipv4-address
>      |     |           |     +--ro oc-bgp-pol:value?          uint16
>      |     |           +--rw oc-bgp-pol:link-bw
>      |     |           |  +--rw oc-bgp-pol:config
>      |     |           |  |  +--rw oc-bgp-pol:asn?         oc-inet:as-number
>      |     |           |  |  +--rw oc-bgp-pol:bandwidth?   oc-types:ieeefloat32
>      |     |           |  +--ro oc-bgp-pol:state
>      |     |           |     +--ro oc-bgp-pol:asn?         oc-inet:as-number
>      |     |           |     +--ro oc-bgp-pol:bandwidth?   oc-types:ieeefloat32
>      |     |           +--rw oc-bgp-pol:route-target
>      |     |           |  +--rw oc-bgp-pol:config
>      |     |           |  |  +--rw oc-bgp-pol:asn?     oc-inet:as-number
>      |     |           |  |  +--rw oc-bgp-pol:value?   uint16
>      |     |           |  +--ro oc-bgp-pol:state
>      |     |           |     +--ro oc-bgp-pol:asn?     oc-inet:as-number
>      |     |           |     +--ro oc-bgp-pol:value?   uint16
>      |     |           +--rw oc-bgp-pol:route-target-ip
>      |     |           |  +--rw oc-bgp-pol:config
>      |     |           |  |  +--rw oc-bgp-pol:ipv4-address?   oc-inet:ipv4-address
>      |     |           |  |  +--rw oc-bgp-pol:value?          uint16
>      |     |           |  +--ro oc-bgp-pol:state
>      |     |           |     +--ro oc-bgp-pol:ipv4-address?   oc-inet:ipv4-address
>      |     |           |     +--ro oc-bgp-pol:value?          uint16
>      |     |           +--rw oc-bgp-pol:route-origin
>      |     |           |  +--rw oc-bgp-pol:config
>      |     |           |  |  +--rw oc-bgp-pol:asn?     oc-inet:as-number
>      |     |           |  |  +--rw oc-bgp-pol:value?   uint16
>      |     |           |  +--ro oc-bgp-pol:state
>      |     |           |     +--ro oc-bgp-pol:asn?     oc-inet:as-number
>      |     |           |     +--ro oc-bgp-pol:value?   uint16
>      |     |           +--rw oc-bgp-pol:route-origin-ip
>      |     |           |  +--rw oc-bgp-pol:config
>      |     |           |  |  +--rw oc-bgp-pol:ipv4-address?   oc-inet:ipv4-address
>      |     |           |  |  +--rw oc-bgp-pol:value?          uint16
>      |     |           |  +--ro oc-bgp-pol:state
>      |     |           |     +--ro oc-bgp-pol:ipv4-address?   oc-inet:ipv4-address
>      |     |           |     +--ro oc-bgp-pol:value?          uint16
>      |     |           +--rw oc-bgp-pol:color
>      |     |              +--rw oc-bgp-pol:config
>      |     |              |  +--rw oc-bgp-pol:match-type?   oc-bgp-types:bgp-color-match-type
>      |     |              |  +--rw oc-bgp-pol:value?        uint32
>      |     |              |  +--rw oc-bgp-pol:flags?        enumeration
>      |     |              +--ro oc-bgp-pol:state
>      |     |                 +--ro oc-bgp-pol:match-type?   oc-bgp-types:bgp-color-match-type
>      |     |                 +--ro oc-bgp-pol:value?        uint32
>      |     |                 +--ro oc-bgp-pol:flags?        enumeration
```


jefftant:
Darren,

Link BW functionality as described in draft-ietf-idr-link-bandwidth is
rather limited, please take a look at draft-ietf-bess-ebgp-dmz which
addresses most of the limitations (transitivity is still upto
implementation though), there are implementation in EOS, IOS-XR and FRR.

Thanks,
Jeff

dplore:
> draft-ietf-bess-ebgp-dmz

Thanks for the updated reference.  I will refer to this in the model description.

dplore:
Last call, this will merge on Jun 9th, 2023

dplore:
@robshakir Now ready for a first look at the yang

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -661,6 +670,297 @@ module openconfig-bgp-policy {
...
+  }
+}
+
+  grouping ext-community-member-config {
+    description
+    "Configuration data for typed BGP extended communities.";
+
+    leaf index {
+      type uint16;
+      mandatory true;
```
earies, line 951: `mandatory`  is a design pattern not taken up for other list keys across OC models to date - it is only softly implied by way of the key being a leafref to one of the config/state nodes but now this would enforce the non-leafref'd node to be mandatory (in this case state)

Would suggest consistency before all else

Both YANG 1.0/1.1 also state that the mandatory statement on key leafs are ignored.  This would also apply here since this node is the ultimate leafref

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -661,6 +670,297 @@ module openconfig-bgp-policy {
...
+    description "Assigned value for this community.";
+  }
+}
+
+  grouping ext-community-member-config {
+    description
+    "Configuration data for typed BGP extended communities.";
+
+    leaf index {
+      type uint16;
```
earies, line 950: What is the full intention of an integer based index here?  Is there reference to any implementation supporting this?

release/models/bgp/openconfig-bgp-policy.yang
dplore: Thanks for this feedback.  Good point on the consistency, and the yang standard, I will remove this.

release/models/bgp/openconfig-bgp-policy.yang
dplore: The index is intended to be a unique identifier for each record in the list.   It is not intended to identify an order for entries (it's not a sequence id).   (Do we need communities in a community set to be ordered?).   For these reasons, I don't expect index to be represented in an equivalent CLI command or parameter.    

This is one of two options I've thought of so far.  I thought this was preferable to introducing a name based key as community-sets already have a name.  Adding another, nested named for each entry in a community-set seems undesirable.   Since yang requires a list that can be configured to have a key, an unkeyed list isn't an option.

I welcome other ideas on how to model the list of community-set members.   I would like to get rid of the clunky, error-prone regexes we are using to constrain strings to unit16,unit32 values.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -661,6 +670,296 @@ module openconfig-bgp-policy {
...
+  leaf asn {
+    type oc-inet:as-number;
+    description "Autonomous system number.";
+  }
+
+  leaf bps {
+    type uint64;
+    description "Bandwidth in bits per second.";
+  }
+}
```
jhaas-pfrc, line 876: The link-bandwidth draft has a number of modeling challenges.  Here's the relevant text from https://datatracker.ietf.org/doc/html/draft-ietf-idr-link-bandwidth-07

> The extended community is optional non-transitive.  The value of the
>    high-order octet of the extended Type Field is 0x40.  The value of
>    the low-order octet of the extended type field for this community is
>    0x04.  The value of the Global Administrator subfield in the Value
>    Field SHOULD represent the Autonomous System of the router that
>    attaches the Link Bandwidth Community.  If four octet AS numbering
>    scheme is used [[RFC6793](https://datatracker.ietf.org/doc/html/rfc6793)], AS_TRANS should be used in the Global
>    Administrator subfield.  The bandwidth of the link is expressed as 4
>    octets in IEEE floating point format, units being bytes (not bits!)
>    per second.  It is carried in the Local Administrator subfield of the
>    Value Field.

Thus, for config purposes you *could* use an as-number type, but it'll be reduced into AS_TRANS by the implementations if it's following the draft.  

Juniper's implementation simply wants a 16-bit number there.

The "bps" leaf is also at odds with the contents of the draft being *bytes* per second.  I suggest addressing that.

The place where things are a significant headache is that the protocol component is carried in floating point format on the wire.  While Juniper's implementation limits the input to a 32-bit integer, a 64-bit integer could eventually be implemented - but it's incompatible from original commit if so.  But regardless of whether the input is 32 or 64-bit at configuration, eventually on the wire it is reduced to a IEEE 32-bit float and the number is coerced into something that no longer matches the input.

Implementations thus often take the integer input, and in operations, display some form of float output or a integer truncated output for the on-the-wire float.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -661,6 +670,296 @@ module openconfig-bgp-policy {
...
+    type uint16;
+    description "Assigned value for this community.";
+  }
+}
+
+grouping ext-community-route-target-ip-config {
+  description "Grouping for route-target-ip extended community
+  config.";
+  leaf ip-address {
+    type oc-inet:ip-address;
```
jhaas-pfrc, line 895: This type is a union of v4 and v6 and isn't appropriate here.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -661,6 +670,296 @@ module openconfig-bgp-policy {
     }
   }
 
+  grouping ext-community-members-top {
+    description
+      "Top-level grouping for members of an external community set";
```
jhaas-pfrc, line 675: extended, not external

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -661,6 +670,296 @@ module openconfig-bgp-policy {
...
+
+    container members {
+        description
+            "Enclosing container for list of extended community members";
+
+        list member {
+            key "index";
+            description
+              "List of members in an external community set.  Each entry
```
jhaas-pfrc, line 684: extended, not external

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -661,6 +670,296 @@ module openconfig-bgp-policy {
...
+
+          container state {
+            config false;
+            description "State for BGP extended community set members";
+
+            uses ext-community-member-config;
+            uses ext-community-member-state;
+          }
+
+          container asn-type {
```
jhaas-pfrc, line 710: Here and following, you have top level containers for communities by their structure.  However, without knowing the type/subtype, you have zero idea what these things are.  For example, route targets can be of these types, but other things may use this structure without being route targets.
My recommendation is that these generic containers are NOT included as they only create ambiguity.

jhaas-pfrc:
A few high level comments on link bandwidth.  As Jeff Tantsura notes above, there is work in BESS in IETF that discusses how to change some of the rules about how transitivity is managed.  That work will eventually run up against discussion in IDR about how transitivity for extended communities needs to be dealt with; i.e. are we even breaking rules for transitivity if an implementation simply originates a non-transitive over an eBGP session and how is that distinct from a bug in the transitivity scoping being enforced.

There's also the headache that Juniper's implementation of link bandwidth is not correct vs. the draft.  Juniper created the original draft, but at the same time has been shipping code that uses TRANSITIVE link-bandwidth.  This creates interop issues and work is ongoing to figure out how to change Juniper's implementation to both provide continuing support to our customers and cover existing transitive use cases while at the same time adding support for the non-transitive one in the existing draft.  This work is being discussed with the current IETF draft authors.

jhaas-pfrc:
High level comment on this proposal: I feel your pain.  

The fundamental issue is that extended communities are an on-the-wire easy thing but in config and display are a unstructured mess.  

From a display standpoint, there's a desire to provide a simple string-based canonical form. 

For the configuration standpoint, trying to get out of regex hell is desirable.  As an example, the ambiguity for a 2-byte AS route target from RFC 4360 and RFC 5668 4-byte ASes I highlighted in the diff provide an example where display ambiguity for on-the-wire encoding can be troublesome to distinguish without better keywords (I think IETF has a bug still in this respect), but can be made fully clear with structured containers to remove such ambiguity.

The challenge is when configuration moves to structured elements that there should be clear coupling between the configuration style and the resultant string display style.  See my comments covering the on-the-wire for link bandwidth being float vs. configuration as integer.

rszarecki:
Regarding bandwidth encoding in OpenConfig.

In contrast to CLI, the user would be OpenConfig client/software. So why not keep baldwidth in on-the-wire format to avoid uncertanity?

release/models/bgp/openconfig-bgp-policy.yang
dplore: The goal by defining this level of containers it to provide a way to identify the 'style' of community being used.  

- asn-type community (2octet ASN rfc4360)
- asn4-type community (4octet ASN rfc5668 section 1) (just added this sub container to clarify)
- ip-type community (rfc5668 section 2)
- link-bw community (draft-ietf-idr-link-bandwidth-07)
- route-target-ip (rfc4360 section 4 using using an IP address)
- route-target (rfc4360 section 4)
- route-origin (rfc4369 section 5)
- route-origin-ip (rfc4369 section 5 using ip address)
- color (rfc9012 and draft-ietf-idr-segment-routing-te-policy)

These containers are similar to using a keyword in a command line community statement.  They go one step further than the example commands below which need to parse strings to derive the 'style'.
From JunOS example has a keyword, but also relies on some string parsing too:
```
set policy-options community origin-as members origin:64511:3
set policy-options community origin-ip members origin:172.16.7.7:4
set policy-options community target-as members target:64511:1
```

These are equivalent to proposed OC:
```
prefix = /routing-policy/defined-sets/bgp-defined-sets/ext-community-sets
$prefix/ext-community-set[name=origin-as]/members/member[index=1]/asn-type/config/asn = "64511"
$prefix/ext-community-set[name=origin-as]/members/member[index=1]/asn-type/config/value = "3"

$prefix/ext-community-set[name=origin-ip]/members/member[index=1]/ip-type/config/asn = "172.16.7.7"
$prefix/ext-community-set[name=origin-ip]/members/member[index=1]/ip-type/config/value = "4"

$prefix/ext-community-set[name=origin-ip]/members/member[index=1]/route-target/config/asn = "64511"
$prefix/ext-community-set[name=origin-ip]/members/member[index=1]/route-target/config/value = "1"
```

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -661,6 +670,296 @@ module openconfig-bgp-policy {
...
+  leaf asn {
+      type oc-inet:as-number;
+      description "Autonomous system number.";
+  }
+
+  leaf value {
+      type    uint16;
+      description "Assigned value for this community.";
+  }
+}
```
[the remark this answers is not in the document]
dplore: Added asn4-type to explicitly handle the 2byte and 4byte ASN scenarios.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -661,6 +670,296 @@ module openconfig-bgp-policy {
...
+
+grouping ext-community-ip-type-config {
+  description "Grouping for ip-type extended community config.";
+  leaf asn {
```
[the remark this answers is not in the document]
dplore: fixed

release/models/bgp/openconfig-bgp-policy.yang
dplore: Updated to use float32 for the bandwidth as suggested.

I will remove the ASN leaf for link bandwidth since it essentially cannot be configured.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -796,6 +805,324 @@ module openconfig-bgp-policy {
...
+
+grouping ext-community-color-config {
+  description "Grouping for color extended community config.";
+  leaf match-type {
+    type oc-bgp-types:bgp-color-match-type;
+    description "Specifies the match criteria between BGP NH and
+    SR policy endpoint";
+  }
+
+  leaf value {
```
jhaas-pfrc, line 1102: https://datatracker.ietf.org/doc/html/draft-ietf-idr-segment-routing-te-policy-26#name-color-extended-community covers changes that are happening to the color extended community.  What this will eventually push this model to is a need to have a second leaf that isn't solely "value", it will need to cover the flags and other things.

dplore: Added `flags` enum below to cover draft-ietf-idr-segment-routing-te-policy-26

rszarecki:
With `oc-bgp-pol:community-member*` deprecated, there is no way to efficiently express:
-  "any community with ASN 1234" - with this PR we would need to list 65535 individual communities.
- " any community with private ASN and value of lower then 1000" - with this PR we would need to list 1,021,977 individual communities.

traditionally this is achived by use of reg-exp.

If we want to keep spitit of strongly typed, structured data, perhaps we shall provide ability to express ASN range and/or value range, instead of just singelton values.

dplore:
> With `oc-bgp-pol:community-member*` deprecated, there is no way to efficiently express:
> 
> * "any community with ASN 1234" - with this PR we would need to list 65535 individual communities.
> * " any community with private ASN and value of lower then 1000" - with this PR we would need to list 1,021,977 individual communities.
> 
> traditionally this is achived by use of reg-exp.
> 
> If we want to keep spitit of strongly typed, structured data, perhaps we shall provide ability to express ASN range and/or value range, instead of just singelton values.

Thanks for catching this.  The regex I'm attempting to remove is the  typedefs which use a string and regex validator in the yang model.   Regex as a user supplied match criteria I think is a feature we need to keep.  

 I added a container and leaf for a user supplied regex.  Please let me know what you think.   There are two concepts to evaluate for this I think.
1. The path/yang structure 
2. The idea that a single, standard POSIX format regex should be used to match against a list of communities.  

Item 2 has two parts which are very important I think.  The first is we should use a standard, testable regex format.  Today each vendor has their own regex format and special rules which are only testable if you have a running binary image from the vendor and send routes.  By using a standard regex format, one can write code or use any standard regex evaluator tool to test how a match works.  

The second part is  I proposed a standard way to format the community list to be matched.  We need this also so the standard regex that will give us the same match behavior no matter which implementation it's pushed to.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -796,6 +807,449 @@ module openconfig-bgp-policy {
...
+
+  grouping community-member-config {
+    description
+    "Configuration data for BGP standard communities.";
+
+    leaf index {
+      type uint16;
+      description
+        "Represents an index of a member entry.";
+    }
```
dplore, line 889: This needs to have a community value.
There should also be a community regex option (list of regexes)

PR #884, opened 2023-06-06
Add vendor-policy-definition container under routing-policy

mjancenko-arista:
### Change Scope

* A new container, `vendor-policy-definition`, will be added to `routing-policy/policy-definitions/policy-definition`. This can be used to define a policy in the routing policy framework that a vendor may provide.

 * Implementation A: [Arista](https://www.arista.com/en/support/toi/eos-4-27-2f/15102-routing-control-functions-language-and-configuration) - The contents of `vendor-policy-definition/config/text` would map to the definition of an RCF function in EOS.

<pre>
  +--rw routing-policy
     <... snip ...>
     +--rw policy-definitions
        +--rw policy-definition* [name]
           +--rw name                        -> ../config/name
           +--rw config
           |  +--rw name?   string
           +--ro state
           |  +--ro name?   string
           <b>+--rw vendor-policy-definition
           |  +--rw config
           |  |  +--rw text?           string
           |  |  +--rw policy-group?   string
           |  +--ro state
           |     +--ro text?           string
           |     +--ro policy-group?   string</b>
           +--rw statements
              <... snip ...>
</pre>

### DIFF
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -40,6 +40,10 @@ module openconfig-routing-policy {
     actions may be multitude of changes to route attributes or a
     final disposition of accepting or rejecting the route.
 
+    Alternatively, policies can be expressed in the policy language
+    defined by a vendor. This type of policy is provided as a text
+    string and interpreted by the vendor's policy framework.
+
     Route policy evaluation:
 
     Policy definitions are referenced in routing protocol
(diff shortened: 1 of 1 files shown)


dplore:
We reviewed this in the OC operators meeting today. 

This proposal is redundant to the routing-policy defined today and is not aligned with the goal of having vendor neutral configuration.  In addition, OpenConfig models seek to model attributes and values versus opaque text strings.

If the current policy models are lacking functionality, proposals could be made to add the needed functionality.

dplore:
As this model change is not aligned with OpenConfig intent, I will close this now.

Issue #886, opened 2023-06-08
In openconfig-macsec model, support rx-late-pkts in macsec-interface-counters grouping. 


m26singhvi:
In openconfig-macsec model macsec-interface-counters  grouping, currently, we have the following counters : 

```
    1.  tx-untagged-pkts
    2. rx-untagged-pkts 
    3. rx-badtag-pkts
    4. rx-unknownsci-pkts 
    5. rx-nosci-pkts 
```
This issue is to support `rx-late-pkts` as well in the same grouping. This counter is already mentioned in 802.1AE-2018 Table 13-7.

PR #887, opened 2023-06-08
Update oc-macsec model to support rx-late-pkts in macsec-interface-co…

m26singhvi:
Update oc-macsec model to support rx-late-pkts in macsec-interface-counters grouping (#886)

### Change Scope

*  In openconfig-macsec model macsec-interface-counters grouping, currently, rx-late-pkts counter is not supported.
*  This change enhances the model to add the support for same.
*  This change is backwards compatible as it is only adding a new leaf. 
*  802.1AE-2018 standard refers to this counter as inPktsLate.

### DIFF
--- a/release/models/macsec/openconfig-macsec.yang
+++ b/release/models/macsec/openconfig-macsec.yang
@@ -18,10 +18,16 @@ module openconfig-macsec {
     "This module defines configuration and state data for
      MACsec IEEE Std 802.1AE-2018.";
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "1.1.0";
   oc-ext:regexp-posix;
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
+
+  revision "2023-06-08" {
+    description
+      "Support rx-late-pkts leaf.";
+    reference "1.1.0";
+  }
 
   revision "2022-04-28" {
     description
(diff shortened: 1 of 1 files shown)


release/models/macsec/openconfig-macsec.yang
```diff
@@ -18,11 +18,17 @@ module openconfig-macsec {
     "This module defines configuration and state data for
      MACsec IEEE Std 802.1AE-2018.";
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "1.0.1";
```
dplore, line 21:
```suggestion
  oc-ext:openconfig-version "1.1.0";
```

Need to update minor version due to addition of leaf.

release/models/macsec/openconfig-macsec.yang
```diff
@@ -18,11 +18,17 @@ module openconfig-macsec {
...
   oc-ext:regexp-posix;
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
+  revision "2023-06-08" {
+    description
+      "Support rx-late-pkts leaf.";
+    reference "1.0.1";
```
dplore, line 29:
```suggestion
    reference "1.1.0";
```

Need to update minor version due to addition of leaf.

dplore (changes requested):
Need to update minor version due to addition of leaf, otherwise looks good.

release/models/macsec/openconfig-macsec.yang
m26singhvi: Thanks for the review. I have updated the minor version in the latest update.

release/models/macsec/openconfig-macsec.yang
m26singhvi: Thanks for the review. I have updated the minor version in the latest update.

m26singhvi: Is this good to merge now ?

dplore:
* [Cisco IOS XE](https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst9300/software/release/16-9/configuration_guide/sec/b_169_sec_9300_cg/macsec_encryption.html#configuring_examples_for_macsec_encryption) implements this counter in a show command

* [Arista EOS](https://www.arista.com/en/um-eos/eos-data-plane-security#xx1211285) supports a show command for late packets

dplore (approved):

PR #891, opened 2023-06-10
Clarify use match-set-options=ALL with regex

### LINKED ISSUE #875: Clarify what "match-set-options=ALL" means when "community-member" is a list of regexp
In `routing-policy/defined-sets/bgp-defined-sets` we have
```
pyang -p third_party/ietf -p release release/models/*/*.yang -f tree --tree-path /routing-policy/defined-sets/bgp-defined-sets
  +--rw routing-policy
     +--rw defined-sets
        +--rw oc-bgp-pol:bgp-defined-sets
           +--rw oc-bgp-pol:community-sets
           |  +--rw oc-bgp-pol:community-set* [community-set-name]
           |     +--rw oc-bgp-pol:community-set-name    -> ../config/community-set-name
           |     +--rw oc-bgp-pol:config
           |     |  +--rw oc-bgp-pol:community-set-name    string
           |     |  +--rw oc-bgp-pol:community-member*     union
           |     |  +--rw oc-bgp-pol:match-set-options?    oc-pol-types:match-set-options-type
           |     +--ro oc-bgp-pol:state
           |        +--ro oc-bgp-pol:community-set-name    string
           |        +--ro oc-bgp-pol:community-member*     union
           |        +--ro oc-bgp-pol:match-set-options?    oc-pol-types:match-set-options-type
           +--rw oc-bgp-pol:ext-community-sets
           |  +--rw oc-bgp-pol:ext-community-set* [ext-community-set-name]
           |     +--rw oc-bgp-pol:ext-community-set-name    -> ../config/ext-community-set-name
           |     +--rw oc-bgp-pol:config
           |     |  +--rw oc-bgp-pol:ext-community-set-name?   string
           |     |  +--rw oc-bgp-pol:ext-community-member*     union
           |     |  +--rw oc-bgp-pol:match-set-options?        oc-pol-types:match-set-options-type
           |     +--ro oc-bgp-pol:state
           |        +--ro oc-bgp-pol:ext-community-set-name?   string
           |        +--ro oc-bgp-pol:ext-community-member*     union
           |        +--ro oc-bgp-pol:match-set-options?        oc-pol-types:match-set-options-type
```

Specifically, `match-set-options=ALL` means "match is true if given value matches all members of the defined set" (see https://github.com/openconfig/public/blob/master/release/models/policy/openconfig-policy-types.yang#L102-L121).

This is straightforward when `leaf-list community-member` is just a list of community values. For all members of the defined set to match the given value, we just need `leaf-list community-member` to be a subset of the BGP community path.

However, it's unclear what `match-set-options=ALL` means when `leaf-list community-member` is a list of regular expressions (see https://github.com/openconfig/public/blob/master/release/models/bgp/openconfig-bgp-policy.yang#L443-L472).

To me this doesn't seem to make much sense semantically. If this is indeed something meaningful we can configure, then a clarification in the description section can be really helpful.

Tagging @dplore because we chatted about this offline.

dplore:
* (M)  openconfig-policy-types.yang

match-set-options=ALL is not supported when used  with oc-bgp-types:bgp-community-regexp-type

Resolves #875 

## Scope of Change 
There is ambiguity in the OC model as to how matches are intended to be performed.  Sorry for the long explanation below, but this hopefully clearly explains it considering OC policy match vs. what NOS are doing today.

I believe the logic issue can be explained this way:

### Match behavior for match-set-options-type=ANY
There is no issue with the ANY behavior. I present this as a case which is well understood as background for the motivation for this PR.

`match-set-options-type=ANY`  The device should compare each item in a set of values with a list of match-set criteria.  If one or more values match one or more of the elements in the match-set, then the match operation is true.   In other words, a logical "OR" operation is used both for comparing a value with each match-set and across all the values.

For example, a prefix contains two communities:  5413:1, 29636:2
and the match-set contains 5413:[01] and 5555:[12] 
`match-set-options-type=ANY` will return a match because 5413:1 matches regex 5413:[01].

###  Match behavior for match-set-options-type=ALL

`match-set-options-type=ALL` requires a logical "AND" operation to be performed.   But it's ambiguous to me exactly how OC expects this to be performed.  There are two options I can see.  
1. The device should compare each item in a list of values with the list of match criteria.  ALL of the match criteria must be true for a given value in the list for a match to be true.
2. The device should compare each item in a list of values with ALL the items in the list of match criteria.  All of the match criteria must match at least one, but not all values in the value list. 
3. The device should compare each item in the list of values with the match criteria.  ALL of the values in the list must match at least one match criteria.   

In option 1, multiple regexes aren't logical because they must have some common/overlapping criteria in order for one value to match multiple regexes.  In this case it is more simple just have just one regex with the desired match criteria.  
This is the motivation for this PR.  However, it is not clear if OC requires this match behavior. 

In option 2, using multiple regexes seems logical.  

In option 3, using multiple regexes seems logical.  

[Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-0/routing/command/reference/rr40crs1book_chapter8.html#wp897883513) supports behavior like option 2, but only for community ranges, not for regex.   

[JunOS's policy match implementation for regex](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-configuring-as-path-regular-expressions-to-use-as-routing-policy-match-conditions.html) specifies that  ANY / OR is implemented when regex is in place.  A match-set-options ALL is not supported when using regex.

[Arista EOS ip extcommunity-list regexp](https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1116761) allows defining a list of regular expressions.  In [a comment below](https://github.com/openconfig/public/pull/891#issuecomment-1634379245), Arista indicates they support option 2 behavior.

[Nokia SR Linux supports policy match using a community set with multiple regex entries](https://documentation.nokia.com/srlinux/22-3/SR_Linux_Book_Files/Data_Model_Reference/srl_nokia-routing-policy_0.html#d4593) but it is not clear what the match behavior is from the documention.   Below @hellt describes that currently ALL is not implemented, but if they did implement it, it would follow option 3.

### DIFF
--- a/release/models/policy/openconfig-policy-types.yang
+++ b/release/models/policy/openconfig-policy-types.yang
@@ -24,7 +24,14 @@ module openconfig-policy-types {
     policy.  It can be imported by modules that contain protocol-
     specific policy conditions and actions.";
 
-  oc-ext:openconfig-version "3.2.3";
+  oc-ext:openconfig-version "3.2.4";
+
+  revision "2023-06-09" {
+    description
+      "Clarify that match-set-options-type ALL is not valid in
+      combination with bgp-community-regexp-type";
+    reference "3.2.4";
+  }
 
   revision "2022-11-08" {
     description
@@ -107,7 +114,8 @@ module openconfig-policy-types {
       }
       enum ALL {
         description "match is true if given value matches all
-        members of the defined set";
+        members of the defined set.  This type is not valid when
+        used with oc-bgp-types:bgp-community-regexp-type";
       }
       enum INVERT {
         description "match is true if given value does not match any


rolandphung (approved):

robshakir:
Can you explain why this is not logical? 

i.e., if I were to write two regular expressions like `5413:[01]` and  `29636:[12]` - could I not expect that communities that are associated with the route match both? i.e., there are multiple communities possible per route.

Do we rather mean that this is not commonly supported?

dplore:
There is ambiguity in the OC model as to how matches are intended to be performed.  Sorry for the long explanation below, but this hopefully clearly explains it considering OC policy match vs. what NOS are doing today.

I believe the logic issue can be explained this way:

## Match behavior for match-set-options-type=ANY
There is no issue with the ANY behavior. I present this as a case which is well understood as background for the motivation for this PR.

`match-set-options-type=ANY`  The device should compare each item in a set of values with a list of match-set criteria.  If one or more values match one or more of the elements in the match-set, then the match operation is true.   In other words, a logical "OR" operation is used both for comparing a value with each match-set and across all the values.

For example, a prefix contains two communities:  5413:1, 29636:2
and the match-set contains 5413:[01] and 5555:[12] 
`match-set-options-type=ANY` will return a match because 5413:1 matches regex 5413:[01].

##  Match behavior for match-set-options-type=ALL

`match-set-options-type=ALL` requires a logical "AND" operation to be performed.   But it's ambiguous to me exactly how OC expects this to be performed.  There are two options I can see.  
1. The device should compare each item in a list of values with the list of match criteria.  All the match criteria must be true for a given value in the list for a match to be true.
2. The device should compare each item in a list of values with ALL the items in the list of match criteria.  All of the match criteria must match at least one, but not all values in the value list. 

In option 1, multiple regexes aren't logical because they must have some common/overlapping criteria in order for multiple values to match more than one regex.  In this case it is far more simple just have just one regex with the desired match criteria.  This is the motivation for this PR.  However, it is not clear if OC requires this match behavior. 

In option 2, using multiple regexes seems logical.  

[Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-0/routing/command/reference/rr40crs1book_chapter8.html#wp897883513) supports behavior like option 2.

[JunOS's policy match implementation for regex](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-configuring-as-path-regular-expressions-to-use-as-routing-policy-match-conditions.html) also supports option 2

[Arista EOS ip extcommunity-list regexp](https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1116761) allows defining a list of regular expressions.  However I cannot find documentation which explains the match behavior in enough detail to know if option 1, 2 or some other behavior is implemented.   @rolandphung can you assist?  

[Nokia SR Linux supports policy match using a community set with multiple regex entries](https://documentation.nokia.com/srlinux/22-3/SR_Linux_Book_Files/Data_Model_Reference/srl_nokia-routing-policy_0.html#d4593) but it is not clear what the match behavior is from the documention.  @hellt can you assist?

robshakir:
There seem to be two discrete sets of functionality there -- thanks for highlighting the alternate @hellt. 

If we say that `ALL` means that "all regexps within the set must be matched", then I think @dplore's analysis covers that case well. Alternatively, with @hellt's envisaged behaviour we're saying `ALL` means "all members of the community list must match one of the regexps specified". 

I don't recall having seen an implementation that would do the "all members must match" option, so I'm OK with not allowing the "ALL" case here, which seems to be consistent with the other places that we say "ALL is not supported" like neighbour sets/prefix sets.

This all being said -- I note that `community-set` and `ext-community-set` seem to be the only place where we define this within the actual set, rather than the match. All the other matches that are in routing policy have the `match-set-options` specified outside of them. I can't remember the history here -- was this a change that was made specifically for these two types? I believe that there was some discussion of something like this previously.

dplore:
I updated my list of options to include the algorithm that @hellt [specifies](https://github.com/openconfig/public/pull/891#issuecomment-1613582956). and moved my options explanation to the first comment in this PR.

I'm awaiting feedback from @earies and @rolandphung before making more changes.

xchu-arista:
@dplore Arista EOS `ip community-list` and `ip extcommunity-list` support match behaviors outlined in both the `match-set-options=ANY` case and `match-set-options=ALL` option 2 case.

When used with regexp type communities however, `match-set-options=ALL` is not supported.

rszarecki:
I can offer different view point here:
Observe that community-set has leaf-list[]. Observe that match-set-type is attribute of community-set. Not a method.
Hence ALL|ANY|INVERT match-set-type applys to elements of list, regardless if element is exact-value or regexp.
Evaluation therefore is:
check if any of path's community matches liaf-list[i] element. This produce arrey of true/false, one per each leaf-list element.
Then apply match-set-type afainst this array.
ALL then AND(array[]) (case 2)
ANY then OR(array[])
INVERT thaen !(OR(array[])).

Please note that communities are transitive attributes, hence in Internet context, one AS may add attribute to route and expect to be carried over internet. Thence logic that requires that ALL communities of path are matching is impractical. We can't know what community will appear next. But we may ask that some communities C1 & C2 and C3 exist simultanously as attributes of routes.

-------------

Above logic will work well for regexp.

PR #894, opened 2023-06-17
Skip yanglint in CI

wenovus:
Looks like yanglint is way too out-of-date now. I'll work on getting the latest version ready so we can run it again and also create a CD for the image itself so that this doesn't happen again.


aashaikh (approved):
LGTM -- thanks for the fix!

PR #898, opened 2023-06-27
Clarify transport/state/remote-port vs. config/neighbor-port

dplore:
This is a description only clarification on the purpose of 
`/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/state/neighbor-port`
versus
`/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/transport/state/remote-port`

state/neighbor-port contains the value of the local device's configured (config/neighbor-port) destination TCP port of a bgp neighbor.

transport/state/remote-port shows the source TCP port of a remote peer which is currently connected.  This could be a dynamic or statically configured peer.  The value might be the same as state/neighbor-port if the local device created the connection.    But if the connection was created by the neighbor, the TCP port will be an ephemeral TCP port number.

### DIFF
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -24,7 +24,13 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "9.4.0";
+  oc-ext:openconfig-version "9.4.1";
+
+  revision "2023-06-27" {
+    description
+      "Clarify bgp remote-port description";
+    reference "9.4.1";
+  }
 
   revision "2023-04-01" {
     description
--- a/release/models/bgp/openconfig-bgp-neighbor.yang
+++ b/release/models/bgp/openconfig-bgp-neighbor.yang
@@ -203,7 +203,8 @@ submodule openconfig-bgp-neighbor {
         type oc-inet:port-number;
         default 179;
         description
-          "Destination TCP port number of the BGP peer";
+          "Destination TCP port number of the BGP peer when initiating a
+          session from the local router";
     }
 
     leaf enabled {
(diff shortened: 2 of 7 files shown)


release/models/bgp/openconfig-bgp-neighbor.yang
```diff
@@ -475,7 +481,9 @@ submodule openconfig-bgp-neighbor {
       type oc-inet:port-number;
       description
         "Remote port being used by the peer for the TCP session
-        supporting the BGP session";
+        supporting the BGP session.  This may be the same value
+        as the configured neighbor-port or another TCP port if
+        the peer initiated the TCP session.";
```
wenovus, line 486: This **is expected to** be the same value
as the configured neighbor-port **if the local device initiated the connection** or another TCP port if
the peer initiated the TCP session."; ?

release/models/bgp/openconfig-bgp-neighbor.yang
```diff
@@ -197,7 +203,7 @@ submodule openconfig-bgp-neighbor {
         type oc-inet:port-number;
         default 179;
         description
-          "Port number of the BGP peer";
+          "Destination TCP port number of the BGP peer";
```
wenovus, line 206: "Destination TCP port number of the BGP peer **when initiating a session from the local router**"; ?

In a BGP session, my understanding is that both sides need to configure each other as neighbours, and as result both sides have `neighbor-port` configured (either explicitly or via the default value). `neighbor-port` can therefore only be different from `remote-port` if the connection was initiated by the neighbour via a different port number. When this happens we are ignoring this configuration.

wenovus (approved):

release/models/bgp/openconfig-bgp-common-multiprotocol.yang
```diff
@@ -24,7 +24,13 @@ submodule openconfig-bgp-common-multiprotocol {
...
     multiple contexts.";
 
-  oc-ext:openconfig-version "9.4.0";
+  oc-ext:openconfig-version "9.4.1";
+
+  revision "2023-06-27" {
+    description
+      "Clarify bgp remote-port description";
+    reference "9.4.1";
+  }
```
sachendras, line 33: Should we update the following as well w/ similar explanation for consistency?

/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/transport/state/
local-port
description:
Local TCP port being used for the TCP session supporting the BGP session

sachendras (approved):

Issue #861, opened 2023-04-20
Default values for red/enable-ecn and wred/enable-ecn mean both are effectively enabled.


rgwilton:
In the following model, both enable-ecn and drop have default values that makes the model ambiguous as to whether red or wred is applies to the queue.

Under:
```
     +--rw queue-management-profiles
        +--rw queue-management-profile* [name]
           +--rw wred
           |  +--rw uniform
           |     +--rw config
           |     |  +--rw min-threshold?                  uint64
           |     |  +--rw max-threshold?                  uint64
           |     |  +--rw enable-ecn?                     boolean
           |     |  +--rw drop?                           boolean
           |     |  +--rw weight?                         uint32
           |     |  +--rw max-drop-probability-percent?   oc-types:percentage
           |     +--ro state
           |        +--ro min-threshold?                  uint64
           |        +--ro max-threshold?                  uint64
           |        +--ro enable-ecn?                     boolean
           |        +--ro drop?                           boolean
           |        +--ro weight?                         uint32
           |        +--ro max-drop-probability-percent?   oc-types:percentage
           +--rw red
              +--rw uniform
                 +--rw config
                 |  +--rw min-threshold?   uint64
                 |  +--rw max-threshold?   uint64
                 |  +--rw enable-ecn?      boolean
                 |  +--rw drop?            boolean
                 +--ro state
                    +--ro min-threshold?   uint64
                    +--ro max-threshold?   uint64
                    +--ro enable-ecn?      boolean
                    +--ro drop?            boolean
```

dplore:
Hi @rgwilton ,  is the issue that RED and WRED should be mutually exclusive?

rgwilton:
Sorry for being slow to respond.

This is just one of the examples where there are mutually exclusive features (e.g., only one of them is expected to be present) but the use of YANG defaults means that they are both in effect.

I think that this needs to be fixed, with the possible different approaches being:

1) Remove the default values.
2) Allow the use choice or presence containers to fix the model at the language level (but OpenConfig doesn't usually use these constructs because it prevents operational state from being reported, if it hasn't been configured).
3) For this particular case, rework the model so that the config is merged into a single container (but this can't be a generic solution).  I.e., have a single container called 'red' but that contains all the configuration that you have under 'wred'.  If no weight has been specified then it is red, if a weight has been specified then it is wred.

[...]

PR #900, opened 2023-07-01
Add IPv6 RA config mode

dplore:
### Change Scope
Introduce a new leaf to set one of three different modes for IPv6 router advertisements.  

Deprecate the ipv6 router advertisement config "suppress" leaf since common operational requirements involve setting one of three different values.

### Platform Implementations

 * [Arista EOS](https://www.arista.com/en/um-eos/eos-ipv6#xx1143369) implements `ipv6 nd ra disabled` and `ipv6 nd ra disabled all`
 
 * [Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/ios_xr_sw/iosxr_r3-7/addr_serv/command/reference/ir37ntsk.html#wp1182427) implements `ipv6 nd ra suppress` and `ipv6 nd ra suppress all`

 * [JunOS](https://www.juniper.net/documentation/us/en/software/junos/neighbor-discovery/topics/topic-map/ipv6-neighbor-discovery.html) uses router-advertisement to enable or disable on a per interface basis.  It is disabled by default.  JunOS uses [solicit-router-advertisement-unicast](https://www.juniper.net/documentation/us/en/software/junos/neighbor-discovery/topics/ref/statement/solicit-router-advertisement-unicast.html) to enable/disable support for solicitation.

* [Nokia SRLinux](https://documentation.nokia.com/cgi-bin/dbaccessfilename.cgi/3HE17569AAAA01_V1_SR%20Linux%20R21.6%20Configuration%20Basics.pdf) can enable and disable ipv6 router advertisements.  It is not appear that solicitation is configurable.

### DIFF
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -950,10 +950,9 @@ module openconfig-if-ip {
       type boolean;
       default true;
       description
-        "All ipv6 router advertisement functions are disabled.  The
-        local system will not transmit router advertisement messages and
-        will not respond to router solicitation messages";
-
+        "If set to false, all ipv6 router advertisement functions are
+        disabled.  The local system will not transmit router advertisement
+        messages and will not respond to router solicitation messages";
     }
 
     leaf interval {
@@ -978,9 +977,8 @@ module openconfig-if-ip {
       type boolean;
       default false;
       description
-        "When set to true, router advertisement neighbor discovery messages
-        are not transmitted on this interface.  Responses to router
-        solicitation are transmitted.";
+        "When set to true, router advertisement neighbor discovery
+        messages are not transmitted on this interface.";
     }
 
     leaf mode {
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -946,6 +946,16 @@ module openconfig-if-ip {
     description
       "Configuration parameters for IPv6 router advertisements.";
 
+    leaf enable {
+      type boolean;
+      default true;
+      description
+        "All ipv6 router advertisement functions are disabled.  The
+        local system will not transmit router advertisement messages and
+        will not respond to router solicitation messages";
+
+    }
+
     leaf interval {
       type uint32;
       units seconds;
@@ -975,25 +985,19 @@ module openconfig-if-ip {
 
     leaf mode {
       type enumeration {
-        enum ENABLE {
+        enum ALL {
           description
-            "The system will transmit ipv6 router advertisement messages and
-            respond to route solicitation requests.";
+            "The system will transmit unsolicited router advertisement
+            messages and respond to router solicitation requests.";
         }
-        enum DISABLE {
-          description
-            "All ipv6 router advertisement functions are disabled.  The
-            local system will not transmit router advertisement messages and
-            will not respond to router solicitation messages.";
-        }
-        enum DISABLE_RA {
+        enum DISABLE_UNSOLICITED_RA {
           description
             "Unsolicted router advertisement messages are not transmitted on
             this interface.  Responses to router solicitation messages will
             be transmitted.";
         }
       }
-      default "ENABLE";
+      default "ALL";
       description
         "Mode controls which set of behaviors the local system should perform
         to support ipv6 router advertisements.";
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -48,7 +48,7 @@ module openconfig-if-ip {
 
   revision "2023-06-30" {
     description
-      "Deprecate ipv6 router advertisment config suppress leaf and add config
+      "Deprecate IPv6 router advertisment config suppress leaf and add config
       mode leaf.";
     reference "3.4.0";
   }
@@ -61,7 +61,7 @@ module openconfig-if-ip {
 
   revision "2023-02-06" {
     description
-      "Add ipv6 link-local configuration.";
+      "Add IPv6 link-local configuration.";
     reference "3.2.0";
   }
 
@@ -950,7 +950,7 @@ module openconfig-if-ip {
       type boolean;
       default true;
       description
-        "If set to false, all ipv6 router advertisement functions are
+        "If set to false, all IPv6 router advertisement functions are
         disabled.  The local system will not transmit router advertisement
         messages and will not respond to router solicitation messages";
     }
@@ -998,7 +998,7 @@ module openconfig-if-ip {
       default "ALL";
       description
         "Mode controls which set of behaviors the local system should perform
-        to support ipv6 router advertisements.";
+        to support IPv6 router advertisements.";
       reference "RFC4861: Neighbor Discovery for IP version 6 (IPv6)";
     }
 
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -1,66 +1,55 @@
 module openconfig-if-ip {
 
-yang-version "1";
+  yang-version "1";
 
-// namespace
-namespace "http://openconfig.net/yang/interfaces/ip";
+  // namespace
+  namespace "http://openconfig.net/yang/interfaces/ip";
 
-prefix "oc-ip";
+  prefix "oc-ip";
 
-// import some basic types
-import openconfig-inet-types {
-    prefix oc-inet;
-}
-import openconfig-interfaces {
-    prefix oc-if;
-}
-import openconfig-vlan {
-    prefix oc-vlan;
-}
-import openconfig-yang-types {
-    prefix oc-yang;
-}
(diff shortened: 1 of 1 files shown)


release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -957,11 +964,41 @@ module openconfig-if-ip {
...
+        "When set to true, router advertisement neighbor discovery messages
+        are not transmitted on this interface.  Responses to router
+        solicitation are transmitted.";
+    }
+
+    leaf mode {
+      type enumeration {
+        enum ENABLE {
+          description
+            "The system should not respond to ARP requests that
```
LimeHat, line 980: Wrong description for `ENABLE` enum?

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -957,11 +964,41 @@ module openconfig-if-ip {
...
+        }
+        enum DISABLE {
+          description
+            "All ipv6 router advertisement functions are disabled.  The
+            local system will not transmit neighbor discovery messages and
+            will not respond to router solicitation messages.";
+        }
+        enum DISABLE_RA {
+          description
+            "Router advertisement neighbor discovery messages are not
```
LimeHat, line 992: Would be good to add `Unsolicited` router advertisement ..., since responses to RS also use RA message format

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -957,11 +964,41 @@ module openconfig-if-ip {
...
+        enum ENABLE {
+          description
+            "The system should not respond to ARP requests that
+            do not specify an IP address configured on the local
+            subinterface as the target address.";
+        }
+        enum DISABLE {
+          description
+            "All ipv6 router advertisement functions are disabled.  The
+            local system will not transmit neighbor discovery messages and
```
LimeHat, line 987: I suggest to replace `neighbor discovery` with `Router Advertisement` to avoid potential confusion with neighbor solicit/neighbor advertisement messages

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -957,11 +964,41 @@ module openconfig-if-ip {
...
     leaf suppress {
+      status deprecated;
       type boolean;
       default false;
       description
-        "When set to true, router advertisement neighbor discovery
-        messages are not transmitted on this interface.";
+        "When set to true, router advertisement neighbor discovery messages
+        are not transmitted on this interface.  Responses to router
+        solicitation are transmitted.";
```
LimeHat, line 973: To me, this feels like a change in the existing behavior; my interpretation of the previous description is that all RAs are suppressed (solicited and unsolicited). Is this the intention?

dplore: One of the motivations for this PR is that the existing description of behavior is a bit vague.  My interpretation of the [RFC4861](https://datatracker.ietf.org/doc/html/rfc4861#section-4 ) is that there are four separate messages in ipv6 neighbor discovery.     I think one can interpret the [current](https://github.com/openconfig/public/blob/e2799c0a9f2d9c977e404d8703971d766f64a23a/release/models/interfaces/openconfig-if-ip.yang#L959-L965) description of 'suppress'  strictly to mean never send any section 4.2 messages, would would include both solicited and unsolicited router advertisement messages as you are suggesting.   

I have also heard interpretation that the current suppress leaf should apply only the unsolicited advertisement and that responses to router solicitation are supported.  I think this is based on one interpreting a response to not be an 'advertisement', even though I think one could say this is contrary to the RFC definition.  

Debating is maybe not very useful though as the current leaf cannot meet operational need for options to suppress solicited  and unsolicited separately.

dplore:
Based on discussions for other model changes (#861 and #902)  regarding  style for OC enable/disable defaults and a list of  options, I have refactored this to use an enable leaf at the container level, followed by a list of options which are relevant only if the feature is enabled.

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -44,7 +44,14 @@ module openconfig-if-ip {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "3.3.0";
+  oc-ext:openconfig-version "3.4.0";
+
+  revision "2023-06-30" {
+    description
+      "Deprecate ipv6 router advertisment config suppress leaf and add config
```
robshakir, line 51: `s/ipv6/IPv6/` here and throughout.

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -957,11 +974,34 @@ module openconfig-if-ip {
     }
 
     leaf suppress {
+      status deprecated;
       type boolean;
       default false;
       description
-        "When set to true, router advertisement neighbor discovery
-        messages are not transmitted on this interface.";
+        "When set to true, router advertisement neighbor discovery messages
```
robshakir, line 981: Changing the description of a leaf that is subsequently deprecated seems an interesting choice. If this leaf is to be removed, then should it not just be that the existing description remains the same and there is a new leaf or set leaves that are created that  implement the modified functionality?

robshakir:
Minor nit, the IOS XR documentation link here is for classic IOS/XE I believe. I think the relevant docs for suppression of RAs are [here](https://www.cisco.com/c/en/us/td/docs/ios_xr_sw/iosxr_r3-7/addr_serv/command/reference/ir37ntsk.html#wp1182427).

release/models/interfaces/openconfig-if-ip.yang
LimeHat: > Debating is maybe not very useful though as the current leaf cannot meet operational need for options to suppress solicited and unsolicited separately.

My comment was more in the context of preserving backwards-compatibility for the leaf: if we clarifying the intended behavior, it seems that some of the implementations will become incompatible with the new definition

dplore: Ok, point well taken!  I have changed the PR to keep the original description for the  suppress leaf.  If there are differences in interpretation of this deprecated leaf it's best resolved by moving to the new model.

release/models/interfaces/openconfig-if-ip.yang
dplore: Reverted this change

dplore:
> Minor nit, the IOS XR documentation link here is for classic IOS/XE I believe. I think the relevant docs for suppression of RAs are [here](https://www.cisco.com/c/en/us/td/docs/ios_xr_sw/iosxr_r3-7/addr_serv/command/reference/ir37ntsk.html#wp1182427).

updated the PR description with the correct IOS XR link

LimeHat:
Thanks for the changes Darren, lgtm.

One more nit about the documentation references: I cannot comment on behalf of Juniper, but based on the command description it seems that `solicit-router-advertisement-unicast` modifies only how solicited RAs are sent (to a unicast vs all-nodes multicast address; both options are allowed in [section 6.2.6 of RFC4861](https://datatracker.ietf.org/doc/html/rfc4861#section-6.2.6)). I don't see anything about suppression of  unsolicited RAs.

rszarecki (approved):

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -957,13 +973,35 @@ module openconfig-if-ip {
     }
 
     leaf suppress {
+      status deprecated;
       type boolean;
       default false;
       description
         "When set to true, router advertisement neighbor discovery
         messages are not transmitted on this interface.";
```
rszarecki, line 981: This leaf is deprecated, but I thik it is stil valid to explain expectation.
Pls. consider adding to description" ..on this interface, neither as unsolicited or as responce to Router Solicitation"

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -939,6 +946,15 @@ module openconfig-if-ip {
     description
       "Configuration parameters for IPv6 router advertisements.";
 
+    leaf enable {
+      type boolean;
+      default true;
+      description
+        "If set to false, all IPv6 router advertisement functions are
+        disabled.  The local system will not transmit router advertisement
+        messages and will not respond to router solicitation messages";
```
xavier-contreras, line 955: nit: add fullstop.

rolandphung (approved):

release/models/interfaces/openconfig-if-ip.yang
dplore: Thanks for the thought here.  I also at first thought updating the deprecated leaf description would be useful.  This was debated lightly here:  https://github.com/openconfig/public/pull/900#discussion_r1254769606

However, thinking through this a bit more, since there are differences in implementations due to the ambiguity here I think it's good enough to leave the description as is.  This is because implementations aren't expected to change their behavior for a deprecated leaf and rather should implement the new leaf and enum in this PR.

robshakir (approved):

PR #904, opened 2023-07-05
Fix typo in ospf lsdb experimental-te leaf description

### LINKED ISSUE #877: file openconfig-ospfv2-lsdb.yang, leaf experimental-te, word ture is not correct in description
file openconfig-ospfv2-lsdb.yang, at leaf experimental-te, description should true, not ture, please modify this, thanks.

dplore:
Fixes #877


robshakir (changes requested):
As this is a submodule, the versions in the parent module and other submodules need to be updated too.

robshakir (approved):

PR #907, opened 2023-07-10
Bias voltage thresholds

jnelson-sj:
### Change Scope

* This change adds support for getting transceiver optical monitoring thresholds for power supply voltage and laser bias current.  Each of these thresholds has a high value and a low value. The transceiver is considered to be in an alarmed state if the relevant parameter is above the high threshold or below the low threshold.
### Platform Implementations

 * Arista EOS: <a href="https://www.arista.com/en/um-eos/eos-transceiver-performance-monitoring#unique_2099204811" rel="nofollow">Documentation for displaying thresholds</a>

## Tree
Changes are in <b>bold</b>

<pre class="notranslate">
module: openconfig-platform-transceiver

  augment /oc-platform:components/oc-platform:component:
    +--rw transceiver
       +--rw config
       +--ro state
       +--rw physical-channels
       +--rw thresholds
          +--ro threshold* [severity]
             +--ro severity    -> ../state/severity
             +--ro state
                +--ro severity?                   identityref
                +--ro laser-temperature-upper?    decimal64
                +--ro laser-temperature-lower?    decimal64
                +--ro output-power-upper?         decimal64
                +--ro output-power-lower?         decimal64
                +--ro input-power-upper?          decimal64
                +--ro input-power-lower?          decimal64
                +--<b>ro laser-bias-current-upper?   decimal64
                +--<b>ro laser-bias-current-lower?   decimal64
                +--<b>ro supply-voltage-upper?       decimal64
                +--<b>ro supply-voltage-lower?       decimal64
  </pre>


dplore:
Will discuss in OC Operators meeting July 18, 2023

dplore:
This was reviewed today.  @oscargdd asked to clarify which attribute should be used for the threshold comparison?  The descriptions for the container or leaf should specify if mix/max/avg/instant should be used.  I expect it is `instant`, but it is good to explicitly define this in the description for the container (as I think all the leaves would operate on the `instant` value).

dplore (changes requested):
This was reviewed today. @oscargdd asked to clarify which attribute should be used for the threshold comparison? The descriptions for the container or leaf should specify if mix/max/avg/instant should be used. I expect it is instant, but it is good to explicitly define this in the description for the container (as I think all the leaves would operate on the instant value).

jnelson-sj:
Added comment in the threshold container:  "Each threshold is compared to the instant value of the measured parameter corresponding to the threshold";

dplore (approved):

PR #908, opened 2023-07-11
bootz skip ownership check

xw-g:


release/models/system/openconfig-system-bootz.yang
```diff
@@ -38,15 +44,34 @@ module openconfig-system-bootz {
...
 
+  grouping validate-ownership {
+    description
+      "Configuration related to ownership validation";
+    leaf validate-ownership {
+      type boolean;
+      default true;
+      description
+        "This field controls if the device should validate the ownership in
+        the next bootz workflow";
```
dplore, line 71:
```suggestion
        "This field controls if the device should validate the ownership
        voucher as described in the bootz bootstrapping service API
        flow.";
```

release/models/system/openconfig-system-bootz.yang
```diff
@@ -19,10 +19,16 @@ module openconfig-system-bootz {
     service running on a network device.";
```
dplore, line 20: Can you also add a reference field to https://github.com/openconfig/bootz ?

release/models/system/openconfig-system-bootz.yang
xw-g: I denied your suggestions, because we need to call it out that it's for next bootz workflow that runs.

release/models/system/openconfig-system-bootz.yang
xw-g: done.

release/models/system/openconfig-system-bootz.yang
dplore: I was attempting to be more precise on two dimensions:
- This leaf refers specifically to the [bootz bootstrapping service](https://github.com/openconfig/bootz#api-flow) which is a specific step within the bootz API Flow.  (although admittedly, the bootz specification specifies a section header as API Flow but then also uses the words 'workflow' occasionally inside the API Flow section, I guess API Flow and Workflow are the same thing?). 

- This setting should be honored for every invocation of the [bootstrapping service](https://github.com/openconfig/bootz#api-flow), not just the 'next' workflow?  I guess since the configuration is persistent, the 'next' workflow could be considered all future workflows.

xw-g: The parent container is Bootz, so the all the leaves here is about the Bootz spec.

dplore:
@marcushines @xw-g do you want to continue with this PR?  If yes please review

PR #912, opened 2023-07-13
Fix CI push and change pyangbind to run against its master branch as a compatibility report check

wenovus:
* inline comments not supported
* decrease runtime for pyangbind

Prerequisite PR: https://github.com/openconfig/models-ci/pull/84

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -52,8 +52,7 @@ steps:
     git clone git@github.com:openconfig/models-ci.git /go/src/github.com/openconfig/models-ci
     cd /go/src/github.com/openconfig/models-ci
     # Modify the major version to update models-ci version.
-    #branch=$(git tag -l 'v9*' | sort -V | tail -1)
-    branch=local-pyangbind-plugin
+    branch=$(git tag -l 'v10.*' | sort -V | tail -1)
     git checkout $branch
   volumes:
   - name: 'ssh'
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -91,7 +91,7 @@ steps:
     -skipped-validators=confd,yanglint
     -extra-pyang-versions=2.2.1
     -branch=$BRANCH_NAME
-    #-compat-report=pyangbind
+    -compat-report=pyangbind
   secretEnv: ['GITHUB_ACCESS_TOKEN']
   env:
   - 'GOPATH=/go'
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -293,7 +292,7 @@ steps:
     path: /go
   id: 'compat-report'
 
-timeout: 600s
+timeout: 900s
```
dplore, line 295: wow, cloudbuild is taking more than 10 minutes?  That is too slow for a CI check.  @bstoll is this latency  expected?

wenovus: The latency is from pyangbind's code generator trying to generate all the code in OC. Given it's python I think it makes sense.

The generation is already being done in parallel:
https://github.com/openconfig/models-ci/blob/3f6a71eba144cc3aa4b72ab5110e5064644d8b9d/cmd_gen/main.go#L301C14-L301C14

wenovus: Alright, for some reason if I use a local version of pyangbind instead of the one downloaded into the Python virtual env cache for pyang's `--plugindir` it halves the time.

Total build time is now 7min

wenovus:
@robshakir Are you ok with having pyangbind in the compat-report, or should we fix that first?

cloudbuild.yaml
```diff
@@ -89,8 +90,8 @@ steps:
     -pr-number=$_PR_NUMBER
     -skipped-validators=confd,yanglint
     -extra-pyang-versions=2.2.1
-    #-compat-report=yanglint
     -branch=$BRANCH_NAME
+    -compat-report=pyangbind
```
robshakir, line 94: Are we now making pyangbind non-blocking?

wenovus: There is an error in the BFD model: https://gist.github.com/OpenConfigBot/f2158a52bc907ed4e690582d47f2af61#gistcomment-4628801

I want to make it unblocking until we resolve it.

wenovus: Opened an issue in pyangbind: https://github.com/robshakir/pyangbind/issues/302

cloudbuild.yaml
wenovus: I will now keep pyangbind running against its main branch instead of what's published on pip

dplore (approved):

PR #914, opened 2023-07-17
Add bytes-sampled field to sflow sampling model

pandeytejas:
### Change Scope

* release/models/sampling/openconfig-sampling-sflow.yang
  * Added a new field bytes-sampled to indicate an approximate estimate of the total bytes sampled.


dplore:
Please provide references for implementations.  At a glance, I don't see counters for this from a few major implementations, but I might have overlooked their implementation.

dplore:
@pandeytejas please do fix the miscellaneous checks and add references, I'd like to see this one get merged.

PR #925, opened 2023-07-25
Update openconfig-network-instance-policy.yang

rszarecki:
### Change Scope

* Expanding description of /routing-policy/policy-definitions/policy-definition/statements/statement/conditions/match-protocol-instance/config/protocol-name leaf by providing bit more context of expected values and relation to other configuration elements.
* 
### Platform Implementations

N/A


dplore (approved):
LGTM

PR #928, opened 2023-07-26
Temporal qos buffer size definition

rszarecki:
### Change Scope

Adding buffer management leafs for minimum buffer allocation and maximum static shared buffer alocatione expressed in microseconds. Thie enables same buffer menegermet profile to be reusable across: 
 - interface of different speed 10GE vs 100GE vs. 400GE
 - LAG interfaces of different speed
 It also keep queueing latency consistent on LAG which capacity changes due to some member link failures, repairs and augmentation.

The actual memory allocated to physical interface is calculated as product:\
\
              ***temporal_leaf * 10^(-6) * PHY_interface_speed * share***\
                    where ***share*** is calculated as ***queue_WRR_weight/Sum(all WRR_weights)***
```
module: openconfig-qos
  +--rw qos
     +--rw buffer-allocation-profiles
        +--rw buffer-allocation-profile* [name]
           +--rw queues
              +--rw queue* [name]
                 +--rw config
                 |  +--rw name?                                  -> ../../../../../../queues/queue/config/name
                 |  ...
                 |  +--rw dedicated-buffer-temporal?             uint64
                 |  ...
                 |  +--rw static-shared-buffer-limit-temporal?   uint32
                 |  +...
                 +--ro state
                    +--ro name?                                  -> ../../../../../../queues/queue/config/name
                    ...
                    +--ro dedicated-buffer-temporal?             uint64
                    ...
                    +--ro static-shared-buffer-limit-temporal?   uint32
                    ...
```

The change is backward compatible.

### Platform Implementations

 * ARISTA EOS [link](https://www.arista.com/en/support/toi/eos-4-23-0f/14326-maximum-latency-tail-drop-thresholds)
 ```
(config)#qos profile latency
(config-qos-profile-latency)#tx-queue 3
(config-qos-profile-latency-txq-3)#latency maximum <1-50000> microseconds
```
 * JUNIPER JUNOS [link](https://www.juniper.net/documentation/us/en/software/junos/cos/topics/ref/statement/buffer-size-edit-cos.html)
 ```
[schedulers]
{
    scheduler-name {
        buffer-size  temporal <microseconds>;
    }
}
 ```
 * CISCO IOS-XR [link](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/qos/b-qos-cr-cisco8000/b-qos_chapter_01.html#wp3222517277)
 ```
queue-limit <value> us
 ```

### DIFF
--- a/release/models/qos/openconfig-qos-interfaces.yang
+++ b/release/models/qos/openconfig-qos-interfaces.yang
@@ -369,13 +369,13 @@ submodule openconfig-qos-interfaces {
     leaf dropped-pkts {
       type oc-yang:counter64;
       description
-        "Number of packets dropped by the queue due to overrun, that is tail-drop or WRED drop";
+        "Number of packets dropped by the queue due to overrun, that is tail-drop or AMQ (RED, WRED, etc) indudeced drops as indicated by the attached queue-management-profile";
     }
 
     leaf dropped-octets {
       type oc-yang:counter64;
       description
-        "Number of octets dropped by the queue due to overrun, that is tail-drop or WRED drop";
+        "Number of octets dropped by the queue due to overrun, that is tail-drop or AMQ (RED, WRED, etc) indudeced drops as indicated by the attached queue-management-profile";
     }
   }
 
--- a/release/models/qos/openconfig-qos-interfaces.yang
+++ b/release/models/qos/openconfig-qos-interfaces.yang
@@ -25,13 +25,7 @@ submodule openconfig-qos-interfaces {
     configuration and operational state associated with
     interfaces.";
 
-  oc-ext:openconfig-version "0.10.0";
-
-  revision "2023-07-13" {
-    description
-      "Add support for aged-packet, aged-byte conters";
-    reference "0.10.0";
-  }
+  oc-ext:openconfig-version "0.9.1";
 
   revision "2023-04-25" {
     description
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: release/models/qos/openconfig-qos-interfaces.yang [hunk 1]
```diff
@@ -362,13 +369,13 @@ submodule openconfig-qos-interfaces {
     leaf dropped-pkts {
       type oc-yang:counter64;
       description
-        "Number of packets dropped by the queue due to overrun";
+        "Number of packets dropped by the queue due to overrun, that is tail-drop or WRED drop";
```
dplore, line 372: I think overrun clearly is tail-drop.   But I am not sure that a queue overrun is the same as a WRED drop.  Do we need to include this clarification in this PR?  I suggest this clarification is a separate PR as I think this deserves some additional impact analysis. (ie: do 2+ platforms support this interpretation?  Do we need this clarification?)

dplore (changes requested):

release/models/qos/openconfig-qos-interfaces.yang
rszarecki: There is no other counter that counts WRED dropped packets. And for sure we want dropped packets to be counted.
Unless we want to define new counters, abouve is only resonable interpretation - IMHO.
If we are going to define new drop couters we should have explicit Tail-drop and WRED-drop. Perhaps at this time we should introduce support for "colored" WRED (like 3 WRED slopes per queue - follow IETF PHBs AFx.1, AFx.2, AFx.3) and corresponding drop counters.
IMHO making existing and only per-queue drop counters (`dropped-pkts`, `dropped-octets`) and aggregated couter for drop regardless of cause the best option for now.

dplore: OC Operators comment that this description update should also mention that the queue-management profile determines if RED/WRED is configured.  For example:

"Number of packets dropped by the queue due to overrun, that is tail-drop or RED/WRED if indicated by the queue-management-profile"

release/models/qos/openconfig-qos-interfaces.yang
```diff
@@ -369,13 +369,13 @@ submodule openconfig-qos-interfaces {
     leaf dropped-pkts {
       type oc-yang:counter64;
       description
-        "Number of packets dropped by the queue due to overrun, that is tail-drop or WRED drop";
+        "Number of packets dropped by the queue due to overrun, that is tail-drop or AMQ (RED, WRED, etc) indudeced drops as indicated by the attached queue-management-profile";
```
dplore, line 372: Suggestion to follow style guide on line length and spelling correction. 

```suggestion
        "Number of packets dropped by the queue due to overrun, that is tail-drop
        or AMQ (RED, WRED, etc) induced drops as indicated by the attached
        queue-management-profile";
```

dplore (changes requested):
Thanks for the description update.  Added suggestion to follow style guide on line length and spelling correction.

dplore (approved):
LGTM

PR #905, opened 2023-07-07
Add multicast counters to ip interface, remove physical counters from subinterface

dplore:
### Change Scope

* Refactor counters for interfaces and subinterfaces to use a common grouping for counters used for both interfaces and subinterfaces
* Add multicast-pkt counters to common interface counters
* Add resets counter for interface grouping
* Deprecate in-fcs-error, in-unknown-protosand  carrier transitions from subinterfaces
* Removed TODO item regarding additional error counters
* This change is backwards compatible

```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--ro state
        |  +--ro counters
        |  |  +--ro in-octets?             oc-yang:counter64
        |  |  +--ro in-pkts?               oc-yang:counter64
        |  |  +--ro in-unicast-pkts?       oc-yang:counter64
        |  |  +--ro in-broadcast-pkts?     oc-yang:counter64
        |  |  +--ro in-multicast-pkts?     oc-yang:counter64
        |  |  +--ro in-errors?             oc-yang:counter64
        |  |  +--ro in-discards?           oc-yang:counter64
        |  |  +--ro out-octets?            oc-yang:counter64
        |  |  +--ro out-pkts?              oc-yang:counter64
        |  |  +--ro out-unicast-pkts?      oc-yang:counter64
        |  |  +--ro out-broadcast-pkts?    oc-yang:counter64
        |  |  +--ro out-multicast-pkts?    oc-yang:counter64
        |  |  +--ro out-discards?          oc-yang:counter64
        |  |  +--ro out-errors?            oc-yang:counter64
        |  |  +--ro last-clear?            oc-types:timeticks64
        |  |  +--ro in-unknown-protos?     oc-yang:counter64
        |  |  +--ro in-fcs-errors?         oc-yang:counter64
        |  |  +--ro carrier-transitions?   oc-yang:counter64
        |  |  +--ro resets?                oc-yang:counter64.       <-- added
        +--rw subinterfaces
        |  +--rw subinterface* [index]
        |     +--ro state
        |     |  +--ro counters
        |     |     +--ro in-octets?             oc-yang:counter64
        |     |     +--ro in-pkts?               oc-yang:counter64
        |     |     +--ro in-unicast-pkts?       oc-yang:counter64
        |     |     +--ro in-broadcast-pkts?     oc-yang:counter64
        |     |     +--ro in-multicast-pkts?     oc-yang:counter64        <-- added
[...]

PR #933, opened 2023-07-28
Revert "Add multicast counters to ip interface, remove physical counters from subinterface"

dplore:
Reverts openconfig/public#905

#905 unintentionally removed counters from IP subinterfaces.    A new PR will be introduced to fix this.

FYI, this breaking change was detected by the following [featureprofiles CI check](https://github.com/openconfig/featureprofiles/actions/runs/5695540733/job/15439270834?pr=1946).  We will look into exposing public model breaking changes via checks like this.


robshakir:
revert lgtm.

robshakir (approved):

dplore:
CI fail check is expected for version numbering because this is reverting to older version for
openconfig-if-ip.yang and openconfig-interfaces.yang

PR #936, opened 2023-08-01
Add breaking change check

wenovus:
The check fails if there is a backward-incompatible change that isn't accompanied by a major version update (if not at v0).
This check is **not** exhaustive, currently it checks for the following:
* leaf deletion
* leaf type change

There are other types of backward-incompatible changes not caught, e.g.
* string pattern/number range restrictions becoming more restrictive.

Note: The "breaking"/"non-breaking" labels are still "dumb", i.e. they're determined by whether there are any major YANG version updates.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -53,6 +53,7 @@ steps:
     cd /go/src/github.com/openconfig/models-ci
     # Modify the major version to update models-ci version.
     branch=$(git tag -l 'v11.*' | sort -V | tail -1)
+    branch=fix-ygot
     git checkout $branch
   volumes:
   - name: 'ssh'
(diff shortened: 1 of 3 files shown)


cloudbuild.yaml
```diff
@@ -52,7 +52,8 @@ steps:
     git clone git@github.com:openconfig/models-ci.git /go/src/github.com/openconfig/models-ci
     cd /go/src/github.com/openconfig/models-ci
     # Modify the major version to update models-ci version.
-    branch=$(git tag -l 'v10.*' | sort -V | tail -1)
+    branch=$(git tag -l 'v11.*' | sort -V | tail -1)
+    branch=c880ed2
```
dplore, line 56: Is this intentional?

wenovus: Oh no it's not, let me fix that.

dplore (approved):

PR #938, opened 2023-08-09
Update string type keys to use interface-id type

rolandphung:
### Change Scope
- Fix certain keys that were typed to `string` when that should be `interface-id` type.
- The goal is to make these `interface-ref` and `interface-id` usage to be more consistent
    - `interface-id` underlying type is `string` so there shouldn't be any issues

Paths updated:
* /bfd/interfaces/interface/id
* /network-instances/network-instance/mpls/signaling-protocols/segment-routing/interfaces/interface/interface-id
* /network-instances/network-instance/interfaces/interface/id
* /network-instances/network-instance/protocols/protocol/ospfv2/areas/area/interfaces/interface/id

### DIFF
--- a/release/models/bfd/openconfig-bfd.yang
+++ b/release/models/bfd/openconfig-bfd.yang
@@ -26,7 +26,13 @@ module openconfig-bfd {
     "An OpenConfig model of Bi-Directional Forwarding Detection (BFD)
     configuration and operational state.";
 
-  oc-ext:openconfig-version "0.2.5";
+  oc-ext:openconfig-version "0.2.6";
+
+  revision "2023-08-09" {
+    description
+      "Update interface key to use interface-id type";
+    reference "0.2.6";
+  }
 
   revision "2023-02-06" {
     description
@@ -194,7 +200,7 @@ module openconfig-bfd {
       "Top-level per-interface configuration parameters for BFD.";
 
     leaf id {
-      type string;
+      type oc-if:interface-id;
       description
         "A unique identifier for the interface.";
     }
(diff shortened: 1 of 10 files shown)


dplore:
Yes, this should be a non-breaking change since the underlying type is still string.  Will review at Aug 15 OC Operators meeting

earies:
> Yes, this should be a non-breaking change since the underlying type is still string. Will review at Aug 15 OC Operators meeting

For this comment to be true, a set of guidelines of what the OpenConfig community feels is backwards compatible vs. not is in order since this (and many previous changes to the model set) do not abide by YANG 1.0/1.1 rules on backwards compatibility.  I think history of versioning would show it's a bit subjective in nature to date.

An underlying type can be the same but various forms of codegen (and whatever uses the generated code) can very well break.  In addition, a set of rules can be implicitly imposed by changing types even if the underlying type remains identical.

dplore:
We do know in this case that ygot will not break.  In ygot all types get recursively resolved to their built-in type.  @wenovus for any additional thoughts or comments.

wenovus:
This change is backwards-compatible for ygot since it uses the same types to represent built-in types and derived types resolving to the same ones. RFC7951 and gnmi.proto do not mention derived YANG types at all, so my conclusion is that this same rule also applies from a marshalling perspective.

From a YANG perspective since `interface-id` is an unrestrained string then there is no concern regarding type restriction either.

@JoseIgnacioTamayo or @xavier-contreras on how pyangbind treats YANG derived types, and whether this change would break users.

dplore (approved):

PR #939, opened 2023-08-11
Add hashing model

atmanmehta:
### Change Scope

* Add hashing model to OC. There will be leaf-refs added to interfaces and system models to apply the policies for LAG and L3 ECMP.
* Change is backward compatible.

### DIFF
--- a/release/models/hashing/openconfig-hashing.yang
+++ b/release/models/hashing/openconfig-hashing.yang
@@ -1,0 +1,164 @@
+module openconfig-hashing {
+    yang-version "1";
+
+    // namespace
+    namespace "http://openconfig.net/yang/hashing";
+
+    prefix "oc-hashing";
+
+    // import some basic types
+    import openconfig-extensions { prefix oc-ext; }
+    import openconfig-packet-match { prefix oc-pkt-match; }
+
+    // meta
+    organization "OpenConfig working group";
+
+    contact
+      "OpenConfig working group
+      netopenconfig@googlegroups.com";
+
+    description
+      "Model for managing hashing policies that would be referenced by the
+      interfaces model.
+
+
+    oc-ext:openconfig-version "0.1.0";
+
+    revision "2023-08-08" {
+        description "Initial hashing model.";
+        reference "0.1.0";
+    }
+
+    // OpenConfig specific extensions for module metadata.
+    oc-ext:regexp-posix;
+    oc-ext:catalog-organization "openconfig";
+    oc-ext:origin "openconfig";
+
+    grouping hashing-inputs {
+        description
+          "Top level container for inputs to be used for hashing policies.";
+
+        container config {
+            description
+              "Configurable items at the hashing inputs level.";
+            uses oc-pkt-match:ipv4-protocol-fields-config;
+            uses oc-pkt-match:ipv6-protocol-fields-config;
+            uses oc-pkt-match:transport-fields-config;
+        }
+
+        container state {
+            config false;
+            description
+              "Operational state data at the hashing inputs level.";
+            uses oc-pkt-match:ipv4-protocol-fields-config;
+            uses oc-pkt-match:ipv6-protocol-fields-config;
+            uses oc-pkt-match:transport-fields-config;
+        }
+    }
+
+    grouping hashing-policy-config {
+        description
+          "Configuration data for hashing policies.";
+
+        leaf name {
+            type string;
+            description
+              "The name of the hashing policy.
+
+              When a configured user-controlled policy is created by the
+              system, it is instantiated with the same name in the
+              /policies/policy[name]/state list.";
+        }
+
+        leaf algorithm {
+            default CRC;
+            type enumeration {
+                enum CRC {
+                    description
+                      "CRC hashing algorithm.";
+                }
+                enum CRC_32LO {
+                  description
+                    "CRC_32LO hashing algorithm.";
+                }
+                enum CRC_32HI {
+                  description
+                    "CRC_32HI hashing algorithm.";
+                }
+                enum CRC_CCITT {
+                  description
+                    "CRC using CCITT polynomial based hash algorithm.";
+                }
+                enum CRC_XOR {
+                  description
+                    "Combination of CRC and XOR based hash algorithm.";
+                }
+                enum XOR {
+                  description
+                    "XOR-based hash algorithm.";
+                }
+                enum RANDOM {
+                  description
+                    "Random-based hash algorithm.";
+                }
+            }
+            description
+              "The algorithm used when calculating a hash based on packet headers";
+        }
+
+        leaf offset {
+            type uint64;
+            description
+              "The offset used when calculating the hash algorithm";
+        }
+
+        leaf seed {
+            type uint64;
+            description
+              "The seed used to initialize the hash algorithm";
+        }
+
+        uses hashing-inputs;
+    }
+
+    grouping hashing-top {
+        description
+          "Top level grouping for hashing configuration and operational state
+          data.";
+
+        container policies {
+            description
+              "Top level container for hashing, including configuration and
+              state data.";
+
+            list policy {
+                key "name";
+
+                description
+                  "The list of named policies to be used on the device.";
+
+                leaf name {
+                    type leafref {
+                        path "../config/name";
+                    }
+                    description
+                      "References the name of the hashing policy.";
+                }
+                container config {
+                    description
+                      "Configurable items at the global hash policy level.";
+                    uses hashing-policy-config;
+                }
+                container state {
+                    config false;
+                    description
+                      "Operational state data at the global hash policy
+                      level.";
+
+                    uses hashing-policy-config;
+                }
+            }
+        }
+    }
+}
+
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ This repository is primarily for publishing the models, documents, and other
 material developed by the OpenConfig operators group.
 
 For information about how to contribute to OpenConfig models, please
-see [External Submissions to OpenConfig](doc/external-contributions-guide.md).
+see [External Submissions to OpenConfig](doc/contributions-guide.md).
 
 Feedback and suggestions to improve OpenConfig models is welcomed on the
 [public mailing list](https://groups.google.com/forum/?hl=en#!forum/netopenconfig),
--- a/release/models/optical-transport/openconfig-transport-types.yang
+++ b/release/models/optical-transport/openconfig-transport-types.yang
@@ -24,9 +24,9 @@ module openconfig-transport-types {
 
   oc-ext:openconfig-version "0.16.0";
 
-  revision "2022-09-20" {
+  revision "2022-10-18" {
     description
-      "Add ETH_400GBASE_PSM4 PMD type";
+      "Add ETH_400GMSA_PSM4 PMD type";
     reference "0.16.0";
   }
 
@@ -1050,9 +1050,9 @@ module openconfig-transport-types {
     description "Ethernet compliance code: 400GBASE_DR4";
   }
 
-  identity ETH_400GBASE_PSM4 {
+  identity ETH_400GMSA_PSM4 {
     base ETHERNET_PMD_TYPE;
-    description "Ethernet compliance code: 400GBASE_PSM4";
+    description "Ethernet compliance code: 400GMSA_PSM4";
   }
 
   identity ETH_UNDEFINED {
--- a/release/models/optical-transport/openconfig-transport-types.yang
+++ b/release/models/optical-transport/openconfig-transport-types.yang
@@ -22,7 +22,13 @@ module openconfig-transport-types {
     "This module contains general type definitions and identities
     for optical transport models.";
 
-  oc-ext:openconfig-version "0.15.0";
+  oc-ext:openconfig-version "0.16.0";
+
+  revision "2022-09-20" {
+    description
+      "Add ETH_400GBASE_PSM4 PMD type";
+    reference "0.16.0";
+  }
 
   revision "2021-07-29" {
     description
@@ -1042,6 +1048,11 @@ module openconfig-transport-types {
   identity ETH_400GBASE_DR4 {
     base ETHERNET_PMD_TYPE;
     description "Ethernet compliance code: 400GBASE_DR4";
+  }
+
+  identity ETH_400GBASE_PSM4 {
+    base ETHERNET_PMD_TYPE;
+    description "Ethernet compliance code: 400GBASE_PSM4";
   }
 
   identity ETH_UNDEFINED {


earies:
Can you provide implementation references as is in the PR template and not delete that section

https://github.com/openconfig/public/blob/master/.github/pull_request_template.md

Especially with respect to configurable algorithms, seeds, policies and the application of these policies to other specific constructs

atmanmehta:
> 

@dplore mentioned he had some implementations references that he would help add here.

dplore:
## Implementations 
* [Arista EOS - port-channel load-balance](https://www.arista.com/en/um-eos/eos-port-channels-and-lacp#xx1151429) - defines a variety of configuration items which are specific to different hardware platforms.  They include references to hash policies with attributes including hash polynomials, hash seeds, and hash fields
* [JunOS - per flow load balancing on hash values](https://www.juniper.net/documentation/us/en/software/junos/sampling-forwarding-monitoring/topics/concept/policy-configuring-per-flow-load-balancing-based-on-hash-values.html) defines a way to configure a hash seed which is limited to a small set of hardware platforms.  [forwarding-options/hash-key](https://www.juniper.net/documentation/us/en/software/junos/interfaces-ethernet-switches/topics/topic-map/switches-interface-load-balancing.html#id-configuring-load-balancing-on-a-lag-link)  allow configuration of per-prefix and mac address based balancing.  
* [Cisco IOS-XR bundle-hash](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/Interfaces/b-interfaces-hardware-component-cr-8000/m-link-bundling-commands-8k.html) allows configuring the hash key based on a list of fixed options  
  * (L3/3-tuple or L4/7-tuple) (L3,L4):
  * Single SA/DA pair (IPv4,IPv6) or range (IPv4 only): S/R [S]:
* [Nokia SR Linux system/load-balancing/hash-options](https://documentation.nokia.com/srlinux/22-3/SR_Linux_Book_Files/Configuration_Basics_Guide/configb-systmmgmt.html#ariaid-title13) allows hash seed and some hash key configuration support.

earies:
Thx for the references @dplore - looking across the implementation references, this PR is appearing to cater to a single implementation by enforcing user-defined hash policies with a superset (looks like some design patterns are off a bit here too w/ nested config/config and state/config containers).  

I can't speak for all implementations but it appears various are going to be system wide parameters with only a subset of what is defined in here that is applicable.

Would be curious to hear from @rgwilton, @hellt, @jsterne on this one

jhaas-pfrc:
@earies has already addressed some of my concerns that this is overly platform specific.  Subsets of the parameters may be able to be generically modeled, but in many circumstances it seems likely that they'd be platform or even line-card specific.  Given OC's dislike of if-feature, there's not a clean way to model such things generically.

A more specific bit of feedback is that the "algorithm" section is likely very implementation specific.  "CRC" against what fields of the forwarding keys?  What flavor of CRC?  Etc.  

For such a field, it'd be more likely to be able to be genericized if the type was an identity and it was implementation specific with explicit documentation as to what fields are being impacted.

For scenarios where a given algorithm was common across implementations, standardizing an identity in the base model would be better than multiple flavors of the same.

hellt:
adding @limehat

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,176 @@
...
+
+    grouping hashing-inputs {
+        description
+          "Top level container for inputs to be used for hashing policies.";
+
+        container config {
+            description
+              "Configurable items at the hashing inputs level.";
+            container ipv4 {
+               uses oc-pkt-match:ipv4-protocol-fields-config;
```
dplore, line 45: This grouping configures an ipv4 address value which I am not sure is what is needed?

Is the intent to select if ipv4 source and/or destination address fields should be used  for hash computation?  If yes, probably something like this will work:

hash-field-modes/hash-field-mode/ipv4 =  enum(none, src_address, dst_address, both, target_defined)
hash-field-modes/hash-field-mode/ipv6 =  enum(none, src_address, dst_address, both,  flow_label,  target_defined)
hash-field-modes/hash-field-mode/ipv6 =  enum(none, src_address, dst_address, both,  target_defined)

hash-field-modes/hash-field-mode/transport_ports =  enum(none, srcport, dstport, both,  target_defined)
hash-field-modes/hash-field-mode/transport_ports =  enum(none, srcport, dstport, both,  target_defined)

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,176 @@
...
+                }
+            }
+            description
+              "The algorithm used when calculating a hash based on packet headers";
+        }
+
+        leaf offset {
+            type uint64;
+            description
+              "The offset used when calculating the hash algorithm";
```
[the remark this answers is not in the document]
atmanmehta: Updated the description. Please let me know if it should be reworded further.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,176 @@
...
+            description
+              "Configurable items at the hashing inputs level.";
+            container ipv4 {
+               uses oc-pkt-match:ipv4-protocol-fields-config;
+            }
+            container ipv6 {
+                uses oc-pkt-match:ipv6-protocol-fields-config;
+            }
+            container transport {
+                uses oc-pkt-match:transport-fields-config;
```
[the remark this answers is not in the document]
atmanmehta: Refactored the entire hashing-inputs grouping as discussed offline. It should more accurately reflect the intent now.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,176 @@
...
+    oc-ext:origin "openconfig";
+
+    grouping hashing-inputs {
+        description
+          "Top level container for inputs to be used for hashing policies.";
+
+        container config {
+            description
+              "Configurable items at the hashing inputs level.";
+            container ipv4 {
```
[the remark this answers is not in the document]
atmanmehta: Would it be acceptable to start with this basic set of inputs and augment in the future as required?

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,176 @@
...
+    // OpenConfig specific extensions for module metadata.
+    oc-ext:regexp-posix;
+    oc-ext:catalog-organization "openconfig";
+    oc-ext:origin "openconfig";
+
+    grouping hashing-inputs {
+        description
+          "Top level container for inputs to be used for hashing policies.";
+
+        container config {
```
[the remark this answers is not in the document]
atmanmehta: Refactored this entire sub-tree. This comment is now obsolete.

release/models/hashing/openconfig-hashing.yang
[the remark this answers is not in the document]
dplore: Yes, it seems it's easy to add additional 'hash-field-modes' as non-breaking changes to this model.

dplore:
Thanks for the changes @atmanmehta .  They address all the issues so far.  I have added this PR to the [OC Community meeting agenda](https://docs.google.com/document/d/11sXUF2zujhbPs_bt-RFJhh2DcIbEvnOpd9GypE-u3cc/edit#heading=h.xzit9nc89pp) and will present there to promote greater visibility.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+        leaf ipv6 {
+            type hash-field-type-ipv6;
+            description "The IPv6 fields that should be used to
+                compute the hash.";
+        }
+        leaf transport-ports {
+            type hash-field-type-transport;
+            description "The transport layer fields that should be used to
+                compute the hash.";
+        }
```
rszarecki, line 142: I would like to see additionaly:
IPv4_in_IPv4 and IPv6-in-IPv4 where  each lement is again list of outher and inner fields:
ipv4_ipv4/outherfields/outherfield type hash-field-type-ipv4
ipv4_ipv4/innerfields/innerfield type hash-field-type-ipv4
ipv6_ipv4/outherfields/outherfield/ipv4 hash-field-type-ipv6
ipv6_ipv4/innerfields/innerfield/ipv6 hash-field-type-ipv6

---------
Similiarly fot  ip-over-GRE over IPv4/IPv6,
grev4/outherfields/outherfield type hash-field-type-ipv4
grev4/payloads/payload enum(ipv4 ipv6 mpls Eth) // use fields deviend for basic v4,v6,mpls,eth depending on GRE protocol. if not on list 
grev6/outherfields/outherfield type hash-field-type-ipv6
grev6/payloads/payload enum(ipv4 ipv6 mpls Eth) // use fields deviend for basic v4,v6,mpls,eth depending on GRE protocol. if not on list 

--------
Also for MPLS
mpls/labels - number of labels to use
mpls/base enum bottom XOR top default top
mpls/payloads/payload enum (ipv4 ipv6 PWE3 ipv4_grev4 ipv6_grev6 ipv4_ipv4 ipv6_ipv4 ...)

----------
Also Eth- usefull for PWE3 services or L2 switching devices
eth/fields/fild enum (srcMAC,dstMAC,Vlan-ID, EtherType/payload)
eth/payloads/payload enum ((ipv4 ipv6 PWE3 ipv4_grev4 ipv6_grev6 ipv4_ipv4 ipv6_ipv4 ...)

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+              When a configured user-controlled policy is created by the
+              system, it is instantiated with the same name in the
+              /policies/policy[name]/state list.";
+        }
+
+        leaf algorithm {
+            type string;
+            description
+              "The algorithm used when calculating a hash based on packet headers";
+        }
```
rszarecki, line 163: I would rather define
components/component/integrated-circuit/hash-algoritms/hash-algo/index
components/component/integrated-circuit/hash-algoritms/hash-algo/state/index
components/component/integrated-circuit/hash-algoritms/hash-algo/state/description
ALL of ABOVE of READ ONLY. to be populated by NOS/implemetation. Self-documentation.

Then here just reference index of above. And make a statement that if omited, then target_default applies.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+
+        leaf offset {
```
rszarecki, line 165: if this is rotation, then call it "rotation".

Also it is worth to document what if this value is higher then hash-algo output. For example 
if leaf value is 2^20 (that is 1,048,576) and hash is CRC-32 what it men rotate 32 bits 1,048,576 ? which direction? I can rotate by at most 31 is not?

Is ther any implementation that alows for configuration of rotation? If not what is the utility?

release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -432,6 +438,14 @@ module openconfig-interfaces {
...
     }
 
+    leaf hashing-policy {
+        type leafref {
+            path "/oc-hashing:policies/oc-hashing:policy/oc-hashing:name";
+        }
+        description "The hashing policy to be used when hashing packets coming
+            in on the interface.";
+    }
+
```
rszarecki, line 448: I only see that Arista allows for some hash configuration on per ingress interfaces basis. And it is not clear form referenced documentation on which platform/chipset.

rszarecki (commented):
In addition to all comments, considering havy dependency on actula INTEGRATED-CIRCUIT (ASIC, NPU) I would suggest that policy shouyld be attachable to:
/components/component/linecard/config/hashing-policy (chassis may have mix of ASICs)
/components/component/integrated-circuit/config/hashing-policy

Considering above, perhaps instead of `/system/config/hashing-policy` we should use `/components/component/chassis/config/hashing-policy` ?

Then all atachment,but per-ingress-interface, are under `/component/components/...`. And per-interface I do not see to be widly supported anyway.

rszarecki:
There is one more area not explored.
When hash-function output is calculated it is a n-bit value. Say 16-bit. There is a mapping function from 2^16 potential values to N (weighted-)load-balanced  forwarding-hops.
This could be 2 stage process with fixed buckets in between (say 1024). So 2^16 is mapped to 1000 bucket which then are mapped to N fwd-nh.
Some implementation alows to interfere with it. 2 simple example would be use 10 LSB of hash output or 10 MSB.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+            description
+                "The number of bits to rotate the output of the hash algorithm by
+                in order to get the final computed hash value";
+        }
+
+        leaf seed {
+            type uint64;
+            description
+              "The seed used to initialize the hash algorithm";
+        }
```
rszarecki, line 176: I would ass 2 boolena leafs:
symmetric - ensured that traffic of bi-dir session produce same hash. (e.g. src/dst fields are ordered before feeding hash function)
persistent - ensures that if fwd-nh (e.g. LAG member) fails, flows that was on other members are not re-hashed. only affected flows are re-distributed among remaining fwd-nh.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+
```
rszarecki, line 177:
```suggestion
        leaf symmetric {
            type boolean;
            default false;
            description
              "Hash is sematrical - returns same value regardless of traffic direction of same flow/session. Also regardless of ingress interface";
        }
        leaf persistent {
            type boolean;
            default false;
            description
              "In case of output inteface/forwarding-nh failure/removal, only flows that have used this interface/nexthop are re-hashed";
        }
```

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+                    hash.";
+            }
+            enum TARGET_DEFINED {
+                description "Use the default fields that are used by
+                    the device for calculating the hash.";
+            }
+            enum FLOW_LABEL {
+                description "Use the flow label in the IPv6 header
+                    to calculate the hash.";
+            }
```
rszarecki, line 90:
```suggestion
            }
             enum INGRESS_SUBINTERFACE {
                description "Use the ingress subinterface identifired in the calculation of the hash.";
            }
```

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+            enum BOTH {
+                description "Use both the IPv4 source and
+                    destination addresses in the calculation of the
+                    hash.";
+            }
+            enum TARGET_DEFINED {
+                description "Use the default fields that are used by
+                    the device for calculating the hash.";
+            }
```
rszarecki, line 58:
```suggestion
            }
            enum INGRESS_SUBINTERFACE {
                description "Use the ingress subinterface identifired in the calculation of the hash.";
            }
```

rszarecki:
The reference to [Cisco IOS-XR bundle-hash](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/Interfaces/b-interfaces-hardware-component-cr-8000/m-link-bundling-commands-8k.html) is wrong. This is just utility that alows execute hash against mocked-up packet and see what is result.

release/models/interfaces/openconfig-interfaces.yang
dplore: The intent here is that this is a list of policies.  We'll update the description to not reference 'packets coming in the interface'.      gPINS will add support for multiple policies and Arista EOS is an existing implementation

release/models/hashing/openconfig-hashing.yang
dplore: will move this to a boolean leaf

release/models/hashing/openconfig-hashing.yang
dplore: will move to a boolean leaf

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+        }
+        description "The trasport layer fields that should be used to
+            compute the hash.";
+    }
+
+    grouping hashing-inputs {
+        description
+            "Top level container for inputs to be used for hashing policies.";
+
+        leaf ipv4 {
```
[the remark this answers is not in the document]
dplore: Good idea.  We'll change this from a leaf with enums to several boolean leafs, one for each option so they can bet set in whatever combination a user wants.  We'll also add the ingress interface as a leaf here.

release/models/hashing/openconfig-hashing.yang
dplore: For this PR we will only introduce the simple field selection which is targeted for gPINS release and is at least partially supported by the referenced implementations in this PR.

It is possible to  add additional encapsulations in the future without breaking changes to the model by adding additional leafs to this hashing-inputs grouping.  The additional leafs may be added to the public OC model in the future, or they can be vendor augments.

Since there are many constraints in hardware as to which encapsulations may be supported, especially for more complex use cases, we should leave open the option that vendors are expected to augment the model with the specific scenarios they support.  These are often custom or one-off items that are not generally available and may also not be formally documented, which is a good reason to use a vendor provided augment.  These 'custom' options may also not be very 'configurable'.  For example, hardware may support only using a subset of the bits of some fields (ipv6 comes to mind).  Which bits are used may not be configurable or even exposed to the user through the model, config or even the SDK.  Using a vendor augment, a vendor can expose the ability to turn on or off these capabilities.

release/models/hashing/openconfig-hashing.yang
dplore: This is a good idea, will add it.  Probably use 'name' instead of index though.

release/models/hashing/openconfig-hashing.yang
dplore: References to implementation needed for this?

release/models/hashing/openconfig-hashing.yang
dplore: References to implementation needed for this?

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+    }
+
+    grouping hashing-top {
+        description
+          "Top level grouping for hashing configuration and operational state
+          data.";
+
+        container policies {
```
robshakir, line 185: This creates `/policies` which seems really ambiguous -- given that we have `/routing-policy` too. I suggest that this is something that we should probably have under `/system/hashing-policies`.

release/models/interfaces/openconfig-interfaces.yang
robshakir: Can you link to the documentation that shows this being configurable per ingress /interface/ please? It's not clear to me that this is widely supported.

release/models/system/openconfig-system.yang
```diff
@@ -47,7 +48,12 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.17.1";
+  oc-ext:openconfig-version "0.17.2";
+
+  revision "2023-08-28" {
+    description "Add leaf for referencing a hashing policy.";
+    reference "0.17.2";
```
robshakir, line 55: Given that this is implemented at this point, this module should be `1.x.y.` at least. This also should not be a patch change.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+            description
+              "The name of the hashing policy.
+
+              When a configured user-controlled policy is created by the
+              system, it is instantiated with the same name in the
+              /policies/policy[name]/state list.";
+        }
+
+        leaf algorithm {
+            type string;
```
robshakir, line 160: This seems extremely open as a definition. What does this string mean? What are the possible values, and how do we validate them? It should at the very least be an `identity` that has a common set of values that are widely implemented, and then a subsequent set of potential extensions that can be added to the same base identity if there are hardware/vendor specific options that are needing to be represented here.

release/models/hashing/openconfig-hashing.yang
robshakir, line 165: Can we cite where these are actually configurable? the references in the PR comments don't seem to show this is something that is generally supported.

In the case that we are looking to add extensions that are for specific types of devices that might expose this -- or that this is something that is to be supported in the future, please look at the pattern of using a `-ext` module.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+            enum DST_ADDRESS {
+                description "Use only the IPv4 destination address in the
+                    calculation of the hash.";
+            }
+            enum BOTH {
+                description "Use both the IPv4 source and
+                    destination addresses in the calculation of the
+                    hash.";
+            }
+            enum TARGET_DEFINED {
```
robshakir, line 55: How does one test this configuration parameter?

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+
+    oc-ext:openconfig-version "0.1.0";
+
+    revision "2023-08-08" {
+        description "Initial hashing model.";
+        reference "0.1.0";
+    }
+
+    // OpenConfig specific extensions for module metadata.
+    oc-ext:regexp-posix;
```
robshakir, line 32: This should not be needed any more - since all patterns should use the POSIX regexp extension.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+    oc-ext:catalog-organization "openconfig";
+    oc-ext:origin "openconfig";
+
+    typedef hash-field-type-ipv4 {
```
robshakir, line 36: For this typedef, please consider:

a) whether this should be an `identity` so that it is more extensible by other modules.
b) providing references as to the possible configurable values across multiple vendors.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+            enum TARGET_DEFINED {
+                description "Use the default fields that are used by
+                    the device for calculating the hash.";
+            }
+        }
+        description "The IPv4 fields that should be used to
+            compute the hash.";
+    }
+
+    typedef hash-field-type-ipv6 {
```
robshakir, line 64: Ditto the considerations for the IPv4 typedef.

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+                description "Use both the source and
+                    destination ports from the transport header in
+                    the calculation of the hash.";
+            }
+            enum TARGET_DEFINED {
+                description "Use the default fields that are used by
+                    the device for calculating the hash.";
+            }
+        }
+        description "The trasport layer fields that should be used to
```
robshakir, line 120:
```suggestion
        description "The transport layer fields that should be used to
```

robshakir (changes requested):

release/models/hashing/openconfig-hashing.yang
sallylsy: We have changed the emun to the list of leaf. Please see in https://github.com/openconfig/public/pull/959/files

release/models/system/openconfig-system.yang
sallylsy: I marked the version to `1.18.0`.

release/models/hashing/openconfig-hashing.yang
sallylsy: Changes made in https://github.com/openconfig/public/pull/959

release/models/hashing/openconfig-hashing.yang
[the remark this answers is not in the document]
dplore: New PR #959 responds to this

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,238 @@
...
+    grouping hashing-inputs {
+        description
+            "Top level container for inputs to be used for hashing policies.";
+
+        leaf ipv4 {
+            type hash-field-type-ipv4;
+            description "The IPv4 fields that should be used to
+                compute the hash.";
+        }
+        leaf ipv6 {
```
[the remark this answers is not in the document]
dplore: #959 addresses this

release/models/hashing/openconfig-hashing.yang
dplore: #959 proposes a solution for this.  Please comment on #959

release/models/interfaces/openconfig-interfaces.yang
dplore: Per interface is not widely supported.  gPINS plans to add support for this to Sonic.    Should we include this leaf in the model or only augment /interfaces in a gPINS/SONIC yang?  Please follow up with comments at #959

release/models/hashing/openconfig-hashing.yang
dplore: It was decided to remove this leaf

release/models/hashing/openconfig-hashing.yang
dplore: addressed in #959

release/models/hashing/openconfig-hashing.yang
dplore: addressed in #959

release/models/hashing/openconfig-hashing.yang
dplore: Proposed solution in #959 requiring the implementation to publish a list of algorithms that are implementation defined.  Please comment on #959

release/models/hashing/openconfig-hashing.yang
dplore: this leaf is no longer present in #959

release/models/hashing/openconfig-hashing.yang
dplore: See proposal in #959

release/models/hashing/openconfig-hashing.yang
dplore: A test would require target specific knowledge to create a test for this particular enum.  Use of this enum is expected when the target contains some default or hardcoded selection of fields which are not described by the model.  

For example, an asic may select only a portion of the bits in the src and dest ip address fields and port number fields.  Which bits are used may not be exposed and may not be configurable by the user, by the NOS or by the ASIC SDK.  Yet they are still "target defined" and don't match the other enums presented here.  

If a target defined mode isn't available on an implementation, then the implementation should return an UNIMPLEMENTED (12) error.

dplore: We can update the description for this enum in #959

dplore: Continue the discussion here as desired: https://github.com/openconfig/public/pull/959/files#r1337998777

dplore:
Please refer to #959 for comments. Closing this PR as it is now stale.

PR #943, opened 2023-08-17
Add config and state for individual Ethernet port priority in a LAG

marius-ore:
### Change Scope
* Add `port-priority` config and state for Ethernet interface that is part of a logical aggregation / LAG
* Add /lacp/interfaces/interface/members/member/state/partner-port-priority
* This change is not backwards compatible due to breaking change of changing type for list.

### Platform Implementations

 * Arista, section "Configuring Port Priority": https://www.arista.com/en/um-eos/eos-port-channels-and-lacp
 * Cisco, section "Configuring the LACP System ID and Port Priority": https://www.cisco.com/c/en/us/td/docs/ios/12_2sb/feature/guide/gigeth.html
 * JunOS: https://www.juniper.net/documentation/us/en/software/junos/interfaces-ethernet/topics/ref/statement/port-priority-edit-interfaces.html

### Tree view
```
module: openconfig-lacp
  +--rw lacp
     +--rw config
     |  +--rw system-priority?   uint16
     +--ro state
     |  +--ro system-priority?   uint16
     +--rw interfaces
        +--rw interface* [name]
           +--rw name       -> ../config/name
           +--rw config
           |  +--rw name?              oc-if:base-interface-ref
           |  +--rw interval?          lacp-period-type
           |  +--rw lacp-mode?         lacp-activity-type
           |  +--rw system-id-mac?     oc-yang:mac-address
           |  +--rw system-priority?   uint16
           +--ro state
           |  +--ro name?              oc-if:base-interface-ref
           |  +--ro interval?          lacp-period-type
           |  +--ro lacp-mode?         lacp-activity-type
           |  +--ro system-id-mac?     oc-yang:mac-address
           |  +--ro system-priority?   uint16
           +--rw members
              +--rw member* [interface]
                 +--rw interface    -> ../config/interface                          <-- changed
                 +--rw config                                                              <-- new
                 |  +--rw interface?       oc-if:base-interface-ref.            <-- new
                 |  +--rw port-priority?   uint16                                <-- new
                 +--ro state
                    +--ro interface?               oc-if:base-interface-ref
                    +--ro port-priority?           uint16
                    +--ro activity?                lacp-activity-type
                    +--ro timeout?                 lacp-timeout-type
                    +--ro synchronization?         lacp-synchronization-type
                    +--ro aggregatable?            boolean
                    +--ro collecting?              boolean
                    +--ro distributing?            boolean
                    +--ro system-id?               oc-yang:mac-address
                    +--ro oper-key?                uint16
                    +--ro partner-id?              oc-yang:mac-address
                    +--ro partner-key?             uint16
                    +--ro port-num?                uint16
                    +--ro partner-port-num?        uint16                                    <-- new
                    +--ro partner-port-priority?   uint16
                    +--ro last-change?             oc-types:timeticks64
                    +--ro counters
                       +--ro lacp-in-pkts?               oc-yang:counter64
                       +--ro lacp-out-pkts?              oc-yang:counter64
                       +--ro lacp-rx-errors?             oc-yang:counter64
                       +--ro lacp-tx-errors?             oc-yang:counter64
                       +--ro lacp-unknown-errors?        oc-yang:counter64
                       +--ro lacp-errors?                oc-yang:counter64
                       +--ro lacp-timeout-transitions?   oc-yang:counter64

```


release/models/interfaces/openconfig-if-aggregate.yang
```diff
@@ -204,12 +213,63 @@ module openconfig-if-aggregate {
...
+        path "/oc-if:interfaces/oc-if:interface/oc-if:name";
+      }
+      description
+        "Specify the logical aggregate interface to which
+        this interface belongs";
+    }
+
+    leaf port-priority {
+      type uint16;
+      default "32768";
```
earies, line 239: Suggest dropping the default - this is per-implementation and not defined in the LACP specification

release/models/lacp/openconfig-lacp.yang
```diff
@@ -26,7 +26,13 @@ module openconfig-lacp {
     managing aggregate interfaces.   It works in conjunction with
     the OpenConfig interfaces and aggregate interfaces models.";
 
-  oc-ext:openconfig-version "1.2.0";
+  oc-ext:openconfig-version "1.2.1";
```
earies, line 29: 1.3.0

release/models/interfaces/openconfig-if-aggregate.yang
```diff
@@ -204,12 +213,63 @@ module openconfig-if-aggregate {
...
+        uses aggregation-ethernet-config;
+      }
+
+      container state {
+        config false;
+        description
+          "Operational state for interface's aggregation parameters";
+
+
+        uses aggregation-ethernet-config;
```
earies, line 268: There can be protocol agnostic aggregation parameters then there is also LACP specific.  For what is being discussed here, we are talking about LACP so would likely propose either a LACP container or prefixing of nodes.  Container organization is likely cleaner/expansible

marius-ore: So should I just rename the new container at `/interfaces/interface/ethernet` from `aggregation` to `lacp`? Would the leaf `aggregate-id` still belong there then?

robshakir: IMHO, if this is LACP specific, then the better change is to have this be stored under `/lacp`. Our pattern is that the hierarchy is protocol-centric, and interfaces don't become a collecting ground for all parameters.

We'll have to fix making sure that the `config false` is applied at the right point in the tree. @dplore -- this will need us to discuss whether this is a backwards incompatible change (making a node that was `config false` now `config true` but not changing the `config` state of any particular leaf).

If there are generic statistics to do with aggregation, then `/interfaces/interface/aggregation` would be preferred -- but I'm not sure that I'm immediately aware of any.

release/models/interfaces/openconfig-if-aggregate.yang
```diff
@@ -204,12 +213,61 @@ module openconfig-if-aggregate {
...
         this interface belongs";
     }
   }
 
+  grouping aggregation-ethernet-config {
+    description
+      "Adds configuration items for dedicated aggregation container
+      for Ethernet interfaces belonging to a logical aggregate / LAG";
+
+    leaf aggregate-id {
```
robshakir, line 228: We actually explicitly made this not part of a Ethernet configuration to support the idea of bundles of non-Ethernet interfaces.

(Specifically, networks that I've worked on have used POS bundles, and this is relatively widely supported -- `bundle-POS` on XR, `as` on JUNOS. Of course, good old POS isn't that popular these days, and no-one has yet extended OpenConfig to support SONET/SDH interfaces at all AFAIK. This being said, it still would be possible, and given we've already made the choice to accommodate it, I'm not clear we should change it now.)

release/models/lacp/openconfig-lacp.yang
```diff
@@ -245,6 +251,18 @@ module openconfig-lacp {
         timestamp in nanoseconds relative to the Unix Epoch
         (Jan 1, 1970 00:00:00 UTC).";
     }
+
+    leaf port-priority {
```
robshakir, line 255: Per my other comment, I definitely don't think the pattern of having this configuration on `/interfaces` is correct. We should have the `config` and `state` of this leaf mirror each other within the sibling `config` and `state` containers (intended config, applied config in state) rather than having the split across the modules.

release/models/interfaces/openconfig-if-aggregate.yang
dplore: If we use a  `/lacp/interfaces/interface` approach then we are decoupled with the L1/L2 transport.

release/models/lacp/openconfig-lacp.yang
dplore: `/lacp/interfaces/interface` approach should also address this

release/models/interfaces/openconfig-if-aggregate.yang
marius-ore: My understanding is that `/lacp/interfaces/interface/config` refers to aggregate interfaces, whereas LACP data for individual Ethernet member interfaces is under `/lacp/interface/interfaces/members/member`. There is currently no config container under the latter, that is why I initially proposed to add the container `/lacp/interface/interfaces/members/member/config`.

@dplore, do you specifically mean that we should consider adding `port-priority` under `/lacp/interface/interface/config`? Because I'm not sure what the interpretation of this container would be then, since `port-priority` is Ethernet interface-specific, while `/lacp/interface/interface/config` refers to aggregate interfaces.

So for example, if we had aggregate interface `Port-Channel1` with `Ethernet1` and `Ethernet2` as member interfaces, `Port-Channel1` would appear  as a key under `/lacp/interfaces/interface`, whereas `Ethernet1` and `Ethernet2` would appear as keys under `/lacp/interfaces/interface/members/member`. So if I wanted to give `Ethernet1` a lower priority within `Port-Channel1`, I'm not sure how I would implement that with `port-priority` being a leaf in `/lacp/interfaces/interface/config`, since that container can configure `Port-Channel1` but not `Ethernet1`.

release/models/interfaces/openconfig-if-aggregate.yang
marius-ore: @dplore would you mind expanding a bit on the `/lacp/interfaces/interface` approach? As detailed in my other comment, I'm not sure how individual Ethernet interface `port-priority` fits at `/lacp/interfaces/interface/config`, since that container configures aggregate interfaces. I need a way to configure the priority of individual Ethernet interfaces in a LAG, that's why I initially proposed to add a config container under `/lacp/interfaces/interface/members/member`. But I'm happy to try out any other approach

release/models/interfaces/openconfig-if-aggregate.yang
robshakir: I think it does need to be `/lacp/interfaces/interface/members/member/config` given that its per member configuration. This was the `config true` container I was suggesting we would add.

marius-ore: Ok great! I noticed that there are two `config false`: one under `/lacp/interfaces/interface/members` and one under `/lacp/interfaces/interface/members/member/state`:
```
    container members {
      config false;
      [...]

      list member {
        key "interface";
        [...]

        leaf interface {
          [...]
        }

        container state {

          config false;
```

So in order to create the new container `/lacp/interfaces/interface/members/member/config` with `config true`, the line `config false` from `/lacp/interfaces/interface/members` would need to be removed. Changing an existing container from `config false` to `config true` would make this a breaking change, right?

earies: Yes - it would be backwards incompatible but likely varied/minimal impact to existing implementations imo

I guess I'll offer another viewpoint for consideration back to the `/interfaces` subtree.  It's really a design pattern/view at the end of the day

Now, _most_ implementations that support link aggregation will usually have the configuration reference construct to the parent virtual interface at the member

* IOS-XR: https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5xx/l2vpn/66x/b-l2vpn-cg-66x-ncs540/b-l2vpn-cg-63x-ncs5xx_chapter_0100.html
* EOS: https://www.arista.com/en/um-eos/eos-port-channels-and-lacp
* JUNOS: https://www.juniper.net/documentation/us/en/software/junos/interfaces-security-devices/topics/topic-map/security-interface-config-aggregated-ethernet.html
* Linux bonding w/ iproute2: http://www.uni-koeln.de/~pbogusze/posts/LACP_configuration_using_iproute2.html

But the opposite is possible and very well a mix depending on bootstrap mechanisms w/ Linux bonding or teaming (conf files, init declarations, etc..)

Now, the `ethernet` container under `interfaces` is probably not the right anchor point as @robshakir pointed out as this ties aggregation parameters specifically to ethernet and there are other possibilities for other link types albeit more and more rare these days - especially in relation to OpenConfig (where its likely many other features/subtrees are non-existant for the use-cases today)

But I think the `aggregation` container would very well be a good anchor imo where you can have protocol agnostic (or other) coupled with a sub-container for `lacp` specific that could be gated by a `when ...` back to the interface `type`

This also provides a known consistent location to target for members (`/interfaces/interfaces/aggregation`).  If we add other LAG parameters in the future they could be dispersed into other subtrees for protocol agnostic or protocol based.

It would however create the split for state as it is today.... much like viewing LACP related state under `show interfaces...` vs. something like `show lacp interfaces ...` - I think this is also a common split among implementations where the latter provides the necessary state detail and the prior only provides a reference to the parent

dplore: So we have `/lacp/interfaces/interface/members/member/config` as a new container to be added (config true) for providing lacp configuration which is per-interface.    

This PR should also deprecate existing /interfaces/interface/ethernet/(config|state)/aggregate-id as originally proposed and add aggregate-id in the new container  as `/lacp/interfaces/interface/members/member/config/aggregate-id`.

My interpretation is does not need to be treated as  a breaking change because this is adding a new container, not changing the attributes of an existing container.  In other words, the existing `/lacp/interfaces/interface/members` container can stay as config false, I think?

The existing `/interfaces/interface/aggregation/config/` container remains for protocol agnostic link aggregation configuration which is applied to the 'virtual', aggregated interface.

release/models/interfaces/openconfig-if-aggregate.yang
dplore: I agree with the approach to use the members container.  Using the approach at https://github.com/openconfig/public/pull/943#discussion_r1302002147 will move (deprecate) the aggregate-id out of the Ethernet configuration and resolve this comment.

release/models/lacp/openconfig-lacp.yang
dplore: Addressing https://github.com/openconfig/public/pull/943#discussion_r1302002147 will also resolve this conversation

release/models/interfaces/openconfig-if-aggregate.yang
earies: It's backwards incompatible due to the following which is inherited to child nodes

```
  grouping aggregation-lacp-members-top {
    description
      "Top-level grouping for aggregate members list";

    container members {
      config false;
      description
        "Enclosing container for the list of members interfaces of
        the aggregate. This list is considered operational state
        only so is labeled config false and has no config container";
```

But as I mentioned, highly unlikely to be a big issue among implementations.  the `config false` needs removal at the container first.

release/models/interfaces/openconfig-if-aggregate.yang
earies: Would just want to make sure to preserve the `aggregate-id` for the state node since it would be beneficial to enumerate interfaces to figure out which parent they are a member of.

release/models/interfaces/openconfig-if-aggregate.yang
dplore: Can the container `members` with list `member` be config false while a child container `members/member/config` is config true?  I don't see this prohibited in RFC6020.

earies: > Can the container `members` with list `member` be config false while a child container `members/member/config` is config true? I don't see this prohibited in RFC6020.

No - any `config false` at a hierarchy means all subhierarchies inherit and cannot be overridden back to `config true`

From RFC 6020/7950

```
When a node is tagged with "config false", its subhierarchy is flagged as state data.
```

Compilation errors from pyang/yanglint:

```
error: config true cannot be set when the parent is config false

Configuration node cannot be child of any state data node. (Path "/a:a/b/config".)
```

The RFC verbiage could maybe be a bit more verbose but if you consider state only subtrees where some configuration hierarchy would be nested deep into the tree, this would conflict w/ many internal behaviors that would not permit writing of the parent object/node (in this case the members container, member list, key, etc..)

In addition - the list key is also currently a leafref to the state node and would need to change to the config node if proceeding w/ this change

```
           +--ro members
              +--ro member* [interface]
                 +--ro interface    -> ../state/interface
```

marius-ore: Ok, thanks everyone for the additional details. I am almost ready to submit a new version of the PR, adding the new `/lacp/interfaces/interface/members/member/config`, but I still have some questions remaining. A summary of my intended changes:

1. Add the `config` container with the following leaves:
```
module: openconfig-lacp
  +--rw lacp
     +--rw interfaces
        +--rw interface* [name]
           +--rw members
              +--rw member* [interface]
                 +--rw config
                    +--rw interface       oc-if:base-interface-ref
                    +--rw port-priority   uint16
```
2. Remove the `config false` from `/lacp/interfaces/interface/members`. This is necessary because otherwise, as @earies pointed out, the child container `/lacp/interfaces/interface/members/member/config` cannot be `config true`.
3. Change the key to list `/lacp/interfaces/interface/members/member` from `state/interface` to `config/interface`:
```
  +--rw lacp
     +--rw interfaces
        +--rw interface* [name]
           +--rw members
              +--rw member* [interface]
-                 +--ro interface    -> ../state/interface
+                 +--rw interface    -> ../config/interface
```
4. Deprecate `/interfaces/interface/ethernet/(config|state)/aggregate-id` and move it to the new `/lacp/interfaces/interface/members/member/(config|state)`. It’s on this last point that I still have some questions:

      a. What is the meaning of `aggregate-id` at this new path? Because the information about the aggregate interface is already contained in the path. For example if I, want to add `Ethernet1` to `Port-Channel1`, I would run something like`update /lacp/interfaces/interface[name=Port-Channel1]/members/member[interface=Ethernet1]/config/interface “Ethernet1”`. So in this case the the value of `/lacp/interfaces/interface[name=Port-Channel1]/members/member[interface=Ethernet1/config/aggregate-id` would be `Port-Channel1`.
      
      So then should `aggregate-id` be a leafref pointing to `/lacp/interfaces/interface/name`?

      b. With the deprecation of the leaf `/interfaces/interface/ethernet/(config|state)/aggregate-id`, how will an Ethernet interface be added to a `static` (non-LACP) aggregate interface? The new location of the leaf under the `/lacp` container does not seem like the right place for config that is explicitly non-LACP.

marius-ore: Added a new commit reverting my current changes and adding instead `/lacp/interfaces/interface/members/member/config` with `config true`. Haven't yet made any changes (deprecate or move) to `/interfaces/interface/ethernet/(config|state)/aggregate-id` -- will wait for someone to confirm that it's okay to do. I'm still concerned about how moving that attribute to `/lacp/interfaces/interface/members/member/(config|state)` will affect configuring static/non-LACP aggregate interfaces

release/models/interfaces/openconfig-if-aggregate.yang
marius-ore: Happy to keep `/interfaces/interface/ethernet/state/aggregate-id`, but one thing I wonder is if I deprecate `/interfaces/interface/ethernet/config/aggregate-id`, how else would one configure a physical interface to be part of a static/non-LACP aggregate? Doesn't seem to me like `/lacp/interfaces/interface/members/member/config` would be the right place for that

release/models/lacp/openconfig-lacp.yang
```diff
@@ -234,6 +250,18 @@ module openconfig-lacp {
...
+      description
+        "Member interface's priority in its aggregate interface.";
+    }
+
+    leaf partner-port-priority {
+      type uint16;
+      description
+        "Member interface partner's priority in its aggregate interface.";
+    }
+
```
dplore, line 264: Remove these leafs.  

(Note: Instead of explicitly adding these leaves here, the preferred style is to include the 'config' grouping in your state container.  This helps ensure all config leaves have equivalent state leaves)

release/models/lacp/openconfig-lacp.yang
```diff
@@ -310,28 +338,31 @@ grouping aggregation-lacp-members-statistics {
...
             "Reference to aggregate member interface";
         }
 
+        container config {
+          description
+            "Operational state data for aggregate members";
+
+          uses aggregation-lacp-members-config;
+        }
+
```
dplore, line 365: Also add `uses aggregation-lacp-members-config;` to the [state container below](https://github.com/openconfig/public/pull/943/files#diff-c9bb1c5ddd819c0196804f71d00fb6245a0eb3996186b677ae2d9f617ef78405R367-R376)

dplore (changes requested):

release/models/interfaces/openconfig-if-aggregate.yang
dplore: After re-reading, I agree, we shouldn't deprecate `/interfaces/interface/ethernet/config/aggregate-id` after all.  As you state, it is needed to configure ethernet ports into a LAG independent of whether or not LACP is configured.

release/models/lacp/openconfig-lacp.yang
marius-ore: Used config grouping with `interface` and `port-priority` for both `config` and `state`. But kept `partner-port-priority` under state as it is not a configurable parameter.

marius-ore:
I think I've completed all the changes suggested so far. Any other feedback, or is the code good to merge?

dplore:
I added the tree view to the description for the PR.    

Summary of what was added:

module: openconfig-lacp
  +--rw lacp
     +--rw interfaces
        +--rw interface* [name]
           +--rw members
              +--rw member* [interface]
                 +--rw interface    -> ../config/interface
                 +--rw config
                 **|  +--rw port-priority?   uint16**
                 +--ro state
                    **+--ro partner-port-priority?   uint16**

dplore (approved):
LGTM, last call for this change

dplore:
Note, this is considered a breaking change because the list key for members was changed to point at the new config/interface leaf instead of the prior state/interface leaf.

release/models/lacp/openconfig-lacp.yang
```diff
@@ -26,7 +26,14 @@ module openconfig-lacp {
     managing aggregate interfaces.   It works in conjunction with
     the OpenConfig interfaces and aggregate interfaces models.";
 
-  oc-ext:openconfig-version "1.2.0";
+  oc-ext:openconfig-version "2.0.0";
```
dplore, line 29:
```suggestion
  oc-ext:openconfig-version "1.3.0";
```

release/models/lacp/openconfig-lacp.yang
```diff
@@ -26,7 +26,14 @@ module openconfig-lacp {
...
+
+  revision "2023-08-28" {
+    description
+      "Add config container for aggregation member and port-priority leaf under
+       both config and state";
+    reference "2.0.0";
```
dplore, line 35:
```suggestion
    reference "1.3.0";
```

dplore (commented):
Since this is now a non-breaking change, I've suggested a new, minor version number increment

dplore:
@robshakir for your review

robshakir (dismissed):
LGTM, one indentation issue and then I think that this is a backwards incompatible change (as discussed with @dplore).

robshakir (approved):

PR #949, opened 2023-08-31
add NEXT-STATEMENT to policy-result-type

dplore:
### Change Scope

* Add `NEXT-STATEMENT` to the typedef for `route-policy-result`.
* Use `NEXT-STATEMENT` as the default.
* This does not change the intended functionality of routing policies.  It is intended to
clarify the expected behavior if there is a match for a policy, but no policy-result is
specified.

This change adds a default where there previously was none.  So if an implementation performs some other action than  `NEXT-STATEMENT` by default, it could be a breaking change.  

### Platform Implementations
This result type corresponds to the the following  policy match actions in existing implementations.

 * JunOS implements [next term](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-configuring-actions-in-routing-policy-terms.html) which is equivalent to `NEXT-STATEMENT`
 *  Cisco implements [pass](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/routing/configuration/guide/rc40xr12k_chapter7.html) which is similar to `NEXT-STATEMENT`.   It differs in that `pass` means both accept-route and continue.  In OC, `NEXT-STATEMENT` only means move to the next action in the policy.  
 *  Arista EOS implements [continue](https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1150191) which is a superset of `NEXT-STATEMENT`.    Arista's continue statement has additional parameters not contemplated in this PR.

### DIFF
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -90,7 +90,7 @@ module openconfig-routing-policy {
 
   revision "2023-10-11" {
     description
-      "Add NEXT_STATEMENT and NEXT_POLICY policy-result-type enums.";
+      "Add NEXT_STATEMENT policy-result-type enums.";
     reference "3.4.0";
   }
 
@@ -187,11 +187,6 @@ module openconfig-routing-policy {
       enum NEXT_STATEMENT {
         description "Any modifications of the route are preserved and the evaluation of the policy will continue to the
         next statement.";
-      }
-      enum NEXT_POLICY {
-        description "Evaluation of statements in current policy will
-        stop.  Any modifications to the route are preserved and
-        evaluation proceeds to the next policy.";
       }
     }
     description
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -59,11 +59,7 @@ module openconfig-routing-policy {
 
     If the action statement has the NEXT_STATEMENT policy result, all the
     defined actions are exectued and policy evaluation proceeds to the next
-    statement.  If the action statement has NEXT-POLICY as the policy result,
-    there is no final disposition of the route, all the defined actions are
-    executed, evaluation of statements within the current policy stops and
-    evaluation proceeds to the next policy in the chain, if any.  The
-    NEXT_STATEMENT is the default policy result action.
+    statement.  The NEXT_STATEMENT is the default policy result action.
 
     If the condition is not satisfied, then evaluation proceeds to
     the next policy statement.  If none of the policy statement
@@ -189,6 +185,7 @@ module openconfig-routing-policy {
         next statement.";
       }
     }
+    default NEXT_STATEMENT;
     description
       "Type used to specify route disposition in
       a policy chain";
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -55,8 +55,12 @@ module openconfig-routing-policy {
     statement has either accept-route or reject-route actions, policy
     evaluation of the current policy definition stops, and no further
     policy definitions in the chain are evaluated.  If the action
-    statement has the continue action, policy evaluation proceeds to
-    the next statement. Continue is the default policy result action.
+    statement has the NEXT-STATEMENT action, policy evaluation proceeds
+    to the next statement. If the action statement has the NEXT-POLICY
+    action, evaluation of statements within the current policy stops.
+    Any modifications to a route are preserved and evaluation proceeds
+    to the next policy, if any.  The NEXT-STATEMENT is the default
+    policy result action.
 
     If the condition is not satisfied, then evaluation proceeds to
     the next policy statement.  If none of the policy statement
@@ -177,12 +181,17 @@ module openconfig-routing-policy {
         description "Policy rejects the route and evaluation of the
         current policy definition stops.";
       }
-      enum CONTINUE {
+      enum NEXT-STATEMENT {
         description "Evaluation of the policy will continue to the
         next statement.";
       }
+      enum NEXT-POLICY {
+        description "Evaluation of statements in current policy will
+        stop.  Any modifications to the route are preserved and
+        evaluation proceeds to the next policy.";
+      }
     }
-    default CONTINUE;
+    default NEXT-STATEMENT;
     description
       "Type used to specify route disposition in
       a policy chain.";
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -51,16 +51,18 @@ module openconfig-routing-policy {
     Evaluation of each policy definition proceeds by evaluating its
     corresponding individual policy statements in order.  When a
     condition statement in a policy statement is satisfied, the
-    corresponding action statement is executed.  If the action
-    statement has a final disposition configured as policy result, either accept-route or reject-route,
-    evaluation of the current policy definition stops after having executed all the defined actions , and no further policy statements are evaluated. In case there is a policy chain, no further 
-    policy definitions in the chain are evaluated.  If the action
-    statement has the NEXT_STATEMENT action, policy evaluation proceeds
-    to the next statement. If the action statement has the NEXT-POLICY
-    action, evaluation of statements within the current policy stops.
-    Any modifications to a route are preserved and evaluation proceeds
-    to the next policy, if any.  The NEXT_STATEMENT is the default
-    policy result action.
+    corresponding action statement is executed.
+
+    If the action statement has either accept-route or reject-route actions,
+    policy evaluation of the current policy definition stops, all
+    modifications to the route are applied and no further statements in the
+    policy are evaluated. In case there is a policy chain, no further policy
+    definitions in the chain are evaluated.  If the action statement has
+    the NEXT_STATEMENT action, policy evaluation proceeds to the next
+    statement.  If the action statement has the NEXT-POLICY action all
+    the modifications to the route are applied, evaluation of statements
+    within the current policy stops and the next policy in the chain is
+    evaluated.
 
     If the condition is not satisfied, then evaluation proceeds to
     the next policy statement.  If none of the policy statement
@@ -83,13 +85,7 @@ module openconfig-routing-policy {
     the remaining conditions (using a modified route if the
     subroutine performed any changes to the route).";
 
-  oc-ext:openconfig-version "3.4.0";
-
-  revision "2023-08-31" {
-    description
-      "Add 'continue' to policy-result-type.";
-    reference "3.4.0";
-  }
+  oc-ext:openconfig-version "3.3.0";
 
   revision "2022-05-24" {
     description
@@ -191,10 +187,9 @@ module openconfig-routing-policy {
         evaluation proceeds to the next policy.";
       }
     }
-    default NEXT_STATEMENT;
     description
       "Type used to specify route disposition in
-      a policy chain.";
+      a policy chain";
   }
 
 
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -181,11 +181,11 @@ module openconfig-routing-policy {
         description "Policy rejects the route and evaluation of the
         current policy definition stops.";
       }
-      enum NEXT-STATEMENT {
+      enum NEXT_STATEMENT {
         description "Evaluation of the policy will continue to the
         next statement.";
       }
-      enum NEXT-POLICY {
+      enum NEXT_POLICY {
         description "Evaluation of statements in current policy will
         stop.  Any modifications to the route are preserved and
         evaluation proceeds to the next policy.";
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -53,7 +53,7 @@ module openconfig-routing-policy {
     condition statement in a policy statement is satisfied, the
     corresponding action statement is executed.  If the action
     statement has a final disposition configured as policy result, either accept-route or reject-route,
-    evaluation of the current policy definition stops, and no further
+    evaluation of the current policy definition stops after having executed all the defined actions , and no further policy statements are evaluated. In case there is a policy chain, no further 
     policy definitions in the chain are evaluated.  If the action
     statement has the NEXT_STATEMENT action, policy evaluation proceeds
     to the next statement. If the action statement has the NEXT-POLICY
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -56,10 +56,10 @@ module openconfig-routing-policy {
     accept-route or reject-route, evaluation of the current policy definition
     stops, and no further policy statements are evaluated. In case there is a
     policy chain, no further policy definitions in the chain are evaluated.
-    
+
     If the action statement has the NEXT_STATEMENT policy result, all the
     defined actions are exectued and policy evaluation proceeds to the next
-    statement. If the action statement has NEXT-POLICY as the policy result,
+    statement.  If the action statement has NEXT-POLICY as the policy result,
     there is no final disposition of the route, all the defined actions are
     executed, evaluation of statements within the current policy stops and
     evaluation proceeds to the next policy in the chain, if any.  The
@@ -86,7 +86,13 @@ module openconfig-routing-policy {
     the remaining conditions (using a modified route if the
     subroutine performed any changes to the route).";
 
-  oc-ext:openconfig-version "3.3.0";
+  oc-ext:openconfig-version "3.4.0";
+
+  revision "2023-10-11" {
+    description
+      "Add NEXT_STATEMENT and NEXT_POLICY policy-result-type enums.";
+    reference "3.4.0";
+  }
 
   revision "2022-05-24" {
     description
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -58,7 +58,7 @@ module openconfig-routing-policy {
     policy chain, no further policy definitions in the chain are evaluated.
 
     If the action statement has the NEXT_STATEMENT policy result, all the
-    defined actions are exectued and policy evaluation proceeds to the next
+    defined actions are executed and policy evaluation proceeds to the next
     statement.  The NEXT_STATEMENT is the default policy result action.
 
     If the condition is not satisfied, then evaluation proceeds to
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -54,7 +54,9 @@ module openconfig-routing-policy {
     corresponding action statement is executed.  If the action
     statement has either accept-route or reject-route actions, policy
     evaluation of the current policy definition stops, and no further
-    policy definitions in the chain are evaluated.
+    policy definitions in the chain are evaluated.  If the action
+    statement has the continue action, policy evaluation proceeds to
+    the next statement. Continue is the default policy result action.
 
     If the condition is not satisfied, then evaluation proceeds to
     the next policy statement.  If none of the policy statement
@@ -77,7 +79,13 @@ module openconfig-routing-policy {
     the remaining conditions (using a modified route if the
     subroutine performed any changes to the route).";
 
-  oc-ext:openconfig-version "3.3.0";
+  oc-ext:openconfig-version "3.4.0";
+
+  revision "2023-08-31" {
+    description
+      "Add 'continue' to policy-result-type.";
+    reference "3.4.0";
+  }
 
   revision "2022-05-24" {
     description
@@ -162,15 +170,22 @@ module openconfig-routing-policy {
   typedef policy-result-type {
     type enumeration {
       enum ACCEPT_ROUTE {
-        description "Policy accepts the route";
+        description "Policy accepts the route and evaluation of the
+        current policy definition stops.";
       }
       enum REJECT_ROUTE {
-        description "Policy rejects the route";
+        description "Policy rejects the route and evaluation of the
+        current policy definition stops.";
+      }
+      enum CONTINUE {
+        description "Evaluation of the policy will continue to the
+        next statement.";
       }
     }
+    default CONTINUE;
     description
       "Type used to specify route disposition in
-      a policy chain";
+      a policy chain.";
   }
 
 
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -182,7 +182,7 @@ module openconfig-routing-policy {
         current policy definition stops.";
       }
       enum NEXT_STATEMENT {
-        description "Evaluation of the policy will continue to the
+        description "Any modifications of the route are preserved and the evaluation of the policy will continue to the
         next statement.";
       }
       enum NEXT_POLICY {
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -55,11 +55,11 @@ module openconfig-routing-policy {
     statement has either accept-route or reject-route actions, policy
     evaluation of the current policy definition stops, and no further
     policy definitions in the chain are evaluated.  If the action
-    statement has the NEXT-STATEMENT action, policy evaluation proceeds
+    statement has the NEXT_STATEMENT action, policy evaluation proceeds
     to the next statement. If the action statement has the NEXT-POLICY
     action, evaluation of statements within the current policy stops.
     Any modifications to a route are preserved and evaluation proceeds
-    to the next policy, if any.  The NEXT-STATEMENT is the default
+    to the next policy, if any.  The NEXT_STATEMENT is the default
     policy result action.
 
     If the condition is not satisfied, then evaluation proceeds to
@@ -191,7 +191,7 @@ module openconfig-routing-policy {
         evaluation proceeds to the next policy.";
       }
     }
-    default NEXT-STATEMENT;
+    default NEXT_STATEMENT;
     description
       "Type used to specify route disposition in
       a policy chain.";
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -52,7 +52,7 @@ module openconfig-routing-policy {
     corresponding individual policy statements in order.  When a
     condition statement in a policy statement is satisfied, the
     corresponding action statement is executed.  If the action
-    statement has either accept-route or reject-route actions, policy
+    statement has a final disposition configured as policy result, either accept-route or reject-route,
     evaluation of the current policy definition stops, and no further
     policy definitions in the chain are evaluated.  If the action
     statement has the NEXT_STATEMENT action, policy evaluation proceeds
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -52,15 +52,18 @@ module openconfig-routing-policy {
     corresponding individual policy statements in order.  When a
     condition statement in a policy statement is satisfied, the
     corresponding action statement is executed.  If the action
-    statement has a final disposition configured as policy result, either accept-route or reject-route,
-    evaluation of the current policy definition stops after having executed all the defined actions , and no further policy statements are evaluated. In case there is a policy chain, no further 
-    policy definitions in the chain are evaluated.  If the action
-    statement has the NEXT_STATEMENT policy result, all the defined actions are exectued and policy evaluation proceeds
-    to the next statement. If the action statement has the NEXT-POLICY
-    action, evaluation of statements within the current policy stops.
-    Any modifications to a route are preserved and evaluation proceeds
-    to the next policy, if any.  The NEXT_STATEMENT is the default
-    policy result action.
+    statement has a final disposition configured as policy result, either
+    accept-route or reject-route, evaluation of the current policy definition
+    stops, and no further policy statements are evaluated. In case there is a
+    policy chain, no further policy definitions in the chain are evaluated.
+    
+    If the action statement has the NEXT_STATEMENT policy result, all the
+    defined actions are exectued and policy evaluation proceeds to the next
+    statement. If the action statement has NEXT-POLICY as the policy result,
+    there is no final disposition of the route, all the defined actions are
+    executed, evaluation of statements within the current policy stops and
+    evaluation proceeds to the next policy in the chain, if any.  The
+    NEXT_STATEMENT is the default policy result action.
 
     If the condition is not satisfied, then evaluation proceeds to
     the next policy statement.  If none of the policy statement
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -86,7 +86,7 @@ module openconfig-routing-policy {
 
   revision "2023-10-11" {
     description
-      "Add NEXT_STATEMENT policy-result-type enums.";
+      "Add NEXT_STATEMENT policy-result-type enum value.";
     reference "3.4.0";
   }
 


wenovus (approved):
I think continue is a good name for it. If we choose to add "NEXT_POLICY" in the future this still LGTM.

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -54,7 +54,13 @@ module openconfig-routing-policy {
     corresponding action statement is executed.  If the action
     statement has either accept-route or reject-route actions, policy
     evaluation of the current policy definition stops, and no further
```
oscargdd, line 56:
```suggestion
    evaluation of the current policy definition stops after having executed all the defined actions , and no further policy statements are evaluated. In case there is a policy chain, no further 
```

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -54,7 +54,13 @@ module openconfig-routing-policy {
...
-    policy definitions in the chain are evaluated.
+    policy definitions in the chain are evaluated.  If the action
```
oscargdd, line 57:
```suggestion
    policy definitions in the chain are evaluated.  If the actions
```

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -54,7 +54,13 @@ module openconfig-routing-policy {
...
+    statement has the NEXT_STATEMENT action, policy evaluation proceeds
```
oscargdd, line 58:
```suggestion
    statement has the NEXT_STATEMENT policy result, all the defined actions are exectued and policy evaluation proceeds
```

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -54,7 +54,13 @@ module openconfig-routing-policy {
...
+    to the next statement. If the action statement has the NEXT-POLICY
+    action, evaluation of statements within the current policy stops.
+    Any modifications to a route are preserved and evaluation proceeds
```
oscargdd, line 61:
```suggestion
    Any modifications to the route performed by actions during the evaluation are preserved and evaluation proceeds
```

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -54,7 +54,13 @@ module openconfig-routing-policy {
...
+    to the next policy, if any.  The NEXT_STATEMENT is the default
```
oscargdd, line 62:
```suggestion
    to the next policy in the chain, if any.  The NEXT_STATEMENT is the default
```

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -162,15 +174,27 @@ module openconfig-routing-policy {
...
+        description "Policy accepts the route and evaluation of the
+        current policy definition stops.";
       }
       enum REJECT_ROUTE {
-        description "Policy rejects the route";
+        description "Policy rejects the route and evaluation of the
+        current policy definition stops.";
+      }
+      enum NEXT_STATEMENT {
+        description "Evaluation of the policy will continue to the
```
oscargdd, line 185:
```suggestion
        description "Any modifications of the route are preserved and the evaluation of the policy will continue to the
```

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -54,7 +54,13 @@ module openconfig-routing-policy {
     corresponding action statement is executed.  If the action
     statement has either accept-route or reject-route actions, policy
     evaluation of the current policy definition stops, and no further
-    policy definitions in the chain are evaluated.
+    policy definitions in the chain are evaluated.  If the action
+    statement has the NEXT_STATEMENT action, policy evaluation proceeds
+    to the next statement. If the action statement has the NEXT-POLICY
+    action, evaluation of statements within the current policy stops.
```
oscargdd, line 60:
```suggestion
    as policy result, there is no final disposition of the route, all the defined actions are exectued, evaluation of statements within the current policy stops, so no further statements are evaluated. 
```

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -54,7 +54,13 @@ module openconfig-routing-policy {
     corresponding action statement is executed.  If the action
     statement has either accept-route or reject-route actions, policy
     evaluation of the current policy definition stops, and no further
-    policy definitions in the chain are evaluated.
+    policy definitions in the chain are evaluated.  If the action
+    statement has the NEXT_STATEMENT action, policy evaluation proceeds
+    to the next statement. If the action statement has the NEXT-POLICY
```
oscargdd, line 59:
```suggestion
    to the next statement. If the actions statement has the NEXT-POLICY
```

oscargdd (changes requested):
Please find some suggestions to clarify that the all the actions are executed before jumping to the next statement or policy.

robshakir:
What is the operational use case that we're envisaging for `NEXT_POLICY` here? I can see that this is potentially supported in multiple vendor implementations (but the multi-vendor justification here seems a bit tenuous, can we put together an example of two vendor language policies that actually allow for `NEXT_POLICY`-esque behaviour? It seems like you would need to do `continue 100` where `100` is a statement that just falls through to the next policy, but this would imply that you have policy chaining -- so a better example to demonstrate that this actually possible is needed.)

In a similar vein, what's the case where there is ambiguity w.r.t `NEXT_STATEMENT`. Today, if there isn't a "terminating" action i.e., ACCEPT/REJECT this is the default (as we are saying that this is a NOOP) is there some specific reason that we're trying to make it more explicit?

robshakir (changes requested):
Blocking for additional background to be described in this PR.

rszarecki:
What is expected behavior if all matching statement has NEXT_STATEMENT action?

wenovus:
> What is expected behavior if all matching statement has NEXT_STATEMENT action?

Then the default apply-policy action is applied. This is either REJECT or ACCEPT.

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -54,7 +54,13 @@ module openconfig-routing-policy {
     corresponding action statement is executed.  If the action
```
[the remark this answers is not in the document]
dplore: Each statement only has one action?  (ACCEPT, REJECT, NEXT_STATEMENT, NEXT_POLICY)

rszarecki:
IOS-XR 'pass' is not same as 'NEXT-STATEMENT'. It Carrie's semantics of "conditionally accepted" unless explicitly dropped in subsequent statements.
IOS-XR default depends on other actions taken. If any route attribute is modified, implicit action is to accept route and stop policy evaluation. If there is no modification, route is **dropped**.

release/models/policy/openconfig-routing-policy.yang
dplore: I feel that action (singular) is appropriate as only one action can be configured.

release/models/policy/openconfig-routing-policy.yang
dplore: I like this and made a modification that is very similar to what you suggest.

rszarecki:
> Then the default apply-policy action is applied (default-import-policy default-export-policy leaves). This is either REJECT or ACCEPT.

Yes, but...
1. what if this leaves are omitted/not specified?
2. the description of this leaves is bit unclear: "explicitly set a default policy if no policy definition in the export policy chain is satisfied." In short "definition is not satisfied". Is it applicable:
  -  ONLY when `import-policy`/`export-policy` are not present XOR 
  - if non of statements in `import-policy`/`export-policy` policy chain is matched XOR
  - if on of matched statement have `ACCEPT`/`REJECT` action 
  - or some combination of above.

release/models/policy/openconfig-routing-policy.yang
dplore: I think there is just one action per statement?

release/models/policy/openconfig-routing-policy.yang
dplore: added very similar text.

wenovus:
> IOS-XR 'pass' is not same as 'NEXT-STATEMENT'. It Carrie's semantics of "conditionally accepted" unless explicitly dropped in subsequent statements. IOS-XR default depends on other actions taken. If any route attribute is modified, implicit action is to accept route and stop policy evaluation. If there is no modification, route is **dropped**.

That's interesting. it looks like [Cisco's](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/routing/configuration/guide/rc40xr12k_chapter7.html#con_1143816) and [Arista's](https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1150191) policy evaluation may be quite different from [Juniper's](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-routing-policies-chain-evaluation-method.html), which looks the most similar to OpenConfig to me:

Cisco: "If a policy modifies a route, by default the policy accepts the route.", "If a policy does not modify the route, by default the route is dropped."
Arista: "The evaluation of routes whose settings are the same as match command parameters normally ends and the statement's set commands are applied to the route."

I didn't look deeper into Arista, but in Cisco, whether a particular policy matched a route changes the default behaviour of the route. aka. when mapped to OpenConfig, the default-(import/export)-policy changes depending on whether a particular policy matched and had a `pass` statement. This is not expressible in OpenConfig without a meaningful modification to the policy.

> > Then the default apply-policy action is applied (default-import-policy default-export-policy leaves). This is either REJECT or ACCEPT.
> 
> Yes, but...
> 
> 1. what if this leaves are omitted/not specified?
> 2. the description of this leaves is bit unclear: "explicitly set a default policy if no policy definition in the export policy chain is satisfied." In short "definition is not satisfied". Is it applicable:
> 
> * ONLY when `import-policy`/`export-policy` are not present XOR
> * if non of statements in `import-policy`/`export-policy` policy chain is matched XOR
> * if on of matched statement have `ACCEPT`/`REJECT` action
> * or some combination of above.

So the default policy type has a YANG default value: https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-inter-instance-policies-apply-policy-config-default-import-policy

Per the standard this means even if not configured, it will still have the value of REJECT.

release/models/policy/openconfig-routing-policy.yang
dplore: added

dplore:
> > IOS-XR 'pass' is not same as 'NEXT-STATEMENT'. It Carrie's semantics of "conditionally accepted" unless explicitly dropped in subsequent statements. IOS-XR default depends on other actions taken. If any route attribute is modified, implicit action is to accept route and stop policy evaluation. If there is no modification, route is **dropped**.

`NEXT-STATEMENT` is similar to (or maybe exactly the same as) Cisco `pass` and Arista `continue`.   NEXT-STATEMENT also evaluates to: "apply these changes and continue to evaluate."  A later statement in the policy could still be reject route or accept route.     Please read the updated/clarified description at the beginning of the yang and let me know if this is clear?   (Thanks @oscargdd  for additional content for clarification)   


> Cisco: "If a policy modifies a route, by default the policy accepts the route.", "If a policy does not modify the route, by default the route is dropped." Arista: "The evaluation of routes whose settings are the same as match command parameters normally ends and the statement's set commands are applied to the route."

In OC I think one could produce this behavior using`ACCEPT-ROUTE` as the policy-result-type for every statement.  This may be useful as a way to translate from Arista and Cisco into OC.

dplore:
> What is the operational use case that we're envisaging for `NEXT_POLICY` here? I can see that this is potentially supported in multiple vendor implementations (but the multi-vendor justification here seems a bit tenuous, can we put together an example of two vendor language policies that actually allow for `NEXT_POLICY`-esque behaviour? It seems like you would need to do `continue 100` where `100` is a statement that just falls through to the next policy, but this would imply that you have policy chaining -- so a better example to demonstrate that this actually possible is needed.)

Today in Google, we have policies which make use, in some cases extensively,  of both of these policy-result-type semantics.  I will refer them to you offline.

Regarding ambiguity in the current model:  In the current model, there is only a description at the beginning of the module which says, "If the condition is not satisfied, then evaluation proceeds to the next policy statement."

But what if the condition is satisfied and there is no `policy-result-type`? The current yang allows this, but doesn't define what should happen.

wenovus:
> > Cisco: "If a policy modifies a route, by default the policy accepts the route.", "If a policy does not modify the route, by default the route is dropped."
> 
> In OC I think one could produce this behavior using`ACCEPT-ROUTE` as the policy-result-type for every statement. This may be useful as a way to translate from Arista and Cisco into OC.

I believe this does not work in all cases,
e.g.
```
policy A matches on 10/8, modify & pass.
policy B matches on 10.10/16, drop.
```
In this case, changing `pass` to `accept` would alter the behaviour of the policy. The correct translation is to reverse policy B & A, and then change `pass` to `accept`.

However as you pointed out Arista has a `continue` statement, which may satisfy the concern of showing support in two different vendor implementations.

dplore:
We don't have explicit support of NEXT-POLICY defined by multiple implementations, but we do for NEXT-STATEMENT.

rszarecki:
> > > IOS-XR 'pass' is not same as 'NEXT-STATEMENT'. It Carrie's semantics of "conditionally accepted" unless explicitly dropped in subsequent statements. IOS-XR default depends on other actions taken. If any route attribute is modified, implicit action is to accept route and stop policy evaluation. If there is no modification, route is **dropped**.
> 
> `NEXT-STATEMENT` is similar to (or maybe exactly the same as) Cisco `pass` and Arista `continue`. NEXT-STATEMENT also evaluates to: "apply these changes and continue to evaluate." A later statement in the policy could still be reject route or accept route. Please read the updated/clarified description at the beginning of the yang and let me know if this is clear? (Thanks @oscargdd for additional content for clarification)

Lets take an example.
We do have RFC8212 compliant implementation.
The we do have EBGP export  have policy : {statement 1 {match 10.0.0/24, pass} statement2 {match 11/8, drop}}. Let evaluate route 10.0.0/24.
it matches statement 1 and is marked as "accepted at the end of policy", then we evaluate statement 2 which do not matches. Hence we are at the end of policy and route is **accepted**.
 
Wherhas in OpenConfig NEXT_STATEMENT (and juniper next-term):
 The we do have EBGP export  have policy : {statement 1 {match 10.0.0/24, next_statement } statement2 {match 11/8, drop}}. Let evaluate route 10.0.0/24.
it matches statement 1 but there is no decission to accept or reject, statemet 2 is not matching. Here we have protocol default (rfc8212) that **reject** 10.0.0/24.

rszarecki:
> Cisco: "If a policy modifies a route, by default the policy accepts the route.", "If a policy does not modify the route, by default the route is dropped."

example:
----------
Cisco:
statement 1 match 10.0.0/24-/32; pass (no modification)
statement 2 match 10.0/16-/32; drop
statement 3 match 10/8-/32 ; pass

the 10.0.0/24 will be dropped.

OpenConfig , JUNIPER
statement 1 match 10.0.0/24-/32 ; ACCEPT_ROUTE (note cisco has no such action)
statement 2 match 10.0/16-/32 ; REJECT_ROUTE
statement 3 match 10/8-/32; ACCEPT_ROUTE

the 10.0.0/24 will be accepted.

dplore:
> Cisco: statement 1 match 10.0.0/24-/32; pass (no modification) statement 2 match 10.0/16-/32; drop statement 3 match 10/8-/32 ; pass
> 
> the 10.0.0/24 will be dropped.

I think this is not correct?  pass is "accept and continue". This policy would accept 10.0.0/24.  Ref: [cisco ios xr route policy disposition](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/routing/configuration/guide/rc40xr12k_chapter7.html#con_1143816)

rszarecki:
> I think this is not correct? pass is "accept and continue". This policy would accept 10.0.0/24. Ref: cisco ios xr route policy disposition
> "When a policy has finished executing, any route that has been modified in the policy or any route that has received a pass disposition in the policy, successfully passes the policy and completes the execution. "

The 10.0.0/24 is matching statement 2 (10.0/16-32, so /16 or any more specific, that is also 10.0.0/24) with explicit drop.

rszarecki (commented):
Please consider replacement of lines 43-87 by following:
```
    Route policy evaluation:
    
    Policy definitions are referenced in routing protocol
    configurations using import and export configuration statements.
    The arguments are members of an ordered list of named policy
    definitions which comprise a policy chain, and optionally, an
    default final dispositinn action for policy chain (i.e., reject
    or accept) applied to routes that does not match conditions of 
    any statement of any policy in chain, or none of matching 
    statement has final dispositinn action (acept or reject).
    If default final dispositinn action for policy chain is not
    explicitly specifird, implicit default action is reject.
    Note: This file contains definition of single policy, wherehas 
    definition of policy chain and default final dispositinn action 
    for policy chain are properties of protocol's policy application,
    and are defiend as part of protocols model. The text in this
    paragraph provides context information.

    Evaluation of each policy definition proceeds by evaluating its
    corresponding individual policy statements in order.  When a
    condition statement in a policy statement is satisfied, the
    corresponding action statement is executed.  If the action
    statement has a final disposition configured as policy result, either
    accept-route or reject-route, evaluation of the current policy definition
    stops, and no further policy statements are evaluated. In case there is a
    policy chain, no further policy definitions in the chain are evaluated.

    If the action statement has the NEXT_STATEMENT policy result, all the
    defined actions are exectued and policy evaluation proceeds to the next
    statement. If statement is last in current policy, and current policy is
    not the last policy of the policy chain, policy evaluation proceeds to 
    the first statement of next policy. If statement is last in current policy,
    and current policy is the last policy of the policy chain, default final 
    disposition action for policy chain is applied.

    If the action statement has the NEXT_POLICY policy result, all the
    defined actions in this stetement are exectued, evaluation of statements
    within the current policy stops and evaluation proceeds to the first 
    statement of next policy in the chain. If statement is last in current
    policy, and current policy is the last policy of the policy chain, default
    final disposition action for policy chain is applied.

    Policy 'subroutines' (or nested policies) are supported by
    allowing policy statement conditions to reference another policy
    definition which applies conditions and actions from the
    referenced policy before returning to the calling policy
    statement and resuming evaluation.  If the called policy
    results in an accept-route (either explicit or by default), then
    the subroutine returns an effective true value to the calling
    policy.  Any other policy result action - reject-route, next_statement, 
    next_policy - returns false.  If the subroutine returns true, 
    the calling policy continues to evaluate the remaining conditions of current
    stement (using a modified route if the subroutine performed any changes to
    the route), and if also satisfied executes statement's action. If the 
    subroutine returns true, evaluation proceeds to next statement or policy.
    ";
```

dplore:
I am changing the scope of this PR back to be only NEXT-STATEMENT since we are only finding 1 instance of NEXT-POLICY functionality.

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -52,9 +52,18 @@ module openconfig-routing-policy {
...
-    statement has either accept-route or reject-route actions, policy
-    evaluation of the current policy definition stops, and no further
-    policy definitions in the chain are evaluated.
+    statement has a final disposition configured as policy result, either
+    accept-route or reject-route, evaluation of the current policy definition
+    stops, and no further policy statements are evaluated. In case there is a
+    policy chain, no further policy definitions in the chain are evaluated.
+
+    If the action statement has the NEXT_STATEMENT policy result, all the
+    defined actions are exectued and policy evaluation proceeds to the next
```
wenovus, line 61:
```suggestion
    defined actions are executed and policy evaluation proceeds to the next
```

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -52,9 +52,18 @@ module openconfig-routing-policy {
...
+    statement.  If the action statement has NEXT-POLICY as the policy result,
+    there is no final disposition of the route, all the defined actions are
+    executed, evaluation of statements within the current policy stops and
+    evaluation proceeds to the next policy in the chain, if any.  The
+    NEXT_STATEMENT is the default policy result action.
```
wenovus, line 66: still talking about NEXT-POLICY

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -77,7 +86,13 @@ module openconfig-routing-policy {
     the remaining conditions (using a modified route if the
     subroutine performed any changes to the route).";
 
-  oc-ext:openconfig-version "3.3.0";
+  oc-ext:openconfig-version "3.4.0";
+
+  revision "2023-10-11" {
+    description
+      "Add NEXT_STATEMENT policy-result-type enums.";
```
wenovus, line 93:
```suggestion
      "Add NEXT_STATEMENT policy-result-type enum value.";
```

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -162,10 +177,16 @@ module openconfig-routing-policy {
...
       enum REJECT_ROUTE {
-        description "Policy rejects the route";
+        description "Policy rejects the route and evaluation of the
+        current policy definition stops.";
+      }
+      enum NEXT_STATEMENT {
+        description "Any modifications of the route are preserved and the evaluation of the policy will continue to the
+        next statement.";
       }
     }
```
wenovus, line 191: Add a default statement for this typedef?

release/models/policy/openconfig-routing-policy.yang
dplore: thanks for catching

release/models/policy/openconfig-routing-policy.yang
dplore: done

wenovus (approved):

robshakir (approved):
LGTM.

TANTRAN1:
Hello Darren, hello all. Sorry to react late. Here in Orange group  we have our +20 IP backbone networks worldwide, which are  built with various vendors routers (mainly: Cisco XR, Juniper, Nokia, Huawei). For more than 20 years, in our engineering guidelines history, the routers configurations templates  converged to obtain most generic and most re-usable config as possible. In this spirit,  we have a wide usage of  NEXT-STATEMENT, and also  NEXT-POLICY.
Nokia and Juniper  OSs  , for example, allow to configure  chained policies.

When automating the production using openconfig yang we would like the possibility to keep following behavior:

example:

export  route-policy  [  policy1   policy2   policy3]

if inside policy1, even statement x1   has result accept, we would
need keep the possibility  to continue   the evaluation of policy2   , then finally the evaluation of policy 3
so need  necessarily  to configure "next policy"  as policy-result  choice inside entry statement.
To justify the need of usage of NEXT-POLICY in Openconfig, the spirit of chained policies for Orange Group rules, is to write them to have most generic ones.  Our idea is to "dedicate" one policy  to one generic task   . Doing this, we ensure the clarity of each policy-map,  our engineers of all Orange  understand  immediately the function of each single policy with its name.
Each policy is seen as a basic brick, so our network engineers   can elaborate the policies (made of several chained basic policies)  adapted to their current use-case.

policy 1   :  task A
policy 2 :  task  B
policy 3 : task C

These kind of configuration of chained policies are widely used among +50000 backbone routers inside Orange networks.
I provide examples extracted from Operational backbone routers from one Orange Affiliate from Europe :

NOKIA:
           group "GRP-IBGP-RR"
                description "Groupe iBGP IPv4 vers Route-Reflector"
                family ipv4
                min-route-advertisement 1
                type internal
                export "POCY-CONN2BGP" "POCY-STATIC2BGP" "POCY-RR-OUT"

            policy-statement "POCY-CONN2BGP"
                entry 5
                    description "ACPT-CONNECTED"
                    from
                        protocol direct
                    exit
                    action next-entry
                        community add "CMNY-CONNECTED"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 35
                    description "ACTP-ASBC"
                    from
                        protocol direct
                        prefix-list "PRFX-RESEAUX-ASBC"
                    exit
                    action next-entry
                        community replace "CMNY-CONNECTED-ASBC"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 40
                    description "ACTP-PAS-RADIUS_VAL"
                    from
                        protocol direct
                        prefix-list "PRFX-PAS-RADIUS"
                    exit
                    action next-entry
                        community add "CMNY-PAS-RADIUS"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 45
                    description "ACTP-PAS-ADMIN_VAL"
                    from
                        protocol direct
                        prefix-list "PRFX-PAS-ADMIN"
                    exit
                    action next-entry
                        community add "CMNY-PAS-ADMIN"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 60
                    description "ACTP-PAS-FW"
                    from
                        protocol direct
                        prefix-list "PRFX-PAS-FW"
                    exit
                    action next-entry
                        community add "CMNY-PAS-FW"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 65
                    description "ACPT-CDN-AKA"
                    from
                        protocol direct
                        prefix-list "PRFX-AKAMAI-CONN"
                    exit
                    action next-entry
                        community add "CMNY-CDN-AKAMAI-CONN"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 1000
                    description "Passage a la policy suivante"
                    action next-policy
                    exit
                exit
            exit
            policy-statement "POCY-STATIC2BGP"
                entry 65
                    description "ACPT-CDN-AKA"
                    from
                        protocol static
                        tag 65
                    exit
                    action next-entry
                        community add "CMNY-CDN-AKAMAI"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 66
                    description "ACPT-CDN-AKA-Lo_HAS"
                    from
                        protocol static
                        tag 66
                    exit
                    action next-entry
                        community add "CMNY-CDN-AKAMAI-RBCI"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 70
                    description "STATIC-ASBC"
                    from
                        protocol static
                        tag 140
                    exit
                    action next-entry
                        community add "CMNY-STATIC-ASBC"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 100
                    description "ACTP-PAS-RADIUS"
                    from
                        protocol static
                        tag 421
                    exit
                    action next-entry
                        community add "CMNY-PAS-RADIUS"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 105
                    description "ACTP-PAS-ADMIN"
                    from
                        protocol static
                        tag 420
                    exit
                    action next-entry
                        community add "CMNY-PAS-ADMIN"
                        local-preference 310000
                        tag 1
                    exit
                exit
                entry 1000
                    description "Passage a la policy suivante"
                    action next-policy
                    exit
                exit
            exit
            policy-statement "POCY-RR-OUT"
                entry 10
                    description "ACPT-NON-BGP : direct, statique, .."
                    from
                        tag 1
                    exit
                    action accept
                        origin igp
                        metric set 0
                        next-hop-self
                    exit
                exit
                entry 15
                    description "BLACKHOLE-DYNAMIQUE-RBCI"
                    from
                        protocol bgp
                        community "CMNY-BLACKHOLE"
                    exit
                    action accept
                        community add "CMNY-TRANSPARENCE-BGP"
                    exit
                exit
                entry 20
                    description "TRANSPARENCE-BGP"
                    from
                        protocol bgp
                        community "CMNY-TRANSPARENCE-BGP"
                    exit
                    action accept
                    exit
                exit
                entry 1000
                    description "ACPT-EBGP"
                    from
                        protocol bgp
                    exit
                    action accept
                        next-hop-self
                    exit
                exit
                default-action drop
                exit
            exit

TANTRAN1:
In these Juniper config extract, we have 3 chained policies, each policy has got a determined scope,   "action next-policy" is useful.

ORANGE operational network config extract JUNIPER:    same spirit,  chained policies used in context of L3VPN routing-instance  (-> oc : network-instance )

CUST-VRF-LDAFRA_BVPN {
        instance-type vrf;
        interface xe-2/2/1.1913;
        route-distinguisher 3215:409023530;
        vrf-import [ GEN-POL-IN-REJECT-INET6 GEN-POL-IN-RADEQUA CUST-POL-IN-LDAFRA_BVPN GEN-POL-BOTH-REJECT ];
        vrf-export [ GEN-POL-OUT-VPN-STATIC-to-MPBGP GEN-POL-OUT-VPN-EXPORT CUST-POL-OUT-LDAFRA_BVPN GEN-POL-BOTH-ACCEPT ];
        routing-options {
            static { .....................


    policy-statement GEN-POL-OUT-VPN-STATIC-to-MPBGP {
        term 10 {
            from {
                protocol [ static access ];
                tag 110;
            }
            then next policy;
        }
        term 20 {
            from {
                protocol [ static access ];
                tag 181;
            }
            then {
                local-preference 200;
                community add GEN-COM-10283-20077;
                community add GEN-COM-10283-120;
                next policy;
            }
        }
        term 30 {
            from {
                protocol [ static access ];
                tag 182;
            }
            then {
                local-preference 100;
                community add GEN-COM-10283-20066;
                community add GEN-COM-10283-110;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 33 {
            from {
                protocol [ static access ];
                tag 800;
            }
            then {
                local-preference 80;
                community add GEN-COM-10283-20080;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 34 {
            from {
                protocol [ static access ];
                tag 180;
            }
            then {
                local-preference 80;
                community add GEN-COM-10283-20080;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 35 {
            from {
                protocol [ static access ];
                tag 183;
            }
            then {
                local-preference 100;
                community add GEN-COM-10283-20066;
                community add GEN-COM-10283-110;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 37 {
            from {
                protocol [ static access ];
                tag 184;
            }
            then {
                local-preference 80;
                community add GEN-COM-10283-20066;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 40 {
            from {
                protocol [ static access ];
                tag 9911;
            }
            then {
                community add GEN-COM-10283-20011;
                next policy;
            }
        }
        term 50 {
            from {
                protocol [ static access ];
                tag 9922;
            }
            then {
                community add GEN-COM-10283-20022;
                next policy;
            }
        }
        term 55 {
            from {
                protocol [ static access ];
                tag 9933;
            }
            then {
                community add GEN-COM-10283-20033;
                next policy;
            }
        }
        term 57 {
            from {
                protocol [ static access ];
                tag 811;
            }
            then {
                local-preference 80;
                community add GEN-COM-10283-20080;
                community add GEN-COM-10283-20011;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 60 {
            from {
                protocol [ static access ];
                tag 261;
            }
            then {
                local-preference 100;
                community add GEN-COM-10283-20066;
                community add GEN-COM-10283-20011;
                community add GEN-COM-10283-110;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 70 {
            from {
                protocol [ static access ];
                tag 262;
            }
            then {
                local-preference 100;
                community add GEN-COM-10283-20066;
                community add GEN-COM-10283-20022;
                community add GEN-COM-10283-110;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 80 {
            from {
                protocol [ static access ];
                tag 271;
            }
            then {
                local-preference 200;
                community add GEN-COM-10283-20077;
                community add GEN-COM-10283-20011;
                community add GEN-COM-10283-120;
                next policy;
            }
        }
        term 90 {
            from {
                protocol [ static access ];
                tag 272;
            }
            then {
                local-preference 200;
                community add GEN-COM-10283-20077;
                community add GEN-COM-10283-20022;
                community add GEN-COM-10283-120;
                next policy;
            }
        }
        term 100 {
            from {
                protocol [ static access ];
                tag 281;
            }
            then {
                community add GEN-COM-10283-20088;
                community add GEN-COM-10283-20011;
                next policy;
            }
        }
        term 110 {
            from {
                protocol [ static access ];
                tag 282;
            }
            then {
                community add GEN-COM-10283-20088;
                community add GEN-COM-10283-20022;
                next policy;
            }
        }
        term 120 {
            from {
                protocol [ static access ];
                tag 291;
            }
            then {
                community add GEN-COM-10283-20099;
                community add GEN-COM-10283-20011;
                next policy;
            }
        }
        term 130 {
            from {
                protocol [ static access ];
                tag 292;
            }
            then {
                community add GEN-COM-10283-20099;
                community add GEN-COM-10283-20022;
                next policy;
            }
        }
        term 140 {
            from {
                protocol [ static access ];
                tag 500;
            }
            then {
                community add GEN-COM-NO-ADVERTISE;
                next policy;
            }
        }
        term 150 {
            from {
                protocol [ static access ];
                tag 2362;
            }
            then {
                community add GEN-COM-10283-2362;
                next policy;
            }
        }
        term 160 {
            from {
                protocol [ static access ];
                tag 50011;
            }
            then {
                community add GEN-COM-10283-50011;
                next policy;
            }
        }
        term 165 {
            from {
                protocol [ static access ];
                tag 50012;
            }
            then {
                community add GEN-COM-10283-50012;
                next policy;
            }
        }
        term 170 {
            from {
                protocol [ static access ];
                tag 50021;
            }
            then {
                community add GEN-COM-10283-50021;
                next policy;
            }
        }
        term 175 {
            from {
                protocol [ static access ];
                tag 50022;
            }
            then {
                community add GEN-COM-10283-50022;
                next policy;
            }
        }
        term 180 {
            from {
                protocol [ static access ];
                tag 50031;
            }
            then {
                community add GEN-COM-10283-50031;
                next policy;
            }
        }
        term 185 {
            from {
                protocol [ static access ];
                tag 50032;
            }
            then {
                community add GEN-COM-10283-50032;
                next policy;
            }
        }
        term 190 {
            from {
                protocol [ static access ];
                tag 50041;
            }
            then {
                community add GEN-COM-10283-50041;
                next policy;
            }
        }
        term 195 {
            from {
                protocol [ static access ];
                tag 50042;
            }
            then {
                community add GEN-COM-10283-50042;
                next policy;
            }
        }
        term 200 {
            from {
                protocol [ static access ];
                tag 50051;
            }
            then {
                community add GEN-COM-10283-50051;
                next policy;
            }
        }
        term 205 {
            from {
                protocol [ static access ];
                tag 50052;
            }
            then {
                community add GEN-COM-10283-50052;
                next policy;
            }
        }
        term 210 {
            from {
                protocol [ static access ];
                tag 50061;
            }
            then {
                community add GEN-COM-10283-50061;
                next policy;
            }
        }
        term 215 {
            from {
                protocol [ static access ];
                tag 50062;
            }
            then {
                community add GEN-COM-10283-50062;
                next policy;
            }
        }
        term 220 {
            from {
                protocol [ static access ];
                tag 251;
            }
            then {
                community add GEN-COM-10283-20055;
                community add GEN-COM-10283-20011;
                next policy;
            }
        }
        term 225 {
            from {
                protocol [ static access ];
                tag 252;
            }
            then {
                community add GEN-COM-10283-20055;
                community add GEN-COM-10283-20022;
                next policy;
            }
        }
        term 10020 {
            from {
                protocol [ static access ];
                tag2 181;
            }
            then {
                local-preference 200;
                community add GEN-COM-10283-20077;
                community add GEN-COM-10283-120;
                next policy;
            }
        }
        term 10030 {
            from {
                protocol [ static access ];
                tag2 182;
            }
            then {
                local-preference 100;
                community add GEN-COM-10283-20066;
                community add GEN-COM-10283-110;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 10040 {
            from {
                protocol [ static access ];
                tag2 9911;
            }
            then {
                community add GEN-COM-10283-20011;
                next policy;
            }
        }
        term 10050 {
            from {
                protocol [ static access ];
                tag2 9922;
            }
            then {
                community add GEN-COM-10283-20022;
                next policy;
            }
        }
        term 10060 {
            from {
                protocol [ static access ];
                tag2 261;
            }
            then {
                local-preference 100;
                community add GEN-COM-10283-20066;
                community add GEN-COM-10283-20011;
                community add GEN-COM-10283-110;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 10070 {
            from {
                protocol [ static access ];
                tag2 262;
            }
            then {
                local-preference 100;
                community add GEN-COM-10283-20066;
                community add GEN-COM-10283-20022;
                community add GEN-COM-10283-110;
                as-path-expand "64544 64544";
                next policy;
            }
        }
        term 10080 {
            from {
                protocol [ static access ];
                tag2 271;
            }
            then {
                local-preference 200;
                community add GEN-COM-10283-20077;
                community add GEN-COM-10283-20011;
                community add GEN-COM-10283-120;
                next policy;
            }
        }
        term 10090 {
            from {
                protocol [ static access ];
                tag2 272;
            }
            then {
                local-preference 200;
                community add GEN-COM-10283-20077;
                community add GEN-COM-10283-20022;
                community add GEN-COM-10283-120;
                next policy;
            }
        }
        term DEFAULT-STATIC {
            from protocol [ static access ];
            then {
                community add GEN-COM-3215-202;
                next policy;
            }
        }
        term NEXT-POLICY {
            then next policy;
        }
    }

    policy-statement GEN-POL-OUT-VPN-EXPORT {
        term 40 {
            from tag 99983;
            then {
                community add CUST-RT-OPERA_OCNADMIN_CLT_TPC-3215_471094000;
                accept;
            }
        }
        term 50 {
            from tag 9385;
            then {
                community add GEN-COM-10283-20112;
                community add GEN-RT-10283-140000;
                next policy;
            }
        }
        term 60 {
            from tag 9386;
            then {
                community add GEN-COM-10283-20186;
                community add GEN-COM-10283-20593;
                community add GEN-RT-10283-3964000;
                accept;
            }
        }
        term 70 {
            from tag 9387;
            then {
                community add GEN-COM-10283-20187;
                community add GEN-COM-10283-20593;
                community add GEN-RT-10283-3964000;
                community add GEN-RT-10283-4014000;
                accept;
            }
        }
        term 80 {
            from tag 9388;
            then {
                community add GEN-COM-10283-20112;
                community add GEN-RT-3215-160000;
                accept;
            }
        }
        term 90 {
            from tag 99986;
            then {
                community add GEN-COM-10283-20066;
                community add CUST-RT-SG_OCNADMIN_CLT_TPC-10283_78351000;
                accept;
            }
        }
        term 95 {
            from tag 99987;
            then {
                community add GEN-COM-10283-20077;
                community add CUST-RT-SG_OCNADMIN_CLT_TPC-10283_78351000;
                accept;
            }
        }
        term 130 {
            from tag 99405;
            then {
                community add GEN-COM-10283-20405;
                community add GEN-RT-10283-294000;
                next policy;
            }
        }
        term 140 {
            from community GEN-COM-10283-20405;
            then {
                community add GEN-RT-10283-294000;
                next policy;
            }
        }
        term 150 {
            from tag 99404;
            then {
                community add GEN-COM-10283-20404;
                community add GEN-RT-10283-244000;
                accept;
            }
        }
        term 160 {
            from community GEN-COM-10283-20404;
            then {
                community add GEN-RT-10283-244000;
                accept;
            }
        }
        term 170 {
            from tag 99408;
            then {
                community add GEN-COM-10283-20408;
                community add GEN-RT-10283-304000;
                accept;
            }
        }
        term 180 {
            from community GEN-COM-10283-20408;
            then {
                community add GEN-RT-10283-304000;
                accept;
            }
        }
        term 190 {
            from tag 99407;
            then {
                community add GEN-COM-10283-20407;
                community add GEN-RT-10283-314000;
                accept;
            }
        }
        term 200 {
            from community GEN-COM-10283-20414;
            then {
                community add GEN-RT-10283-140000;
                next policy;
            }
        }
        term 210 {
            from tag 99415;
            then {
                community add GEN-COM-10283-20415;
                community add GEN-RT-10283-434000;
                next policy;
            }
        }
        term 220 {
            from tag 99413;
            then {
                community add GEN-COM-10283-20413;
                community add GEN-RT-10283-424000;
                accept;
            }
        }
        term 290 {
            from tag 99425;
            then {
                community add GEN-RT-10283-914000;
                next policy;
            }
        }
        term 300 {
            from tag 99426;
            then {
                community add GEN-COM-10283-20426;
                community add GEN-RT-10283-754000;
                accept;
            }
        }
        term 301 {
            from community GEN-COM-10283-20426;
            then {
                community add GEN-RT-10283-754000;
                accept;
            }
        }
        term 400 {
            from community GEN-COM-10283-20104;
            then {
                community add GEN-RT-10283-30000;
                community add GEN-RT-10283-197814000;
                next policy;
            }
        }
        term 410 {
            from community GEN-COM-10283-20105;
            then {
                community add GEN-RT-10283-30000;
                community add GEN-RT-10283-197814000;
                accept;
            }
        }
        term 460 {
            from tag 99997;
            then {
                community add GEN-RT-10283-214000;
                community add GEN-COM-NO-ADVERTISE;
                accept;
            }
        }
        term 480 {
            from tag [ 99996 99998 ];
            then {
                community add GEN-COM-10283-20103;
                community add GEN-RT-10283-44000;
                community add GEN-RT-10283-197814000;
                accept;
            }
        }
        term 500 {
            from tag [ 99995 99999 ];
            then {
                community add GEN-COM-10283-20103;
                community add GEN-RT-10283-30000;
                community add GEN-RT-10283-197814000;
                accept;
            }
        }
        term 600 {
            from tag 99994;
            then {
                community add GEN-COM-10283-20194;
                community add GEN-RT-3215-261144000;
                accept;
            }
        }
        term 10050 {
            from tag2 9385;
            then {
                community add GEN-COM-10283-20112;
                community add GEN-RT-10283-140000;
                next policy;
            }
        }
        term 10190 {
            from tag2 99407;
            then {
                community add GEN-COM-10283-20407;
                community add GEN-RT-10283-314000;
                accept;
            }
        }
        term 10220 {
            from tag2 99413;
            then {
                community add GEN-COM-10283-20413;
                community add GEN-RT-10283-424000;
                accept;
            }
        }
        term 10290 {
            from tag2 99425;
            then {
                community add GEN-RT-10283-914000;
                next policy;
            }
        }
        term NEXT-POLICY {
            then next policy;
        }
    }

rszarecki:
@TANTRAN1 
This PR has been merged half year ago. Please find some explanation and context:
1. Before this merge, OpenConfig supported only ACCEPT_ROUTE or REJECT_ROUTE as disposition. This merge added NEXT_STATEMENT. There is no regression in functionality.
2. OpenConfig as it is today supports policy chaininig at attachment point. specificly `.../apply-policy/ingress-policy` and `.../apply-policy/ingress-policy` are both of `leaf-lists` not just `leafs`.
2. Example you have provided using NOKIA's -  all entries with `next-policy` are the last entries of policy. The same is achivable in OpenConfig via using `NEXT_STATEMENT` in last statement in policy. Or even by not having this last unconditional statement at all. Please read all description in yang file. Specificly https://github.com/openconfig/public/blob/815341544e58548526e7f9bec4e0ed2d615b69ac/release/models/policy/openconfig-routing-policy.yang#L24
3. Indeed JUNOS example uses `next policy` in meny terms. Translating this policy to OpenConfig would be very complex without additinal knowlages . E.g is we know that static route has only one tag, probably we can use `NEXT_STATEMENT`.
4. You mentioned that Orange is using also XR. Can you provide example how IOS XR RPL provides "next policy" logic? I'm not able to find such - AFAIK XR RPL support only "drop", "dome", "pass" (and implicit "done" if neither is provided and route attributes are modified). Or, prehaps Orange can crate policies with desired logic without "next term" construct?


It is philosofical question if we want OpenConfig to be suprset of all vendors capabilities or rather reduced set. If former, you will have very flexible OpenConfig but then each vendor will support only partially. The net result would be that operator will either generate vendor-aware configuration what defeat purpose of vendor agnosticity of OpenConfig. Or will self limit and use only subset of OpenConfig common to all vendors it uses. And this would make high acceptance barier for new vendor. So other form of vendor lock-in.
You can contribute to OpenConfig by Pull Request and add "NEXT_POLICY" action, if you fill strongly you need it for your project. Just ensure that your vendors are going to implement it. Othwerwise it become only never used noise in OpenConfig  model.

TANTRAN1:
Hello rszarecki
I confirm I read carefully   all the contributions inside #949 ...

Juniper & Nokia:  have "next-policy" in their implem.
Huawei VRP: no "next-policy". Just  "goto"   (eq.  next-statement)
Cisco:  no explicit "next-policy".  RPL not chained like Juniper/Nokia can do.  Cisco  RPL can use "apply" which calls other policy from within other policy. But not exact equivalent of chained policies (Nokia/juniper).

I know, The good example of configuration to discuss ,  which requires  "next-policy"  is the ones I provide with JUNIPER   (action next-policy   written in numerous  terms policy-statements, not only the end of policy)

Question "2-cent"  :  is the "next-statement"   put at the last entry of a policy_1,  makes start the evaluation of the policy_2 , even prefix matched conditions of policy_1 ?   do you see a mean to replace effect of "next-policy"  by   usage of "next-statement"  ?

If not,  I maintain the need to re-open the request of having "next-policy" ......  considering the potential impact of rewriting all RPL  for some Nokia/Juniper   customers

Thanks

PR #954, opened 2023-09-07
Add graceful shutdown as BGP_WELL_KNOWN_STD_COMMUNITY

dplore:
### Change Scope

* Add graceful shutdown as a BGP well known community 
* Reference [RFC 8326](https://datatracker.ietf.org/doc/html/rfc8326)

### DIFF
--- a/release/models/bgp/openconfig-bgp-types.yang
+++ b/release/models/bgp/openconfig-bgp-types.yang
@@ -372,7 +372,8 @@ module openconfig-bgp-types {
     description
       "An autonomous system which supports the graceful shutdown
       receiver procedure receiving NLRI tagged with this community
-      will set LOCAL_PREF to a low value for those NLRI.";
+      will set LOCAL_PREF to a low value for those NLRI.  This
+      community has a value of 0xFFFF0000.";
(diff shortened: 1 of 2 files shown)


release/models/bgp/openconfig-bgp-types.yang
```diff
@@ -361,6 +367,15 @@ module openconfig-bgp-types {
     reference "RFC3765";
   }
 
+  identity GRACEFUL_SHUTDOWN {
+    base BGP_WELL_KNOWN_STD_COMMUNITY;
+    description
+      "An autonomous system which supports the graceful shutdown
+      receiver procedure receiving NLRI tagged with this community
+      will set LOCAL_PREF to a low value for those NLRI.";
```
aashaikh, line 375: Could add the value of the community (0xFFFF0000), though we haven't done it consistently in all of these community identities.

aashaikh (approved):
LGTM

release/models/bgp/openconfig-bgp-types.yang
dplore: Good idea, I added.

PR #955, opened 2023-09-08
initial commit

sbarvick:

### Change Scope

* Added base OpenConfig Kubernetes data model with intention of representing what is encoded in manifest files to be invokes with kubectl apply commands.   I have an implementation that does that based on this data model.  It creates the manifest files based on what is configured (using Tail-f ConfD) and invokes them with an action.    
* The action is not included in this data model because the instantiation of the data model should be application specific.   e.g. the data model changes could be handled with other kubectl commands and not necessary 'apply'.
* Kubernetes is very rich so the intention is for this to be the base for additions to fully capture the richness.
* The concentration has been on configuration in the OpenConfig style so the state modeling is largely open.  It also brings up a question for very hierarchical models like this - because the state usually includes the configuration as well as operational state, the top levels of the state hierarchy would include a lot of config if is to include the top level configuration.   I'm sure there are examples and precedent so I'll see how the comments go.
### Platform Implementations

 * Standard Kubernetes administrated through kubectl.

### DIFF
--- a/release/models/kubernetes/openconfig-kubernetes-state.yang
+++ b/release/models/kubernetes/openconfig-kubernetes-state.yang
@@ -1,0 +1,50 @@
+submodule openconfig-kubernetes-state {
+
+  belongs-to openconfig-kubernetes {
+    prefix "oc-k8s";
+  }
+
+  import openconfig-extensions { prefix oc-ext; }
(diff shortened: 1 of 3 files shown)


sbarvick:
Submitted from wrong account.  Need to switch to account that is covered by CLA

PR #956, opened 2023-09-08
Data model for Kubernetes manifest files

sbarvick-dd:
Change Scope

- Added base OpenConfig Kubernetes data model with intention of representing what is encoded in manifest files to be invokes with kubectl apply commands. I have an implementation that does that based on this data model. It creates the manifest files based on what is configured (using Tail-f ConfD) and invokes them with an action.
- The action is not included in this data model because the instantiation of the data model should be application specific. e.g. the data model changes could be handled with other kubectl commands and not necessary 'apply'.
- Kubernetes is very rich so the intention is for this to be the base for additions to fully capture the richness.
- The concentration has been on configuration in the OpenConfig style so the state modeling is largely open. It also brings up a question for very hierarchical models like this - because the state usually includes the configuration as well as operational state, the top levels of the state hierarchy would include a lot of config if is to include the top level configuration. I'm sure there are examples and precedent so I'll see how the comments go.

Platform Implementations

- Standard Kubernetes administrated through kubectl.

### DIFF
--- a/release/models/kubernetes/.spec.yml
+++ b/release/models/kubernetes/.spec.yml
@@ -1,7 +1,7 @@
 - name: openconfig-kubernetes
   docs:
-    - yang/bgp/openconfig-kubernetes-types.yang
-    - yang/bgp/openconfig-kubernetes.yang
+    - yang/kubernetes/openconfig-kubernetes-types.yang
+    - yang/kubernetes/openconfig-kubernetes.yang
   build:
-    - yang/bgp/openconfig-kubernetes.yang
+    - yang/kubernetes/openconfig-kubernetes.yang
   run-ci: true
--- a/release/models/kubernetes/.spec.yml
+++ b/release/models/kubernetes/.spec.yml
@@ -1,0 +1,7 @@
+- name: openconfig-kubernetes
+  docs:
+    - yang/bgp/openconfig-kubernetes-types.yang
+    - yang/bgp/openconfig-kubernetes.yang
+  build:
+    - yang/bgp/openconfig-kubernetes.yang
+  run-ci: true
--- a/release/models/kubernetes/openconfig-kubernetes-state.yang
+++ b/release/models/kubernetes/openconfig-kubernetes-state.yang
@@ -1,0 +1,49 @@
+submodule openconfig-kubernetes-state {
+
+  belongs-to openconfig-kubernetes {
+    prefix "oc-k8s";
+  }
+
+  import openconfig-extensions { prefix oc-ext; }
+  import openconfig-inet-types { prefix oc-inet; }
+
+  // meta
+  organization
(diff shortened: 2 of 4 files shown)


robshakir:
Thanks for the contribution. This is an interesting area to discuss. 

OpenConfig generally has focused on network-device-exposed data models across a variety of different classes of device [0]. We have not generally considered creating data models that allow for the configuration of non-network-device infrastructure such as k8s. The closest to container management that we've explored thus far is [`gnoi.containerz`](https://github.com/openconfig/gnoi/blob/main/containerz/containerz.proto) which is still focused on the API to initialise containers onto a network device, and covers only relatively simple container use cases.

As I understand the use case here, you're looking for a standardised data model that can be used to capture the input that is required to create manifests (i.e., YAML files) in an off-device system (in your case tail-f confd). Do you expect that the YANG models would be used by others, or is this something that is specific to the choice of using a YANG-centric system to create these manifests that might be specific to your deployment?

I worry about creating a data model in YANG in OpenConfig that isn't really abstracting anything -- since AIUI, you're not proposing to create a data model that describes a container manifest that could be translated to, say, docker-swarm or k8s. Perhaps you could describe why you chose OpenConfig as a target here -- do you see advantages of the wider ecosystem that are beyond creating an abstraction?

Thanks,
r.

[0]: the only case that I can think of thus far that isn't device-centric in this way is the catalogue model

sbarvick-dd:
Thanks for interest and very valid question.   The motivation for this effort came from considering the OAM functions of a white box networking system utilizing standard and public Yangs, primarily from the IETF and OpenConfig.    For a white box system, it may happen that overall environment is Docker containers managed by Kubernetes so in that sense, it could be architecturally and functionally appropriate for the OpenConfig platform collection of Yangs.  But, because it was specific to k8s, I made it its own directory after struggling with the question of abstracting or not.  Theoretically, I would like it to be more abstract but it seems that k8s has enough maturity, adoption, and richness of its own that I thought I would just go down the path of making it a data model for the manifest yaml because yaml is NOT a data model.  In fact, I feel that is another opportunity for OpenConfig - to document data models for things that only have yaml configuration serialization (but that is for another time).

dplore:
Do I understand the target scenario for this use case is a network device (or simulated device via confd) running a kubernetes cluster?   Or is the network device a host within a kubernetes cluster with many hosts?

I guess a broad/general question is, does it make more sense to manage kubernetes on a network device using the Kubernetes API and the ecosystem of Kubernetes tools?    What are the use cases that motivate one to manage, even partially, a kubernetes cluster using network device API's?  

Can you provide references for network device implementations which would support a netconf or gNMI interface to define kubernetes clusters that run on device?  You mention a white-box solution; can you provide references to it's implementation/hosting of kubernetes?

PR #959, opened 2023-09-15
Add hashing model

sallylsy:
This PR replaces #939 

### Change Scope

* Introduce a hashing policy which can used to
  * Select an implementation defined algorithm, offset and seed value
  * Select which fields in a packet should be used as an input into a hash function for for load balancing
* Add hashing-policy to interfaces to allow configuration of which algorithm to use on a per interface basis
* Add hashing-policy to `/system` to allow configuration of an algorithm used by all the ports in an implementation
* It is expected that implementations will vary widely in which specific options they support as these are hardware dependent.  The goal is to define a what is operationally useful and supported by at least a subset of hardware.  Other hardware may offer only partial support or even no support.
 
### Implementations 
* [Arista EOS - port-channel load-balance](https://www.arista.com/en/um-eos/eos-port-channels-and-lacp#xx1151429) - defines a variety of configuration items which are specific to different hardware platforms.  They include references to hash policies with attributes including hash polynomials, hash seeds, and hash fields
* [JunOS - per flow load balancing on hash values](https://www.juniper.net/documentation/us/en/software/junos/sampling-forwarding-monitoring/topics/concept/policy-configuring-per-flow-load-balancing-based-on-hash-values.html) has partial support of the configuration.  It defines a way to configure a hash seed on some hardware platforms.  [forwarding-options/hash-key](https://www.juniper.net/documentation/us/en/software/junos/interfaces-ethernet-switches/topics/topic-map/switches-interface-load-balancing.html#id-configuring-load-balancing-on-a-lag-link)  allow configuration of per-prefix and mac address based balancing.    There is no support for configuring different hash algorithms nor to configure per interface.
* [Cisco IOS-XR bundle-hash](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/Interfaces/b-interfaces-hardware-component-cr-8000/m-link-bundling-commands-8k.html) allows configuring the hash key based on a list of fixed options  
  * (L3/3-tuple or L4/7-tuple) (L3,L4).
  * Single SA/DA pair (IPv4,IPv6) or range (IPv4 only).
  * There is no support for configuring different hash algorithms.
* [Nokia SR Linux system/load-balancing/hash-options](https://documentation.nokia.com/srlinux/22-3/SR_Linux_Book_Files/Configuration_Basics_Guide/configb-systmmgmt.html#ariaid-title13) allows hash seed and some hash key  configuration support.   There doesn't appear to be support for configuring different hash algorithms.

* [SONIC ecmp/lag hash](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md#hash) contains configuration for selecting packet fields for hash computation.  These are set at a global level.

### Tree View 
Without augment for vendor

```
module: openconfig-system
  +--rw system
     +--rw oc-hashing:hashing
        +--rw oc-hashing:algorithms
        |  +--rw oc-hashing:vendor
        +--rw oc-hashing:hashing-policies
           +--rw oc-hashing:hashing-policy* [name]
              +--rw oc-hashing:name                -> ../config/name
              +--rw oc-hashing:config
              |  +--rw oc-hashing:name?        string
              |  +--rw oc-hashing:seed?        uint64
              |  +--rw oc-hashing:algorithm?   string
              +--ro oc-hashing:state
              |  +--ro oc-hashing:name?        string
              |  +--ro oc-hashing:seed?        uint64
              |  +--ro oc-hashing:algorithm?   string
              +--rw oc-hashing:hash-field-modes
                 +--rw oc-hashing:config
                 |  +--rw oc-hashing:ingress-interface?   boolean
                 |  +--rw oc-hashing:ip-protocol-type?    boolean
                 +--ro oc-hashing:state
                 |  +--ro oc-hashing:ingress-interface?   boolean
                 |  +--ro oc-hashing:ip-protocol-type?    boolean
                 +--rw oc-hashing:ipv4
                 |  +--rw oc-hashing:config
                 |  |  +--rw oc-hashing:src-addr?   boolean
                 |  |  +--rw oc-hashing:dst-addr?   boolean
                 |  |  +--rw oc-hashing:src-port?   boolean
                 |  |  +--rw oc-hashing:dst-port?   boolean
                 |  +--ro oc-hashing:state
                 |     +--ro oc-hashing:src-addr?   boolean
                 |     +--ro oc-hashing:dst-addr?   boolean
                 |     +--ro oc-hashing:src-port?   boolean
                 |     +--ro oc-hashing:dst-port?   boolean
                 +--rw oc-hashing:ipv6
                    +--rw oc-hashing:config
                    |  +--rw oc-hashing:src-addr?     boolean
                    |  +--rw oc-hashing:dst-addr?     boolean
                    |  +--rw oc-hashing:src-port?     boolean
                    |  +--rw oc-hashing:dst-port?     boolean
                    |  +--rw oc-hashing:flow-label?   boolean
                    +--ro oc-hashing:state
                       +--ro oc-hashing:src-addr?     boolean
                       +--ro oc-hashing:dst-addr?     boolean
                       +--ro oc-hashing:src-port?     boolean
                       +--ro oc-hashing:dst-port?     boolean
                       +--ro oc-hashing:flow-label?   boolean
```

With augment for vendor

```
  +--rw system
     +--rw oc-hashing:hashing
        +--rw oc-hashing:algorithms
        |  +--rw oc-hashing:vendor
        |     +--rw google-pins-hashing:broadcom
        |        +--rw google-pins-hashing:broadcom-family
        |           +--ro google-pins-hashing:hashing-algorithms
        |              +--ro google-pins-hashing:hashing-algorithm* [name]
        |                 +--ro google-pins-hashing:name     -> ../state/name
        |                 +--ro google-pins-hashing:state
        |                    +--ro google-pins-hashing:name?          string
        |                    +--ro google-pins-hashing:description?   string
        +--rw oc-hashing:hashing-policies
           +--rw oc-hashing:hashing-policy* [name]
              +--rw oc-hashing:name                -> ../config/name
              +--rw oc-hashing:config
              |  +--rw oc-hashing:name?              string
              |  +--rw oc-hashing:seed?              uint64
              |  +--rw oc-hashing:algorithm?         string
              |  +--rw google-pins-hashing:offset?   uint64
              +--ro oc-hashing:state
              |  +--ro oc-hashing:name?              string
              |  +--ro oc-hashing:seed?              uint64
              |  +--ro oc-hashing:algorithm?         string
              |  +--ro google-pins-hashing:offset?   uint64
              +--rw oc-hashing:hash-field-modes
                 +--rw oc-hashing:config
                 |  +--rw oc-hashing:ingress-interface?   boolean
                 |  +--rw oc-hashing:ip-protocol-type?    boolean
                 +--ro oc-hashing:state
                 |  +--ro oc-hashing:ingress-interface?   boolean
                 |  +--ro oc-hashing:ip-protocol-type?    boolean
                 +--rw oc-hashing:ipv4
                 |  +--rw oc-hashing:config
                 |  |  +--rw oc-hashing:src-addr?   boolean
                 |  |  +--rw oc-hashing:dst-addr?   boolean
                 |  |  +--rw oc-hashing:src-port?   boolean
                 |  |  +--rw oc-hashing:dst-port?   boolean
                 |  +--ro oc-hashing:state
                 |     +--ro oc-hashing:src-addr?   boolean
                 |     +--ro oc-hashing:dst-addr?   boolean
                 |     +--ro oc-hashing:src-port?   boolean
                 |     +--ro oc-hashing:dst-port?   boolean
                 +--rw oc-hashing:ipv6
                    +--rw oc-hashing:config
                    |  +--rw oc-hashing:src-addr?     boolean
                    |  +--rw oc-hashing:dst-addr?     boolean
                    |  +--rw oc-hashing:src-port?     boolean
                    |  +--rw oc-hashing:dst-port?     boolean
                    |  +--rw oc-hashing:flow-label?   boolean
                    +--ro oc-hashing:state
                       +--ro oc-hashing:src-addr?     boolean
                       +--ro oc-hashing:dst-addr?     boolean
                       +--ro oc-hashing:src-port?     boolean
                       +--ro oc-hashing:dst-port?     boolean
                       +--ro oc-hashing:flow-label?   boolean
```

```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw name                    -> ../config/name
        +--rw config
        |  +--rw name?                                            string
        |  +--rw type                                             identityref
        |  +--rw mtu?                                             uint16
        |  +--rw loopback-mode?                                   oc-opt-types:loopback-mode-type
        |  +--rw hashing-policy?                                  -> /oc-hashing:hashing-policies/hashing-policy/name     <-- added
        +--ro state
        |  +--ro name?                                            string
        |  +--ro type                                             identityref
        |  +--ro mtu?                                             uint16
        |  +--ro loopback-mode?                                   oc-opt-types:loopback-mode-type
        |  +--ro hashing-policy?                                  -> /oc-hashing:hashing-policies/hashing-policy/name     <-- added
```


release/models/hashing/openconfig-hashing.yang
dplore, line 1: I think we should move this file to the release/models/system folder

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,332 @@
...
+                description "Use both the source and
+                    destination ports from the transport header in
+                    the calculation of the hash.";
+            }
+            enum TARGET_DEFINED {
+                description "Use the default fields that are used by
+                    the device for calculating the hash.";
+            }
+        }
+        description "The trasport layer fields that should be used to
```
dplore, line 58:
```suggestion
        description "The transport layer fields that should be used to
```

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,332 @@
...
+            type boolean;
+            description "Use the default fields that are used by
+                the device for calculating the hash.";
+        }
+        leaf ipv6-flow-label {
+            type boolean;
+            description "Use the flow label in the IPv6 header
+                to calculate the hash.";
+        }
+        leaf ipv6-ingress-subinterface {
```
dplore, line 158: I think we can leave this off and instead define a grouping with just one boolean for 'ingress-interface' (vs. subinterface).

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,332 @@
...
+            description "Use both the IPv4 source and
+                destination addresses in the calculation of the
+                hash.";
+        }
+        leaf ipv4-target-define {
+            type boolean;
+            description "Use the default fields that are used by
+                the device for calculating the hash.";
+        }
+        leaf ipv4-ingress-subinterface {
```
dplore, line 101: see comment re: ipv6-ingress-subinterface

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,332 @@
...
+        description
+          "Configuration data for hashing policies.";
+
+        leaf name {
+            type string;
+            description
+              "The name of the hashing policy.
+              When a configured user-controlled policy is created by the
+              system, it is instantiated with the same name in the
+              /policies/policy[name]/state list.";
```
dplore, line 191:
```suggestion
              /system/hashing-policies/hashing-policy/name list.";
```

release/models/hashing/openconfig-hashing.yang
```diff
@@ -0,0 +1,332 @@
...
+        leaf description {
+            type string;
+            description
+                "Description of the supported algorithm.";
+        }
+    }
+    grouping supported-algorithms-top {
+        description
+            "Top-level container of the supported algorithms for supported
+            algorithms.";
```
dplore, line 298:
```suggestion
            "Top-level container of the supported algorithms for supported
            algorithms.  The list of algorithms are expected to be defined by
            the target device.";
```

release/models/hashing/openconfig-hashing.yang
sallylsy:
```
    grouping ingress-interface-top {
        leaf ingress-interface {
            type boolean;
            description "Use the ingress subinterface identifired in the
                calculation of the hash.";
        }
    }

grouping ipv6-fields-config {
    ...
    uses ingress-interface-top;
}
grouping ipv4-fields-config {
    ...
    uses ingress-interface-top;
}
```

dplore: Yes, that is what I'm thinking

release/models/hashing/openconfig-hashing.yang
sallylsy: Done

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,326 @@
...
+                    config false;
+                    uses supported-algorithms-state;
+                }
+            }
+        }
+
+    }
+
+    uses supported-algorithms-top;
+    uses hashing-top;
```
dplore, line 325: replace this with augment statements.  This way hashing adds to  the /system model, while staying independent.  One can choose to implement the hashing model or not without affecting the /system model.

Example:
```
  augment "/oc-system:system" {
    description
      "Hashing algorithms and hashing policies";

    uses supported-algorithms-top;
    uses hashing-top;
  }
```

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,326 @@
...
+            description
+                "Description of the supported algorithm.";
+        }
+    }
+    grouping supported-algorithms-top {
+        description
+            "Top-level container of the supported algorithms for supported
+            algorithms.  The list of algorithms are expected to be defined by
+            the target device.";
+        container supported-algorithms {
```
dplore, line 293: Can you rename the containers to include the word `hashing`.  So we get a path (combined with the augment suggestion below) which reads like:

`/system/hashing-algorithms/hashing-algorithm`

release/models/hashing/openconfig-hashing.yang
dplore: Ah, ingress-interface should be instantiated at:  `/system/hashing-policies/hashing-policy/config/ingress-interface`

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,323 @@
...
+                    calculation of the hash.";
+            }
+            enum BOTH {
+                description "Use both the IPv4 source and
+                    destination addresses in the calculation of the
+                    hash.";
+            }
+            enum TARGET_DEFINED {
+                description "Use the default fields that are used by
+                    the device for calculating the hash.";
```
dplore, line 57:
```suggestion
                description "Use fields that are defined by the device for calculating the hash.  Use of this enum is expected when the target contains some default or hardcoded selection of fields which are not described by this model.  For example, implementation may select only a portion of the bits in the source and destination ip address fields and port number fields.  Which bits are used may not be exposed and may not be configurable by the user, by the OPERATING_SYSTEM or by the INTEGRATED_CIRCUIT.  If a TARGET_DEFINED mode isn't available on an implementation, an UNIMPLEMENTED (12) error should be returned if it is attempted to be used.";
```

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,323 @@
...
+                    transport header in the calculation of the hash.";
+            }
+            enum BOTH {
+                description "Use both the source and
+                    destination ports from the transport header in
+                    the calculation of the hash.";
+            }
+            enum TARGET_DEFINED {
+                description "Use the default fields that are used by
+                    the device for calculating the hash.";
```
dplore, line 117:
```suggestion
                description "Use transport fields that are defined by the device for calculating the hash.  Use of this enum is expected when the target contains some default or hardcoded selection of fields which are not described by this model.  For example, an implementation may select only a portion of the bits in the port number fields or may only support fields for a limited set of transports.  Which transports and bits are used may not be exposed and may not be configurable by the user, by the OPERATING_SYSTEM or by the INTEGRATED_CIRCUIT.  If a TARGET_DEFINED mode isn't available on an implementation, an UNIMPLEMENTED (12) error should be returned if it is attempted to be used..";
```

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,336 @@
...
+    augment "/oc-sys:system" {
+        description
+            "Augment for Hashing algorithms and hashing policies";
+
+        uses hashing-top;
+    }
+
+    augment "/oc-intf:interfaces/oc-intf:interface/oc-intf:config" {
+        description
+            "Augment for configuration data hashing policy on the iinterface.";
```
jhaas-pfrc, line 315: typo in interface

jhaas-pfrc:
Moving to the list of supported algorithms from the enums is a good step, IMO.  What is desired is "how do you know what set of things this platform supports", even though it turns configuration into something that needs to query this operational state to know if you CAN successfully configure it.  The alternative is to publish a module that contains supported items for that platform at that version such that it can be discovered at hello with the idea that this isn't quite dynamic data.

The place where this still is problematic is the strings retain a similar level of ambiguity from the prior comment on the enum.  If your string says "XOR", what does it mean?  What fields does it operate against?  The description in the list can help with the descriptions, certainly.

A potential alternative is a registered set of identities where common patterns are managed within the repository in a fashion similar to IETF's IANA function.  Thus, when a particular behavior is registered in that module, all implementations with that behavior have a common configuration item.  Since it would be an identity, vendor augmentation can happen when necessary.

release/models/system/openconfig-hashing.yang
sallylsy: Done

dplore:
> A potential alternative is a registered set of identities where common patterns are managed within the repository in a fashion similar to IETF's IANA function. Thus, when a particular behavior is registered in that module, all implementations with that behavior have a common configuration item. Since it would be an identity, vendor augmentation can happen when necessary.

We could specify a container which implementations are expected to augment containing a flat list of yang identities for "hash algorithms".      This would be something like a "distributed registry" where the user must resolve which yang files to include in their configuration generation. 

OC first introduced this pattern in the pipeline counters with the container at [/components/component/integrated-circuit/pipeline-counters/drop/vendor](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-integrated-circuit-pipeline-counters-drop-vendor).   

Ref: [Vendor Counter Guide](https://github.com/openconfig/public/blob/master/doc/vendor_counter_guide.md) .  (Note there is also a [PR in progress with more verbose description of the intent for these pipeline counters](https://github.com/openconfig/public/pull/961))

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,336 @@
...
+                        description
+                            "Configurable items at the hashing inputs level.";
+                        uses hashing-inputs;
+                    }
+                    container state {
+                        config false;
+                        description "Operational state data at the hashing
+                            inputs level.";
+                        uses hashing-inputs;
+                    }
```
bgardiner, line 244: Can you help me understand what this operational sate data is?

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,336 @@
...
+                }
+                container hash-field-modes {
+                    description
+                        "Container for specifying inputs to be used when
+                        calculating the hash.";
+                    container config {
+                        description
+                            "Configurable items at the hashing inputs level.";
+                        uses hashing-inputs;
+                    }
```
bgardiner, line 238: It could be very useful to have separate hashing input fields for IPv4 vs. IPv6 packets given the additional entropy in the IPv6 flow label (e.g., L3 src and dest addresses+flow label for IPv6 vs. adding in the L4 src and dst ports for IPv4). If I understand correctly, this is not allowed per the current container structure. Any thoughts on revising to allow this additional flexibility?

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,336 @@
...
+                description "Use both the IPv6 source and
+                    destination addresses in the calculation of the
+                    hash.";
+            }
+            enum TARGET_DEFINED {
+                description "Use the default fields that are used by
+                    the device for calculating the hash.";
+            }
+            enum FLOW_LABEL {
+                description "Use the flow label in the IPv6 header
```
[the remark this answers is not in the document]
dplore: @bgardiner, I don't disagree that these are possibilities, but for OpenConfig, we need to cite publicly available documentation which supports these options.  Do you have some you can help us with?  The issue is that the capabilities of chipsets are often not publicly documented and may in fact require a non-disclosure agreement in order to learn what is supported.  

In addition, often one is not given a choice of how many bits to use.  For example,  `INTEGRATED_CIRCUIT`-1  may support 20 bits, `INTEGRATED_CIRCUIT`-2 may support the lower 16 bits,  `INTEGRATED_CIRCUIT`-3 may support 'every other bit', the realm of potential options is as big as the number of bits in the field and in many cases which bits are used are not configurable, you can choose the one option that is available, or not.  

However, what is publicly documented (as noted in the the references above) is the ability to use the ipv6  flow label (or some portion of it) in the hash computation vs. other ipv6 fields.

If there is an operational use case requiring configuration that requires proprietary, non-disclosure information, then that will have to be communicated outside of the OpenConfig models.  My recommended approach is for an implementation to provide yang which extends (augment/deviate/delete) this OpenConfig model.  See the [comment above](https://github.com/openconfig/public/pull/959#issuecomment-1740077978) on this.   We could include this recommendation in this yang and as an accompanying markdown.  WDYT?

release/models/system/openconfig-hashing.yang
dplore: This is a common pattern in OpenConfig where one can read the applied configuration.  In this case, it is a mirror of the /config container.

release/models/system/openconfig-hashing.yang
dplore: This is possible in the current model, one can configure the ipv4 and the ipv6 fields to different values.

```
           +--rw oc-hashing:hash-field-modes
              +--rw oc-hashing:config
              |  +--rw oc-hashing:ipv4?                hash-field-type-ipv4
              |  +--rw oc-hashing:ipv6?                hash-field-type-ipv6
              |  +--rw oc-hashing:transport-ports?     hash-field-type-transport
              |  +--rw oc-hashing:ingress-interface?   boolean
```

bgardiner: Thanks, understood. But can one make the `transport-ports` input conditional on whether it's ipv4 vs. ipv6?

dplore: Ah yes, @sallylsy this makes sense to me if it does for you too?

sallylsy: Make sense to me. I will make it an enum for ipv4 and ipv6

dplore:
Note: Added reference in the [PR description](https://github.com/openconfig/public/pull/959#issue-1899009062) to include [SONIC ecmp/lag hash](https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/doc/Configuration.md#hash).

sallylsy:
> > A potential alternative is a registered set of identities where common patterns are managed within the repository in a fashion similar to IETF's IANA function. Thus, when a particular behavior is registered in that module, all implementations with that behavior have a common configuration item. Since it would be an identity, vendor augmentation can happen when necessary.
> 
> We could specify a container which implementations are expected to augment containing a flat list of yang identities for "hash algorithms". This would be something like a "distributed registry" where the user must resolve which yang files to include in their configuration generation.
> 
> OC first introduced this pattern in the pipeline counters with the container at [/components/component/integrated-circuit/pipeline-counters/drop/vendor](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-integrated-circuit-pipeline-counters-drop-vendor).
> 
> Ref: [Vendor Counter Guide](https://github.com/openconfig/public/blob/master/doc/vendor_counter_guide.md) . (Note there is also a [PR in progress with more verbose description of the intent for these pipeline counters](https://github.com/openconfig/public/pull/961))

Hashing algorithms subtree as been updated as per the suggestion

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,355 @@
...
+                        level.";
+
+                    uses hashing-policy-config;
+                }
+
+                container hash-field-modes {
+                    description
+                        "Container for specifying inputs to be used when
+                        calculating the hash.";
+                    container ingress-interface {
```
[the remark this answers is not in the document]
sallylsy: Done

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,355 @@
...
+
+    grouping hashing-inputs {
+        description
+            "Top level container for inputs to be used for hashing policies.";
+        
+        leaf protocol-type {
+            type boolean;
+            description
+            "Specification of ipv4 or ipv6.";
+        }
```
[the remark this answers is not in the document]
sallylsy: Done

release/models/system/openconfig-hashing.yang
```diff
@@ -36,11 +36,17 @@ module openconfig-hashing {
     grouping hashing-inputs {
         description
             "Top level container for inputs to be used for hashing policies.";
-        
-        leaf protocol-type {
+
+        leaf ingress-interface {
+            type boolean;
+            description 
+                "Use the ingress subinterface identifired in the
```
dplore, line 43:
```suggestion
                "Include the ingress subinterface identified in the
```

release/models/system/openconfig-hashing.yang
```diff
@@ -36,11 +36,17 @@ module openconfig-hashing {
...
+                calculation of the hash.";
+        }
+        leaf ip-protocol-type {
             type boolean;
             description
-            "Specification of ipv4 or ipv6.";
+                "Specification of ipv4 or ipv6.";
```
dplore, line 49:
```suggestion
                "Include the IP protocol type in the calculation of the hash.";
```

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,359 @@
...
+        container vendor {
+            description
+                "Specify the vendor. Each vendor should have its own set of
+                supported algorithms. For each supported algorithm, a name
+                and a description should be defined. An implementation must
+                augment this model using the schema described in the
+                vendor_counter_guide reference.
+
+                e.g.
+                augment /system/vendor {
```
dplore, line 335:
```suggestion
                augment /system/hashing/vendor {
```

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,359 @@
...
+                "https://github.com/openconfig/public/tree/master/doc/vendor_counter_guide.md";
+        }
+    }
+
+    grouping hashing-top {
+        description
+            "Top-level container for Hashing algorithms and hashing policies";
+        uses supported-algorithms-top;
+        uses hashing-policy-top;
+    }
```
dplore, line 352:
```suggestion
container hashing {
        uses supported-algorithms-top;
        uses hashing-policy-top;
        }
    }
```

release/models/system/openconfig-hashing.yang
```diff
@@ -0,0 +1,359 @@
...
+            description
+                "Description of the supported algorithm.";
+        }
+    }
+    grouping supported-algorithms-top {
+        description
+            "Top-level container of the supported algorithms for supported
+            algorithms.  The list of algorithms are expected to be defined by
+            the target device.";
+        uses vendor-hashing-algorithms;
```
dplore, line 275:
```suggestion
container algorithms {
        uses vendor-hashing-algorithms;
        }
```

dplore (commented):
Couple of extra containers recommended to encapsulate all the hashing configurating into '/system/hashing' and to limit the scope of vendor augments to '/system/hashing/algorithms'

dplore (approved):
LGTM.  This was reviewed again in Nov 21, 2023 OC operators meeting without objection.  Thank you for the model and all the changes.

PR #968, opened 2023-09-28
Candidate Model for Repeated BGP set-communities/set-ext-communities

wenovus:
### Change Scope

* Deprecating the old leaves in favour of leaf-lists with a plural name:
      * community-set-ref (deprecated) -> community-set-refs
      * ext-community-set-ref (deprecated) -> ext-community-set-refs
* This change is backwards-compatible.

Note: [`set-as-path-prepend`](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-actions-bgp-actions-set-as-path-prepend) may also be a candidate for being repeatable if it is supported by multiple platforms.

### Platform Implementations

 * Cisco: [link to documentation](https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-0/routing/command/reference/rr40crs1book_chapter8.html#wp446732862)
 * Juniper: [link to documentation](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-defining-bgp-communities-and-extended-communities-for-use-in-routing-policy-match-conditions.html)

```
leaf added: /openconfig-routing-policy/routing-policy/policy-definitions/policy-definition/statements/statement/actions/bgp-actions/set-community/reference/config/community-set-refs ("openconfig-bgp-policy": openconfig-version 6.1.1 -> 6.2.0)
leaf added: /openconfig-routing-policy/routing-policy/policy-definitions/policy-definition/statements/statement/actions/bgp-actions/set-community/reference/state/community-set-refs ("openconfig-bgp-policy": openconfig-version 6.1.1 -> 6.2.0)
leaf added: /openconfig-routing-policy/routing-policy/policy-definitions/policy-definition/statements/statement/actions/bgp-actions/set-ext-community/reference/config/ext-community-set-refs ("openconfig-bgp-policy": openconfig-version 6.1.1 -> 6.2.0)
leaf added: /openconfig-routing-policy/routing-policy/policy-definitions/policy-definition/statements/statement/actions/bgp-actions/set-ext-community/reference/state/ext-community-set-refs ("openconfig-bgp-policy": openconfig-version 6.1.1 -> 6.2.0)
```

### DIFF
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -1015,7 +1015,7 @@ module openconfig-bgp-policy {
       "Configuration data for referening a extended community-set
       in the set-ext-community action";
 
-    leaf ext-community-set-refs {
+    leaf-list ext-community-set-refs {
       type leafref {
         path "/oc-rpol:routing-policy/oc-rpol:defined-sets/" +
           "oc-bgp-pol:bgp-defined-sets/" +
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -28,7 +28,16 @@ module openconfig-bgp-policy {
     It augments the base routing-policy module with BGP-specific
     options for conditions and actions.";
 
-  oc-ext:openconfig-version "6.1.1";
+  oc-ext:openconfig-version "6.2.0";
+
+  revision "2023-10-03" {
+    description
+      "Deprecate community-set-ref and ext-community-set-ref,
+      add the following leaf-list nodes:
+      * community-set-refs
+      * ext-community-set-refs";
+    reference "6.2.0";
+  }
 
   revision "2023-03-27" {
     description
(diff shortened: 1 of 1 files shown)


oscargdd:
Which is the use case behind this change? With current model you can already set multiple communities. The only thing you cannot do is to set multiple community sets. I think this extra benefit is not worth a breaking change...

wenovus:
> Which is the use case behind this change? With current model you can already set multiple communities. The only thing you cannot do is to set multiple community sets. I think this extra benefit is not worth a breaking change...

That's right, the use case is to be able to set multiple REFERENCE community sets in a single statement.

As you pointed out this does not increase the expressiveness of the current model: it simply allows some conciseness that vendor CLIs currently offer. If multiple REFERENCE community sets are required, then the current model requires multiple statements with the same set of conditions created, which makes the config more verbose compared to a single statement.

One alternative to this PR is to simply change `community-set-ref` to a leaf-list: `/openconfig-routing-policy/routing-policy/policy-definitions/policy-definition/statements/statement/actions/bgp-actions/set-community/reference/config/community-set-ref`. It would only allow a single ADD/REPLACE/REMOVE method, but would allow multiple community sets to be set. However, this is still a breaking change.

wenovus:
Changed PR to leaf-list, deprecating the old leaf in favour of a plural name.

dplore (approved):
Discussed on Oct 3, 2023 OC operator meeting.  Thanks for the change to leaf-list.  This looks good to me and makes policy more efficent to write and easier to read as it is a common operational use case to apply multiple, named community sets (lists).   This change also brings the model closer to the semantics of the two referenced bgp policy languages.

Setting to last-call in 1 week, Oct 10,2023

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -994,7 +1015,21 @@ module openconfig-bgp-policy {
       "Configuration data for referening a extended community-set
       in the set-ext-community action";
 
+    leaf ext-community-set-refs {
```
rgwilton, line 1018: Should this be a leaf-list?

wenovus: Apologies for not double-checking, fixed.

Any YANG tree diff tools should show the type of the node in addition to the path.

dplore:
> @dplore Will this be merged soon?

Thanks for the reminder, done.

PR #971, opened 2023-10-06
add peer_address to bgp-next-hop-type

dplore:
### Change Scope

Add peer-address to `bgp-nexthop-type`.  This enum is efficient to use when setting 
the next hop  in a policy applied to a peer group.  Without it one cannot set the next-hop to the peer address  in a peer group and must instead resort to configuring the next-hop at a per neighbor level which can create a large amount of additional configuration if there are many peers in a group.

### Platform Implementations

 * [JunOS next-hop peer-address](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-configuring-actions-in-routing-policy-terms.html).  Note Juniper only supports next-hop peer-address for import policies.
 * [Cisco IOS XR set next-hop](https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-0/routing/command/reference/rr40crs1book_chapter8.html#wp710515721)
 * [Arista EOS - set route-map next-hop](https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1151397)


rolandphung (approved):
This LGTM on my end.

dplore:
Reviewed in Dec 19, 2023 oc operators meeting without objection.  Last call for this, it will merge on Dec 26, 2023.

Issue #129, opened 2018-02-16
/interfaces/interface/config/mtu type uint16 not compatible with default Linux configuration


gmathon-ovh:
Hi,

Default loopback interface MTU is set to 64*1024 (65536) in [linux kernel since 2012](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/commit/drivers/net/loopback.c?h=v4.15.3&id=0cf833aefaa85bbfce3ff70485e5534e09254773). 
Openconfig defines /interfaces/interface/config/mtu as a uint16, so max value is 65535 (matching maximum IPvX datagram total length?):
https://github.com/openconfig/public/blob/0c955cd2a4d0e0e78e20fe8146fe3e6d4cfda9bc/release/models/interfaces/openconfig-interfaces.yang#L335

When you try to store default linux loopback config following this model, it will fail because the value don't fit uint16 type.

Should the type be changed to allow higher values, or the default MTU default configuration decreased?

robshakir:
Hi,

Apologies, we'd missed this issue. I've opened an internal issue tracking this fix. I don't see a problem with it given that Linux has this implementation, but we'll discuss and push an update.

Thanks,
r.

precla:
Any updates on this?

It is still in the latest [openconfig-if-ip.yang](https://github.com/openconfig/public/blob/323a3e2e43c60aafa2bf78ec7c955c00f6cfac4d/release/models/interfaces/openconfig-if-ip.yang#L457) model.

There is no limitation in the MTU size in "RFC 791: Internet Protocol"

dplore:
Hi @precla,  thanks for the reference and ping on this.

Rob raised a proposal to the internal OpenConfig operators group back in 2018, but it quickly became stale.

What do you think about this update? Look good?

        type uint32 {
          range 1..65536;
        }

precla:
Hi @dplore and thanks for the reply!

I'd say that from a Linux point of view it looks good. Considering that only the [loopback can have 65536](https://elixir.bootlin.com/linux/latest/source/drivers/net/loopback.c#L203).
Whereas all other interfaces can have [max 65535](https://elixir.bootlin.com/linux/v6.5.7/source/include/linux/ip.h#L53).

[...]

PR #972, opened 2023-10-11
Adapt mtu range to support Linux loopback max MTU of 65536

precla:
### Change Scope

* Extend the range of the allowed MTU to 65536 instead of max 65535
* Changes the type from uint16 to uint32 and adds range restriction
### Platform Implementations

 * Linux loopback MTU value: https://elixir.bootlin.com/linux/latest/source/drivers/net/loopback.c#L203

### Other

See discussion in https://github.com/openconfig/public/issues/129


dplore:
Will review on Oct 17 OC Operators call

PR #974, opened 2023-10-12
Default final disposition action of policy chain

rszarecki:
### Change Scope

* Clarification of semantics of default-import-policy,
      default-export-policy leaves.


dplore:
Discussed in Feb 6, 2024 OC Operators meeting.  BGP inheritance is defined here: https://github.com/openconfig/public/blob/35614de36d435c3d5efa5a119bd9aba367c9762f/release/models/bgp/openconfig-bgp.yang#L39-L69

I think your explanations are in line with this, but I think we should not redefine / re-explain it here.

The team suggested if you think it helps, you could add a reference in the routing-policy description to the bgp model description.  

I will close this for now, since it's been open awhile.  If you like to reopen this and change it to a short reference to the BGP model, we can revisit.

PR #982, opened 2023-10-24
Move BGP community and BGP extended community match-set-options back to bgp-conditions

### LINKED ISSUE #518: Inconsistencies in sets - openconfig-bgp-policy.yang
This is regarding an issue we observed in the latest openconfig revision for openconfig-bgp-policy.yang
https://github.com/openconfig/public/blob/master/release/models/bgp/openconfig-bgp-policy.yang

To give a brief overview, sets are treated as independent entities which do not define the execution of the policy. 
Sets were used in the conditional statements which define the policy execution.
But in the latest OC model, a part of the execution logic is moved into the sets(openconfig-bgp-policy.yang) - community-set and ext-community-set are such that they carry match options like ANY, ALL, INVERT along with the set rather than the conditional statement in the policy. (See the tree diff attached between the old OC model and the latest on the right)

This change is not consistent with all the sets in the same file.
community-set and ext-community-set follow the new convention(highlighted in green) whereas as-path-set follows older convention(in the red box) which does not included match-set-options.
Few sets are also part of openconfig-routing-policy.yang which still follows the older convention like as-path-sets.

**Request to reconsider/review this change in openconfig-bgp-policy.yang sets and revert back to the older convention.**

image001

Also, in the newer revision policy execution data is coupled with sets. 
Earlier a set could be used in different policies with different match-set-options, but because of this new change, a set is bound to be used with a single match-set-option only.
This restricts the way a set could be used with different match-set-options.
For example, consider a community-set XYZ, in the older convention, XYZ can be used in various statements with any of the match-set-option.
But in the new convention, XYZ will store the match-set-option within, so the user must always limit the set to be used with that match-set-option.
If we must use XYZ with a different match-set-option, we will have to clone a new set from XYZ and change the match-set-option(losing out on reusability).
This will lead to duplicate sets with different match-set-option.

xuqma:
### Change Scope

* Move `match-set-options` of BGP community and BGP extended community back to `policy-definitions/statements/.../bgp-conditions` for consistency across sets. This fixes #518. 
  
  The original paths will be deprecated, including:
  * `/routing-policy/defined-sets/bgp-defined-sets/community-sets/community-set/config/match-set-options`
  * `/routing-policy/defined-sets/bgp-defined-sets/community-sets/ext-community-set/config/match-set-options`
  * `/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/bgp-conditions/config/community-set`
  * `/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/bgp-conditions/config/ext-community-set`
  
  The new paths are:
  * `/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/bgp-conditions/match-community-set/config/match-set-options`
  * `/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/bgp-conditions/match-ext-community-set/config/match-set-options`
  * `/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/bgp-conditions/match-community-set/config/community-set`
  * `/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/bgp-conditions/match-ext-community-set/config/ext-community-set`

* This change is backwards compatible, as we mark the old paths as deprecated.

### Platform Implementations
* [Arista documentation](https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1150922)
* [Cisco documentation](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/routing/configuration/guide/rc40xr12k_chapter7.html#con_1240966)

### DIFF
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -28,7 +28,18 @@ module openconfig-bgp-policy {
     It augments the base routing-policy module with BGP-specific
     options for conditions and actions.";
 
-  oc-ext:openconfig-version "6.2.0";
+  oc-ext:openconfig-version "6.3.0";
+
+  revision "2023-10-23" {
+    description
+      "Revert revision 2019-02-01.
+      Move match-set-options of BGP community and BGP extended community from
+      defined-sets/bgp-defined-sets/community-set and
+      defined-sets/bgp-defined-sets/ext-community-set back to
+      policy-definitions/statements/.../bgp-conditions
+      for consistency across sets.";
+    reference "6.3.0";
+  }
 
   revision "2023-10-03" {
     description
@@ -244,12 +255,113 @@ module openconfig-bgp-policy {
     }
   }
 
+  grouping match-community-config {
+    description
+      "Configuration data for match conditions on community set";
+
+    leaf community-set {
+      type leafref {
+        path "/oc-rpol:routing-policy/oc-rpol:defined-sets/" +
+          "oc-bgp-pol:bgp-defined-sets/oc-bgp-pol:community-sets/" +
+          "oc-bgp-pol:community-set/oc-bgp-pol:community-set-name";
+      }
+      description "References a defined community set";
+    }
+    uses oc-rpol:match-set-options-group;
+  }
+
+  grouping match-community-state {
+    description
+      "Operational state data for match conditions on community set";
+  }
+
+  grouping match-community-top {
+    description
+      "Top-level grouping for match conditions on community set";
+
+    container match-community-set {
+      description
+        "Match a referenced community set according to the logic
+        defined in the match-set-options leaf";
+
+      container config {
+        description
+          "Configuration data for match conditions on community set";
+
+        uses match-community-config;
+      }
+
+      container state {
+
+        config false;
+
+        description
+          "Operational state data for match conditions on community set";
+
+        uses match-community-config;
+        uses match-community-state;
+      }
+    }
+  }
+
+  grouping match-ext-community-config {
+    description
+      "Configuration data for match conditions on extended community set";
+
+    leaf ext-community-set {
+      type leafref {
+        path "/oc-rpol:routing-policy/oc-rpol:defined-sets/" +
+          "oc-bgp-pol:bgp-defined-sets/oc-bgp-pol:ext-community-sets/" +
+          "oc-bgp-pol:ext-community-set/oc-bgp-pol:ext-community-set-name";
+      }
+      description "References a defined extended community set";
+    }
+    uses oc-rpol:match-set-options-group;
+  }
+
+  grouping match-ext-community-state {
+    description
+      "Operational state data for match conditions on extended community set";
+  }
+
+  grouping match-ext-community-top {
+    description
+      "Top-level grouping for match conditions on extended community set";
+
+    container match-ext-community-set {
+      description
+        "Match a referenced extended community set according to the logic
+        defined in the match-set-options leaf";
+
+      container config {
+        description
+          "Configuration data for match conditions on extended community set";
+
+        uses match-ext-community-config;
+      }
+
+      container state {
+
+        config false;
+
+        description
+          "Operational state data for match conditions on extended
+          community set";
+
+        uses match-ext-community-config;
+        uses match-ext-community-state;
+      }
+    }
+  }
+
   grouping bgp-match-set-conditions {
     description
       "Condition statement definitions for checking membership in a
       defined set";
 
     uses match-as-path-top;
+    uses match-community-top;
+    uses match-ext-community-top;
   }
 
   grouping community-count-config {
@@ -391,6 +503,7 @@ module openconfig-bgp-policy {
     }
 
     leaf community-set {
+      status deprecated;
       type leafref {
         path "/oc-rpol:routing-policy/oc-rpol:defined-sets/" +
           "oc-bgp-pol:bgp-defined-sets/oc-bgp-pol:community-sets/" +
@@ -401,6 +514,7 @@ module openconfig-bgp-policy {
     }
 
     leaf ext-community-set {
+      status deprecated;
       type leafref {
         path "/oc-rpol:routing-policy/oc-rpol:defined-sets/" +
           "oc-bgp-pol:bgp-defined-sets/oc-bgp-pol:ext-community-sets/" +
@@ -477,7 +591,13 @@ module openconfig-bgp-policy {
         removed.";
     }
 
-    uses oc-rpol:match-set-options-group;
+    leaf match-set-options {
+      status deprecated;
+      type oc-pol-types:match-set-options-type;
+      description
+        "Optional parameter that governs the behaviour of the
+        match operation";
+    }
   }
 
   grouping community-set-state {
@@ -553,7 +673,13 @@ module openconfig-bgp-policy {
         removed.";
     }
 
-    uses oc-rpol:match-set-options-group;
+    leaf match-set-options {
+      status deprecated;
+      type oc-pol-types:match-set-options-type;
+      description
+        "Optional parameter that governs the behaviour of the
+        match operation";
+    }
   }
 
   grouping ext-community-set-state {


dplore:
Added to Oct 31 OC Operators review 

Added to Nov 2, 2023 OC community meeting agenda for feedback:
https://docs.google.com/document/d/11sXUF2zujhbPs_bt-RFJhh2DcIbEvnOpd9GypE-u3cc/edit

rszarecki (changes requested):
1) I think we shall consider this as BRAKING change. Yes, old path are not removed (yet) but deprecated. Hence not required to be supported. I can imagine compliant new implementation that supports only new style, hence fail to accept old-style policy config.
2) Under this proposal it is possible (using deprecated) leaf write policy with dual-inversion, or even ALL (match-community-set/config/match-set-options) && INVERT (in deprecated but still legal community-set/config/match-set-options). Which creates quite fragile situation. Perhaps smart server can detect it and return Error. But still not nice possibility.
The concern 2) would be solved if this change would be declared as breaking.
3) This change introduce ambiguity how community+policy could be configured. It is cleare that old is deprecated, but old style is still there and would be accepted.

xuqma:
Thanks for the suggestion! I have bumped the version to `7.0.0` to mark the change as breaking. Please let me know if there are other changes we can make to better reconcile/validate the values at old/new paths.

dplore:
Deprecating the old paths and adding new is part of the definition of non-breaking changes.  The major version increment is not required.   The only function of a 'deprecating' a leaf is notification that it will be deleted in some future, major (breaking change) release.  The leaf is still required to be supported by a  network operating system (NOS).

To facilitate versioning in OC/Yang, a NOS must:
- Assert which [release of OpenConfig](https://github.com/openconfig/public/tags) is supported 
- Assert differences in their model by  publishing additional yang files that contain Yang defined deviations and augments.

1. Say we start with a NOS release 1.0  that supports v2.0.0 of OC without yang deviations/augments.
1. OC release is updated to v2.1.0 with adding leaves and deprecating others.
1. In release 1.1 of the NOS, the OC release is v2.1.0.   This NOS must still support the deprecated leafs.  (They are present in OC v2.1.0).  
      2. If the deprecated leafs are dropped, then that must be done with a NOS published yang deviation.  This would be a breaking change to OC.  


1. In release 2.0 of the NOS, the OC release supported is v3.0.0.   

v3.0.0 of OC is a new major release of OC.  In this scenario some deprecated leaves were deleted, creating a breaking change and need for a major version increment.  The major version increment notifies consumers that it contains  breaking changes in OC and code changes are to be expected to use this OC release.  

In an operational environment, a configuration generation system will need to support v2.2.0 of OC and v3.0.0 of OC.  It will also need to know which NOS support which version in order to generate compatible OC.  In OC we are commiting to making major releases infrequently, such as once per year.   See https://www.openconfig.net/docs/guides/releases/

rszarecki (commented):
Xuquma,
I do have one more concern.

Why this match os not under /conditions/bg-conditions/ as all other that are based on BGP only attributes? You introduced inconsistency this way (while fixing the other inconsistency)

I belive grouping match conditions per-protocol would greatly improve/simplife validation or policy attachment. E.g. if policy w/ bgp-condition are attached to say ISIS, server can easly spot it and generate error.

dplore:
> Thanks Darren for explanation. Hmm. Do I get it right:
> 
> * any new (greenfield) client  MUST support all deprecated path in addition to all other path.
>   AND
> * any new (greenfield) server  MUST support all deprecated path in addition to all other path.
> 
> Or only one of above? Which one? Then indeed this is not breaking change. But I'm not sure it is practical for implementers perspective.

Yes, a server/network operating system release (brand new implementation or pre-existing with upgrade) is expected to support a given release of OC, including deprecated leafs.  In practice, all NOS  have Yang deviations/augments as well as deviations in behavior, some of which are captured in [featureprofiles](https://github.com/openconfig/featureprofiles/blob/main/internal/deviations/deviations.go).  

In practice, a client / network management system must accommodate the deviations for each NOS in a given OC release, plus deal with multiple OC releases, as any operational network will always have some variation of deployed NOS releases and therefore OC releases.

dplore:
> Xuquma, I do have one more concern.
> 
> Why this match os not under /conditions/bg-conditions/ as all other that are based on BGP only attributes? You introduced inconsistency this way (while fixing the other inconsistency)
> 
> I belive grouping match conditions per-protocol would greatly improve/simplife validation or policy attachment. E.g. if policy w/ bgp-condition are attached to say ISIS, server can easly spot it and generate error.

@xuqma I think I agree with @rszarecki here.   The refactored `match-community-set` container should follow the same pattern as `match-as-path-set`.  ie:

[/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/bgp-conditions/match-as-path-set](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-conditions-bgp-conditions-match-as-path-set)

xuqma:
> > Xuquma, I do have one more concern.
> > Why this match os not under /conditions/bg-conditions/ as all other that are based on BGP only attributes? You introduced inconsistency this way (while fixing the other inconsistency)
> > I belive grouping match conditions per-protocol would greatly improve/simplife validation or policy attachment. E.g. if policy w/ bgp-condition are attached to say ISIS, server can easly spot it and generate error.
> 
> @xuqma I think I agree with @rszarecki here. The refactored `match-community-set` container should follow the same pattern as `match-as-path-set`. ie:
> 
> [/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/bgp-conditions/match-as-path-set](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-conditions-bgp-conditions-match-as-path-set)

Good catch! I put the wrong paths in the PR description though the actual model change moved `match-community-set` and `match-ext-community-set` under `bgp-conditions`. I have updated the description. Thanks for the review and apologies for the confusion.

With Darren's help, I was able to generate the model diff using pyang from this PR and confirmed the new paths are part of `bgp-conditions`:

```
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-community-set
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-community-set/config
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-community-set/config/community-set
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-community-set/config/match-set-options
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-community-set/state
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-community-set/state/community-set
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-community-set/state/match-set-options
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-ext-community-set
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-ext-community-set/config
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-ext-community-set/config/ext-community-set
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-ext-community-set/config/match-set-options
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-ext-community-set/state
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-ext-community-set/state/ext-community-set
> /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/conditions/openconfig-bgp-policy:bgp-conditions/match-ext-community-set/state/match-set-options
```

Let me know if you need further info. Thanks!

rszarecki (approved):

rszarecki:
I'm good.

dplore (approved):

PR #983, opened 2023-10-25
Clarify default policy

### LINKED ISSUE #981: What should happen if there is no routing-policy configured?
Routing-policy describes what should happen when a policy is defined and applied. 
 https://github.com/openconfig/public/blob/0bd68159424078c8fbc3724a4bc1ef4e947eb3cf/release/models/policy/openconfig-routing-policy.yang#L45-L49

But there is not an explicit definition of what should happen if no policy is applied.  For example, if a BGP session is configured, but   no policy is configured for that session, I interpret that there is no policy and therefore all routes are accepted.  If a policy is applied to a session, then the rules for that policy are enforced.  (including for example, the [default-import-policy](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-bgp-peer-groups-peer-group-apply-policy-state-default-import-policy) which is `REJECT`)

Do we need a clarification to the description of the routing-policy module?

dplore:
Fixes #981

### DIFF
--- a/release/models/policy/openconfig-routing-policy.yang
+++ b/release/models/policy/openconfig-routing-policy.yang
@@ -80,9 +80,21 @@ module openconfig-routing-policy {
     policy.  Similarly, a reject-route action returns false.  If the
     subroutine returns true, the calling policy continues to evaluate
     the remaining conditions (using a modified route if the
-    subroutine performed any changes to the route).";
+    subroutine performed any changes to the route).
 
-  oc-ext:openconfig-version "3.4.0";
+    If a routing protocol is defined and the related
+    apply-policy/default-(import|export)-policy leaf is not set, the
+    default value for the default-(import|export)-policy leaf must be
+    applied.  See RFC6020 7.6.1 which applies to this model.";
+
+  oc-ext:openconfig-version "3.4.1";
+
+  revision "2023-10-24" {
+    description
+      "Clarify the use of default-(import|export)-policy when no policy
+      is applied to a protocol.";
+    reference "3.4.1";
+  }
 
   revision "2023-10-11" {
     description


release/models/policy/openconfig-routing-policy.yang
```diff
@@ -80,9 +80,21 @@ module openconfig-routing-policy {
     policy.  Similarly, a reject-route action returns false.  If the
     subroutine returns true, the calling policy continues to evaluate
     the remaining conditions (using a modified route if the
-    subroutine performed any changes to the route).";
+    subroutine performed any changes to the route).
 
-  oc-ext:openconfig-version "3.4.0";
+    If a routing protocol is defined and the related
```
sachendras, line 85: Does this also cover for Route redistribution polices using table-connections? https://github.com/openconfig/public/blob/master/doc/network_instance_redistribution.md. OC path for default-policy attachment on table-connections is /network-instances/network-instance/table-connections/table-connection/config/default-import-policy
I think we should clarify here that the default of reject_all is true for table-connections as well if no default-import-policy is defined.

robshakir (approved):

release/models/policy/openconfig-routing-policy.yang
dplore: Yes, it's reject or all contexts where route policy is used.   The route policy model is used in many portions of openconfig and will get updated if/when new protocols are added.  We probably shouldn't try to list them all in the description here.

sachendras (approved):

release/models/policy/openconfig-routing-policy.yang
nraghavendran: does this apply to routing protocols other than BGP? the wording in 85 seems to be generic?

dplore: Good question!  The defaults takes effect everywhere  routing policy is applied.  Today in OC, the following paths reference routing-policy:

/network-instances/network-instance/inter-instance-policies/apply-policy (between vrf's)
/network-instances/network-instance/protocols/protocol/bgp/global/apply-policy.  (bgp)
/network-instances/network-instance/table-connections/table-connection/config/import-policy (between protocols in a vrf)

PR #985, opened 2023-10-26
add uptime and promote system to v1

dplore:
### Change Scope

* Add /system/up-time
* Bump the openconfig-system module to version 1.0
 
### Platform Implementations

 * [Cisco XROS show version](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/system-management/b-system-management/m-hardware-redundancy-commands.html) 
 * [JunOS show system uptime](https://www.juniper.net/documentation/us/en/software/junos/junos-overview/topics/ref/command/show-system-uptime.html)

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -47,7 +47,13 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.17.1";
+  oc-ext:openconfig-version "1.0.0";
+
+revision "2023-10-26" {
+    description
+      "Add up-time leaf and promote module to version 1.0.";
+    reference "1.0.0";
+  }
 
   revision "2023-06-16" {
     description
@@ -369,6 +375,14 @@ module openconfig-system {
         type oc-yang:date-and-time;
         description
           "The current system date and time.";
+    }
+
+    leaf up-time {
+        type oc-types:timeticks64;
+        units "nanoseconds";
+        description
+          "The amount of time since the network operating system was 
+          initialized.";
     }
 
     leaf boot-time {


rolandphung:
Does this leaf make sense to be a monotonicity increasing value? Could this be a timestamp instead? Subscribing to that would be a bit noisy for an end value that doesn't necessarily change.

How often should this leaf be updated?

robshakir:
We've litigated on this leaf for a while, and discussed a few times. I believe we actually originally had `uptime`, and changed it to `boot-time`.

Pros for `boot-time` being the only leaf:
 * `uptime` is a monotonically increasing value, and hence needs to be subscribed to using a `SAMPLE` otherwise it will change every second. (note that of course we have other data that changes every second, or more often, like counters)
 * we likely want to understand what the boot time was even if we have an `uptime` counter.
 * a `number-of-boots` value could be used to tell you when a system rebooted, and would also avoid a monotonically increasing value.

Problems with `boot-time` being the only leaf:
 * requires NTP synchronisation at power-on time to be set accurately, if we do not have this then `boot-time` might be set to an invalid value, or the clock might change in the future to mean that "the time now - `boot-time` != uptime".
 * avoids the need for some new persistent `number-of-boots` counter, that must last across reboots, and might reset if there are filesystem or hardware changes in the same system.

robshakir (approved):

dplore:
Reviewed at Dec 5,2023 OC operators meeting without objection.   Moving to last call, this will merge on Dec 19th

ayelrod:
Hello, can this PR be merged now @dplore? Thanks!

release/models/system/openconfig-system.yang
```diff
@@ -371,6 +377,14 @@ module openconfig-system {
           "The current system date and time.";
     }
 
+    leaf up-time {
+        type oc-types:timeticks64;
+        units "nanoseconds";
```
earies, line 382: This comment can apply to other leafs that are tagged with `units nanoseconds` but would like to clarify the expectations around time precision

1. To accurately represent precision as nanoseconds
2. To represent solely according to the datatype regardless of actual precision

I think what we'd probably find if we audited all implementations across all such leafs, the internal representation may likely be less precise and there will just be a conversion to fulfill the datatype in various cases.  It may give the perception of nanoseconds but may infact just be seconds/centiseconds, etc..

And if we look across the vendor references listed here, it is likely we'd actually find seconds (or less) precision today (across common show commands)

For Linux systems, if `/proc/uptime` is leveraged, kernel sources currently convert this to centiseconds

Are there any vendor or common OSS src references to nanoseconds precision?  Otherwise it seems to fulfill the accurate precision requested here (which I'd be curious on how nanoseconds on uptime would actually be used), it would likely be all new work for each implementation

earies: Also note that the `timeticks64` type defined in `openconfig-types` already hard codes a unit of `nanoseconds` so it is implied that any leaf using that type inherits that unit (this might need revisiting outside of this PR unless we agree its about presentation and not precision)

Quick output from a shipping implementation will yield seconds precision packed into a timeticks64 type dropping the tail `648933887`

```
$ date -d @1704321883
Wed Jan  3 15:44:43 MST 2024
$ date -d @1704321916
Wed Jan  3 15:45:16 MST 2024
```

```
system/processes/process[pid=1736]/state/start-time: 1704321883648933887
system/processes/process[pid=2243]/state/start-time: 1704321916648933887
system/processes/process[pid=1947]/state/start-time: 1704321903648933887
system/processes/process[pid=40]/state/start-time: 1704321850648933887
system/processes/process[pid=1729]/state/start-time: 1704321883648933887
system/processes/process[pid=1945]/state/start-time: 1704321903648933887
system/processes/process[pid=1153]/state/start-time: 1704321855648933887
system/processes/process[pid=1391]/state/start-time: 1704321857648933887
system/processes/process[pid=42]/state/start-time: 1704321850648933887
system/processes/process[pid=1378]/state/start-time: 1704321857648933887
system/processes/process[pid=1085]/state/start-time: 1704321855648933887
system/processes/process[pid=1404]/state/start-time: 1704321857648933887
system/processes/process[pid=1755]/state/start-time: 1704321883648933887
system/processes/process[pid=1472]/state/start-time: 1704321858648933887
system/processes/process[pid=1086]/state/start-time: 1704321855648933887
system/processes/process[pid=1944]/state/start-time: 1704321903648933887
```

dplore: I think we can always say an implementation must send data according to the datatype, even if the implementation cannot support the precision implied.  

There is not a strong operational requirement to maintain uptime with nanosecond precision.  I used nanoseconds to be consistent with the other system time related leafs like boottime and last-configuration-timestamp

dplore: I'll also note that currently OC only has a timeticks64 type we've used throughout the models.  I think if we want to assert precision requirements we'll need a separate effort for this.

PR #987, opened 2023-10-28
Update openconfig-network-instance.yang

sachendras:
Add clarifications to the Table-connections implementation in OC to avoid misinterpretation and maintain consistency between different NOS.

https://github.com/openconfig/public/blob/master/doc/external-contributions-guide.md]

### Change Scope

### Platform Implementations

   implementation output.
   implementation output.


release/models/network-instance/openconfig-network-instance.yang
```diff
@@ -478,7 +478,10 @@ module openconfig-network-instance {
               A connection connecting a source table to a destination
               table implies that routes that match the policy specified
               for the connection are available for the destination
-              protocol to advertise, or match within its policies.";
+              protocol to advertise, or match within its policies.
+              Implemtations of this connection should ensure that the leaked 
+              routes mustn't participate in the route selection process in
+              the destination table";
```
dplore, line 484: Is this what is really intended?  IMO, isn't the purpose of 'redistributing' routes from one protocol to another is to use them?   (subject to policies defined)

sachendras: Well, if you meant, route-redistribution from one protocol to another is to actually make the redistributed route available for advertisements in the destination protocol to peers rather than influencing the forwarding decisions in the local router, then yes thats what I intend to explicitly state here.

We have 2 kinds of implementations that are impacted by Table connections. One kind where there is a separate table for each routing-protocol. Then there is another setup which has a single table for all routing-protocols and the redistributed routes also end up in the same table. For this latter implementation, I guess it is important to explicitly define the expectations of the feature. If the redistributed route ends up being active, the concerns are that we may end up in a race condition.

dplore: Oh, I think I see you're saying the implementation should not insert a RIB or FIB (AFT) entry for redistributed routes?   Is this a more clear way to state the concern?   [This official OC documentation](https://github.com/openconfig/public/blob/0426774f2d7af0dc0a2b913893c8d45644176d65/doc/network_instance_redistribution.md#rib-modeling-approach-used-in-openconfig) includes guidance on multi vs. single RIB.  We could add a reference to this documentation into the yang and elaborate there?

rszarecki: The intent of change is to clearly state that redistributed route MUST NOT take part is active/installed/forwarding route selection. Otherwise we will create cyclic dependency and potentially forwarding loops and/or blackholes.
The unspoken rule is that only active/installed/forwarding route could be source of redistrion. 
This has nothing to do with implementation differences.
Jeff Haas point that term "table-connection" could be understand as inserting redistributed toute to destination's protocol table on which best/active rote/path selection is running. What must NOT be a case. Here is my mental model:
```
protocol's                                                        protocol's
    peer                protocol's peer                               peer   
                                                                             
      |                        |                                        ^    
      |                        |                                        |    
+-----+------------------------+----------------------------------------+--+ 
|+----+----------------+       |      +---------------------------------+-+| 
||+---+    Src Protocol|       |      |           Dst Protocol          | || 
|||                    |       |      |                                 | || 
|||                    |       |      |                                 | || 
|||                    |       |      |  #===========================#  | || 
|||                    |       |      |  |                           |  | || 
|||                    |       +------+->|            RIB            |  | || 
|||  #===============# |              |  |                           |  | || 
|||  |               | |              |  #===========================#  | || 
||+->|      RIB      |-+-------+      |    ^       |   |                | || 
||   |               | |       |      |    |       |   |active          | || 
||   #===============# |       |      |            |   +routes +        | || 
||           |     ^   |              |   flag     |    only*  |        | || 
||           |     |   |    active    |  active    |           v        | || 
||        +--+     |   |    routes    |  routes    |    +-------------+ | || 
||        |        |   |     only     |            |    |    route    | | || 
||        v        |   |              |    |       |    |advertisement|-+ || 
||+---------------+|   |       |      |    |       v    +-------------+   || 
||| protol's best ||   |       |      |    | +----------+      ^          || 
|||route selection||   |       |      |    | | protol's |      |          || 
||+---------------+|   |       |      |    | |   best   |      |          || 
||        |        |   |       |      |    | |  route   |    +-+          || 
||        |        |   |       |      |    | |selection |    |            || 
||        |        |   |       |      |    | +----------+    |            || 
||        |        |   |       v      |    |       |         |            || 
||        |        |   |+------------+|    |       |  /=============\     || 
||        |            ||   table    ||    |       |  |redistributed|     || 
||        |      flag  ||connections |+----+-------+->|     RIB     |     || 
||    +---+     active ||incl. policy||    |       |  |             |     || 
||    |         routes |+------------+|    |       |  \=============/     || 
||    |                |              |    |       |                      || 
||    |            |   |              |    |       |                      || 
||    |            |   |              |    |       |                      || 
||    |            |   |              |    |       |                      || 
|+----+------------+---+              +----+-------+----------------------+| 
|     |            |                       |       |                       | 
|     v            |                       |       v                       | 
|+-------------------------------------------------------------------+     | 
||                      network-instance active                      |     | 
||                          route selection                          |     | 
|+-------------------------------------------------------------------+     | 
|                                  |                                       | 
|                                  v                                       | 
|#===================================================================#     | 
||                                                                   |     | 
||                                FIB                                |     | 
||                                                                   |     | 
|#===================================================================#     | 
|                             Network instance                             | 
+--------------------------------------------------------------------------+ 
                                                                             
* for BGP as destination protocol, the "best-external" and "best-inactive"   
could be included CONDITIONALLY.                                             
The condition being prefix do not exists in virtual "redistributed RIB"      
                                                                             
```
Note, "redistributed routes" are by definition allways baes and active.

doc/network_instance_redistribution.md
```diff
@@ -23,6 +23,7 @@ In numerous use cases an operator wishes to take routes that exist in one protoc
...
 
 In this latter case, the use of an import policy within a particular protocol instance is equivalent to the former - with the only difference being the context in which the policy is defined.   With a table per protocol, an explicit configuration of redistributing routes from protocol A to protocol B is specified.  In the latter case, the target protocol (protocol B) is specified, and a policy is used to select the source protocols from which routes may be imported for advertisement.
+In both cases, it is important that the destination protocol (protocol B) in this case only place the redistributed routes from the source protocol (protocol A in this case) in its RIB for advertisements, but do not use those routes in the route selection process and place them in the FIB (forwarding information base).
```
rszarecki, line 26:
```suggestion
In both cases, it is important that the destination protocol (protocol B) in this case only place the redistributed routes from the source protocol (protocol A in this case) in its "advertisement RIB view", but do not use those routes in the active/installed/forwarding route selection process and do not place them in the FIB (forwarding information base).
```

release/models/network-instance/openconfig-network-instance.yang
```diff
@@ -478,7 +478,11 @@ module openconfig-network-instance {
               A connection connecting a source table to a destination
               table implies that routes that match the policy specified
               for the connection are available for the destination
-              protocol to advertise, or match within its policies.";
+              protocol to advertise, or match within its policies.
+
+            reference
+              "Route Redistribution in OpenConfig Network Instance:
+              https://github.com/openconfig/public/blob/master/doc/network_instance_redistribution.md#interconnection-of-protocol-ribs";
```
dplore, line 485:
```suggestion
              https://github.com/openconfig/public/blob/master/doc/network_instance_redistribution.md#interconnection-of-protocol-ribs";
}
```

release/models/network-instance/openconfig-network-instance.yang
```diff
@@ -478,7 +478,11 @@ module openconfig-network-instance {
               A connection connecting a source table to a destination
               table implies that routes that match the policy specified
               for the connection are available for the destination
-              protocol to advertise, or match within its policies.";
+              protocol to advertise, or match within its policies.
```
dplore, line 481:
```suggestion
              protocol to advertise, or match within its policies.";
```

release/models/network-instance/openconfig-network-instance.yang
```diff
@@ -50,6 +50,12 @@ module openconfig-network-instance {
 
   oc-ext:openconfig-version "4.2.0";
```
dplore, line 51:
```suggestion
  oc-ext:openconfig-version "4.2.1";
```

doc/network_instance_redistribution.md
```diff
@@ -22,7 +22,8 @@ In numerous use cases an operator wishes to take routes that exist in one protoc
...
+In this latter case, the use of an import policy within a particular protocol instance is equivalent to the former - with the only difference being the context in which the policy is defined.   With a table per protocol, an explicit configuration of redistributing routes from protocol A to protocol B is specified.  In the latter case, the target protocol (protocol B) is specified, and a policy is used to select the source protocols from which routes may be imported for advertisement. 
+In both cases, it is important that the destination protocol (protocol B) in this case only place the redistributed routes from the source protocol (protocol A in this case) in its "advertisement RIB view", but do not use those routes in the active/installed/forwarding route selection process and,consequently, do not place them in the FIB (forwarding information base).
```
dplore, line 26: Suggested rewording to attempt to simplify the text but keep the same meaning.  
```suggestion
In both cases,  the destination protocol (protocol B) only installs the redistributed routes from the source protocol (protocol A) in its "advertisement RIB view".  The redistributed routes are not installed in the FIB (forwarding information base) and therefore are not used for forwarding.
```

dplore (approved):

dplore (approved):
Need to fix CI checks

Issue #991, opened 2023-11-03
aft route summary model refactor required


dplore:
@dplore Looks like this introduced an OpenConfig style error that unfortunately the script didn't catch. I have a rough draft modification that will catch this error, but I want to start the conversation on how to fix it:

```
public/release/models/network-instance/openconfig-network-instance.yang:1394 (at public/release/models/aft/openconfig-aft.yang:192): error: Parent node (ipv4-unicast) has multiple children where one is a list, but a list is not allowed to have siblings: list: ipv4-entry, container: summary
public/release/models/network-instance/openconfig-network-instance.yang:1394 (at public/release/models/aft/openconfig-aft.yang:206): error: Parent node (ipv6-unicast) has multiple children where one is a list, but a list is not allowed to have siblings: list: ipv6-entry, container: summary
```

_Originally posted by @wenovus in https://github.com/openconfig/public/issues/783#issuecomment-1792940248_

dplore:
The fix for this is probably to move the summary container to instantiated at  `/afts/ipv4-unicast/summary` and `/afts/ipv6-unicast/summary`

PR #783, opened 2023-01-09
Added new model for route summary

ruoyi-anet:
Initial proposal to model route summary in a network instance. There will be a route summary for IPv4 and IPv6 respectively. Each route summary will be a list of route counts keyed by the protocol type.

### Change Scope

* Adding a new module named `route-summary`.
```
$ pyang -p release/models -f tree release/models/network-instance/openconfig-route-summary.yang
module: openconfig-route-summary

  augment /oc-ni:network-instances/oc-ni:network-instance/oc-ni:afts/oc-ni:ipv6-unicast:
    +--ro summary
       +--ro route-count* [origin-protocol]
          +--ro origin-protocol    -> ../state/origin-protocol
          +--ro state
             +--ro origin-protocol?   identityref
             +--ro total-count
                +--ro fib-count?   oc-yang-types:counter64
  augment /oc-ni:network-instances/oc-ni:network-instance/oc-ni:afts/oc-ni:ipv4-unicast:
    +--ro summary
       +--ro route-count* [origin-protocol]
          +--ro origin-protocol    -> ../state/origin-protocol
          +--ro state
             +--ro origin-protocol?   identityref
             +--ro total-count
                +--ro fib-count?   oc-yang-types:counter64
 ```

The rendered network-instance tree looks like:
<pre>
pyang -f tree -p release/models/*/* --tree-path=/network-instances/network-instance

module: openconfig-network-instance
  +--rw network-instances
     +--rw network-instance* [name]
[... snip ...]
        +--ro afts
        |  +--ro ipv4-unicast
        |  |  +--ro ipv4-entry* [prefix]
        |  |  |  +--ro prefix    -> ../state/prefix
        |  |  |  +--ro state
        |  |  |     +--ro prefix?                                     oc-inet:ipv4-prefix
        |  |  |     +--ro counters
        |  |  |     |  +--ro packets-forwarded?          oc-yang:counter64
        |  |  |     |  +--ro octets-forwarded?           oc-yang:counter64
        |  |  |     |  +--ro packets-forwarded-backup?   oc-yang:counter64
        |  |  |     |  +--ro octets-forwarded-backup?    oc-yang:counter64
        |  |  |     +--ro entry-metadata?                             binary
        |  |  |     +--ro origin-protocol?                            identityref
        |  |  |     +--ro decapsulate-header?                         oc-aftt:encapsulation-header-type
[...]

PR #992, opened 2023-11-03
Revert "Added new model for route summary"

dplore:
Reverts openconfig/public#783 to fix issue identified at #991


wenovus (approved):

wenovus:
I think this should be merged relatively soon since all the other PRs are going to start to break. They will need to merge with this one in order to avoid the Linter failures.

PR #994, opened 2023-11-06
Clarify interface TESTING mode

dplore:
### Change Scope

* Clarify the description of TESTING mode for interfaces.

### Platform Implementations
 * This mode is in development by multiple vendors to support [gNOI link qualification](https://github.com/openconfig/gnoi/blob/main/packet_link_qualification/index.md).
 * An [OpenConfig feature profiles test](https://github.com/openconfig/featureprofiles/tree/e91835ce37a73474f9e7ac59c50c0b78b946ec9f/feature/gnoi/packet_link_qualification/tests/packet_link_qualification_test) exercises this functionality.

### DIFF
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -612,8 +612,9 @@ module openconfig-interfaces {
             "The interface is should be treated as if in admin-down state for
             control plane protocols.  In addition, while in TESTING state the
             device should remove the interface from aggregate interfaces.
-            This state may be set by the device as part of a transaction such
-            as gNOI Link Qualification Service.";
+            An interface transition to the TESTING state based on a qualification 
+            workflow, or internal device triggered action - such as the gNOI Link
+            Qualification service";
             reference
               "gNOI Link Qualification Service
                https://github.com/openconfig/gnoi/blob/main/packet_link_qualification/index.md";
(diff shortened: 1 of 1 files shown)


release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -600,11 +608,15 @@ module openconfig-interfaces {
             "Not ready to pass packets and not in some test mode.";
         }
         enum TESTING {
-          //TODO: This is generally not supported as a configured
-          //admin state, though it's in the standard interfaces MIB.
-          //Consider removing it.
           description
-            "In some test mode.";
+            "The interface is should be treated as if in admin-down state for
```
cfernanz, line 612: remove "is"

release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -600,11 +608,15 @@ module openconfig-interfaces {
...
+            control plane protocols.  In addition, while in TESTING state the
+            device should remove the interface from aggregate interfaces.
+            This state may be set by the device as part of a transaction such
+            as gNOI Link Qualification Service.";
```
robshakir, line 616: Maybe something like:

"An interface may be set into a `TESTING` state based on a qualification workflow, or internal device state - such as those that can be triggered by the gNOI Link Qualification service."

dplore: Modified similar to what you suggested.

PR #995, opened 2023-11-08
Changing type for NTP fields in openconfig-system.yang

kbchandradeep:
Changing type for root-delay, root-dispersion and offset
from uint32, uint64 and uint64 respectively to oc-types:ieeefloat32

* type has been changed for 3 NTP fields in openconfig-system.yang
https://github.com/openconfig/public/blob/master/release/models/system/openconfig-system.yang

 * Implementation A: In NTP RFC 5905, these fields are defined as floating point numbers and implemented as floating point numbers.
 https://datatracker.ietf.org/doc/html/rfc5905

This difference in type has resulted in implementations to deviate these fields as not-supported.

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -753,7 +753,7 @@ module openconfig-system {
     }
 
     leaf root-delay {
-      type uint32;
+      type oc-types:ieeefloat32;
       // TODO: reconsider units for these values -- the spec defines
       // rootdelay and rootdisperson as 2 16-bit integers for seconds
       // and fractional seconds, respectively.  This gives a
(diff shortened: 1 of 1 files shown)


jsterne:
One downside of ieeefloat32 type is that the underlying YANG type is binary. The native output of this in many clients will likely just be a raw byte dump (maybe in hex) and it will be pretty non-intuitive for humans looking at the values (unless the various clients or output functions add specific support for this ieeefloat32 type). Maybe that's not a top priority here but something to consider.

dplore:
Is there some operational use case that cannot be met using the int values in the model?  The model does not need to match the wire format.  The model should be easy to consume, and converting between int and float is trivial to do in code.

There needs to be a strong operational use case to justify making this breaking change.  I'll close this for now.  You may reopen if you have some strong operational use case.

PR #1007, opened 2023-11-28
add model-name

dplore:
### Change Scope

* Add `/components/component/state/model-name`.

### Platform Implementations

 * [Cisco IOS XR - show version](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/system-management/b-system-management/m-hardware-redundancy-commands.html#wp3568776008)
 * [JunOS - show system information](https://www.juniper.net/documentation/us/en/software/junos/junos-overview/topics/ref/command/show-system-information.html)

### DIFF
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -473,8 +473,8 @@ module openconfig-platform {
       type string;
       description
         "A human readable string describing the model of a component.
-        This string typically has a 1:1 relationship with the part-no
-        leaf.";
+        This string is optional and should only be populated if part-no 
+        is also populated.";
     }
 
     leaf clei-code {
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -473,7 +473,7 @@ module openconfig-platform {
       type string;
       description
         "A human readable string describing the model of a component.
-        This string is optional and should only be populated if part-no 
+        This string is optional and should only be populated if part-no
         is also populated.";
     }
 
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,7 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.23.0";
+  oc-ext:openconfig-version "0.24.0";
+
+  revision "2023-11-28" {
+    description
+      "Add model-name";
+    reference "0.24.0";
+  }
 
   revision "2023-02-13" {
     description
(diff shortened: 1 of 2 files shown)


release/models/platform/openconfig-platform.yang
```diff
@@ -463,6 +469,14 @@ module openconfig-platform {
         (field replaceable unit)";
     }
 
+    leaf model-name {
+      type string;
+      description
+        "A human readable string describing the model of a component.
+        This string typically has a 1:1 relationship with the part-no
```
robshakir, line 476: Is there any reason that it /isn't/ the `part-no` leaf? It would be useful to add some clarifying example as to why I would want to use this leaf vs. the `part-no` one, or an example of what it should be populated with.

dplore: Generically speaking part number might be "A123453-1234" and model might be "Acme Router 1000"
Or "psu234-123", model "Acme 1000 high power supply unit".  Some vendors may use the same string for part number the model name.  

In the optical side of the industry, we see:
Part-no --> maps cleanly to part-no
Serial-no --> maps cleanly to serial-no
Model-name --> We don't have a mapping for this and its really just a more human readable version of part-no. So vendors put something like this in `/components/component/state/description`.  In practice, the leaf  `/components/component/state/clei-code` is not implemented.

Example for packet side:

[Juniper mapping between part number and model number](https://supportportal.juniper.net/s/article/Mapping-between-FRU-part-numbers-and-model-numbers?language=en_US) clearly has two different values for model and part number.  

[Cisco 8000 series](https://www.cisco.com/c/en/us/products/collateral/routers/8000-series-routers/datasheet-c78-742571.html) example shows several series names (8000 series, but also 8202 and 8808) which have a one to many relationship with part number. 

In two examples from Arista, they also have a series name which is one to many relationship with part number.  [Arista 7800 series](https://www.arista.com/en/qsg-7800-series/7800-series-overview), [Arista 7280R series](https://www.arista.com/assets/data/pdf/Datasheets/7280R-DataSheet.pdf)

[Nokia 7250 series](https://www.tempestns.com/wp-content/uploads/2020/05/Nokia_7250_IXR-10-IXR-6-IXR-s1-data.pdf) documents product identifiers but it is not clear if these are part numbers or model names or both.

dplore:
Reviewed in Dec 5,2023 OC operators meeting without objection.  Last call for this change, to be merged on Dec 19,2023

robshakir:
Please make sure to address my comment before merging.

release/models/platform/openconfig-platform.yang
dplore: Fixed in https://github.com/openconfig/public/pull/1160

PR #1008, opened 2023-11-30
Allow IP addresses in AFT nexthops to be zoned.

robshakir:
### Change Scope

* Relaxes the regexp on afts/nexthops/nexthop/state/ip-address to allow for zoned IPv4 and IPv6 values.

### Nokia SRL

```
--{ + running }--[  ]--
A:srl# info network-instance mgmt static-routes
    network-instance mgmt {
        static-routes {
            route 2001:db8::/32 {
                next-hop-group nhop
            }
        }
    }
--{ + running }--[  ]--
A:srl# info network-instance mgmt next-hop-groups
    network-instance mgmt {
        next-hop-groups {
            group nhop {
                nexthop 0 {
                    ip-address fe80::1%mgmt0.0
                    resolve false
                }
            }
        }
    }
--{ + running }--[  ]--
A:srl# show network-instance mgmt  route-table ipv6-unicast prefix 2001:db8::/32 detail
--------------------------------------------------------------------------------------------------------------------------------------------------------
IPv6 unicast route table of network instance mgmt
--------------------------------------------------------------------------------------------------------------------------------------------------------
Destination            : 2001:db8::/32
ID                     : 0
Route Type             : static
Route Owner            : static_route_mgr
Origin Network Instance: mgmt
Metric                 : 1
Preference             : 5
Active                 : true
Last change            : 2023-11-30T02:26:48.203Z
Resilient hash         : false
```

### Arista EOS

Configured with `ipv6 route 2001:db8::/32 Po1 fe80::1` this is a scoped route to a single interface.

### Cisco XR

Configured with:

```
router static
	address-family ipv6 unicast
		2001:db8::/32 bundle-ether1 fe80::1
```

this is a statically scoped entry.

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,13 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2023-11-29" {
+    description
+      "Allow a zone to be used in a next-hop IP address.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -300,7 +300,7 @@ submodule openconfig-aft-common {
     }
 
     leaf ip-address {
-      type oc-inet:ip-address;
+      type oc-inet:ip-address-zoned;
       description
         "The IP address of the next-hop system.";
     }
--- a/release/models/aft/openconfig-aft-ethernet.yang
+++ b/release/models/aft/openconfig-aft-ethernet.yang
@@ -20,7 +20,13 @@ submodule openconfig-aft-ethernet {
     "Submodule containing definitions of groupings for the abstract
     forwarding tables for Ethernet.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2023-11-29" {
+    description
+      "Allow a zone to be used in a next-hop IP address.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description
--- a/release/models/aft/openconfig-aft-ipv4.yang
+++ b/release/models/aft/openconfig-aft-ipv4.yang
@@ -20,7 +20,13 @@ submodule openconfig-aft-ipv4 {
     "Submodule containing definitions of groupings for the abstract
     forwarding tables for IPv4.";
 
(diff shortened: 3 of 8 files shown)


robshakir:
> Major YANG version changes in commit [9aaaefa](https://github.com/openconfig/public/commit/9aaaefae75ea03a37f7eb63f0124ee0430e11869):

You're wrong, but that's OK. This is a nuanced change.

 It's a type change, but the type change accepts all the previous values and doesn't change the actual type, so I don't think this is a major version change.

robshakir:
Oh, I didn't think this was a bug! I thought that this was actually that we didn't detect that changing from `string { pattern a; }` to `string { pattern b; }` was backwards compatible if b is more permissive than a. 

Thanks @wenovus !

dplore:
> Oh, I didn't think this was a bug! I thought that this was actually that we didn't detect that changing from `string { pattern a; }` to `string { pattern b; }` was backwards compatible if b is more permissive than a.
> 
> Thanks @wenovus !

I thought the breaking change detection was good.  A type was changed, and further, the major version was not incremented.  I don't expect the check to introspec a regex to know that the new type is actually a superset of the old type and therefore non-breaking.

chrisy:
> Regarding the zoned IP address, is there some reference to this format? I can see that the regex allows IP's like `1.1.1.1%eth0`, but I've never heard of a 'zoned' ip address before?

Zoned addresses have been around for decades, but less common on classic routers. I may be mistaken about their origin but they are essential for, among other things, representing IPv6 link locals on Unix like hosts. IPv4 link locals are not nearly so prevalent but are still a thing.
https://datatracker.ietf.org/doc/html/rfc4007#section-11

wenovus:
> Oh, I didn't think this was a bug! I thought that this was actually that we didn't detect that changing from `string { pattern a; }` to `string { pattern b; }` was backwards compatible if b is more permissive than a.
> 
> Thanks @wenovus !

The comment should've said "No major changes in commit ...", the referenced PR fixes this.

The breaking change check actually did NOT trigger, which happens to be correct in this PR but may not be correct in others. This is because the check currently just looks at the base type. Since both derived types were string types it let it go.

I'll take a look at making this more precise in a future refactor of misc-checks (I want to make use of goyang being able to now analyze two separate versions of openconfig/public YANG trees at once) since it does look like a use case. If you had done the opposite then it should've flagged it. I think the right thing to do here is just to post a comment saying that there **could** be a breaking change since there is a type change even though the base type was the same.

dplore:
Reviewed in Feb 6, 2024 OC Operators meeting.

The following should be added:
Update the ip-address-zoned type to include reference to the RFC 

Update the regex pattern for ipv[4,6]-address-zoned types to make the zone specification optional.  ie: Add a `?` at the end of the regex pattern here: https://github.com/openconfig/public/blob/35614de36d435c3d5efa5a119bd9aba367c9762f/release/models/types/openconfig-inet-types.yang#L156-L157

robshakir:
I'll take a look at making these changes.

PR #1011, opened 2023-12-05
Update openconfig-transport-types.yang

oscargdd:
Aditional PMD types

### Change Scope

Using the openconfig-platform model to discover the transceivers in our network, we discovered some cases that require a new type in Openconfig PMD.
identity ETH_100GBASE_ER4L: Defined in MSA\100GBASE-ER4 Lite 
identity ETH_1GBASE_LX10:  (note that the SFP MSA has also EX, ZX... )
identity ETH_40GBASE_ACU10M (this one and the next I would like to discuss what to do with other distances, I see that there is wide variety that are reported by the devices)
 identity ETH_100GBASE_CU5M

### Platform Implementations

The proposed codes are ethernet standards and available to use in all the platforms using plugables. No proprietary modes are included.

### DIFF
--- a/release/models/optical-transport/openconfig-transport-types.yang
+++ b/release/models/optical-transport/openconfig-transport-types.yang
@@ -22,7 +22,14 @@ module openconfig-transport-types {
     "This module contains general type definitions and identities
     for optical transport models.";
 
-  oc-ext:openconfig-version "0.20.0";
+  oc-ext:openconfig-version "0.21.0";
+
+  revision "2023-08-03" {
+    description
+      "Add new PMDs: ETH_100GBASE_ER4L (MSA\100GBASE-ER4 Lite),
+      ETH_1GBASE_LX10. ETH_40GBASE_ACU10M and ETH_100GBASE_CU5M";
+    reference "0.21.0";
+  }
 
   revision "2023-08-03" {
     description
@@ -989,6 +996,11 @@ module openconfig-transport-types {
       "Ethernet compliance codes (PMD) supported by transceivers";
   }
 
+  identity ETH_1GBASE_LX10 {
+    base ETHERNET_PMD_TYPE;
+    description "Ethernet compliance code: ETH_1GBASE_LX10";
+  }
+
   identity ETH_10GBASE_LRM {
     base ETHERNET_PMD_TYPE;
     description "Ethernet compliance code: 10GBASE_LRM";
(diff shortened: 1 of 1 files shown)


release/models/optical-transport/openconfig-transport-types.yang
```diff
@@ -1109,6 +1131,11 @@ module openconfig-transport-types {
     description "Ethernet compliance code: 100GBASE_DR";
   }
 
+  identity ETH_100GBASE_CU5M {
+    base ETHERNET_PMD_TYPE;
+    description "Ethernet compliance code: 100GBASE_CU5M";
```
dplore, line 1136:
```suggestion
    description "Ethernet compliance code: 100GBASE_CU5M";
```

Should this be `100GBASE-CR4`?  This is more widely used based on searches I did.

release/models/optical-transport/openconfig-transport-types.yang
```diff
@@ -1039,6 +1051,11 @@ module openconfig-transport-types {
     description "Ethernet compliance code: 40GBASE_PSM4";
   }
 
+  identity ETH_40GBASE_ACU10M {
+    base ETHERNET_PMD_TYPE;
+    description "Ethernet compliance code: ETH_40GBASE_ACU10M";
```
dplore, line 1056: Probably should be ETH_40GBASE_ACU

dplore:
Please include a reference for each PMD.  The reference can be to IEEE spec for the PMD or to search results showing a variety of implemenations/transceivers which implement the PMD using a name that matches the identity proposed.

release/models/optical-transport/openconfig-transport-types.yang
oscargdd: Hi!! Yes... seems 100GBASE-CR4 is the standard term.  Found the reference in IEEE: Ehttps://grouper.ieee.org/groups/802/3/bj/public/may12/diminico_01a_0512.pdf

I had put as first version the sample I got from the vendor replied that :-)  Will change to the standard one and add the reference to IEEE.

release/models/optical-transport/openconfig-transport-types.yang
oscargdd: I have doubts in this one... Is it like the previous case and it is IEEE 802.3ba 40GBASE-CR4 ?

release/models/optical-transport/openconfig-transport-types.yang
oscargdd: hI @dplore 100GBASE_CR4 already existed in Openconfig, so I just update the reference. Do you think is is also worth updating the description to say it is a DAC?

release/models/optical-transport/openconfig-transport-types.yang
oscargdd: @dplore 40GBASE_CR4 was already available, so I just update the reference... no need to add a new ACU10M type.

dplore (approved):

PR #1012, opened 2023-12-05
Adding support for operational state representing interface rates.

hyun-arista:
### Change Scope

* Adding support for configuration and operational state representing interface rates under `/interfaces/interface`.  Given an interface, the load-interval (the interval of interface rates calculation in seconds) is configurable.  Also, given an interface, the operational state will expose ingress/egress rates in units of bits-per-second (bps) and packets-per-second (pps).

pyang output:
```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw oc-if-rates:rates
           +--rw oc-if-rates:config
           |  +--rw oc-if-rates:load-interval?   uint16
           +--ro oc-if-rates:state
              +--ro oc-if-rates:load-interval?   uint16
              +--ro oc-if-rates:out-bits-rate?   uint64
              +--ro oc-if-rates:in-bits-rate?    uint64
              +--ro oc-if-rates:out-pkts-rate?   uint64
              +--ro oc-if-rates:in-pkts-rate?    uint64
```

### Platform Implementations

 * Arista `show interfaces counters rate`: [link to documentation](https://www.arista.com/en/um-eos/eos-ethernet-ports#xx1143927)
*  [Cisco IOSXR](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/system_monitoring/command/reference/b-sysmon-cr-asr9k/b-sysmon-cr-asr9k_chapter_0111.html#wp1787771667) supports 600 as the max in 30 second increments.  
* Nokia SRLinux [info from state interface](https://infocenter.nokia.com/public/SRLINUX213R1A/index.jsp?topic=%2Fcom.srlinux.configbasics%2Fhtml%2Fconfigb-interfaces.html) command supports showing a traffic rate, but does not expose a configuration for the `load-interval` over which the rate is averaged.
* JunOS [show interfaces <interface_name> extensive](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/command/show-interfaces-extensive.html) shows a bits per second rate, but does not expose a configuration for the `load-interval` over which the rate is averaged.

### DIFF
--- a/release/models/interfaces/openconfig-if-rates.yang
+++ b/release/models/interfaces/openconfig-if-rates.yang
@@ -1,0 +1,114 @@
+module openconfig-if-rates {
+
+  yang-version "1";
+
+  namespace "http://openconfig.net/yang/interfaces/rates";
+
+  prefix "oc-if-rates";
+
+  import openconfig-extensions { prefix oc-ext; }
+  import openconfig-interfaces { prefix "oc-if"; }
+
+  organization
+    "OpenConfig working group";
+
+  contact
+    "www.openconfig.net";
+
+  description
+    "This module adds configuration and operational state for interface rates.";
+
+  oc-ext:openconfig-version "0.1.0";
+
+  revision 2024-04-02 {
+    description
+      "Augment of /interfaces/interface to include interface rates.";
+    reference
+      "0.1.0";
+  }
+
+  grouping interface-rates-config {
+    description
+      "Grouping of interface rates related configuration";
+
+    leaf load-interval {
+      type uint16 {
+        range "1..600";
+      }
+      units "seconds";
+      default 300;
+      description
+        "The interval of interface rates calculation in seconds";
+    }
+  }
+
+  grouping interface-rates-state {
+    description
+      "Grouping of interface rates with different direction and units";
+
+    leaf load-interval {
+      type uint16;
+      units "seconds";
+      description
+        "The interval of interface rates calculation in seconds";
+    }
+
+    leaf out-bits-rate {
+      type uint64;
+      units "bps";
+      description
+        "The calculated transmitted rate of the interface, measured in bits
+        per second.";
+    }
+
+    leaf in-bits-rate {
+      type uint64;
+      units "bps";
+      description
+        "The calculated received rate of the interface, measured in bits
+        per second.";
+    }
+
+    leaf out-pkts-rate {
+      type uint64;
+      units "pps";
+      description
+        "The calculated transmitted rate of the interface, measured in packets
+        per second.";
+    }
+
+    leaf in-pkts-rate {
+      type uint64;
+      units "pps";
+      description
+        "The calculated received rate of the interface, measured in packets
+        per second.";
+    }
+  }
+
+  augment "/oc-if:interfaces/oc-if:interface" {
+    description
+      "Adds interface rates.";
+
+    container rates {
+      description
+        "Enclosing container for interface rates.";
+
+      container config {
+        description
+          "Enclosing container for interface rates related configuration";
+
+        uses interface-rates-config;
+      }
+
+      container state {
+        config false;
+        description
+          "Enclosing container for operational state representing
+          interface rates.";
+
+        uses interface-rates-state;
+      }
+    }
+  }
+}
--- a/release/models/interfaces/.spec.yml
+++ b/release/models/interfaces/.spec.yml
@@ -8,6 +8,7 @@
     - yang/interfaces/openconfig-if-poe.yang
     - yang/interfaces/openconfig-if-ip-ext.yang
     - yang/interfaces/openconfig-if-8021x.yang
+    - yang/interfaces/openconfig-if-rates.yang
     - yang/vlan/openconfig-vlan.yang
     - yang/interfaces/openconfig-if-tunnel.yang
     - yang/platform/openconfig-platform-port.yang
@@ -26,6 +27,7 @@
     - yang/interfaces/openconfig-if-poe.yang
     - yang/interfaces/openconfig-if-ip-ext.yang
     - yang/interfaces/openconfig-if-8021x.yang
+    - yang/interfaces/openconfig-if-rates.yang
     - yang/vlan/openconfig-vlan.yang
     - yang/interfaces/openconfig-if-tunnel.yang
     - yang/platform/openconfig-platform-port.yang


dplore:
Should there also be an refresh interval which is configurable?  

 (example ref: [Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/system_monitoring/command/reference/b-sysmon-cr-asr9k/b-sysmon-cr-asr9k_chapter_0111.html#wp1787771667)

LimeHat:
Perhaps the type should be uint64 (gauge64?) or uint32.

dplore:
@hyun-arista any response for us?

hyun-arista:
Sorry for the delay on the response; lost track this.

Regarding the load-interval:  Yes, having a load-interval to be configurable with a range of say 0-600 seconds with default of 300 seconds sounds reasonable.

Regarding changing the rate value type from float to int:  I am not sure about this; I don't have a concrete example but wouldn't it be better to have it as accurate as possible?  We can leave it up to the end-user to truncate the fractional part if needed.

dplore:
> Sorry for the delay on the response; lost track this.
> 
> Regarding the load-interval: Yes, having a load-interval to be configurable with a range of say 0-600 seconds with default of 300 seconds sounds reasonable.
> 
> Regarding changing the rate value type from float to int: I am not sure about this; I don't have a concrete example but wouldn't it be better to have it as accurate as possible? We can leave it up to the end-user to truncate the fractional part if needed.

We'd like to avoid introducing more floats into OC unless the underlying data really requires it.  Integer or decimal would be preferred.  uint64 seems good here.  I don't think we really need to see fractional packets and bits per second?  

Here's a link to the latest thread on float in OC yang models: https://github.com/openconfig/public/issues/1042

hyun-arista:
> > Sorry for the delay on the response; lost track this.
> > Regarding the load-interval: Yes, having a load-interval to be configurable with a range of say 0-600 seconds with default of 300 seconds sounds reasonable.
> > Regarding changing the rate value type from float to int: I am not sure about this; I don't have a concrete example but wouldn't it be better to have it as accurate as possible? We can leave it up to the end-user to truncate the fractional part if needed.
> 
> We'd like to avoid introducing more floats into OC unless the underlying data really requires it. Integer or decimal would be preferred. uint64 seems good here. I don't think we really need to see fractional packets and bits per second?
> 
> Here's a link to the latest thread on float in OC yang models: #1042

Thanks for letting me know.  I wasn't aware of this issue.  I'll submit an update of the PR early next week (adding a config container to the rates container and adding a load-interval leaf + changing the rate value type to uint64).  Thanks!

hyun-arista:
The proposal is updated as follows:
- Added configuration for `load-interval` (type: uint16, units: "seconds", range:  "0..600", and default: 300)
- Changed the type of state leaves from float to uint64.

Thanks!

release/models/interfaces/openconfig-if-rates.yang
```diff
@@ -0,0 +1,107 @@
...
+      "0.1.0";
+  }
+
+  grouping interface-rates-config {
+    description
+      "Grouping of interface rates related configuration";
+
+    leaf load-interval {
+      type uint16 {
+        range "0..600";
```
LimeHat, line 36: While `600` is the common max. value in the existing implementations, I don't think it should be used in the OC yang as the maximum.

Also, is `0` a valid value? If so, please add the meaning of "0" to the description.

LimeHat (changes requested):

release/models/interfaces/openconfig-if-rates.yang
dplore: @hyun-arista I suggest using range 1...600

@LimeHat if you can show a evidence of a range longer than 600, please provide it and we can use it.  I couldn't find a SRLinux reference?  

If not, 600 seems good enough for now.  I would expect increasing the range limit in the future should be non-breaking as the model becomes more permissive.

Repeating and adding references here:
*  [Cisco IOSXR](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/system_monitoring/command/reference/b-sysmon-cr-asr9k/b-sysmon-cr-asr9k_chapter_0111.html#wp1787771667) supports 600 as the max in 30 second increments.  
* [JunOS](https://www.juniper.net/documentation/us/en/software/junos/interfaces-security-devices/cli-reference/topics/ref/statement/load-interval-edit-interfaces.html) is at 180 max in 10 second increments.  
* [Arista EOS](https://www.arista.com/en/um-eos/eos-ethernet-ports) does not include details on the values supported.

LimeHat: @dplore I don't have a reference for an implementation that supports >600s as of today, but what is the purpose of the limit?

E.g. there's no implementation that supports 60k MTU or `max(uint32)` is-is hello-interval, and we don't limit those values.

dplore: It's true, OC is not very consistent about using ranges.  Having a smaller surface area to support and/or test can be useful.  IMO, it's feasible to model range constraints when there are one or more of:
- an externally defined specification 
- 2 or more implementation references
- rough consensus

I'll bring this up on the OC operators call Apr 9, 2024 and see if the operator group has enough of an opinion to say this should be included or excluded.

LimeHat: Thank you, Darren. 
Indeed, it makes a lot of sense in the first case (an externally defined spec) since all other values are invalid/wrong.

I see two issues with enforcing an arbitrary limits for values that are not limited by the standards:
1. any implementer that is willing to go above the limit is forced to make an upstream model change first, wait until it is accepted (if it is accepted), and wait for the next OC release, and only then it can be properly implemented.
2. it is hard to model these limits in a consistent way. 

Could you elaborate on the point about the surface area? If the OC doesn't cap this at 600, it doesn't mean that every implementation should support higher values. In fact, even with 600, there will be differences between the implementations (as the JunOS example shows)

s19nal: we discussed at todays operators meeting - I don't have any objection to a restriction that covers a sensible operational range.  Suggest we use a range 1..600 for now and as Darren suggests, we could - given an operational use case in future - extend the range without it being a breaking change.

LimeHat: Can we have some general guidance on when the ranges should be used? I'd suggest to update the OC style guide.

Otherwise, it seems pretty arbitrary (when the values are not restricted by a standard), and there's no clarity on how the "sensible operational range" is defined and in what cases. Why 600 is sensible, and 700 is not?

I'm not too concerned about the specific use case, but I do have a general concern about the consistency and the implications of imposing arbitrary restrictions.

hyun-arista:
I've updated the PR in the following way:
- Changed the range of `load-interval` to 1..600.
- Added a mirroring state leaf for the `load-interval` config leaf.
Thank you!

release/models/interfaces/openconfig-if-rates.yang
dplore: I agree that in this specific case, specifying a range is not much of a concern.  There are not obvious conflicts with this range.  Operationally it's not useful to average the counters for an interface over a very long period of time.    How long? Well we've said no need to support a value greater than 10 minutes. 

As far as a general rule for all of OC, I am not sure it is feasible?  If anyone has a proposal it would be a good pull request to review!

dplore (approved):
LGTM. Last call for comments.  This will merge on April 22, 2024

release/models/interfaces/openconfig-if-rates.yang
LimeHat: >As far as a general rule for all of OC, I am not sure it is feasible?

It's better to be feasible because this PR sets an interesting precedent.  
My proposal would be, "don't use ranges when there's no corresponding standard that limits the values", which is in line with the current usage I think. 

Also, is changing the range in the future can be considered non-breaking? For example, if we extend this range to 900 and there's a system (e.g. an NMS) that does a strict validation and relies on the previous range.

earies:
Just catching up on this PR now...

> [JunOS](https://www.juniper.net/documentation/us/en/software/junos/interfaces-security-devices/cli-reference/topics/ref/statement/load-interval-edit-interfaces.html) is at 180 max in 10 second increments.

Note: This is an invalid reference for current JUNOS/EVO platforms (very old EOL platform, different feature-set).  Thx for digging up refs @dplore but ideally the submitter puts this in the PR submission that already states this vs. ignoring the template (all too common I see on submissions) _or_ if ever unsure - feel free to tag other implementor github handles.

For reference - we do not currently support this feature in JUNOS/EVO as of today.

Other than that a few questions/comments:

* It appears we are bringing a long standing feature that's use has primarily been on-box diagnosis (run a command to give the operator an idea of calculated rates over the last sample period) - useful but another view of data already present that can be calculated off sibling/child leafs correct?  If that is the only use then why additional nodes in a data-model and not handle this in downstream renderers?
* The categorization is flat leafs at the `state` container - at a minimum I'd recommend a container such as `rates` to group  (but I'd like to understand the prior point first)

dplore:
> 
> * It appears we are bringing a long standing feature that's use has primarily been on-box diagnosis (run a command to give the operator an idea of calculated rates over the last sample period) - useful but another view of data already present that can be calculated off sibling/child leafs correct?  If that is the only use then why additional nodes in a data-model and not handle this in downstream renderers?

We discussed this point in the OC Operator's review.  The consensus was that we see a strong need to support on-box debugging, including configuring a custom the interface counter rate interval (`load-interval` in this model).

We didn't have a deep discussion on the actual bits/pkt rate counters themselves, but there was no objection to including these.  Agreed the values are typically derived off device by a NMS from the octet/pkt counts.

earies:
> > * It appears we are bringing a long standing feature that's use has primarily been on-box diagnosis (run a command to give the operator an idea of calculated rates over the last sample period) - useful but another view of data already present that can be calculated off sibling/child leafs correct?  If that is the only use then why additional nodes in a data-model and not handle this in downstream renderers?
> 
> We discussed this point in the OC Operator's review. The consensus was that we see a strong need to support on-box debugging, including configuring a custom the interface counter rate interval (`load-interval` in this model).
> 
> We didn't have a deep discussion on the actual bits/pkt rate counters themselves, but there was no objection to including these. Agreed the values are typically derived off device by a NMS from the octet/pkt counts.

Understood for the on-box debuggability aspect but I see this as something that doesn't need to be expanded in the data-model directly as it is presentation of other nodes depending on some configured value and underlying feature (specific interface rate feature at that).  It's a question of data-model node growth vs. handling this in presentation.

The underlying feature is essentially a cache of sampled data over a period of time on which calculations can be ran so if there is support on-box of such a cache, a configurable load-interval should not even be necessary but a query to the cache with a value of time could return such results - not limited to "interface rates"

Essentially the equivalent of a lightweight on-box TSDB that maybe can selectively cache certain datapoints over a limited period of time is likely the better solution here imo.

dplore:
I don't think you're wrong Ebben.  But seeing as we have two references that have full support, this is enough to move this forward.  

I noted that JunOS [show interface extensive](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/command/show-interfaces-extensive.html) but does not expose what the "load-interval" is.

dplore:
Discussed in April 30, 2024 OC operators meeting without objection.

PR #1014, opened 2023-12-07
add hex-string-prefixed typedef

### LINKED ISSUE #903: Ambiguous key-ids in keychain
The keychain model has been updated with the key-id being a union:

```
  grouping keychain-key-config {
    description "This grouping defines key-chain key parameters.";

    leaf key-id {
      type union {
        type oc-yang:hex-string {
          length "1..64";
        }
        type uint64;
      }
      description
        "Identifier for the key within the keychain.";
    }
```

The problem with this is that for some values (e.g., a hex value that happens to contain no letters), and for some encodings then it will be ambiguous as to whether the key-id is meant to represent a hex string or an integer.  E.g., with both the XML encoding and IETF YANG JSON encoding, these will both be encoded as strings.  I.e., all keys will look to the server like they are hex strings.

Any implementations that treat MAC Sec keys (hex strings) differently from other integer based ids for keychains will likely have problems mapping this.

dplore:
Fixes #903 

Deprecate the type hex-string and introduce a new type to replace it named `hex-string-prefixed`.  

The current PR replaces hex-string with hex-string-prefixed where it is used in a union with an integer, as that is the root issue.  

`0x` is used as a prefix to disambiguate when a hex-string is a union with other string types.

### Impact of change
Summary:

The change to release/models/keychain/openconfig-keychain.yang is a breaking change of a v0 model.

The change to release/models/policy/openconfig-policy-types.yang is adding hex-string-prefixed to the end of the union, allowing the non-breaking introduction of the new type into this v3.x model.

Details:

`hex-string` type is used in a union type in
  typedef tag-type which appears in the following leaves:

[/network-instances/network-instance/protocols/protocol/static-routes/static/config/set-tag](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-static-routes-static-config-set-tag)	

[/network-instances/network-instance/protocols/protocol/static-routes/static/state/set-tag](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-static-routes-static-state-set-tag)	

[/network-instances/network-instance/protocols/protocol/local-aggregates/aggregate/config/set-tag](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-local-aggregates-aggregate-config-set-tag)	

[/network-instances/network-instance/protocols/protocol/local-aggregates/aggregate/state/set-tag](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-local-aggregates-aggregate-state-set-tag)

[/routing-policy/policy-definitions/policy-definition/statements/statement/actions/set-tag/inline/config/tag](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-actions-set-tag-inline-config-tag)	

[/routing-policy/policy-definitions/policy-definition/statements/statement/actions/set-tag/inline/state/tag](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-actions-set-tag-inline-state-tag)

And also as a union in the keychain model:

[/keychains/keychain/keys/key/config/key-id](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-keychain.html#keychains-keychain-keys-key-config-key-id)

[/keychains/keychain/keys/key/state/key-id](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-keychain.html#keychains-keychain-keys-key-state-key-id)


release/models/types/openconfig-yang-types.yang
```diff
@@ -32,7 +32,13 @@ module openconfig-yang-types {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.3.1";
+  oc-ext:openconfig-version "1.0.0";
```
chrisy, line 35: Is this really a major bump? Nothing is using the typedef (yet).

release/models/types/openconfig-yang-types.yang
```diff
@@ -32,7 +32,13 @@ module openconfig-yang-types {
...
+
+  revision "2023-12-07" {
+    description
+      "Add hex-string-colon typedef";
+    reference "0.3.1";
```
chrisy, line 40: You forgot to update this version. Little surprised the CI didn't catch that 0.3.1 is duplicated.

chrisy (dismissed):

dplore:
@m26singhvi ready for your review

release/models/types/openconfig-yang-types.yang
```diff
@@ -117,6 +124,16 @@ module openconfig-yang-types {
       "A string consisting of a hexadecimal characters.";
   }
 
+  typedef hex-string-prefixed {
```
m26singhvi, line 127: I think this can be named better. Not a great suggestion but how about rfc6991-hex-string ?

m26singhvi (commented):
Overall looks good to me.  Deprecating hex-string is really a big change. It will be good if someone closer to the openconfig model approves this.

m26singhvi:
@dplore - any updates on this ?

marcushines (approved):

dplore:
Last call for this PR.  This will merge on July 1, 2024

dplore:
@chrisy addressed all your comments and expanded the PR to update the locations where hex-string is used.

release/models/keychain/openconfig-keychain.yang
```diff
@@ -194,7 +200,7 @@ module openconfig-keychain {
 
     leaf key-id {
       type union {
-        type oc-yang:hex-string {
+        type oc-yang:hex-string-prefixed {
```
earies, line 203: Isn't this now going to pose an issue for the `length 1..64` restriction?

`0x` is now a mandatory prefix and counts towards the string length

release/models/policy/openconfig-policy-types.yang
```diff
@@ -165,6 +172,7 @@ module openconfig-policy-types {
     type union {
       type uint32;
       type yang:hex-string;
+      type oc-yang:hex-string-prefixed;
```
earies, line 175: In the case above, there is a replacement of the type whereas here is an addition preserving the previous.  Are we removing any ambiguity by keeping the union of the old + new here?

dplore: updated

release/models/keychain/openconfig-keychain.yang
dplore: updated

dplore:
Addressed all comments, reset last call to Jul 23, 2024

release/models/keychain/openconfig-keychain.yang
earies: Thx - I think the min length would probably need adjustment in that 3 is now the bare minimum (Just `0x` is insufficent to represent a hex-string unless we assume `0x0`).  The pattern regex will cover the type check irregardless.

It seems each usage can define a separate upper bounds but the lower bounds should always have a consistent minimum.

The typedef could carry something like the following:

`range 3..max;`

Whereas each use could override but maintain the min

`range 3..66;`
`range 3..254;`

dplore: Thanks for the catch in the min and good idea to add the other length constraints.  I've updated.

robshakir (approved):

release/models/keychain/openconfig-keychain.yang
```diff
@@ -194,13 +200,19 @@ module openconfig-keychain {
...
+          length "3..66";
+        }
+       type oc-yang:hex-string {
```
robshakir, line 206: Suggest to remove this, since otherwise, we'll fall through and "123" will still be parsed as a `hex-string`.

dplore:
Note breaking change made to the v0.x keychain model.  This is labeled non-breaking as by policy, v0 models are expected to have breaking changes made.  This will change not trigger a major version increment in the OC release.

Issue #1022, opened 2023-12-20
Negative clock offsets are not representable in the NTP model


nkitchen:
NTP clock offsets can be negative, but the current model uses uint64 to represent them:

> https://github.com/openconfig/public/blob/master/release/models/system/openconfig-system.yang#L783-L784

This issue overlaps with #848, which also recognizes the problem with uint offsets, but this issue does not take the position that the type should necessarily be decimal64, nor does it cover root-delay and root-dispersion.

nkitchen:
As described in [RFC 5905](https://datatracker.ietf.org/doc/html/rfc5905#section-8), the range of the offset that can be computed from the timestamps in NTP packets is &plusmn;34 years.

If the `offset` node had type int64 with microsecond units, its range would be &plusmn;2<sup>63</sup> / 10<sup>6</sup>  seconds, which is about &plusmn;292,277 years -- so we don't need millisecond units to get sufficient range.

(Actually, even nanosecond units would work: They would give us a range of &plusmn;292 years.)

PR #1023, opened 2023-12-21
Change type of NTP offset to int64

nkitchen:
### Change Scope

* This change modifies the type of `/system/ntp/servers/server/state/offset` to be signed.
* The change is not backward compatible.
### Platform Implementations

 * NTP Reference Implementation: https://github.com/ntp-project/ntp/blob/9c75327c3796ff59ac648478cd4da8b205bceb77/ntpd/ntp_proto.c#L1934

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -51,7 +51,8 @@ module openconfig-system {
 
   revision "2023-12-20" {
     description
-      "Change type of NTP offset to int64, and clarify its description.";
+      "Change NTP offset, root-delay, and root-dispersion to int64 nanoseconds,
+      and update their descriptions for accuracy and clarity.";
     reference "1.0.0";
   }
 
@@ -759,36 +760,29 @@ module openconfig-system {
     }
 
     leaf root-delay {
-      type uint32;
-      // TODO: reconsider units for these values -- the spec defines
-      // rootdelay and rootdisperson as 2 16-bit integers for seconds
-      // and fractional seconds, respectively.  This gives a
-      // precision of ~15 us (2^-16).  Using milliseconds here based
-      // on what implementations typically provide and likely lack
-      // of utility for less than millisecond precision with NTP
-      // time sync.
-      units "milliseconds";
+      type int64;
+      units "nanoseconds";
       description
-        "The round-trip delay to the server, in milliseconds.";
+        "The total round-trip delay to the reference clock, in nanoseconds.";
       reference
         "RFC 5905 - Network Time Protocol Version 4: Protocol and
-        Algorithms Specification";
+        Algorithms Specification, Section 7.3";
     }
 
     leaf root-dispersion {
-      type uint64;
-      units "milliseconds";
+      type int64;
+      units "nanoseconds";
       description
-        "Dispersion (epsilon) represents the maximum error inherent
-        in the measurement";
+        "The maximum error inherent in the measurement, accumulated over the
+        stratum levels from the reference clock.";
       reference
         "RFC 5905 - Network Time Protocol Version 4: Protocol and
-        Algorithms Specification";
+        Algorithms Specification, Section 4";
     }
 
     leaf offset {
       type int64;
-      units "milliseconds";
+      units "nanoseconds";
       description
         "Estimate of the current time offset from the peer.  This is
         the time difference of the peer's clock minus the local clock.";
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -47,7 +47,13 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.17.1";
+  oc-ext:openconfig-version "1.0.0";
+
+  revision "2023-12-20" {
+    description
+      "Change type of NTP offset to int64, and clarify its description.";
+    reference "1.0.0";
+  }
 
   revision "2023-06-16" {
     description
@@ -781,11 +787,14 @@ module openconfig-system {
     }
 
     leaf offset {
-      type uint64;
+      type int64;
       units "milliseconds";
       description
         "Estimate of the current time offset from the peer.  This is
-        the time difference between the local and reference clock.";
+        the time difference of the peer's clock minus the local clock.";
+      reference
+        "RFC 5905 - Network Time Protocol Version 4: Protocol and
+        Algorithms Specification, Section 8";
     }
 
     leaf poll-interval {


dplore:
Additional references attached.  These both show ntp offset formatted in milliseconds.  

[Cisco IOS XR - show ntp status](https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/system_management/command/reference/yr40xr12k_chapter9.html#wp1026718153)

[JunOS - show ntp status](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/command/show-ntp-status.html)

I acknowledge that the wire format specified in the RFC for NTP is float.  OC leafs may or may not match the wire format.  (There is precedent for both scenarios in OC).  What is more important is that we have the same support across multiple implementations.   

We can bring this issue to the OC community meeting to see if there is any consensus on which change is preferred (https://github.com/openconfig/public/pull/995 or #1023 )

dplore:
After reviewing in the OC community meeting, there was no strong objection or consensus on #1022 or #995.

A few people did point out that float32 can be confusing to use in yang since it gets encoded on the wire as a binary value that needs to be unmarshalled from ieee754 representation.  Since there is no 1:1 mapping between any standard type and the actual ntp value for offset on the wire, I prefer we go with the more simple representation of int64 proposed here in #1022.  I am moving this to last-call, to be merged on Jan 28, 2024.

dplore (approved):

kbchandradeep:
@dplore 
Thanks for the update.
Example values of offset, root delay and root dispersion mentioned in below doc are -124.051 msec, 174.060 msec and 172.37 msec respectively
https://www.cisco.com/c/en/us/td/docs/routers/xr12000/software/xr12k_r4-0/system_management/command/reference/yr40xr12k_chapter9.html#wp1026718153

Example values of offset, root delay and root dispersion mentioned in below doc are -2.852 msec, 23.010 msec and 5.112 msec respectively
https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/command/show-ntp-associations.html

All the 3 fields are decimal values in milliseconds.
If the type for the fields is "int" with units as "milliseconds", are we not truncating the decimal part and only streaming the whole number part? The value displayed in cli would be different from the value streamed out and would this be fine?
Any reason for not having the type as decimal64?

dplore:
We avoid using decimal because of the complexity of dealing with encoding of decimal strings.  A simple alternative is to use microsecods or nanoseconds instead.

@nkitchen would you like to use nanoseconds as it is quite common throughput openconfig.  As [you pointed out](https://github.com/openconfig/public/issues/1022#issuecomment-1865299815), nanoseconds still gives us sufficient range to cover the NTP spec.

earies:
I think this suffices to keep it as a `int64` w/ units normalized to `nanoseconds` for all such leafs where the conversion looks as such (might warrant a common typedef)

```
$ ntpq -pn
     remote           refid      st t when poll reach   delay   offset  jitter
==============================================================================
*2610:20:6f15:15 .NIST.           1 u  354 1024  377   52.399   -0.342   2.174

$ ntpq -pn | grep ^* | awk '{print $8}' | xargs echo "10^6 *" | bc
52399000.000
$ ntpq -pn | grep ^* | awk '{print $9}' | xargs echo "10^6 *" | bc
-342000.000
$ ntpq -pn | grep ^* | awk '{print $10}' | xargs echo "10^6 *" | bc
2174000.000
```

nkitchen:
I updated the PR to use nanoseconds and to include root-delay and root-dispersion. I included description updates from #1024 -- I don't think it makes sense to keep those changes separate any more.

dplore:
@nkitchen do you want this to be a breaking change?  If yes, we will hold on merging this until shortly before the next major release of OC.  This would likely be in December 2024.

If you would like to merge this right away, we need to make this a minor release and non-breaking change by deprecating the current leafs and introducing new leafs.  

Perhaps adding leafs with `-ns` as the suffix to indicate nanoseconds would be clean enough?   I appreciate any other thoughts here.

earies:
> @nkitchen do you want this to be a breaking change? If yes, we will hold on merging this until shortly before the next major release of OC. This would likely be in December 2024.
> 
> If you would like to merge this right away, we need to make this a minor release and non-breaking change by deprecating the current leafs and introducing new leafs.
> 
> Perhaps adding leafs with `-ns` as the suffix to indicate nanoseconds would be clean enough? I appreciate any other thoughts here.

I think we're going to keep running into this situation and while strict backwards compatibility rules are good for everyone, we are going to be faced w/ the dilemma of a published node (that may or may not be implemented/adopted) that is just incorrect and the best way to correct it (or remove) is to break the model in a backwards incompatible fashion.

Strict API compatibility rules would mean you do what you suggest - introduce a new node, mark the old as deprecated then we are faced w/ a weird nomenclature conundrum (foo -> foo2, offset -> offset-ns).

Aside from constantly floating out the question to the community that participates if breaking it would be a big deal or not (which at times is known, other times not so much as most folks consume telemetry data w/o realizing the modeling aspects), if we stick to strict backwards compatibility, I think we'll need to develop some guidance as to examples of how you might name a parallel new node that deprecates the old, how descriptions should be updated to the deprecated node that point to the new, etc...

This also highlights the emphasis to get the initial design as good and correct as possible.  This will never be perfect but we have had plenty of situations where things creep in that are just flat out incorrect and while folks might consume and depend off of them, the question is often _how_ if it is not truly correct but may be good enough...

Timeline-wise, I'd vote for your suggestion @dplore but I think some guidance needs to be put together as this will keep popping up

nkitchen:
> > @nkitchen do you want this to be a breaking change? If yes, we will hold on merging this until shortly before the next major release of OC. This would likely be in December 2024.
[...]
> 
> I think we're going to keep running into this situation and while strict backwards compatibility rules are good for everyone, we are going to be faced w/ the dilemma of a published node (that may or may not be implemented/adopted) that is just incorrect and the best way to correct it (or remove) is to break the model in a backwards incompatible fashion.

Agreed...in this case, it's impossible for any vendor to implement `.../server/state/offset` correctly, so it doesn't make sense to me to be careful about compatibility in the usual way.

Someone could argue that `root-delay` and `root-dispersion` were implementable, so we should be more careful about changing them incompatibly. But those values are not worth anything without a meaningful offset. It's the single most important attribute of an NTP server's state.

Is it possible for this to be treated specially, so that the breaking change can be accepted without waiting a year?

If that's really not an option, then I will revise it into a non-breaking change. I don't think it makes sense to wait a year to have a working NTP-state model.

dplore:
We can choose to treat this as a bug fix since we are fixing something that is broken.  

I will  bring this up as a topic at the Feb 1, 2024 community meeting.  As this is one of those situations where neither solution is great, I'm mostly looking for any evidence that moving forward with merging this now as a bugfix is going to cause major problems.

earies:
> I'm mostly looking for any evidence that moving forward with merging this now as a bugfix is going to cause major problems.

My only comment here to these cases (and they've varied in magnitude over the years) is that from an implementation standpoint, I'm in full agreement - fix the bug, don't carry it forward or prolong - e.g. do the right thing

Consumer-wise however - there is a wide variety of folks that consume telemetry data unbeknownst to the underlying data-model set or compatibility rules which means any backwards incompatible change of any kind is a problem for them whether or not the data is correct and/or ultimately used/acted upon.  It creates a potential break in an existing workflow/pipeline.  This is also the reason why vendor implementations have likely carried forward many years of technical debt from stricter internal protocol to minimize breakage across upgrades, etc...

This latter point has been very painful for implementors to have to handle/field so I think its worth getting these "rules" down into some higher level documentation surrounding at a minimum the YANG data-modeling portion but this can apply across all OpenConfig sub-projects that each may carry different rulesets

nkitchen:
> My only comment here to these cases (and they've varied in magnitude over the years) is that from an implementation standpoint, I'm in full agreement - fix the bug, don't carry it forward or prolong - e.g. do the right thing
> 
> Consumer-wise however - there is a wide variety of folks that consume telemetry data unbeknownst to the underlying data-model set or compatibility rules which means any backwards incompatible change of any kind is a problem for them whether or not the data is correct and/or ultimately used/acted upon. It creates a potential break in an existing workflow/pipeline.

@earies I think you make good points about the value of clear guidance on good practices for compatibility.

I don't think I quite understand your position with respect to this particular change. Do you think that the disruption to existing consumers is a big enough problem that it outweighs the reasons to make the breaking fix?

earies:
> > My only comment here to these cases (and they've varied in magnitude over the years) is that from an implementation standpoint, I'm in full agreement - fix the bug, don't carry it forward or prolong - e.g. do the right thing
> > Consumer-wise however - there is a wide variety of folks that consume telemetry data unbeknownst to the underlying data-model set or compatibility rules which means any backwards incompatible change of any kind is a problem for them whether or not the data is correct and/or ultimately used/acted upon. It creates a potential break in an existing workflow/pipeline.
> 
> @earies I think you make good points about the value of clear guidance on good practices for compatibility.
> 
> I don't think I quite understand your position with respect to this particular change. Do you think that the disruption to existing consumers is a big enough problem that it outweighs the reasons to make the breaking fix?

I support doing the right thing here which is to make the breaking fix.  Everything else is unknown to true impact and am just saying we need to manage that better all around (guidance, expectations, BCP) going forward as it is unknown to many consumers of OpenConfig modeled data today.

We've already been doing this for many years (e.g. https://github.com/openconfig/public/pull/604) - just that in trying to firm things up w/ slower deprecation and carefully introducing backwards incompatibilities, we are likely going to need exceptions to the rules and figure out how to manage them accordingly

jsterne:
This is turning into a more general discussion of change management (and should maybe be moved to a separate issue?) but it is a good discussion.
I can see that we may need and want to accept "exceptions" (e.g. for a functional bug fix like this). 
Would we still bump the major version for these exceptions?
A few things we should consider (some of this reiterates what Ebben was saying):
- clearly let users of OC models know that these exceptions can occur and are part of the normal process of using OC models
- have a strategy for documenting the exceptions
- set some sort of criteria for when to accept exceptions (and we should expect evolution and debate over those criteria over time as the user community gains experience dealing with them).
- what set of people will approve the exceptions? What input is needed for that?

dplore:
The rough consensus in the community is to merge this as a breaking change and include in the next release of OC as a major release. 

Last call for comments, this will merge on Feb 15, 2023

dplore (approved):

PR #1024, opened 2023-12-21
Fix descriptions of root-delay and root-dispersion

nkitchen:
* Remove the comment about reconsidering units of root-delay. As long as it's uint32, milliseconds are the smallest possible units that allow it to have sufficient range.  (The root delay values in NTP packets have a maximum value of about 18 hours.  The range of a uint32 as milliseconds is 2^32 * 0.001 seconds = about 1193 hours. That's big enough, by far.  The range of a uint32 as microseconds is about 1.193 hour = too small.)
 * The old description of root-delay matched the (single-server = non-"root") delay, but the *root* delay is accumulated across strata.
 * The old description of root-dispersion was not clearly differentiated from the (single-server = non-"root") dispersion.

This change suggests the question: Does it make sense for these leaf nodes to be the root values, rather than the single-server values? I think the answer is yes, because the root delay and dispersion are included among the fields in NTP packets.

### Change Scope

* This change affects only node descriptions. It is backwards compatible.
### Platform Implementations

* N/A

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -47,7 +47,14 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.17.1";
+  oc-ext:openconfig-version "0.17.2";
+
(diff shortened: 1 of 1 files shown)


nkitchen:
I included these description updates in #1023, so this PR is no longer needed.

PR #1027, opened 2023-12-27
Add link bandwidth community using schema string regex style

dplore:
### Change Scope

* This is an alternative to #883 which expands on the current OC model for BGP communities which uses strings with regex patterns to identify the different community types.  The purpose of this PR is to solicit feedback to see if this approach is more desirable than #883 .   

### Platform Implementations

 * Cisco IOS XR implements a command like [set extcommunity bandwidth (2906:45750000)](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/asr9k-r6-4/routing/configuration/guide/b-routing-cg-asr9000-64x/b-routing-cg-asr9000-64x_chapter_01111.html#concept_xsj_2q1_pdb
) with a user specified bandwidth in a route policy similar to this PR.

 * JunOS implements link bandwidth via configuration of the [bgp community with a user specified bandwidth](https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/topic-map/load-balancing-bgp-session.html#d29e113__d29e120) in a route policy list similar to this PR

* [Arista EOS](https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1418621) enables link bandwidth using a per neighbor command.   Bandwidths can optionally be set using a [route-map](https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1153691) and a percent format.  I also found [this EOS support article](https://www.arista.com/en/support/toi/eos-4-24-1f/14534-bgp-ucmp) which shows support for an explicit bandwidth in bits/second.

* [FRR](https://docs.frrouting.org/en/latest/bgp.html#weighted-ecmp-using-bgp-link-bandwidth) implements link bandwidth via configuration applied to a policy.  Bandwidth is automatically calculated and not configurable.


release/models/bgp/openconfig-bgp-types.yang
```diff
@@ -595,6 +601,25 @@ module openconfig-bgp-types {
                 '429[0-3][0-9]{6}|42[0-8][0-9]{7}|4[0-1][0-9]{8}|[1-3][0-9]{9}|'  +
                 '[1-9][0-9]{1,8}|[0-9])$';
       }
+      type string {
+        // Extended Link Bandwidth Community
+        // bandwidth:<2 byte asn>:<bandwidth value in bytes/sec>
+        pattern 'bandwidth:'                                                      +
```
sulrich, line 607: nit, we think this might be better served as `link-bandwidth` vs. `bandwidth`.

release/models/bgp/openconfig-bgp-types.yang
```diff
@@ -595,6 +601,25 @@ module openconfig-bgp-types {
...
+                '(6553[0-5]|655[0-2][0-9]|65[0-4][0-9]{2}'                        +
+                '|6[0-4][0-9]{3}|[1-5][0-9]{4}|[1-9][0-9]{1,3}|[0-9]):'           +
+                '(429496729[0-5]|42949672[0-8][0-9]|4294967[0-1][0-9]{2}'         +
+                '|429496[0-6][0-9]{3}|42949[0-5][0-9]{4}|4294[0-8][0-9]{5}|'      +
+                '429[0-3][0-9]{6}|42[0-8][0-9]{7}|4[0-1][0-9]{8}|[1-3][0-9]{9}|'  +
+                '[1-9][0-9]{1,8}|[0-9])';
+        oc-ext:posix-pattern '^bandwidth:'                                        +
```
sulrich, line 614: ibid

dplore:
Closing this in favor of #883

dplore:
After feedback from implementors, it seems this approach in #1027 can be implemented with low effort quickly to add basic support for BGP link-bw to OC.   

#883 will continue to be pursued for implementation as the long term direction for creating BGP community sets.

release/models/bgp/openconfig-bgp-types.yang
dplore: Done

dplore:
Since this new type is based on a string I modified this to a more human readable format using bits/sec and an optional suffix of `k`, `m` or `g` to represent kilo, mega of gigabits per second.

release/models/bgp/openconfig-bgp-types.yang
```diff
@@ -595,6 +601,27 @@ module openconfig-bgp-types {
...
+                '|429496[0-6][0-9]{3}|42949[0-5][0-9]{4}|4294[0-8][0-9]{5}|'      +
+                '429[0-3][0-9]{6}|42[0-8][0-9]{7}|4[0-1][0-9]{8}|[1-3][0-9]{9}|'  +
+                '[1-9][0-9]{1,8}|[0-9])(k|m|g)?';
+        oc-ext:posix-pattern '^link-bandwidth:'                                   +
+                '(6553[0-5]|655[0-2][0-9]|65[0-4][0-9]{2}'                        +
+                '|6[0-4][0-9]{3}|[1-5][0-9]{4}|[1-9][0-9]{1,3}|[0-9])$'           +
+                '(429496729[0-5]|42949672[0-8][0-9]|4294967[0-1][0-9]{2}'         +
+                '|429496[0-6][0-9]{3}|42949[0-5][0-9]{4}|4294[0-8][0-9]{5}|'      +
+                '429[0-3][0-9]{6}|42[0-8][0-9]{7}|4[0-1][0-9]{8}|[1-3][0-9]{9}|'  +
+                '[1-9][0-9]{1,8}|[0-9])(k|m|g)?';
```
robshakir, line 622: I think this needs a `$` on the end.

robshakir, line 622: Should the `k|m|g` be upper-case vs. lower-case? It seems to me like it would be preferable to be upper-case.

release/models/bgp/openconfig-bgp-types.yang
```diff
@@ -595,6 +601,27 @@ module openconfig-bgp-types {
...
+        // bandwidth:<2 byte asn>:<bandwidth value in bits/sec,
+        //           optionally with kilo/mega/giga suffix>
+        // Example: link-bandwidth:20:100m
+        pattern 'link-bandwidth:'                                                 +
+                '(6553[0-5]|655[0-2][0-9]|65[0-4][0-9]{2}'                        +
+                '|6[0-4][0-9]{3}|[1-5][0-9]{4}|[1-9][0-9]{1,3}|[0-9]):'           +
+                '(429496729[0-5]|42949672[0-8][0-9]|4294967[0-1][0-9]{2}'         +
+                '|429496[0-6][0-9]{3}|42949[0-5][0-9]{4}|4294[0-8][0-9]{5}|'      +
+                '429[0-3][0-9]{6}|42[0-8][0-9]{7}|4[0-1][0-9]{8}|[1-3][0-9]{9}|'  +
+                '[1-9][0-9]{1,8}|[0-9])(k|m|g)?';
```
robshakir, line 615: Same comment here w.r.t upper-case for `k|m|g`.

robshakir (commented):
This seems reasonable as a lower-touch/backwards-compatible way to add this support. I don't think that it obviates the need for #883 -- so we can continue the discussion here.

release/models/bgp/openconfig-bgp-types.yang
dplore: Done

release/models/bgp/openconfig-bgp-types.yang
dplore: Agreed, capital is slightly more intuitive as it's canonical for 'Kilo, Mega, Giga' vs lowercase which might get confused with 'milli' per the [SI standard units](https://en.wikipedia.org/wiki/Metric_prefix).  I allowed k and K as SI formally doesn't appear to classify 'K', only 'k'.

dplore:
This was reviewed in Jan 23, 2024 OC operators meeting without objection

robshakir (approved):

PR #1028, opened 2023-12-28
Multipath link bandwidth enabled

rszarecki:
### Change Scope

This PR provides configuration hirarchy that allows to enable/disable honoring link-bandwidth extended community when BGP multipath is forming RIB/FIB entries. 

It is done by by link-bandwidth-ext-community dedicated container, so more attributes can be added in future, without breaking model. (e.g. precission/scalaling/reference factors. Or zero value handling).

```
  +--rw network-instances
        +--rw protocols
              +--rw bgp
                 +--rw global
                    +--rw afi-safis
                          +--rw use-multiple-paths
                             +--rw config
                             |  +--rw enabled?   boolean
                             +--ro state
                             |  +--ro enabled?   boolean
                             +--rw ebgp
                             |  +--rw link-bandwidth-ext-community         <<<<<<<<
                             |  |  +--rw config
                             |  |  |  +--rw enabled?   boolean
                             |  |  +--ro state
                             |  |     +--ro enabled?   boolean
                             ...
                             +--rw ibgp
                                +--rw link-bandwidth-ext-community         <<<<<<<<
                                |  +--rw config
                                |  |  +--rw enabled?   boolean
                                |  +--ro state
                                |     +--ro enabled?   boolean
```
same structure repeats for all levels:
```
/network-instances/network-instance/protocols/protocol/bgp/global/use-multiple-paths/ebgp/link-bandwidth-ext-community/config/enabled
/network-instances/network-instance/protocols/protocol/bgp/global/use-multiple-paths/ebgp/link-bandwidth-ext-community/state/enabled
/network-instances/network-instance/protocols/protocol/bgp/global/use-multiple-paths/ibgp/link-bandwidth-ext-community/config/enabled
/network-instances/network-instance/protocols/protocol/bgp/global/use-multiple-paths/ibgp/link-bandwidth-ext-community/state/enabled
/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/afi-safis/afi-safi/use-multiple-paths/ebgp/link-bandwidth-ext-community/config/enabled
/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/afi-safis/afi-safi/use-multiple-paths/ebgp/link-bandwidth-ext-community/state/enabled
/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/afi-safis/afi-safi/use-multiple-paths/ibgp/link-bandwidth-ext-community/config/enabled
/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/afi-safis/afi-safi/use-multiple-paths/ibgp/link-bandwidth-ext-community/state/enabled
/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/use-multiple-paths/ebgp/link-bandwidth-ext-community/config/enabled
/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/use-multiple-paths/ebgp/link-bandwidth-ext-community/state/enabled
/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/use-multiple-paths/ibgp/link-bandwidth-ext-community/config/enabled
/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/use-multiple-paths/ibgp/link-bandwidth-ext-community/state/enabled
```

This leaf has no effect if BGP multi-path is disabled or if maximum-path attribute of BGP multi-path value is set to 1.

Implementation that do support only one state (e.g. implementation do not support BGP wECMP at all) should still support this leaf and accept configuration if value is set to FALSE, while reject configuration when value is set to TRUE.

### Platform Implementations

 * Arista: [`ucmp mode 1`](https://www.arista.com/en/support/toi/eos-4-24-1f/14534-bgp-ucmp#enabling-ucmp-locally) 

 * FRRrouting: [bgp bestpath bandwidth ignore](https://docs.frrouting.org/en/latest/bgp.html#controlling-link-bandwidth-processing-on-the-receiver)

* Cumulus: [net add bestpath bandwidth ignore](https://docs.nvidia.com/networking-ethernet-software/cumulus-linux-41/Layer-3/Unequal-Cost-Multipath/#set-default-values-for-ucmp-routes)

* Juniper JUNOS, Cisco IOS-XR, Nokia SRL - do not allow to ignore/disable link-bandwidth community. They always instantiate wECMP if all multipath members have  link-bandwidth community attached, or ECMP if at least one do not have it attached.

### DIFF
--- a/release/models/bgp/openconfig-bgp-common.yang
+++ b/release/models/bgp/openconfig-bgp-common.yang
@@ -691,7 +691,7 @@ submodule openconfig-bgp-common {
       Local Administrator subfield of link-bandwidth extended
       community [draft-ietf-idr-link-bandwidth-07].
       This leaf has no effect if BGP multi-path is disabled or
-      if maximum-path attribute of BGP multi-path value is set 
+      if maximum-path attribute of BGP multi-path value is set
       to 1";
     }
   }
--- a/release/models/bgp/openconfig-bgp-common.yang
+++ b/release/models/bgp/openconfig-bgp-common.yang
@@ -685,7 +685,7 @@ submodule openconfig-bgp-common {
       extended community in pultipath RIB/FIB formation";
     leaf enable {
       type boolean;
-      default "false";
+      default "true";
       description
       "When set to TRUE, BGP multiplepath shall distributed traffic
       load among contributing routes proportionally to value of
--- a/release/models/bgp/openconfig-bgp-common.yang
+++ b/release/models/bgp/openconfig-bgp-common.yang
@@ -24,7 +24,14 @@ submodule openconfig-bgp-common {
     may be application to a subset of global, peer-group or neighbor
     contexts.";
 
-  oc-ext:openconfig-version "9.5.0";
+  oc-ext:openconfig-version "9.6.0";
+
+  revision "2023-12-28" {
+    description
+      "Add support for controling use of link-bandwidth extended
+      community for BGP multipath.";
+    reference "9.6.0";
+  }
 
   revision "2023-11-01" {
     description
@@ -627,6 +634,21 @@ submodule openconfig-bgp-common {
       container ibgp {
         description
           "Multipath parameters for iBGP";
+        container link-bandwidth-ext-community {
+          description
+            "Usage of DMZ Link-Bandwidth extended community";
+          container config {
+            description
+            "Configuration parameters relating to usage of link-bandwidth"
+            uses bgp-common-use-multiple-paths-link-bandwidth-config;
+          }
+          container state {
+            config false;
+            description
+            "State information relating to usage of link-bandwidth";
+            uses bgp-common-use-multiple-paths-link-bandwidth-config;
+          }
+        }
         container config {
           description
             "Configuration parameters relating to iBGP multipath";
@@ -639,6 +661,21 @@ submodule openconfig-bgp-common {
           uses bgp-common-use-multiple-paths-ibgp-config;
         }
       }
+    }
+  }
+
+  grouping bgp-common-use-multiple-paths-link-bandwidth-config {
+    description
+      "Parameters controlling usage of  of DMZ Link-Bandwidth
+      extended community in pultipath RIB/FIB formation";
+    leaf enable {
+      type boolean;
+      default "false";
+      description
+      "When set to TRUE, BGP multiplepath shall distributed traffic
+      load among contributing routes proportionally to value of
+      Local Administrator subfield of link-bandwidth extended
+      community [draft-ietf-idr-link-bandwidth-07]";
     }
   }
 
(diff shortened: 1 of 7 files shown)


dplore:
Should we not specify a default?   Since Arista EOS is disabled by default and the other vendors are enabled by default we could leave out a default value, which means the default, if any, is up to the device.   We can further give guidance in the description for this leaf to always set the desired value to ensure the desired behavior occurs.

nandanarista:
> Should we not specify a default? Since Arista EOS is disabled by default and the other vendors are enabled by default we could leave out a default value, which means the default, if any, is up to the device. We can further give guidance in the description for this leaf to always set the desired value to ensure the desired behavior occurs.

+1 to not having a default. 
Depending on how a h/w platform implements wECMP, enabling wECMP can lead to more resource utilization, so seems like a somewhat unsafe default=True in that regard.

rszarecki:
I’m fine with no default in this case.
But in this case implementation must support leaf.
- Do not reject config if leaf value is same as implementation default.
- reject config if value is uncupported.



--------------
Rafal Szarecki

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -642,6 +679,21 @@ submodule openconfig-bgp-common {
     }
   }
 
+  grouping bgp-common-use-multiple-paths-link-bandwidth-config {
+    description
+      "Parameters controlling usage of  of DMZ Link-Bandwidth
+      extended community in pultipath RIB/FIB formation";
+    leaf enable {
+      type boolean;
+      default "true";
```
dplore, line 688: Please remove default.

dplore (commented):
looks good except the default should be removed as discussed.  You may update the description to indicate the expected behavior when setting this leaf.

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -642,6 +679,21 @@ submodule openconfig-bgp-common {
...
+      description
+      "When set to TRUE, BGP multiplepath shall distributed traffic
```
nandanarista, line 690: i'm assuming that when set to TRUE, wECMP only applies if `use-multiple-paths/config/enabled` and `use-multiple-paths/{i|e}bgp/config/maximum-paths` is > 1, otherwise setting this leaf to true has no effect?

In EOS, `ucmp mode 1` [1] has no effect unless `maximum-paths <n>` where n > 1 is also true

[1] https://www.arista.com/en/support/toi/eos-4-24-1f/14534-bgp-ucmp
```
‘ucmp mode 1 [<M> [<deviation%>]]’ is a mandatory command along with ‘maximum-paths <max-paths> [ecmp <max-ecmp>]’ command to enable UCMP for received routes at the DUT locally.
```

nandanarista (commented):
Thanks for putting this together, Rafal!
one nit and one question

dplore:
This was reviewed at the Jan 9th OC Operators meeting without objection.   Setting last call to Jan 16, 2024

dplore:
@rszarecki this is ready except for the version conflicts.  Can you fix for us?

rszarecki:
@dplore 
version conflicts fixed.

dplore (approved):
LGTM, thank you @rszarecki !

PR #1032, opened 2024-01-05
Update description for forwarding-viable

sachendras:
Add more clarity/rules for the forwarding-viable leaf.

1. When a controller "voluntarily" changes the state of all member-links on a LAG bundle or the bundle itself to forwarding-viable=False, it expects that the port/bundle is not used to transmit data-plane traffic even when that is the port of last resort.

a. This requirement can be easily be achieved by an implementation if they follow the rule that, when a port/bundle is set to forwarding-viable=false, do not transmit L3 packets. This is true for both control-plane or data-plane packets. As a result, it would be a graceful transition for L3 traffic on that connection because, the other end of the connection will have its L3 protocol's dead-interval/hold-down timer expired leading for it to migrate away from the connection eventually.

2. I feel minimum-link calculation for aggregate bundle should be ignored for member ports whose forwarding-viable=false. Minimum-Links add unnecessary complications on the controller end. Also, that will result in an implementation to disable a bundle if the min-links requirement is not met and that is not the expectation. Expectation is that, the implementation must continue to Receive and Transmit L2 and only Receive L3 packets on such ports.

3. Finally, in order to remove confusions around IS-IS, the pull calls out that IS-IS should be treated as L3 protocols from this leaf's perspective.

More details in [go/forwarding_viable_expected_behavior](http://goto.google.com/forwarding_viable_expected_behavior)

### DIFF
--- a/release/models/interfaces/openconfig-if-sdn-ext.yang
+++ b/release/models/interfaces/openconfig-if-sdn-ext.yang
@@ -47,9 +47,8 @@ module openconfig-if-sdn-ext {
         This is used by an external programming entity to disable an interface
         (usually part of an aggregate) for the purposes of forwarding
         traffic. This allows a logical aggregate to continue to be
-        used with partial capacity, for example.  Note that setting
-        `forwarding-viable = false` is not equivalent to
-        administratively disabling the interface.
+        used with partial capacity. Setting `forwarding-viable = false` is not 
+        equivalent to administratively disabling the interface.
         Some rules to follow when a port/bundle is set for 
         Forwarding-viable=False:
           1. Given that L2 protocols like LACP should be allowed to be 
--- a/release/models/interfaces/openconfig-if-sdn-ext.yang
+++ b/release/models/interfaces/openconfig-if-sdn-ext.yang
@@ -52,13 +52,46 @@ module openconfig-if-sdn-ext {
         administratively disabling the interface -- in particular, the
         interface is expected to participate in L2 protocols such as
         LLDP or LACP even if it blocked from forwarding traffic.
-        Some rules to follow when a port/bundle is set for Forwarding-viable=False,
-          1. Given that L2 protocols like LACP should be allowed to be transmitted and received on such ports, a device mustn't consider such ports in the calculation for minimum number of links required for an aggregate bundle when such a configuration exists on the box. Hence, the device musn't disable the bundle because the minimum number of member links requirement is not met for the bundle.
-          2. While it is required that L2 protocols like LLDP, LACP, P4RT must be allowed for transmit and receive on such ports/bundles, CLNS protocol is an exception and must be treated as per the requirement for L3 protocols below.
-          3. L3 packets, control-plane as well as data-plane must be received but not transmitted on such ports/bundles. By doing so, expectation is that the other end of the connection ceases to use its respective ports for transmit and receive when either the dead-timer or the hold-down timer of its L3 protocol expires. 
-          4. An implementation should follow rule#3 even when the subject port/bundle is the last resort of communication for the device.";
+        Some rules to follow when a port/bundle is set for 
+        Forwarding-viable=False:
+          1. Given that L2 protocols like LACP should be allowed to be 
+          transmitted and received on such ports, a device mustn't consider 
+          such ports in the calculation for minimum number of links required 
+          for an aggregate bundle when such a configuration exists on the box.
+          Hence, the device musn't disable the bundle because the minimum
+          number of member links requirement is not met for the bundle.
+          2. While it is required that L2 protocols like LLDP, LACP, P4RT must
+          be allowed for transmit and receive on such ports/bundles, CLNS
+          protocol is an exception and must be treated as per the requirement
+          for L3 protocols below.
+          3. L3 packets, control-plane as well as data-plane must be received
+          but not transmitted on such ports/bundles. By doing so, expectation
+          is that the other end of the connection ceases to use its respective
+          ports for transmit and receive when either the dead-timer or the
+          hold-down timer of its L3 protocol expires. 
+          4. An implementation should follow rule#3 even when the subject
+          port/bundle is the last resort of communication for the device.";
     }
   }
+
+You, 46 min
+how did u do that?
+
+Rahul Unnikrishnan, 45 min
+manually
+i can see the column width in text editor
+
+You, 45 min
+ok, ... 🙂
+what is the expected width
+?
+
+Rahul Unnikrishnan, 44 min
+i've seen Dan Ameme changing my python script width to <80 columns.
+so i did that here too
+also the max width previously in the file was 79. so setting it to <79 or less seemed safe
+
+
 
   augment "/oc-if:interfaces/oc-if:interface/oc-if:config" {
     description
--- a/release/models/interfaces/openconfig-if-sdn-ext.yang
+++ b/release/models/interfaces/openconfig-if-sdn-ext.yang
@@ -21,7 +21,13 @@ module openconfig-if-sdn-ext {
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
-  oc-ext:openconfig-version "0.1.0";
+  oc-ext:openconfig-version "0.2.0";
+
+  revision 2024-02-21 {
+    description
+      "Initial revision.";
+    reference "0.2.0";
+  }
 
   revision 2021-03-30 {
     description
--- a/release/models/interfaces/openconfig-if-sdn-ext.yang
+++ b/release/models/interfaces/openconfig-if-sdn-ext.yang
@@ -51,30 +51,25 @@ module openconfig-if-sdn-ext {
         equivalent to administratively disabling the interface.
         Some rules to follow when an interface or aggregate interface is set for 
         Forwarding-viable=False:
-          1. When forwarding-viable is FALSE `/interfaces/interface/state/oper-status` should 
-          reflect the L2 status, which when UP should be treated as such for 
+          1. When forwarding-viable is FALSE `/interfaces/interface/state/oper-status` 
+          should reflect the L2 status, which when UP should be treated as such for 
           aggregate interface `min-links` checks.
 
           2. L2 protocols like LLDP, LACP must be allowed for transmit and 
           receive on such ports/bundles, IS-IS PDUs should be handled as per the 
           requirements for L3 packets below.
 
-          3. L3 packets, control-plane as well as data-plane must be received
-          but not transmitted on such ports/bundles. Since the port is receiving 
+          3. For L3 packets, control-plane as well as data-plane packets must be 
+          received but not transmitted on such ports/bundles. Since the port is receiving 
           L3 packets, inflight data-plane traffic will continue to be delivered to its 
-          destination post FIB lookup. Also, received control-plane traffic should be 
-          handled normally. Given that the router is not allowed to transmit L3 control-plane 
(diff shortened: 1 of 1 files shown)


release/models/interfaces/openconfig-if-sdn-ext.yang
```diff
@@ -51,7 +51,26 @@ module openconfig-if-sdn-ext {
...
+          1. Given that L2 protocols like LACP should be allowed to be 
+          transmitted and received on such ports, a device mustn't consider 
+          such ports in the calculation for minimum number of links required 
+          for an aggregate bundle when such a configuration exists on the box.
+          Hence, the device musn't disable the bundle because the minimum
+          number of member links requirement is not met for the bundle.
+          2. While it is required that L2 protocols like LLDP, LACP must
+          be allowed for transmit and receive on such ports/bundles, CLNS
+          protocol is an exception and must be treated as per the requirement
+          for L3 protocols below.
```
LimeHat, line 66: I'd argue that CLNS _is_ a layer-3 protocol.

But to keep things simple and straightforward, maybe we can just state that IS-IS is considered a L3 protocol for the purposes of this definition?

sachendras: Done

release/models/interfaces/openconfig-if-sdn-ext.yang
```diff
@@ -51,7 +51,25 @@ module openconfig-if-sdn-ext {
...
+        Forwarding-viable=False:
+          1. Given that L2 protocols like LACP should be allowed to be 
+          transmitted and received on such ports, a device mustn't consider 
+          such ports in the calculation for minimum number of links required 
+          for an aggregate bundle when such a configuration exists on the box.
+          Hence, the device musn't disable the bundle because the minimum
+          number of member links requirement is not met for the bundle.
+          2. L2 protocols like LLDP, LACP must be allowed for transmit and 
+          receive on such ports/bundles, IS-IS should be handled as per the 
+          requirements for L3 protocols below.
```
dplore, line 65: This can be removed, it is redundant to text in lines 53-54.

sachendras: I actually updated the section above. Want to keep the rules section clean.

release/models/interfaces/openconfig-if-sdn-ext.yang
```diff
@@ -47,11 +47,26 @@ module openconfig-if-sdn-ext {
...
+        Forwarding-viable=False:
+          1. Given that L2 protocols like LACP should be allowed to be 
+          transmitted and received on such ports, a device mustn't consider 
+          such ports in the calculation for minimum number of links required 
+          for an aggregate bundle when such a configuration exists on the box.
+          Hence, the device mustn't disable the bundle because the minimum
+          number of member links requirement is not met for the bundle.
+          2. L2 protocols like LLDP, LACP must be allowed for transmit and 
+          receive on such ports/bundles, IS-IS should be handled as per the 
+          requirements for L3 protocols below.
```
rszarecki, line 62:
```suggestion
          requirements for L3 packets below.
```

release/models/interfaces/openconfig-if-sdn-ext.yang
```diff
@@ -47,11 +47,26 @@ module openconfig-if-sdn-ext {
...
+        Forwarding-viable=False:
+          1. Given that L2 protocols like LACP should be allowed to be 
+          transmitted and received on such ports, a device mustn't consider 
+          such ports in the calculation for minimum number of links required 
+          for an aggregate bundle when such a configuration exists on the box.
+          Hence, the device mustn't disable the bundle because the minimum
+          number of member links requirement is not met for the bundle.
+          2. L2 protocols like LLDP, LACP must be allowed for transmit and 
+          receive on such ports/bundles, IS-IS should be handled as per the 
+          requirements for L3 packets below.
```
rszarecki, line 62:
```suggestion
          receive on such ports/bundles, IS-IS PDUs should be handled as per the 
          requirements for L3 packets below.
```

release/models/interfaces/openconfig-if-sdn-ext.yang
```diff
@@ -47,11 +47,26 @@ module openconfig-if-sdn-ext {
...
+          Hence, the device mustn't disable the bundle because the minimum
+          number of member links requirement is not met for the bundle.
+          2. L2 protocols like LLDP, LACP must be allowed for transmit and 
+          receive on such ports/bundles, IS-IS PDUs should be handled as per the 
+          requirements for L3 packets below.
+          3. L3 packets, control-plane as well as data-plane must be received
+          but not transmitted on such ports/bundles. By doing so, expectation
+          is that the other end of the connection ceases to use its respective
+          ports for transmit and receive when either the dead-timer or the
+          hold-down timer of its L3 protocol expires.
```
dplore, line 67: We should define what the expectation is for received L3 packets.  I would guess the device is supposed to forward L3 packets, right?    If yes, then stating something like the following might make this more clear?

4. Received L3 packets should be evaluated by the data plane as usual.  For example, when a dataplane packet is received, a lookup should be performed and the packet forwarded as usual.  If an L3 control plane packet is received, it should be processed as usual.

sachendras: updated as per the internal doc, ptal.

release/models/interfaces/openconfig-if-sdn-ext.yang
```diff
@@ -47,11 +47,34 @@ module openconfig-if-sdn-ext {
...
         (usually part of an aggregate) for the purposes of forwarding
         traffic. This allows a logical aggregate to continue to be
-        used with partial capacity, for example.  Note that setting
-        `forwarding-viable = false` is not equivalent to
-        administratively disabling the interface -- in particular, the
-        interface is expected to participate in L2 protocols such as
-        LLDP or LACP even if it blocked from forwarding traffic.";
+        used with partial capacity. Setting `forwarding-viable = false` is not 
+        equivalent to administratively disabling the interface.
+        Some rules to follow when a port/bundle is set for
```
dplore, line 52: Here and throughout I suggest using the OpenConfig terms interface and aggregate : 

```suggestion
        Some rules to follow when an interface or aggregate interface is set for 
```

release/models/interfaces/openconfig-if-sdn-ext.yang
```diff
@@ -47,11 +47,34 @@ module openconfig-if-sdn-ext {
...
+        used with partial capacity. Setting `forwarding-viable = false` is not 
+        equivalent to administratively disabling the interface.
+        Some rules to follow when a port/bundle is set for 
+        Forwarding-viable=False:
+          1. Given that L2 protocols like LACP should be allowed to be 
+          transmitted and received on such ports, a device mustn't consider 
+          such ports in the calculation for minimum number of links required 
+          for an aggregate bundle when such a configuration exists on the box.
+          Hence, the device mustn't disable the bundle because the minimum
+          number of member links requirement is not met for the bundle.
```
dplore, line 59: would it be simpler to say 

```
When forwarding-viable is FALSE the interface oper-state should still reflect the L2 status, which when UP and should be treated as such for aggregate interface `min-links` checks.
```

sachendras: Done

release/models/interfaces/openconfig-if-sdn-ext.yang
```diff
@@ -47,11 +47,31 @@ module openconfig-if-sdn-ext {
...
+          destination post FIB lookup. Also, received control-plane traffic should be 
+          handled normally. Given that the router is not allowed to transmit L3 control-plane 
+          packets on such a port with its FV set to false, if this is a singleton 
+          port, OR if all member interfaces in a bundle are ever set to be FV=False, 
+          expectations are that the remote end of the connection will have its L3 protocol's 
+          dead-interval OR hold-down timer expired taking down the adjacency/peering on 
+          that connection.
+
+          4. An implementation should follow rule#3 even when the subject
+          port/bundle is the last resort of communication for the device.";
```
dplore, line 74:
```suggestion
          4. An implementation should follow rule #3 even when the subject
          interface is the last resort of communication for the device.";
```

release/models/interfaces/openconfig-if-sdn-ext.yang
nachikethas: Hi Sach, this statement appears tailored for a protocol like IS-IS. The peer does not _have to_ stop sending traffic to this device on the non-viable ports based on a timeout even if all the trunk members are not-viable. So, can we modify the statement to talk only about DUT and not enforce expectations on the peer? Specifically,
1. Stop L3 transmit for both control and data-plane packets.
2. L3 forwarding should work as usual on Rx without any timeout considerations. The peer can choose to continue sending this packets here indefinitely.

If you prefer, you can call out IS-IS as an example, where the timeout will happen.

sachendras: @nachikethas I have updated rule#3 to include other use cases, ptal

release/models/interfaces/openconfig-if-sdn-ext.yang
```diff
@@ -47,11 +47,34 @@ module openconfig-if-sdn-ext {
...
+          dead-interval or Hold-down timer of L3 protocols like IS-IS/BGP on the 
+          peer router may expire taking down the adjacency or peering on that 
+          connection. However, the peer may still continue to use the bundle for 
+          forwarding traffic towards the subject router if the peer's forwarding 
+          table is programmed by an external programming entity. In this scenario, 
+          the subject router will continue to recieve traffic indefinitely on 
+          such a bundle.
+
+          4. An implementation should follow rule #3 even when the subject
+          interface is the last resort of communication for the device.";
```
dplore, line 77:
```suggestion
        used with partial capacity. Setting `forwarding-viable = false` is not
        equivalent to administratively disabling the interface.
        Some rules to follow when an interface or aggregate interface is set for
        Forwarding-viable=False:
          1. Aggregate interface '/interfaces/interface/aggregation/state/min-links'
           checks should be evaluated based on
          `/interfaces/interface/state/oper-status`.  'min-links' should not be
          affected by the use of forwarding viable.

          2. L2 protocols like LLDP and LACP must be processed normally on
          transmit and receive on such ports/bundles.  IS-IS PDUs should be
          handled as per the requirements for L3 packets below.

          3. L3 packets must not be transmitted on the interface.  
          
          4. Received L3 packets must be processed normally.  Received data-plane
          traffic will continue to forwarded to its destination post FIB lookup.
          Received control-plane traffic must also be processed normally. 
          
          5. It is possible that the dead-interval or hold-down timer of L3
          protocols like IS-IS/BGP on the peer router may expire taking down the
          adjacency or peering on that connection. However, the peer may still
          continue to transmit packets which are received by the local device.  
          These received packet should continue to be processed normally as
          per rule #4 above.

          For example, if the peer's forwarding table is programmed using gRIBI
          by an external controller, the local device will continue to receive
          packets.

          6. An implementation should follow rule #3 even when the subject
          interface on the local device is the last resort of communication for a
          given destination.  For example, the only nexthop for a destination is
          an aggregate interface which has all member interfaces set to
          forwarding-viable = false.  In this scenario all L3 packets for that
          destination will be dropped.";
```

dplore (approved):
The text makes sense to me as is.  I've suggested some formatting and style changes which I hope makes it easier to read.  PTAL.

dplore (approved):
LGTM!  @nachikethas can you please give a final review?

nachikethas (approved):

PR #1038, opened 2024-01-26
Add `gre` container under next-hops aft entry state. Add `src-ip`, `dst-ip` and `ttl` under `gre` aft entry state for telemetry.

romeyod:
* (M) aft/openconfig-aft-ethernet.yang
* (M) aft/openconfig-aft-ipv4.yang
* (M) aft/openconfig-aft-ipv6.yang
* (M) aft/openconfig-aft-mpls.yang
* (M) aft/openconfig-aft-pf.yang
* (M) aft/openconfig-aft-state-synced.yang
* (M) aft/openconfig-aft.yang

Change Scope
  - Add `gre` container under next-hops aft entry state. 
  - Add `src-ip`, `dst-ip` and `ttl` under `gre` aft entry state for telemetry.

Platform Implementations
  - Arista [documentation](https://www.arista.com/en/um-eos/eos-nexthop-groups#xx1147395)

Relevant pyang output:
```
  +--rw network-instances
        +--ro afts
        |  +--ro next-hops
        |     +--ro next-hop* [index]
        |        +--ro index            -> ../state/index
        |        +--ro state
        |        |  +--ro index?                       uint64
        |        |  +--ro programmed-index?            uint64
        |        |  +--ro ip-address?                  oc-inet:ip-address
        |        |  +--ro mac-address?                 oc-yang:mac-address
        |        |  +--ro pop-top-label?               boolean
        |        |  +--ro pushed-mpls-label-stack*     oc-mplst:mpls-label
        |        |  +--ro encapsulate-header?          oc-aftt:encapsulation-header-type
        |        |  +--ro decapsulate-header?          oc-aftt:encapsulation-header-type
        |        |  +--ro origin-protocol?             identityref
        |        |  +--ro lsp-name?                    string
        |        |  +--ro counters
        |        |  |  +--ro packets-forwarded?   oc-yang:counter64
        |        |  |  +--ro octets-forwarded?    oc-yang:counter64
        |        |  +--ro vni-label?                   oc-evpn-types:evi-id
        |        |  +--ro tunnel-src-ip-address?       oc-inet:ip-address
        |        |  +--ro oc-aftni:network-instance?   oc-ni:network-instance-ref
        |        +--ro ip-in-ip
        |        |  +--ro state
        |        |     +--ro src-ip?   oc-inet:ip-address
        |        |     +--ro dst-ip?   oc-inet:ip-address
        |        +--ro gre
        |        |  +--ro state
        |        |     +--ro src-ip?   oc-inet:ip-address
        |        |     +--ro dst-ip?   oc-inet:ip-address
        |        |     +--ro ttl?      uint8
        |        +--ro interface-ref
        |           +--ro state
        |              +--ro interface?      -> /oc-if:interfaces/interface/name
        |              +--ro subinterface?   -> ../interface]/subinterfaces/subinterface/index
```

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -225,7 +225,9 @@ submodule openconfig-aft-common {
           description
             "When specified, the packet has an GRE
             (Generic Routing Encapsulation)header applied to
-            it before forwarding to the specified next-hop.";
+            it before forwarding to the specified next-hop.
+            encapsulate-header leaf should be set to GRE for this
+            to apply";
 
           container state {
             config false;
@@ -412,10 +414,7 @@ submodule openconfig-aft-common {
       "GRE encapsulation applied on a next-hop";
 
     leaf src-ip {
-      type union {
-        type oc-inet:ip-address;
-        type string;
-      }
+      type oc-inet:ip-address;
       description
         "Source IP address to use for the encapsulated packet. This
          may be expressed as either an IP address or reference
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -418,7 +418,7 @@ submodule openconfig-aft-common {
       type oc-inet:ip-address;
       description
         "The source IP address for the GRE encapsulation may be expressed
-        using this leaf (src-ip) or if may be derived from 
+        using this leaf (src-ip) or if may be derived from
         '../../interface-ref/state/subinterface'";
     }
 
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,15 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2024-01-26" {
+    description
+      "Add gre container under next-hops aft entry state.
+       Add src-ip, dst-ip and ttl under gre aft entry state
+       for telemetry.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description
@@ -213,6 +221,20 @@ submodule openconfig-aft-common {
           }
         }
 
+        container gre {
+          description
+            "When specified, the packet has an GRE
+            (Generic Routing Encapsulation)header applied to
+            it before forwarding to the specified next-hop.";
+
+          container state {
+            config false;
+            description
+              "State parameters relating to GRE encapsulation.";
+            uses aft-common-entry-nexthop-gre-state;
+          }
+        }
+
         uses oc-if:interface-ref-state;
       }
     }
@@ -382,6 +404,37 @@ submodule openconfig-aft-common {
       type oc-inet:ip-address;
       description
         "Destination IP address to use for the encapsulated packet.";
+    }
+  }
+
+  grouping aft-common-entry-nexthop-gre-state {
+    description
+      "GRE encapsulation applied on a next-hop";
+
+    leaf src-ip {
+      type union {
+        type oc-inet:ip-address;
+        type string;
+      }
+      description
+        "Source IP address to use for the encapsulated packet. This
+         may be expressed as either an IP address or reference
+         to the name of an interface.";
+    }
+
+    leaf dst-ip {
+      type oc-inet:ip-address;
+      description
+        "Destination IP address to use for the encapsulated packet.";
+    }
+
+    leaf ttl {
+      type uint8;
+      description
+        "This leaf reflects the configured/default TTL value that is used in the
+         outer header during packet encapsulation. When this leaf is not set,
+         the TTL value of the inner packet is copied over as the outer packet's
+         TTL value during encapsulation.";
     }
   }
 
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -29,7 +29,8 @@ submodule openconfig-aft-common {
     description
       "Add gre container under next-hops aft entry state.
        Add src-ip, dst-ip and ttl under gre aft entry state
-       for telemetry.";
+       for telemetry. Updated description for
+       tunnel-src-ip-address properties under next-hops.";
     reference "2.5.0";
   }
 
@@ -293,7 +294,7 @@ submodule openconfig-aft-common {
     leaf tunnel-src-ip-address {
       type oc-inet:ip-address;
       description
-        "Where applicable this represents the tunnel source ip address.
+        "Where applicable this represents the vxlan tunnel source ip address.
         For VXLAN this represents the source VTEP ip address";
     }
   }
@@ -411,7 +412,7 @@ submodule openconfig-aft-common {
 
   grouping aft-common-entry-nexthop-gre-state {
     description
-      "GRE encapsulation applied on a next-hop";
+      "GRE encapsulation applied on a IPv4 and IPv6 next-hop.";
 
     leaf src-ip {
       type oc-inet:ip-address;
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -417,9 +417,9 @@ submodule openconfig-aft-common {
     leaf src-ip {
       type oc-inet:ip-address;
       description
-        "Source IP address to use for the encapsulated packet. This
-         may be expressed as either an IP address or reference
-         to the name of an interface.";
+        "The source IP address for the GRE encapsulation may be expressed
+        using this leaf (src-ip) or if may be derived from 
+        '../../interface-ref/state/subinterface'";
     }
 
     leaf dst-ip {
--- a/release/models/aft/openconfig-aft-ethernet.yang
+++ b/release/models/aft/openconfig-aft-ethernet.yang
@@ -20,7 +20,15 @@ submodule openconfig-aft-ethernet {
     "Submodule containing definitions of groupings for the abstract
     forwarding tables for Ethernet.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2024-01-26" {
+    description
+      "Add gre container under next-hops aft entry state.
+      Add src-ip, dst-ip and ttl under gre aft entry state
+      for telemetry.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description
--- a/release/models/aft/openconfig-aft-ipv4.yang
+++ b/release/models/aft/openconfig-aft-ipv4.yang
@@ -20,7 +20,15 @@ submodule openconfig-aft-ipv4 {
     "Submodule containing definitions of groupings for the abstract
     forwarding tables for IPv4.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2024-01-26" {
+    description
+      "Add gre container under next-hops aft entry state.
+      Add src-ip, dst-ip and ttl under gre aft entry state
+      for telemetry.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description
--- a/release/models/aft/openconfig-aft-ipv6.yang
+++ b/release/models/aft/openconfig-aft-ipv6.yang
@@ -20,7 +20,15 @@ submodule openconfig-aft-ipv6 {
     "Submodule containing definitions of groupings for the abstract
     forwarding tables for IPv6.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2024-01-26" {
+    description
+      "Add gre container under next-hops aft entry state.
+      Add src-ip, dst-ip and ttl under gre aft entry state
+      for telemetry.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description
--- a/release/models/aft/openconfig-aft-mpls.yang
+++ b/release/models/aft/openconfig-aft-mpls.yang
@@ -21,7 +21,15 @@ submodule openconfig-aft-mpls {
     "Submodule containing definitions of groupings for the abstract
     forwarding table for MPLS label forwarding.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2024-01-26" {
+    description
+      "Add gre container under next-hops aft entry state.
+      Add src-ip, dst-ip and ttl under gre aft entry state
+      for telemetry.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description
--- a/release/models/aft/openconfig-aft-pf.yang
+++ b/release/models/aft/openconfig-aft-pf.yang
@@ -28,7 +28,15 @@ submodule openconfig-aft-pf {
     fields other than the destination address that is used in
     other forwarding tables.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2024-01-26" {
+    description
+      "Add gre container under next-hops aft entry state.
+      Add src-ip, dst-ip and ttl under gre aft entry state
+      for telemetry.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description
--- a/release/models/aft/openconfig-aft-state-synced.yang
+++ b/release/models/aft/openconfig-aft-state-synced.yang
@@ -16,7 +16,15 @@ submodule openconfig-aft-state-synced {
     "Submodule containing definitions of groupings for the state
     synced signals corresponding to various abstract forwarding tables.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2024-01-26" {
+    description
+      "Add gre container under next-hops aft entry state.
+      Add src-ip, dst-ip and ttl under gre aft entry state
+      for telemetry.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description
--- a/release/models/aft/openconfig-aft.yang
+++ b/release/models/aft/openconfig-aft.yang
@@ -42,7 +42,15 @@ module openconfig-aft {
     is referred to as an Abstract Forwarding Table (AFT), rather than
     the FIB.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.5.0";
+
+  revision "2024-01-26" {
+    description
+      "Add gre container under next-hops aft entry state.
+      Add src-ip, dst-ip and ttl under gre aft entry state
+      for telemetry.";
+    reference "2.5.0";
+  }
 
   revision "2023-09-26" {
     description


release/models/aft/openconfig-aft-common.yang
```diff
@@ -385,6 +407,37 @@ submodule openconfig-aft-common {
...
+  grouping aft-common-entry-nexthop-gre-state {
+    description
+      "GRE encapsulation applied on a next-hop";
+
+    leaf src-ip {
+      type union {
+        type oc-inet:ip-address;
+        type string;
+      }
+      description
```
robshakir, line 419: Given that this is state, can we not just simplify this down to only using the IP address? The implementation can resolve what address is being used if needs be.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -385,6 +407,37 @@ submodule openconfig-aft-common {
...
   }
 
+  grouping aft-common-entry-nexthop-gre-state {
+    description
+      "GRE encapsulation applied on a next-hop";
+
+    leaf src-ip {
+      type union {
+        type oc-inet:ip-address;
+        type string;
```
robshakir, line 417: If this is going to reference an L3 interface (I don't think it should, see other comment) then this needs to be an `interface-ref` which cannot be in this union, so will need a restructure to be able to reference an interface, subinterface tuple.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -385,6 +407,37 @@ submodule openconfig-aft-common {
     }
   }
 
+  grouping aft-common-entry-nexthop-gre-state {
+    description
+      "GRE encapsulation applied on a next-hop";
+
+    leaf src-ip {
```
robshakir, line 414: To be clear here, GRE over both v6 and v4 are intended to be covered by this same modelling? That sounds good to me, but I just wanted to verify.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -385,6 +407,37 @@ submodule openconfig-aft-common {
...
+        type string;
+      }
+      description
+        "Source IP address to use for the encapsulated packet. This
+         may be expressed as either an IP address or reference
+         to the name of an interface.";
+    }
+
+    leaf dst-ip {
+      type oc-inet:ip-address;
```
robshakir, line 426: Is it possible to set this to be a link-local address for a v6 destination such that we should use the `-zoned` type?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -213,6 +221,20 @@ submodule openconfig-aft-common {
           }
         }
 
+        container gre {
```
robshakir, line 224: Should we add a description here that `encapsulate-header` should be set to `GRE` for this to apply?

robshakir (commented):
Added a few comments, but this LGTM.

One thing that I'll note is that we're going to end up with a few different places that `src-ip` and `dst-ip` leaves are duplicated for X-over-IP encaps. I'm not worried about that since I think its logical to think "this is GRE over IP and the dest of the IP header is X", but I just wanted to highlight this here.

This also LGTM for how it would be represented in gRIBI.

robshakir:
Doesn't need to be handled in this PR -- but in reviewing this I had a thought. If we were doing MPLS over GRE then we'll specify something in the MPLS-related leaves and then subsequently in the GRE container. `encapsulate-header` today is a scalar leaf, so what would its value be if we were doing MPLS over GRE?

There are a few options:

 * make `encapsulate-header` just the outer-most header.
 * deprecate `encapsulate-header` and imply the headers from the specified leaves -- this one probably doesn't work because we then don't know what order they are applied in.
 * have containers that are specific to combinatorial encapsulations -- e.g., `mpls-over-gre` which has the MPLS and GRE leaves in it. This denormalises the model a bit and duplicates definitions.
 * have a new `encapsulate-stack` `leaf-list` which specifies the order that the encapsulations are being applied -- i.e., I can specify `["GRE", "MPLS"]` which indicates that it's GRE+IPv[46] and then MPLS.

WDYT?

krishna-juniper:
https://github.com/openconfig/public/blob/master/release/models/aft/openconfig-aft-common.yang#L272

Description of tunnel-src-ip-address is confusing. It can be used to populate gre tunnel src address.
can we update the description to make it specific to EVPN/VXLAN?

release/models/aft/openconfig-aft-common.yang
romeyod: Makes sense. I will make it just IP address.

release/models/aft/openconfig-aft-common.yang
romeyod: Yes, this is correct. Do you want me to update the `description` with that as well?

release/models/aft/openconfig-aft-common.yang
romeyod: Maybe I didnt understand your question correctly. Please feel free to elaborate.

My understanding of the GRE encapsulation is that source and destination IP address for the GRE header and the network designer needs to make sure the destination is reachable in the underlay network. The GRE encapsulation itself does not care about the `link-local` address as long as the routing table resolves the GRE packet correctly.

release/models/aft/openconfig-aft-common.yang
romeyod: Makes sense. Will add that to the description

romeyod:
> Doesn't need to be handled in this PR -- but in reviewing this I had a thought. If we were doing MPLS over GRE then we'll specify something in the MPLS-related leaves and then subsequently in the GRE container. `encapsulate-header` today is a scalar leaf, so what would its value be if we were doing MPLS over GRE?
> 
> There are a few options:
> 
> * make `encapsulate-header` just the outer-most header.
> * deprecate `encapsulate-header` and imply the headers from the specified leaves -- this one probably doesn't work because we then don't know what order they are applied in.
> * have containers that are specific to combinatorial encapsulations -- e.g., `mpls-over-gre` which has the MPLS and GRE leaves in it. This denormalises the model a bit and duplicates definitions.
> * have a new `encapsulate-stack` `leaf-list` which specifies the order that the encapsulations are being applied -- i.e., I can specify `["GRE", "MPLS"]` which indicates that it's GRE+IPv[46] and then MPLS.
> 
> WDYT?

IMHO a new `encapsulate-stack` `leaf-list` seems like the best path forward at first glance.

romeyod:
> https://github.com/openconfig/public/blob/master/release/models/aft/openconfig-aft-common.yang#L272
> 
> Description of tunnel-src-ip-address is confusing. It can be used to populate gre tunnel src address. can we update the description to make it specific to EVPN/VXLAN?

Makes sense to update the documentation. Not sure if we should couple that with this PR. Maybe file a issue and @robshakir and company to start a separate discussion.

release/models/aft/openconfig-aft-common.yang
earies: > are leafrefs supported in unions now? I didnt find any existing examples of it.

All OC models are 1.0 based - this is dependent on a entire model-set shift to 1.1+

romeyod: > > are leafrefs supported in unions now? I didnt find any existing examples of it.
> 
> All OC models are 1.0 based - this is dependent on a entire model-set shift to 1.1+

Thanks for confirming this. As per @robshakir feedback (`Given that this is state, can we not just simplify this down to only using the IP address? The implementation can resolve what address is being used if needs be.`) I have made this IP address only

release/models/aft/openconfig-aft-common.yang
romeyod: description update. Let me know if anything else needs to be updated as well.

wenovus:
Merging into main branch for pyangbind check fix.

release/models/aft/openconfig-aft-common.yang
robshakir: agreed -- so if we wanted to set the destination IP to be something that was link local -- because the routing table _could_ resolve this, then we might want to be able to say `fe80::1%eth0` here, if this is the case then we should use [`oc-inet:ip-address-zoned`](https://github.com/openconfig/public/blob/master/release/models/types/openconfig-inet-types.yang#L296) which allows us to specify that.

i'm torn a little on whether this is needed, because we also have `interface-ref` which specifies the egress interface, so in this case, it might be redundant information.

robshakir:
w.r.t 

> > https://github.com/openconfig/public/blob/master/release/models/aft/openconfig-aft-common.yang#L272
> > Description of tunnel-src-ip-address is confusing. It can be used to populate gre tunnel src address. can we update the description to make it specific to EVPN/VXLAN?
> 
> Makes sense to update the documentation. Not sure if we should couple that with this PR. Maybe file a issue and @robshakir and company to start a separate discussion.

let's continue this discussion here: https://github.com/openconfig/public/issues/1040

release/models/aft/openconfig-aft-common.yang
romeyod: @robshakir my understanding from your other comment ( `Given that this is state, can we not just simplify this down to only using the IP address? The implementation can resolve what address is being used if needs be.` ) was we want to not have `interface-ref` and only `ip-address`

Let me know if it still makes sense to keep it as `ip-address` or make it [oc-inet:ip-address-zoned](https://github.com/openconfig/public/blob/master/release/models/types/openconfig-inet-types.yang#L296)

romeyod: @robshakir I agree that with respect to `dst-ip` this case is redundant given `interface-ref`

I am not very familiar with `link-local` addresses but I would expect this `dst-ip` to be the destination IP address programmed in the GRE header as the state. [This](https://github.com/openconfig/public/pull/1038#discussion_r1467855575) comment of yours from earlier implies the implementation can resolve what address is being used if needs be.

But I might be missing context given by limited knowledge of  `link-local` addresses

dplore: If we clearly specify that use of a link-local address also requires specifying an interface-ref then I think it's redundant to using zoned-ip's. (which to save others from looking this up, a zoned IP merely concatenates an ip-address and interface name together).

release/models/aft/openconfig-aft-common.yang
dplore: Please do update the description since the question was asked.  It's worth it since the question came up here.  :)

dplore:
> > Doesn't need to be handled in this PR -- but in reviewing this I had a thought. If we were doing MPLS over GRE then we'll specify something in the MPLS-related leaves and then subsequently in the GRE container. `encapsulate-header` today is a scalar leaf, so what would its value be if we were doing MPLS over GRE?
> > There are a few options:
> > 
> > * make `encapsulate-header` just the outer-most header.
> > * deprecate `encapsulate-header` and imply the headers from the specified leaves -- this one probably doesn't work because we then don't know what order they are applied in.
> > * have containers that are specific to combinatorial encapsulations -- e.g., `mpls-over-gre` which has the MPLS and GRE leaves in it. This denormalises the model a bit and duplicates definitions.
> > * have a new `encapsulate-stack` `leaf-list` which specifies the order that the encapsulations are being applied -- i.e., I can specify `["GRE", "MPLS"]` which indicates that it's GRE+IPv[46] and then MPLS.
> > 
> > WDYT?
> 
> IMHO a new `encapsulate-stack` `leaf-list` seems like the best path forward at first glance.

Agreed, ideally we have some option that allows exposing any number of nested encapsulations.  I like this option because it keeps the afts table flat.  

Do you mind refactoring this PR to implement leaf-list encapsulate-stack? ie:

```
  +--rw network-instances
        +--ro afts
        |  +--ro next-hops
        |     +--ro next-hop* [index]
        |        x--ro ip-in-ip     (deprecated)
        |        |  x--ro state     (deprecated)
        |        |     x--ro src-ip?   oc-inet:ip-address  (deprecated)
        |        |     x--ro dst-ip?   oc-inet:ip-address  (deprecated)
        **|        +--ro encapsulate
        |        |  +--ro state
        |        |     +--ro stack*?   oc-types:encap-types
        |        |     +--ro src-ip?   oc-inet:ip-address
        |        |     +--ro dst-ip?   oc-inet:ip-address
        |        |     +--ro ttl?      uint8
        |        +--ro interface-ref
        |           +--ro state
        |              +--ro interface?      -> /oc-if:interfaces/interface/name
        |              +--ro subinterface?   -> ../interface]/subinterfaces/subinterface/index**
 ```

**Comment by romeyod** [comment]
> > > Doesn't need to be handled in this PR -- but in reviewing this I had a thought. If we were doing MPLS over GRE then we'll specify something in the MPLS-related leaves and then subsequently in the GRE container. `encapsulate-header` today is a scalar leaf, so what would its value be if we were doing MPLS over GRE?
> > > There are a few options:
> > > 
> > > * make `encapsulate-header` just the outer-most header.
> > > * deprecate `encapsulate-header` and imply the headers from the specified leaves -- this one probably doesn't work because we then don't know what order they are applied in.
> > > * have containers that are specific to combinatorial encapsulations -- e.g., `mpls-over-gre` which has the MPLS and GRE leaves in it. This denormalises the model a bit and duplicates definitions.
> > > * have a new `encapsulate-stack` `leaf-list` which specifies the order that the encapsulations are being applied -- i.e., I can specify `["GRE", "MPLS"]` which indicates that it's GRE+IPv[46] and then MPLS.
> > > 
> > > WDYT?
> > 
> > 
> > IMHO a new `encapsulate-stack` `leaf-list` seems like the best path forward at first glance.
> 
> Agreed, ideally we have some option that allows exposing any number of nested encapsulations. I like this option because it keeps the afts table flat.
> 
> Do you mind refactoring this PR to implement leaf-list for the list of encaps? Something like:
> 
> ```
> module: openconfig-network-instance
>   +--rw network-instances
>      +--rw network-instance* [name]
>         +--ro afts
>         |  +--ro next-hops
>         |     +--ro next-hop* [index]
>         |        x--ro ip-in-ip     (deprecated)
>         |        |  x--ro state     (deprecated)
>         |        |     x--ro src-ip?   oc-inet:ip-address  (deprecated)
>         |        |     x--ro dst-ip?   oc-inet:ip-address  (deprecated)
>         |        +--ro encapsulate
>         |        |  +--ro state
>         |        |     +--ro stack*?   oc-types:encap-types   
>         |        |     +--ro src-ip?   oc-inet:ip-address
>         |        |     +--ro dst-ip?   oc-inet:ip-address
>         |        |     +--ro ttl?      uint8
>         |        +--ro interface-ref
>         |           +--ro state
>         |              +--ro interface?      -> /oc-if:interfaces/interface/name
>         |              +--ro subinterface?   -> ../interface]/subinterfaces/subinterface/index**
> ```

This will need to a LIST of the entire encap header state since `src-ip`, `dst-ip`, `ttl`, `tos` (in the future) could be part unique per encap, right? Something like:

```
|        +--ro encapsulate
|        |  +--ro state
|        |     +--ro stack*?   oc-types:encap-header  
|        |      |     +--ro encap?   oc-types:encap-type
|        |      |     +--ro src-ip?   oc-inet:ip-address
|        |      |     +--ro dst-ip?   oc-inet:ip-address
|        |      |     +--ro ttl?         uint8
|        |      |     +--ro tos?       uint8
```

@robshakir @dplore let me know your thoughts on this.

Not sure if we can make each of the "encap-header" definition different per encap type i.e. `aft-common-entry-nexthop-gre-state` `aft-common-entry-nexthop-ip-in-ip-state` etc and they are still be part of "encap-header" List. This will make the documentation per "encap-header" definition easier as well.

**Comment by dplore** [comment]
How about  defining containers for each encap with it's own set of custom state leaves?    Since encaps can be quite different, I am not sure it is feasible to have a list of encaps, but rather models containers for each encap with typed leaves.

```
  +--rw network-instances
        +--ro afts
        |  +--ro next-hops
        |     +--ro next-hop* [index]
        |        +--ro ip-in-ip   
        |        |  +--ro state   
        |        |     +--ro <state leafs added here>
        |        +--ro <my other encap container>
        |        |  +--ro state   
        |        |     +--ro <my other encap state leafs added here>
```

**Comment by romeyod** [comment]
> How about defining containers for each encap with it's own set of custom state leaves? Since encaps can be quite different, I am not sure it is feasible to have a list of encaps, but rather models containers for each encap with typed leaves.
> 
> ```
>   +--rw network-instances
>      +--rw network-instance* [name]
>         +--ro afts
>         |  +--ro next-hops
>         |     +--ro next-hop* [index]
>         |        +--ro ip-in-ip   
>         |        |  +--ro state   
>         |        |     +--ro <state leafs added here>
>         |        +--ro <my other encap container>
>         |        |  +--ro state   
>         |        |     +--ro <my other encap state leafs added here>
> ```

@dplore Maybe I am being a bit dense here. Can you please simplify your suggestion for me?

The original intent of the PR was to add gre container under next-hops aft entry state. Add src-ip, dst-ip and ttl under gre aft entry state for telemetry. This was being achieved by using `uses` but is there a different way to typed leaves each of them?

The only reason the List/stack on encap can into the discussion was as a suggestion/food for thought by @robshakir and we were planning to take it up later.

I know this PR has dragged on for a while so I am just TLDRing it for any other reader/party that comes across it :)

File: release/models/aft/openconfig-aft-common.yang (hunk 4 above)
Comment (dplore) [reply release/models/aft/openconfig-aft-common.yang:425]: So in short, I think a zoned IP is not needed (redundant) because we have 
`network-instances/network-instance/afts/next-hops/next-hop/gre/state/dst-ip`

and these (preexisting paths):
`network-instances/network-instance/afts/next-hops/next-hop/interface-ref/state/interface`
`network-instances/network-instance/afts/next-hops/next-hop/interface-ref/state/subinterface`

**Comment by dplore** [comment]
> > https://github.com/openconfig/public/blob/master/release/models/aft/openconfig-aft-common.yang#L272
> > Description of tunnel-src-ip-address is confusing. It can be used to populate gre tunnel src address. can we update the description to make it specific to EVPN/VXLAN?
> 
> Makes sense to update the documentation. Not sure if we should couple that with this PR. Maybe file a issue and @robshakir and company to start a separate discussion.

Please add this description change.  It should be in this PR because this PR introduces a `src-ip` leaf which could be confused with the existing `tunnel-src-ip-address`.  Clarifying that `tunnel-src-ip-address` is not used for gre in the formal description is good.  This will cause the git history to also point to this PR for context.

**Comment by dplore** [comment]
> > How about defining containers for each encap with it's own set of custom state leaves? Since encaps can be quite different, I am not sure it is feasible to have a list of encaps, but rather models containers for each encap with typed leaves.
> > ```
> >   +--rw network-instances
> >      +--rw network-instance* [name]
> >         +--ro afts
> >         |  +--ro next-hops
> >         |     +--ro next-hop* [index]
> >         |        +--ro ip-in-ip   
> >         |        |  +--ro state   
> >         |        |     +--ro <state leafs added here>
> >         |        +--ro <my other encap container>
> >         |        |  +--ro state   
> >         |        |     +--ro <my other encap state leafs added here>
> > ```
> 
> @dplore Maybe I am being a bit dense here. Can you please simplify your suggestion for me?
> 
> The original intent of the PR was to add gre container under next-hops aft entry state. Add src-ip, dst-ip and ttl under gre aft entry state for telemetry. This was being achieved by using `uses` but is there a different way to typed leaves each of them?
> 
> The only reason the List/stack on encap can into the discussion was as a suggestion/food for thought by @robshakir and we were planning to take it up later.
> 
> I know this PR has dragged on for a while so I am just TLDRing it for any other reader/party that comes across it :)

Ah yes, let's consider this comment thread out of scope for this PR.  The current PR scope for GRE only is ok as is and does not need to consider refactoring how any/all tunnel encapsulations are modeled in OC.

**Review by dplore — CHANGES_REQUESTED** [review]
Please add a description change for `tunnel-src-ip-address` as per the issue:

https://github.com/openconfig/public/blob/master/release/models/aft/openconfig-aft-common.yang#L272

I think it is appropriate for this description change to be in this PR because this PR introduces a new `src-ip` leaf which could be confused with the existing `tunnel-src-ip-address`. Clarifying that `tunnel-src-ip-address` is only used for vxlan in the formal description is good.  

The git history will point to this PR and our mention of #1040 for additional context.

Other than this, LGTM.  This will be reviewed at the Apr 30, 2024 OC Operators meeting

**Comment by romeyod** [comment]
> Please add a description change for `tunnel-src-ip-address` as per the issue:
> 
> https://github.com/openconfig/public/blob/master/release/models/aft/openconfig-aft-common.yang#L272
> 
> I think it is appropriate for this description change to be in this PR because this PR introduces a new `src-ip` leaf which could be confused with the existing `tunnel-src-ip-address`. Clarifying that `tunnel-src-ip-address` is only used for vxlan in the formal description is good.
> 
> The git history will point to this PR and our mention of #1040 for additional context.
> 
> Other than this, LGTM. This will be reviewed at the Apr 30, 2024 OC Operators meeting

Thansk @dplore. I have updated the PR with the requested changes. Let me know if any other action is needed.

File: release/models/aft/openconfig-aft-common.yang [hunk 6]
```diff
@@ -385,6 +410,34 @@ submodule openconfig-aft-common {
...
+  grouping aft-common-entry-nexthop-gre-state {
+    description
+      "GRE encapsulation applied on a IPv4 and IPv6 next-hop.";
+
+    leaf src-ip {
+      type oc-inet:ip-address;
+      description
+        "Source IP address to use for the encapsulated packet. This
+         may be expressed as either an IP address or reference
+         to the name of an interface.";
```
dplore, line 422: Suggestion to make it more clear how to use reference to an interface.
```suggestion
        "The source IP address for the GRE encapsulation may be expressed
        using this leaf (src-ip) or if may be derived from 
        '../../interface-ref/state/subinterface'";
```

dplore (commented):
One more comment on the description for src-ip and I think it's ready to review at the April 30th OC operators meeting.

romeyod:
> One more comment on the description for src-ip and I think it's ready to review at the April 30th OC operators meeting.

Thanks @dplore
Committed the suggestion. I am assuming the `Changes requested` blocking item will be resolved after the OC operators meeting

dplore (approved):
Thanks, I don't see any additional changes needed.  This is ready to merge, pending Apr 29, 2024 OC Operators review.

dplore:
Reviewed in Apr 30 OC operators meeting without objection.

Issue #5, opened 2016-01-24
Incorrect default value for leaf unnumbered  in openconfig-interfaces.yang


qzhengyy:
Hi,

  leaf unnumbered (in the grouping subinterfaces-config) is referring to "/interfaces/subinterfaces/index", which is referring to 
"/interfaces/subinterfaces/config/index", which is a leaf of type uint32.

leaf unnumbered has a default value "false", and it doesn't fit to uint32

Just wondering, do have any tools to validate the YANG models you published?

Thanks
Qing

aashaikh:
Yep, this is a bug -- thanks for the catch.  The unnumbered leaf was a boolean leaf in a previous version and the default value didn't get updated when we changed the type.  this will be fixed in the next revision of the interfaces model, which we expect to push soon.

As for validation, we rely on pyang for compiling/validating the models, and actual operator usage in NMSes to further test and evaluate the models.  I expect that the unnumbered leaf hasn't been used much in generating configs so far so wasn't caught before.

PR #1041, opened 2024-01-30
Add rules for controller-card power-off

sachendras:
Add rules for controller-card power-off. Internal doc go/controller_card_gnmi_shutdown for more context.


release/models/platform/openconfig-platform.yang
```diff
@@ -531,6 +539,36 @@ module openconfig-platform {
...
+      type oc-types:timeticks64;
+      units "nanoseconds";
+      description
+        "For components that have redundant roles (e.g. two
+        supervisors in a device, one as primary the other as
+        secondary), this reports the time of the last change of
+        the component's role. The value is the timestamp in
+        nanoseconds relative to the Unix Epoch (Jan 1, 1970 00:00:00 UTC).";
+
+    }
```
dplore, line 570: This should be removed.  It looks like a cut-paste error?  It is already defined at https://github.com/openconfig/public/pull/1041/files#diff-f787893a2471f7ec1785fa010e4981f339dc68a144cffb802df9afcdf02a30b5R582

release/models/platform/openconfig-platform.yang
```diff
@@ -531,6 +539,36 @@ module openconfig-platform {
...
+    leaf last-poweroff-time {
+      type oc-types:timeticks64;
+      units "nanoseconds";
+      description
+        "This reports the time of the last shutdown of the component. 
+         The value is the timestamp in nanoseconds relative to the Unix Epoch (Jan 1, 1970 00:00:00 UTC). 
+         If a system is rebooted and after the configuration is loaded, say the controller-card is identified 
+         to be shutdown in configuration, the card should be powered-off and 'last-power-off-time' 
+         leaf should be recorded.";
+    }
```
dplore, line 558: keep line length to < 70 characters per https://github.com/openconfig/public/blob/master/doc/openconfig_style_guide.md#line-length

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1180,59 @@ module openconfig-platform {
 
       container config {
         description
-          "Configuration data for controller card components. Note that disabling
-          power to the primary supervisor should be rejected, and the operator is
-          required to perform a switchover first.";
+          "Configuration data for controller card components.
```
dplore, line 1183: correct line line to < 70 characters

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1180,59 @@ module openconfig-platform {
...
+          1. Only a CONTROLLER_CARD in state redundant-role SECONDARY will honor 
+             config/power-admin-state set to POWER_DISABLED. If the controller-card's 'redundant-role' 
+             is 'PRIMARY', and it's 'config/power-admin-state' is set to 'POWER_DISABLED`, the NOS should allow 
+             the configuration. However, the change should take effect only on the next reboot. 
+             Following scenarios in context
+          
+             Scenario#1: Let's say controller0 is PRIMARY and controller1 is SECONDARY and controller1 
+             receives an operator driven change of config/power-admin-state = POWER_DISABLED, then 
+             controller0 will power-off controller1 immediately.  The leaf state/power-admin-state should also 
+             be POWER_DISABLED. The boolean flag state/is-powered-off should also show up as TRUE.
```
dplore, line 1196:
```suggestion
             be POWER_DISABLED. ```

File: release/models/platform/openconfig-platform.yang [hunk 5]
```diff
@@ -1142,9 +1180,59 @@ module openconfig-platform {
...
+             the configuration. However, the change should take effect only on the next reboot. 
+             Following scenarios in context
+          
+             Scenario#1: Let's say controller0 is PRIMARY and controller1 is SECONDARY and controller1 
+             receives an operator driven change of config/power-admin-state = POWER_DISABLED, then 
+             controller0 will power-off controller1 immediately.  The leaf state/power-admin-state should also 
+             be POWER_DISABLED. The boolean flag state/is-powered-off should also show up as TRUE. 
+             The state/last-poweroff-reason/trigger should show as USER_INITIATED. Also the NOS can update the 
+             leaf /state/last-poweroff-reason/details. The leaf state/last-poweroff-time should record the time 
+             when the card was powered-off. Following paths:
```
dplore, line 1199:
```suggestion
             The 'state/last-poweroff-reason/trigger' should show as USER_INITIATED.  The NOS may optionally
             update '/state/last-poweroff-reason/details'.   'state/last-poweroff-time' should record the time 
             when the card was powered-off.  For example:
```

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1180,59 @@ module openconfig-platform {
...
+
+             /components/component[name=controller0]/state/redundant-role, PRIMARY
+             /components/component[name=controller1]/state/redundant-role, SECONDARY
+             /components/component[name=controller0]/controller-card/config/power-admin-state, POWER_ENABLED
+             /components/component[name=controller1]/controller-card/state/power-admin-state, POWER_DISABLED
+             /components/component[name=controller1]/controller-card/state/is-powered-off, TRUE
+             /components/component[name=controller1]/state/last-poweroff-reason/trigger, USER_INITIATED
+             /components/component[name=controller1]/state/last-poweroff-reason/details, "User initiated Shutdown"
+             /components/component[name=controller0]/state/last-poweroff-time, 1706590714681765937
```
dplore, line 1208:
```suggestion
             /components/component[name=controller0]/state/redundant-role, PRIMARY
             /components/component[name=controller0]/controller-card/config/power-admin-state, POWER_ENABLED
             /components/component[name=controller0]/controller-card/state/power-admin-state, POWER_ENABLED
             /components/component[name=controller1]/state/redundant-role, SECONDARY
             /components/component[name=controller1]/controller-card/config/power-admin-state, POWER_DISABLED
             /components/component[name=controller1]/controller-card/state/power-admin-state, POWER_DISABLED
             /components/component[name=controller1]/state/last-poweroff-reason/trigger, USER_INITIATED
             /components/component[name=controller1]/state/last-poweroff-reason/details, "User initiated Shutdown"
             /components/component[name=controller1]/state/last-poweroff-time, 1706590714681765937
```

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1180,59 @@ module openconfig-platform {
...
+
+             Scenario#2: If controller0 is PRIMARY and controller1 is SECONDARY and if controller0 is set to 
+             config/power-admin-state = POWER_DISABLED by an operator, then controller0 will stay powered-on until the next reboot. 
+             The leaf state/power-admin-state should show as POWER_DIABLED. The boolean flag however, state/is-powered-off 
+             should show as FALSE. The state/last-poweroff-reason/trigger should show as USER_INITIATED. Also the NOS can 
+             update the leaf /state/last-poweroff-reason/details. The leaf state/last-poweroff-time should record the time 
+             when the card powers-off post the next reboot. Following paths
```
dplore, line 1215:
```suggestion
             Scenario#2: First, controller0 is PRIMARY and controller1 is SECONDARY.  Next,
             controller0 'config/power-admin-state' = POWER_DISABLED is set.  controller0
             will stay powered-on until controller0 'state/redundant-role' changes to 
             SECONDARY or on the next reboot. 
             
             controller0 'config/power-admin-state' will show POWER_DISABLED.  controller0
             'state/power-admin-state' will show POWER_ENABLED.   This indicates the
             difference between the intended state of POWER_DISABLED and the current
             operational state of POWER_ENABLED.  
             
             When controller0 'state/power-admin-state' changes to POWER_ENABLED', then 
             'state/last-poweroff-reason/trigger' should show as USER_INITIATED.  The system
             may update '/state/last-poweroff-reason/details' with an informational message.
             'state/last-poweroff-time' should record the time when the controller0 powers-off. 
             
             For example:
```

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1180,59 @@ module openconfig-platform {
...
+
+             /components/component[name=controller0]/state/redundant-role, PRIMARY
+             /components/component[name=controller1]/state/redundant-role, SECONDARY
+             /components/component[name=controller0]/controller-card/config/power-admin-state, POWER_DISABLED
+             /components/component[name=controller1]/controller-card/state/power-admin-state, POWER_ENABLED
+             /components/component[name=controller0]/controller-card/state/is-powered-off, FALSE
+             /components/component[name=controller0]/state/last-poweroff-reason/trigger, USER_INITIATED
+             /components/component[name=controller0]/state/last-poweroff-reason/details, "User initiated Shutdown"
+             /components/component[name=controller0]/state/last-poweroff-time, 1706590714681765937
```
dplore, line 1224:
```suggestion
             At the time POWER_DISABLED is set:
             
             /components/component[name=controller0]/state/redundant-role, PRIMARY
             /components/component[name=controller0]/controller-card/config/power-admin-state, POWER_DISABLED
             /components/component[name=controller0]/controller-card/state/power-admin-state, POWER_ENABLED
             /components/component[name=controller1]/state/redundant-role, SECONDARY
             /components/component[name=controller1]/controller-card/state/power-admin-state, POWER_ENABLED
             
             After controller0 transitions to redundant-role SECONDARY:

             /components/component[name=controller0]/state/redundant-role, SECONDARY
             /components/component[name=controller0]/controller-card/config/power-admin-state, POWER_DISABLED
             /components/component[name=controller0]/controller-card/state/power-admin-state, POWER_DISABLED
             /components/component[name=controller0]/state/last-poweroff-reason/trigger, USER_INITIATED
             /components/component[name=controller0]/state/last-poweroff-reason/details, "User initiated Shutdown"
             /components/component[name=controller0]/state/last-poweroff-time, 1706590714681765937
             /components/component[name=controller1]/state/redundant-role, PRIMARY
             /components/component[name=controller1]/controller-card/state/power-admin-state, POWER_ENABLED

```

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1180,59 @@ module openconfig-platform {
...
+
+          2. A controller-card which is in state/redundant-role=SECONDARY and is config/power-admin-state=POWER_DISABLED should 
+            remain powered off, even after a reboot. It is possible that after a reboot, both the controller cards are powered-on. 
+            However as soon as the configuration is loaded the NOS should power-off the subject controller-card.
```
dplore, line 1228: Suggesting we keep with 'system' as this is regularly used in openconfig models.  
```suggestion
            However as soon as the configuration is loaded the system should power-off the subject controller-card.
```

dplore (changes requested):

release/models/platform/openconfig-platform-common.yang
```diff
@@ -232,9 +240,16 @@ submodule openconfig-platform-common {
...
-        "Enable or disable power to the component";
+        "This leaf would reflect the configuration in config/power-admin-state leaf. 
+         In other words, this leaf reflects what the NOS expects the state of the component should be.";
+    }
+
+    leaf is-powered-off {
+      type oc-platform-types:boolean;
+      default TRUE;
+      description
+        "This leaf reports the state of a component if it is powered-off or powered-on.";
```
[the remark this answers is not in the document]
sachendras: Well, the idea for this is that `/components/component/<component-type-here>/state/power-admin-state` shows what is configured in `/components/component/<component-type-here>/config/power-admin-state`. Whereas, `/components/component/controller-card/state/is-powered-off` will show the actual state. As we have seen that it is not necessary for the `config/power-admin-state` to always be same as the actual state of the component.

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1180,59 @@ module openconfig-platform {
...
+
+          3. A CONTROLLER_CARD in redundant-role SECONDARY and state/power-admin-state = POWER_DISABLED cannot transition to redundant-role PRIMARY. 
+             If the PRIMARY CONTROLLER_CARD goes down, the device will be offline.
+
+          4. When a controller card boots up and loads it’s configuration, power-admin-state may be set to POWER_DISABLED. The CONTROLLER_CARD should 
+             then power off and never enter into any controller card primary/secondary election process. This also means that an implementation  
+             shouldn't start a controller card election process until the configuration is loaded and consumed.";
```
LimeHat, line 1235: > This also means that an implementation  
             shouldn't start a controller card election process until the configuration is loaded and consumed.";

This assumes/enforces a particular NOS/HW architecture, which isn't valid for all vendors/implementations.

In some implementations, primary/standby selection happens earlier in the boot process, before the full config is loaded (and, often, only a single card is responsible for managing the running-config).

release/models/platform/openconfig-platform.yang
sachendras: How about now? Wondering if it is okay to keep the paths as is?

release/models/platform/openconfig-platform-common.yang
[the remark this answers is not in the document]
dplore: `state/power-admin-state` shows the current operational state.
`config/power-admin-state` shows the configuration value.  

These can return different values.

Throughout openconfig, if one wants to see what is configured, one can `gnmi.Get` the `config` container/leafs.   If you want to see what is operationally active, one should `gnmi.Subscribe` to a `state` container/leafs

release/models/platform/openconfig-platform.yang
sachendras: can you please ping me? sachendras@google

release/models/platform/openconfig-platform-common.yang
```diff
@@ -20,6 +20,14 @@ submodule openconfig-platform-common {
...
 
+  oc-ext:openconfig-version "0.25.0";
+
+  revision "2024-01-30" {
+    description
+      "Added leaf is-powered-off to component-power-management group";
+    reference "0.25.0";
+  }
+
   oc-ext:openconfig-version "0.24.0";
```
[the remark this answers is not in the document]
sachendras: done

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1180,59 @@ module openconfig-platform {
...
+             /components/component[name=controller0]/controller-card/config/power-admin-state, POWER_ENABLED
+             /components/component[name=controller1]/controller-card/state/power-admin-state, POWER_DISABLED
+             /components/component[name=controller1]/controller-card/state/is-powered-off, TRUE
+             /components/component[name=controller1]/state/last-poweroff-reason/trigger, USER_INITIATED
+             /components/component[name=controller1]/state/last-poweroff-reason/details, "User initiated Shutdown"
+             /components/component[name=controller0]/state/last-poweroff-time, 1706590714681765937
+
+             Scenario#2: If controller0 is PRIMARY and controller1 is SECONDARY and if controller0 is set to 
+             config/power-admin-state = POWER_DISABLED by an operator, then controller0 will stay powered-on until the next reboot. 
+             The leaf state/power-admin-state should show as POWER_DIABLED. The boolean flag however, state/is-powered-off
```
iitrgaurav: Please correct the typo "POWER_DIABLED"

release/models/platform/openconfig-platform.yang
LimeHat: ack, will reach out tomorrow

release/models/platform/openconfig-platform.yang
dplore: Hi, I don't see any changes in the line length?  Did you push your change?

release/models/platform/openconfig-platform.yang
dplore: Friendly reminder for this

release/models/platform/openconfig-platform.yang
dplore: Thanks for the update!

release/models/platform/openconfig-platform-common.yang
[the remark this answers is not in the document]
sachendras: Updated, ptal!

release/models/platform/openconfig-platform.yang
sachendras: done!

release/models/platform/openconfig-platform.yang
sachendras: Hopefully now it shows up?

release/models/platform/openconfig-platform.yang
sachendras: Corrected!

release/models/platform/openconfig-platform.yang
sachendras: ack!

release/models/platform/openconfig-platform.yang
sachendras: Done!

release/models/platform/openconfig-platform.yang
```diff
@@ -531,6 +539,36 @@ module openconfig-platform {
...
+    }
+
+    leaf last-poweroff-time {
+      type oc-types:timeticks64;
+      units "nanoseconds";
+      description
+        "This reports the time of the last shutdown of the component. 
+         The value is the timestamp in nanoseconds relative to the Unix Epoch (Jan 1, 1970 00:00:00 UTC). 
+         If a system is rebooted and after the configuration is loaded, say the controller-card is identified 
+         to be shutdown in configuration, the card should be powered-off and 'last-power-off-time'
```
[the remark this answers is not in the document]
sachendras: Updated, ptal!

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,19 +1180,102 @@ module openconfig-platform {
 
       container config {
         description
-          "Configuration data for controller card components. Note that disabling
-          power to the primary supervisor should be rejected, and the operator is
-          required to perform a switchover first.";
+          "Configuration data for controller card components. 
+          Note that when a controller-card's
```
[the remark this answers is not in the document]
sachendras: Moved to a MD doc, ptal!

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
+## Objective:
+
+This document introduces the concept of persistent shutdown of controller-cards in switches/routers with dual controller-cards, outlines rules to determine its use and proposes new telemetry counters to support the feature.
+
+
+## Possible use cases for the gNMI approach:
```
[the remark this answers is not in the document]
dplore: The openconfig yang is not gnmi specific (ie: many people use netconf for example)

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+
+1. Operator suspects faulty system behavior due to faulty hardware and wants to take it offline so they can revisit the issue in a maintenance window. But while this happens, the operator also wants to ensure that the card stays disabled across reboots so that the subject card does not take a primary role after reboot.
+2. It is possible that the faulty controller card is in a boot loop and the operator finds it best to disable the card to prevent undesired state. Expectations are also that the state persists even post a system reboot.
```
dplore, line 11: This reads as an example of item 1.  Indent this?  Maybe add another example where a CPU is overheating if left powered on for more than a few minutes.

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+3. Operator feels it important to keep the card shutdown to prevent unexpected outcomes post the physical swap of the faulty/alerting card so the operator can online the card in a controlled environment post replacement.
+4. Following an RMA, if a new issue arises or the original problem remains, operators may instruct field techs to leave the replacement card installed. This allows for replacement upon arrival or controlled troubleshooting. In these cases, the ability to control the card's operational state is advantageous.
+
+
+## Proposed Rules
```
dplore, line 16:
```suggestion
## CONTROLLER_CARD power and redundancy requirements 
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+
+Idea here is that the subject controller-card can be disabled using the OC-Path [/components/component/controller-card/config/power-admin-state](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-controller-card-config-power-admin-state) by setting its value to POWER\_DISABLED. Following rules:
```
dplore, line 18:
```suggestion
Electrical power for a CONTROLLER_CARD can be turned off using the OC-Path [/components/component/controller-card/config/power-admin-state](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-controller-card-config-power-admin-state) and setting its value to POWER\_DISABLED.  The following rules regarding CONTROLLER_CARD redundancy and changes to `power-admin-state` should be followed by the device:
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+
+1. Only a CONTROLLER_CARD in state redundant-roleSECONDARY will honor config/power-admin-state set toPOWER_DISABLED. If the controller-card's'redundant-role' is 'PRIMARY', and it's 'config/power-admin-state' is set to 'POWER_DISABLED', the NOS must allow the configuration. However, the change must take effect only on the next reboot. Following scenarios in context
```
dplore, line 22:
```suggestion
1. Only a CONTROLLER_CARD in `state/redundant-role` 'SECONDARY' will honor a change in `config/power-admin-state` to POWER_DISABLED. If the controller-card's `redundant-role` is 'PRIMARY', and its `config/power-admin-state` is set to 'POWER_DISABLED', the NOS must allow the configuration. However, the `state/power-admin-state` should remain as 'POWER_ENABLED'.  A change in `state/power-admin-state` must take effect only on the next reboot or if the CONTROLLER_CARD becomes 'SECONDARY'. Examples of scenarios include:
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+             
+`Scenario#1:`
```
dplore, line 24:
```suggestion
### Scenario 1 - Power off a secondary CONTROLLER_CARD
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+/components/component[name=controller0]/controller-card/state/power-admin-state, POWER_ENABLED
+/components/component[name=controller1]/state/redundant-role, SECONDARY
+/components/component[name=controller1]/controller-card/config/power-admin-state, POWER_DISABLED
+/components/component[name=controller1]/controller-card/state/power-admin-state, POWER_DISABLED
+/components/component[name=controller1]/state/last-poweroff-reason/trigger, USER_INITIATED
+/components/component[name=controller1]/state/last-poweroff-reason/details, "User initiated Shutdown"
+/components/component[name=controller1]/state/last-poweroff-time, 1706590714681765937
+```
+
+`Scenario#2:`
```
dplore, line 40:
```suggestion
### Scenario 2 - Power off primary CONTROLLER_CARD
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+    
+If controller-card0 is PRIMARY and controller-card1 is SECONDARY and if controller-card0 is set to config/power-admin-state = POWER_DISABLED by an operator, then controller-card0 will stay powered-on until the next reboot. The leaf state/power-admin-state must show as POWER_ENABLED. The state/last-poweroff-reason/trigger should show as USER_INITIATED. Also the NOS can update the leaf /state/last-poweroff-reason/details. The leaf state/last-poweroff-time should record the time when the card powers-off post the next reboot. For example:
```
dplore, line 42:
```suggestion
If controller-card0 is PRIMARY and controller-card1 is SECONDARY and if controller-card0 is set to config/power-admin-state = POWER_DISABLED by an operator, then controller-card0 will stay powered-on until the next reboot. `state/power-admin-state` must show as POWER_ENABLED. If a reboot of the PRIMARY CONTROLLER_CARD occurs,  `state/last-poweroff-time` should record the time when the card powers-off, `state/last-poweroff-reason/trigger` should show as USER_INITIATED and `/state/last-poweroff-reason/details` may be updated. For example: 
```

dplore: Trigger and and details should only be updated after a power-off occurs (after a reboot in this scenario)

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+2. Since the operation relies on configuration, it is possible that the failure scenario may kick in before the configuration takes effect post a reboot.
+
+    **Response:**
+
+
+	Response here is the same as "1" above.
+
+
+
+3. Both controller-cards are functional and the the secondary controller card is attempted to be shutdown
```
dplore, line 95:
```suggestion
3. Both controller-cards are functional and the the secondary controller card is attempted to be powered off.
```

Here and throughout, use a consistent term such as 'power off' and 'power on' for referring to power-admin-state.  Shutdown might be confused with some other configuration.

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+## Concerns and possible failure scenarios
+
+
+
+1. If a PRIMARY card malfunctions and ends up in a bootloop, would this approach help?
+
+    **Response:**
+
+
+    In this scenario the expectations are that the implementation takes steps to initiate a controller-card switchover operation. Therefore, the standby controller card takes over the PRIMARY role and the system stabilizes allowing for gRPC connections to be established. In this situation if the operator pushes a gNMI configuration to shutdown the SECONDARY card, the PRIMARY card must be able to power-off the SECONDARY controller-card. In this scenario the implementation (depending on their architecture) may also initiate a shutdown of the faulty card from the new PRIMARY card.
```
dplore, line 84:
```suggestion
    In this scenario the expectations are that the implementation takes steps to initiate a controller-card switchover operation. Therefore, the standby controller card takes over the PRIMARY role and the system stabilizes allowing for gRPC connections to be established. In this situation if the operator pushes a configuration to shutdown the SECONDARY card, the PRIMARY card must be able to power-off the SECONDARY controller-card. In this scenario the implementation (depending on their architecture) may also initiate a shutdown of the faulty card from the new PRIMARY card.
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+    Since the lone controller-card is PRIMARY, please follow Rule#1 above.
+
+6. Faulty card is shutdown and then replaced with a working card but device still has the config of `/components/component[controller-card#]/controller-card/config/power-admin-state = POWER_DISABLED`
+
+    **Response:**
+
+
+    This is as per expectations so inserting a new card into the chassis can happen anytime. However, bringing it online happens only in a controlled environment using a gNMI config push operation of `/components/component[controller-card#]/controller-card/config/power-admin-state = POWER_ENABLED`
```
dplore, line 124:
```suggestion
    This is as per expectations so inserting a new card into the chassis can happen anytime. However, bringing it online happens only in a controlled environment using a configuration change to set `/components/component[controller-card#]/controller-card/config/power-admin-state = POWER_ENABLED`
```

release/models/platform/openconfig-platform-common.yang
```diff
@@ -232,9 +238,12 @@ submodule openconfig-platform-common {
       type oc-platform-types:component-power-type;
       default POWER_ENABLED;
       description
-        "Enable or disable power to the component";
+        "When set to POWER_DISABLED, the component should be shut down by removing
+        electrical power.  This is intended to be used to prevent the component
+        from becoming active even after a reboot of the system.  A component may
+        not honor power-admin-state depending on rules defined in the description
+        of the component config container.";
```
dplore, line 245: I suggest if a component does not support this leaf, the device should return an error.

release/models/platform/openconfig-platform-types.yang
```diff
@@ -218,6 +224,28 @@ module openconfig-platform-types {
     uses oc-types:min-max-time;
   }
 
+  grouping component-last-poweroff-reason {
+    description
+      "Common grouping for recording the reason of a component's
+      power-off state";
+
+    leaf trigger {
+      type component-last-poweroff-reason-trigger;
```
dplore, line 233: this type is not defined.  Did you want to do something like:

      type identityref {
        base COMPONENT_POWEROFF_REASON;
      }

dplore (changes requested):

release/models/platform/openconfig-platform-common.yang
sachendras: This basically considers that the implementation does support the leaf. However, if the component is a controller-card and if it happens to be PRIMARY, then it will take the configuration of POWER_DISABLED for "config/power-admin-state" but it will not power-off because it is the PRIMARY. So in this case "state/power-admin-state" will still be "POWER_ENABLED"

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+## Possible use cases for the gNMI approach:
+
+
+
+1. Operator suspects faulty system behavior due to faulty hardware and wants to take it offline so they can revisit the issue in a maintenance window. But while this happens, the operator also wants to ensure that the card stays disabled across reboots so that the subject card does not take a primary role after reboot.
+2. It is possible that the faulty controller card is in a boot loop and the operator finds it best to disable the card to prevent undesired state. Expectations are also that the state persists even post a system reboot.
+3. Operator feels it important to keep the card shutdown to prevent unexpected outcomes post the physical swap of the faulty/alerting card so the operator can online the card in a controlled environment post replacement.
+4. Following an RMA, if a new issue arises or the original problem remains, operators may instruct field techs to leave the replacement card installed. This allows for replacement upon arrival or controlled troubleshooting. In these cases, the ability to control the card's operational state is advantageous.
```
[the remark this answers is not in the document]
sachendras: removed #4

doc/controller-card_poweroff.md
sachendras: Done, ptal!

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+   
+   b. CPU overheating, memory errors or hardware problems with certain components of the PRIMARY card may require the operator to proactively shutdown the card and not let it be online during regular operation until RMA is completed.
+3. Operator feels it important to keep the card shutdown to prevent unexpected outcomes post the physical swap of the faulty/alerting card so the operator can online the card in a controlled environment post replacement.
+
+
+## CONTROLLER_CARD power and redundancy requirements 
+
+Electrical power for a CONTROLLER_CARD can be turned off using the OC-Path [/components/component/controller-card/config/power-admin-state](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-controller-card-config-power-admin-state) and setting its value to POWER\_DISABLED.  The following rules regarding CONTROLLER_CARD redundancy and changes to `power-admin-state` should be followed by the device:
```
dplore, line 19:
```suggestion
Electrical power for a CONTROLLER_CARD can be configured off using the OC-Path [/components/component/controller-card/config/power-admin-state](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-controller-card-config-power-admin-state) and setting its value to POWER\_DISABLED.  The following rules regarding CONTROLLER_CARD redundancy and changes to `power-admin-state` should be followed by the device:
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
+## Controller Card Power Control
+
+This document describes operational use cases, rules and telemetry for power control of `CONTROLLER-CARD` components in network devices using OpenConfig.
+
+
+## Operational use cases for CONTROLLER_CARD power control
+
+1. Operator suspects a network device issue is occurring due to faults on a CONTROLLER_CARD. The operator wants to quickly disable the CONTROLLER_CARD and later, troubleshoot the issue in a maintenance window. The operator also wants to ensure that the card stays disabled even if the device is rebooted. Some example scenarios below:
+   
+   a. It is possible that the faulty controller card is in a boot loop and the operator finds it best to disable the card to prevent undesired state. Expectations are also that the state persists even post a system reboot.
```
dplore, line 10:
```suggestion
   a. It is possible that the faulty controller card is in a boot loop and the operator finds it best to disable the card to prevent undesired state. Expectations are also that the power remains disabled  even after a system reboot.
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+   
+   b. CPU overheating, memory errors or hardware problems with certain components of the PRIMARY card may require the operator to proactively shutdown the card and not let it be online during regular operation until RMA is completed.
```
dplore:
```suggestion
   b. CPU overheating, memory errors or other hardware problems with the PRIMARY card may require the operator to proactively shutdown the card and not let it be online during regular operation until replacement of the hardware is completed.
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+
+1. Only a CONTROLLER_CARD in `state/redundant-role` 'SECONDARY' will honor a change in `config/power-admin-state` to POWER_DISABLED. If the controller-card's `redundant-role` is 'PRIMARY', and its `config/power-admin-state` is set to 'POWER_DISABLED', the NOS must allow the configuration. However, the `state/power-admin-state` should remain as 'POWER_ENABLED'.  A change in `state/power-admin-state` must take effect only on the next reboot or if the CONTROLLER_CARD becomes 'SECONDARY'. Examples of scenarios include:
+             
+### Scenario 1 - Power off a secondary CONTROLLER_CARD
+
+Let's say controller-card0 is PRIMARY and controller-card1 is SECONDARY and controller-card1 receives an operator driven change of config/power-admin-state = POWER_DISABLED, then controller-card0 will power-off controller-card1 immediately.  The leaf state/power-admin-state for controller-card1 must also be POWER_DISABLED. The 'state/last-poweroff-reason/trigger' should show as USER_INITIATED. The NOS may optionally update /state/last-poweroff-reason/details. state/last-poweroff-time should record the time when the card was powered-off. For example:
```
dplore, line 27:
```suggestion
Let's say controller-card0 is PRIMARY and controller-card1 is SECONDARY and controller-card1 receives an operator driven change of config/power-admin-state = POWER_DISABLED, then controller-card0 will power-off controller-card1 immediately.  The leaf state/power-admin-state for controller-card1 must also be POWER_DISABLED. The 'state/last-poweroff-reason/trigger' should show as USER_INITIATED. The NOS may optionally update state/last-poweroff-reason/details. state/last-poweroff-time should record the time when the card was powered-off. For example:
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+
+3. Both controller-cards are functional and the the secondary controller card is attempted to be powered off.
+
+    **Response:**
+
+
+    The shutdown can be initiated by pushing the command [/components/component[controller-card#]/controller-card/config/power-admin-state = POWER\_DISABLED](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-controller-card-config-power-admin-state) to the box. As a result, the secondary controller card whose [/components/component/state/redundant-role](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-state-redundant-role) is SECONDARY
+
+
+    will be shutdown and its [/components/component[controller-card#]/controller-card/state/power-admin-state](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-controller-card-state-power-admin-state) will change to "POWER\_DISABLED". This configuration is expected to stay sticky in the box and when the device reboots, the subject controller-card is expected to remain Powered-off.
```
dplore, line 104:
```suggestion
    will be shutdown and its [/components/component[name="my_secondary_controller-card"]/controller-card/state/power-admin-state](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-controller-card-state-power-admin-state) will change to "POWER\_DISABLED".  If the configuration is saved and when the device reboots, the subject controller-card is expected to remain powered off.
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+    Since the lone controller-card is PRIMARY, please follow Rule#1 above.
+
+6. Faulty card is shutdown and then replaced with a working card but device still has the config of `/components/component[controller-card#]/controller-card/config/power-admin-state = POWER_DISABLED`
+
+    **Response:**
+
+
+    This is as per expectations so inserting a new card into the chassis can happen anytime. However, bringing it online happens only in a controlled environment using a configuration change to set `/components/component[controller-card#]/controller-card/config/power-admin-state = POWER_ENABLED`
```
dplore, line 125:
```suggestion
    The new card should remain powered off.  Power should only be enabled on the new card by a configuration change to set `/components/component[controller-card#]/controller-card/config/power-admin-state = POWER_ENABLED`
```

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,132 @@
...
+
+
+7. Both controller-cards present during the initial boot operation. However, the controller card that is configured as "POWER\_DISABLED" boots up sooner than the other card and takes over the ACTIVE role.
+
+    **Response:**
+
+
+    The controller cards should never take PRIMARY/SECONDARY roles unless the configuration is fully loaded and processed. Hence, once the configuration is loaded, the card configured to be "POWER\_DISABLED" should shutdown. The only exception is when it is the only controller card in the chassis. In which case, the controller card can still load irrespective of its power-admin-state being POWER\_DISABLED. In this situation though, the implementation must send a Syslog message of the severity "Warning" to inform the Operator about the situation.
```
dplore, line 133: This exception doesn't apply to this scenario because we've said `Both controller-cards present during the initial boot operation`

However, we could add another scenario below where:

Two cards are present, secondary is powered off.  The primary card is removed and the system goes offline.  If the device is rebooted (for example via externally removing and reapplying power ), what should happen?

dplore (commented):
Several nits for wording and one new open question .

doc/controller-card_poweroff.md
sachendras: done

doc/controller-card_poweroff.md
sachendras: `This exception doesn't apply to this scenario because we've said Both controller-cards present during the initial boot operation`

Well, not necessarily for all vendors today config load happens before the PRIMARY/SECONDARY election process. Sequence is indeterministic today between config load operation and election process and hence the scenario here is very well a possibility and is orthogonal to the presence of both cards or not. We are calling out this scenario to ensure expectations are clear.

We can definitely add the scenario you mentioned.

dplore: The scenario is defined as two controller cards are present, but the the exception notes is:  `only exception is when it is the only controller card in the chassis. `.  Just a logic issue / should be broken out to a separate scenario IMO.

sachendras: I added Rule#5. Ptal

dplore: LGTM.  The rule reference is a little ambiguous.  There are a couple of things in the document labeled `5.` and it's not clear which one is a rule versus a scenario.  One way to fix this would be to use outline numbered headings like:

1. Item
1.1 Item subheading
2. Another item
2.2 Another subheading

release/models/platform/openconfig-platform-types.yang
```diff
@@ -218,6 +224,31 @@ module openconfig-platform-types {
...
+      power-off state";
+
+    leaf trigger {
+      type identityref {
+        base COMPONENT_POWEROFF_REASON;
+      }
+      type component-last-poweroff-reason-trigger;
+      description        
+        "Records the generic triggers, e.g. user or system
+        initiated the power-off.";
```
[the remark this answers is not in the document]
sachendras: done

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1173,16 @@ module openconfig-platform {
...
+          "Configuration data for controller card components. 
+          A controller-card can be configured for persistent 
+          powered-off mode using the config/power-admin-state leaf.
+          The reference path below defines rules for such a 
+          configuration.";
+
+        reference
+          "Rules for Persistent power-off configuration in 
+          controller-cards:
+          https://github.com/openconfig/public/blob/master/doc/controller-card_poweroff.md";
```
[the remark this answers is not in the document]
sachendras: done!

release/models/platform/openconfig-platform-types.yang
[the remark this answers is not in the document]
sachendras: Done!

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,143 @@
...
+/components/component[name=controller1]/state/redundant-role, PRIMARY
+/components/component[name=controller1]/controller-card/config/power-admin-state, POWER_ENABLED
+/components/component[name=controller1]/controller-card/state/power-admin-state, POWER_ENABLED
+```
+
+2. A controller-card which is in state/redundant-role=SECONDARY and is config/power-admin-state=POWER_DISABLED should remain powered off, even after a reboot. It is possible that after a reboot, both the controller cards are powered-on. However as soon as the configuration is loaded, the system should power-off the subject controller-card.
+
+3. A CONTROLLER_CARD in redundant-role SECONDARY and state/power-admin-state = POWER_DISABLED cannot transition to redundant-role PRIMARY. If the PRIMARY CONTROLLER_CARD goes down, the device will be offline.
+
+4. When a controller card boots up and loads it’s configuration, power-admin-state may be set to POWER_DISABLED. The CONTROLLER_CARD should then power off and never enter into any controller card primary/secondary election process. This also means that an implementation shouldn't start a controller card election process until the configuration is loaded and consumed.
```
LimeHat, line 73: If both controller cards have been set as POWER_DISABLED, does that mean that both cards should be powered off?

Also, there's a slight contradiction with the next rule (# 5). 
>The CONTROLLER_CARD should then power off and never enter into any controller card primary/secondary election process.

(only if another card is detected in the system)

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,143 @@
...
+
+5. If the chassis has a single controller card and it was configured for config/power-admin-state=POWER_DISABLED, as per Rule#1 above, its state/power-admin-state will stay as POWER_ENABLED given that its state/redundant-role=PRIMARY. However during the next reboot, if the chassis continues to have only one controller-card then it must ignore the configuration of config/power-admin-state=POWER_DISABLED and continue to bootup as PRIMARY. In this situation though, the implementation must send a Syslog message of the severity "Warning" to inform the Operator about the situation. Post reboot, the config/power-admin-state=POWER_DISABLED will stay as is and state/power-admin-state will continue to show as POWER_ENABLED. In this case, state/last-poweroff-time, state/last-poweroff-reason/trigger and state/last-poweroff-reason/details must not be updated
```
LimeHat, line 75: This means that the faulty card cannot be disabled if you encounter a failure in communication between the primary and the secondary cards (persistent split brain scenario)

If this is acceptable, I think that should be explicitly mentioned in the text.

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,143 @@
...
+/components/component[name=controller0]/controller-card/state/power-admin-state, POWER_DISABLED
+/components/component[name=controller0]/state/last-poweroff-reason/trigger, USER_INITIATED
+/components/component[name=controller0]/state/last-poweroff-reason/details, "User initiated Shutdown"
+/components/component[name=controller0]/state/last-poweroff-time, 1706590714681765937
+/components/component[name=controller1]/state/redundant-role, PRIMARY
+/components/component[name=controller1]/controller-card/config/power-admin-state, POWER_ENABLED
+/components/component[name=controller1]/controller-card/state/power-admin-state, POWER_ENABLED
+```
+
+2. A controller-card which is in state/redundant-role=SECONDARY and is config/power-admin-state=POWER_DISABLED should remain powered off, even after a reboot. It is possible that after a reboot, both the controller cards are powered-on. However as soon as the configuration is loaded, the system should power-off the subject controller-card.
```
LimeHat, line 69: Perhaps the term "reboot" is used somewhat ambiguously throughout the text, but here it is stated that the card _should_ remain powered off, but may be powered-on temporarily. And at the end of the document, in the scenario # 8, you provide an example where a reboot (through power cycling) must initiate a new boot cycle.

>If the device is rebooted (for example via externally removing and reapplying power ), what should happen?
>    If during reboot the device continues to have a single controller-card, then Rule#5 above should be followed.

LimeHat (changes requested):

doc/controller-card_poweroff.md
dplore: To increase clarity, I recommend replacing 
`It is possible that after a reboot, both the controller cards are powered-on` with 
`During a cold-boot, some implementations may power up both CONTROLLER_CARDS.`

sachendras: Why only cold boot? We are not asking implementations to maintain state even during warm boot. Therefore it is possible that both controller-cards may turn-on simultaneously post a reboot (cold or warm). Expectations are that the implementation loads the configuration on both cards, the card w/ `power-admin-state=POWER_DISABLED` will be powered-off unless it is the only card.

LimeHat: In my view, the ambiguity is that in order to follow the rules that you specified, e.g. the one I quoted:

>If the device is rebooted (for example via externally removing and reapplying power ), what should happen?
If during reboot the device continues to have a single controller-card, then Rule#5 above should be followed.

there must be a way to "reboot" (in your terms) the card that was previously powered-off via the config, and it cannot stay powered off during that time.

But rule 2 says otherwise:
> A controller-card which is in state/redundant-role=SECONDARY and is config/power-admin-state=POWER_DISABLED should remain powered off, even after a reboot.

if the card remains powered off, it will never know if its a single card in the system, or there's another card. So it cannot remain powered off.

doc/controller-card_poweroff.md
LimeHat: I didn't ask to add all hypothetical situations :-)

There are a few challenges that I see, and why I asked about this specifically.

In one of the stated objectives, you say

> Operator suspects a network device issue is occurring due to faults on a CONTROLLER_CARD. The operator wants to quickly disable the CONTROLLER_CARD and later, troubleshoot the issue in a maintenance window. The operator also wants to ensure that the card stays disabled even if the device is rebooted.

So, let's say, an operator has a problem where split brain scenario has occurred - he definitely will want to disable the failed card. But now he cannot truly disable it, if the failed card is not able to see the other card, and thinks that it's primary. Because it follows the rules specified in this document (the only card is not disabled immediately, and post reboot has to check if it's the only card and boot successfully if it is)

Thus, I think it is important to call out that the rules specified here explicitly disallow using the `POWER_DISABLED` configuration to mitigate the split-brain issue if it occurs.

And no, I don't agree that
 > Possibility of such a command to power-off a controller-card using configuration do not change how the implementation should otherwise deal with operational challenges like a Split-brain scenario.

This document
a) enforces a particular architecture of NOS boot process, and active-standby selection (which, in my opinion, should not be standardized by OpenConfig)
b) says that power-off rules cannot be used to mitigate the split brain in certain conditions, where another set of rules w.r.t a single card scenario behavior may produce a completely different result.

doc/controller-card_poweroff.md
LimeHat: The contradiction is in the wording.

> When a controller card boots up and loads it’s configuration, power-admin-state may be set to POWER_DISABLED. The CONTROLLER_CARD should then power off and never enter into any controller card primary/secondary election process.

rule#4 says to load the config, and if the POWER_DISABLED is detected, it _*never*_ enters the election process.
But then rule#5 redeclares that; the controller card will do the election just fine if the other card is not detected. So it's not never.

doc/controller-card_poweroff.md
sachendras: As per the offline discussions, @LimeHat seems like the requirement is understood. How an implementation implements the change is left on them.

doc/controller-card_poweroff.md
sachendras: `So, let's say, an operator has a problem where split brain scenario has occurred - he definitely will want to disable the failed card. But now he cannot truly disable it, if the failed card is not able to see the other card, and thinks that it's primary.`

***Response:** 
- At first, we are also assuming that the operator is able to use push config to the box in a split brain scenario.

`Thus, I think it is important to call out that the rules specified here explicitly disallow using the POWER_DISABLED configuration to mitigate the split-brain issue if it occurs.`

***Response***
I strongly feel that is implicit in the rule already, a Primary controller-card will not be turned off during run time (in a split-brain scenario both cards tend to be PRIMARY and hence none can be turned-off if somehow the operator manages to configure one). Also, if it deems itself as the lone controller-card during reboot, it will continue with the boot process (therefore in the split-brain scenario both cards will continue to boot. However the NOS is expected to have proper safety in place to prevent such a situation at boot). Calling out specific operational failure example in the rules will be repetitive and unproductive. But, this is covered in the `Concerns and possible failure scenarios` section below.

a) enforces a particular architecture of NOS boot process, and active-standby selection (which, in my opinion, should not be standardized by OpenConfig)

**Response**
For this process to work as expected, it is important that the implementation learns what is in the config prior to starting the election.

`b) says that power-off rules cannot be used to mitigate the split brain in certain conditions, where another set of rules w.r.t a single card scenario behavior may produce a completely different result.`

**Response**
Right, that is the expectation and I dont see any ambiguity there. Attached flowchart must help as well. Reiterating the rules in context
- Primary controller card cannot be powered-off, only the secondary-controller card can be powered-off during runtime.  During Split-brain situation when both cards are Primary at run-time, and somehow if the operator can configure one card as disabled, it will stay up because it deems itself as PRIMARY.
- During the boot process, if the controller-card is the only card, then the POWER_DISABLED configuration must be ignored and the boot process should continue. Now if during the boot process there is a split brain situation then the implementation MUST have proper controls in place for protection. Introduction of config/power-admin-state = POWER_DISABLED/POWER_ENABLED is orthogonal to such an expectation for protection from the implementation.

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,154 @@
...
+
+5. If the chassis has a single controller card and it was configured for config/power-admin-state=POWER_DISABLED, as per Rule#1 above, its state/power-admin-state will stay as POWER_ENABLED given that its state/redundant-role=PRIMARY. However during the next warm reboot, if the chassis continues to have only one controller-card then it must ignore the configuration of config/power-admin-state=POWER_DISABLED and continue to bootup as PRIMARY. In this situation though, the implementation must send a Syslog message of the severity "Warning" to inform the Operator about the situation. Post reboot, the config/power-admin-state=POWER_DISABLED will stay as is and state/power-admin-state will continue to show as POWER_ENABLED. In this case, if it was a warm reboot then the following must not be updated, state/last-poweroff-time, state/last-poweroff-reason/trigger and state/last-poweroff-reason/details. The same leaves must be updated for cold reboots.
```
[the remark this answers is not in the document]
sachendras: Removed bullet#6. I dont think we need that.

doc/controller-card_poweroff.md
```diff
@@ -0,0 +1,154 @@
...
+
+
+8. Say we have a situation whereby, A dual controller card device is functional, secondary is powered off by changing power-admin-state to POWER\_DISABLED. Later, the PRIMARY card is removed and the system goes offline. If the device is rebooted (for example via externally removing and reapplying power ), what should happen?
+
+    **Response:**
+   
+    If during reboot the device continues to have a single controller-card, then Rule#5 above should be followed.
+
+9. According to Rule#5, a controller-card must ignore the configuration of `config/power-admin-state=POWER_DISABLED` and must continue with its boot process if it is the only controller-card in the chassis. This also means that, under split-brain situations when the communication between the controller-cards is broken, the option to use config/power-admin-state=POWER_DISABLED for one of the cards may not help post reboot?
```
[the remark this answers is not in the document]
sachendras: `rule https://github.com/openconfig/public/issues/5 only applies to single controller card.`
**Response: ** Right. But the concern is that in a Split-brain scenario both cards think they are the only cards in the box and try to become Primary. More details in the question below.

`Does this mean a scenario where two controller cards are present and both are set to redundancy-state PRIMARY`
**Response: ** `redundant-role` is a state leaf right? IIUC we cant set the `redundant-role` for a controller-card. Having said that, what if the communication between the 2 controller-cards is broken for reasons like software bug, configuration error etc. In that case both cards will consider themselves as the only controller-card in the box and will try to gain Primary role. I will add a definition for Split-brain scenario.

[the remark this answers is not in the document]
dplore: Thanks, just looking for the doc to define "split-brain" so it is clear.  Your explanation here makes sense, but it should be in the markdown.  :)

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1172,15 @@ module openconfig-platform {
...
-          required to perform a switchover first.";
+          "Configuration data for controller card components.
+          A controller-card can be configured for persistent
+          powered-off mode using the config/power-admin-state leaf.
+          The reference path below defines rules for such a
+          configuration.";
+
+        reference
+          "Rules around power-off configuration in controller-cards:
+          https://github.com/openconfig/public/blob/master/doc/controller-card_poweroff.md";
```
dplore, line 1183: Nit:  I think this description and reference should be moved to the controller-card model file (since this file is for power control for any type of component).

This would be a good location to add it I think: 
https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform-controller-card.yang#L58-L65

release/models/platform/openconfig-platform.yang
```diff
@@ -531,6 +539,27 @@ module openconfig-platform {
...
+    leaf last-poweroff-time {
+      type oc-types:timeticks64;
+      units "nanoseconds";
+      description
+        "This records the last time a component was directly powered
+         down. The value is a Unix Epoch timestamp (nanoseconds since
+         Jan 1, 1970 00:00:00 UTC). Component power-off can be:
+         - USER_INITIATED
+         - SYSTEM_INITIATED
+         - POWER_FAILURE
```
LimeHat, line 558: If `power_failure` is not a controlled shutdown (e.g. in a system with many PSUs, some of the PSUs fail and system initiates a proactive, graceful shutdown of the component while there's still enough power to keep the control modules up), then it will not be possible to record the `power_failure` event. 

In other words, it is possible to keep track of only the controller shutdowns

doc/controller-card_poweroff.md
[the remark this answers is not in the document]
sachendras: accepted #5

doc/controller-card_poweroff.md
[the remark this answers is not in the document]
sachendras: done

release/models/platform/openconfig-platform.yang
sachendras: Sure, not a immediate requirement for support atm but I would like to capture it here.

release/models/platform/openconfig-platform-types.yang
```diff
@@ -482,6 +516,29 @@ module openconfig-platform-types {
...
+    description
+      "The component powered-off due to power failure.";
+  }
+
+  identity POWEROFF_SYSTEM_INITIATED {
+    base COMPONENT_POWEROFF_REASON;
+    description
+      "The component is powered-off by system.";
+  }
+
```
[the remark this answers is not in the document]
sachendras: done!

release/models/platform/openconfig-platform.yang
sachendras: done

release/models/platform/openconfig-platform-controller-card.yang
```diff
@@ -73,6 +79,22 @@ module openconfig-platform-controller-card {
...
+        description
+          "Configuration data for controller card components.
+          A controller-card can be configured for persistent
+          powered-off mode using the config/power-admin-state leaf.
+          The reference path below defines rules for such a
+          configuration.";
+
+        reference
+          "Rules around power-off configuration in controller-cards:
+          https://github.com/openconfig/public/blob/master/doc/controller-card_poweroff.md";
```
dplore, line 93: This is a duplicate augment.  To fix it you can remove this augment clause and move the description and reference to replace lines 66-68

sachendras: Done

release/models/platform/openconfig-platform.yang
```diff
@@ -1142,9 +1172,8 @@ module openconfig-platform {
 
       container config {
         description
-          "Configuration data for controller card components. Note that disabling
-          power to the primary supervisor should be rejected, and the operator is
-          required to perform a switchover first.";
+          "Configuration data for controller card components
+          power-admin-state.";
```
dplore, line 1176:
```suggestion
          "Configuration data for controller card components.";
```

dplore (commented):
One last small nit on the description for the config container in the platform model.  Otherwise, LGTM

dplore (approved):
LGTM.  Setting last call for April 22, 2024.

@LimeHat in case you had more comments to make?

dplore:
Reviewed in April 16, 2024 OC Operators meeting.  There is ambiguity about what should happen if a user attempts to power off BOTH CONTROLLER_CARDs.  This should be updated in the documentation and state diagram.

sachendras:
> I think there should be some additional text about the "if there's only one card in the chassis, it should ignore the power-disabled state" rule (rule#5).
> 
> Operational use cases listed at the beginning of this document do not seem to require that behavior.

Done!

sachendras:
> Reviewed in April 16, 2024 OC Operators meeting. There is ambiguity about what should happen if a user attempts to power off BOTH CONTROLLER_CARDs. This should be updated in the documentation and state diagram.

Done!

doc/controller-card_poweroff.md
```diff
@@ -77,7 +77,9 @@ After controller-card0 transitions to redundant-role SECONDARY:
...
    
 
-5. On boot (cold or warm), if the chassis has a single controller card and it is configured for config/power-admin-state=POWER_DISABLED, as per Rule#1 above state/power-admin-state should be POWER_ENABLED given that a single CONTROLLER_CARD should have state/redundant-role=PRIMARY. The system must log a message using severity "Warning" to inform the Operator about the situation.
+5. On boot (cold or warm), if the chassis has a single controller card and it is configured for config/power-admin-state=POWER_DISABLED, it Must continue with the boot process ignoring the configuration. As per Rule#1 above, the controller-card  Must have it's state/power-admin-state as POWER_ENABLED given that a single CONTROLLER_CARD will have it's state/redundant-role=PRIMARY. In this case too, the system must log a message using severity "Warning", to inform the Operator about the situation.
```
dplore, line 80: Nit:

```suggestion
5. On boot (cold or warm), if the chassis has a single controller card and it is configured for config/power-admin-state=POWER_DISABLED, it must continue with the boot process ignoring the configuration. As per Rule#1 above, the controller-card  must have it's state/power-admin-state as POWER_ENABLED given that a single CONTROLLER_CARD will have it's state/redundant-role=PRIMARY. In this case too, the system must log a message using severity "Warning", to inform the Operator about the situation.
```

doc/img/controller_card.png
dplore, line 1: Nit:  might be easier to read if the new condition is a 'decision' in the state machine after the "is redundant_role_primary" decision.

dplore (approved):
Two nits for your to consider, but LGTM

PR #1043, opened 2024-02-01
add bgp comm regex-match

dplore:
### Change Scope

* Add OC type for posix extended regular expressions.
* Change the union type for BGP community sets to use posix extended regular expression.
* Clarify how the regex is used when matching communities.

Rather than using various implementation specific regex implementations
1. A  posix regex format is proposed to be used for BGP community sets.
2. An algorithm for BGP communities to be matched is proposed to make it clear how the regex-match leaf will behave.

The result is a vendor neutral method for writing regex for BGP communities which can be tested.  Testing is possible because the regex implementation is standard and implemented in many open source development and testing tools versus the proprietary regex that are present today.  

This is an incremental step towards an end state for bgp community set modeling described in https://github.com/openconfig/public/pull/883

### Tree Representation
Since this is only a type and description change, there is no change to the OC tree.   Affected paths are:
- `/routing-policy/defined-sets/bgp-defined-sets/community-sets/community-set/config/community-member`
- `/routing-policy/defined-sets/bgp-defined-sets/ext-community-sets/ext-community-set/config/ext-community-member`

### Platform Implementations

Regex based pattern matching for community is widely implemented but most use a proprietary format for the regex.  References are included for comparison of BGP community regex matching behavior.

- [Cisco IOS XR community matching](https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-0/routing/command/reference/rr40crs1book_chapter8.html#wp1911706197) explicitly supports regex evaluation for one community at a time.  
- [JunOS community matching using regex](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-bgp-communities-extended-communities-evaluation-in-routing-policy-match-conditions.html)

### DIFF
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -596,7 +596,7 @@ module openconfig-bgp-policy {
     leaf-list community-member {
       type union {
         type oc-bgp-types:bgp-std-community-type;
-        type oc-types:posix-eregexp;
+        type oc-bgp-types:bgp-community-regexp-type;
         type oc-bgp-types:bgp-well-known-community-type;
       }
       description
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -8,6 +8,7 @@ module openconfig-bgp-policy {
   prefix "oc-bgp-pol";
 
   // import some basic types
+  import openconfig-types { prefix oc-types; }
   import openconfig-inet-types { prefix oc-inet; }
   import openconfig-routing-policy {prefix oc-rpol; }
   import openconfig-policy-types { prefix oc-pol-types; }
@@ -28,7 +29,12 @@ module openconfig-bgp-policy {
     It augments the base routing-policy module with BGP-specific
     options for conditions and actions.";
 
-  oc-ext:openconfig-version "6.3.0";
+  oc-ext:openconfig-version "6.4.0";
+
+  revision "2024-01-31" {
+    description
+    "Add regex-match to community-set.";
+  }
 
   revision "2023-10-23" {
     description
@@ -589,6 +595,14 @@ module openconfig-bgp-policy {
         be removed unless match-set-options is INVERT which will
         reverse this to mean that anything that does not match will be
         removed.";
+    }
+
+    leaf regex-match {
+      type oc-types:posix-eregexp;
+      description "A regex string used to match BGP communities.
+      This regex should be matched against a string which lists
+      BGP communities each represented as [0-9]+:[0-9]+ and space
+      separated.  The BGP community entries are not order dependent.";
     }
 
     leaf match-set-options {
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -676,7 +676,7 @@ module openconfig-bgp-policy {
 
     leaf-list ext-community-member {
       type union {
-        type oc-bgp-types:bgp-std-community-type;
+        type oc-bgp-types:bgp-ext-community-type;
         type oc-types:posix-eregexp;
       }
       description
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -8,13 +8,12 @@ module openconfig-bgp-policy {
   prefix "oc-bgp-pol";
 
   // import some basic types
-  import openconfig-types { prefix oc-types; }
   import openconfig-inet-types { prefix oc-inet; }
   import openconfig-routing-policy {prefix oc-rpol; }
   import openconfig-policy-types { prefix oc-pol-types; }
   import openconfig-bgp-types { prefix oc-bgp-types; }
   import openconfig-extensions { prefix oc-ext; }
-
+  import openconfig-types { prefix oc-types; }
 
   // meta
   organization
@@ -29,11 +28,13 @@ module openconfig-bgp-policy {
     It augments the base routing-policy module with BGP-specific
     options for conditions and actions.";
 
-  oc-ext:openconfig-version "6.4.0";
+  oc-ext:openconfig-version "7.0.0";
 
   revision "2024-01-31" {
     description
-    "Add regex-match to community-set.";
+      "Update community-sets/members/member union type by replacing
+      the bgp-regex type with posix-eregexp.";
+    reference "7.0.0";
   }
 
   revision "2023-10-23" {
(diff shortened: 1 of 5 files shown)


wenovus:
@dplore based on the fact it's not reproducible locally and that random threads are being killed, I'm guessing it's because since all the models were running in parallel, the build machine went OOM and decided to kill one of the pyangbind threads.

Making ygnmi run after pyangbind (empirically these two were the ones conflicting with one another) seems to resolve this issue. But if this happens let me know and I'll make the processes more serial.

In the future I think it makes sense to move to one of the high memory private worker pool nodes. It doubles the build cost but will give us 4x more memory.

----

Please review https://github.com/openconfig/models-ci/pull/98 and then I will fix up cloudbuild.yaml.

wenovus:
@dplore fixed up now.  For any other affected PRs, they will need to merge from the main branch to see the new results.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -578,17 +585,26 @@ module openconfig-bgp-policy {
...
         reverse this to mean that anything that does not match will be
-        removed.";
+        removed.
+
+        For MATCH operations the posix-eregexp type should be evaluated
+        against each community associated with a prefix one community
+        at a time.  Communities must be represented as strings in formats
+        conforming to oc-bgp-types:bgp-std-community-type.  For example:
+        - `1000:1000` for a standard community
+        - `route-origin:1000:1000` for the origin type extended community,
```
rolandphung, line 606: Should this example be in the extended community description?

Alternatively should this regex format excerpt be documented in one place, like a `bgp-community-eregexp` (or separate `std`/`ext`/`large`) type?

It might be important to be able to document the exact format of all possible extended community values.

dplore:
@earies @rgwilton @hellt I would appreciate your feedback as well.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -660,17 +676,21 @@ module openconfig-bgp-policy {
 
     leaf-list ext-community-member {
       type union {
-        type oc-bgp-types:bgp-ext-community-type;
-        type oc-bgp-types:bgp-community-regexp-type;
+        type oc-bgp-types:bgp-std-community-type;
```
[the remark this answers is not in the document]
dplore: Yes, this was unintentional.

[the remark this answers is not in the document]
dplore: fixed

release/models/bgp/openconfig-bgp-policy.yang
dplore: Thanks for this.  I moved the extended community example to `leaf-list ext-community-member`.

Regarding example extended community values, one could say that [oc-bgp-types:bgp-ext-community-type](https://github.com/openconfig/public/blob/5d24889a323a84fd077cbb83c2c7b5fb4896a6fe/release/models/bgp/openconfig-bgp-types.yang#L441-L643) does this.   Take a look there and see if you agree that it is satisfactory.

dplore:
Last call, this will merge on March 8th.

release/models/bgp/openconfig-bgp-policy.yang
rolandphung: I think that covers it well, though the main description does not cover `link-bandwidth:`. Perhaps this change could update that description?

jhaas-pfrc:
The fundamental request here is "use full ieee extended regex".  
IETF chose to go with a restricted subset of this to cover what vendors generally ship.  For the most part, the identified subsets are unsupported for a few reasons:

1. CLI syntaxes don't permit some of the patterns - i.e. it introduces ambiguity in a cli context or potentially introduces restricted or unsafe characters like backslash.
2. backrefs can break match caching, even when they're extremely helpful.  This means a potentially heavy hit to routing pipelines.  

NOTE: regular expressions are generally the worst performing policy element in policy engines across all vendors.  Vendors spend a lot of energy avoiding such performance hits or letting customers shoot themselves in the foot too much.

My recommendation would be for OC to support some form of the IETF restrictions for maximal compatibility across vendors.

From https://datatracker.ietf.org/doc/html/draft-ietf-idr-bgp-model:

> 
> 
>   typedef bgp-community-regexp-type {
>     type string;
>     description
>       "Type definition for communities specified as regular
>        expression patterns.
> 
>        A compliant implementation of this type MUST accept a POSIX.2
>        Extended Regular Expression excluding the following features:
>        - character class expressions (Section 9.3.5)
>        - collating symbols (Section 9.3.5)
>        - equivalence classes (Section 9.3.5)
>        - back-reference expressions (Section 9.3.6)
> 
>        Implementations MAY accept additional forms of regular
>        expressions as long as the minimally compliant form documented
>        above is accepted.";
>     reference
>       "IEEE Std 1003.1-2017: The Open Group Base Specifications Issue
>        7, 2018 edition.";
>   }

dplore:
> The fundamental request here is "use full ieee extended regex". IETF chose to go with a restricted subset of this to cover what vendors generally ship. For the most part, the identified subsets are unsupported for a few reasons:
> 
> 1. CLI syntaxes don't permit some of the patterns - i.e. it introduces ambiguity in a cli context or potentially introduces restricted or unsafe characters like backslash.
> 2. backrefs can break match caching, even when they're extremely helpful.  This means a potentially heavy hit to routing pipelines.


Thanks for this feedback, it's very helpful.  I've added a constraint and reference to the IETF draft.

release/models/bgp/openconfig-bgp-policy.yang
dplore: Done!

release/models/bgp/openconfig-bgp-types.yang
```diff
@@ -673,12 +682,15 @@ module openconfig-bgp-types {
...
-    // take.
-    type oc-types:std-regexp;
+    type oc-types:posix-eregexp;
     description
       "Type definition for communities specified as regular
-      expression patterns";
+      expression patterns.  The regular expression must be a
+      POSIX extended regular expression with some limitations
+      which are commonly found in device implementations described
+      in draft-ietf-idr-bgp-model";
```
robshakir, line 691:
```suggestion
      in draft-ietf-idr-bgp-model.";
```

robshakir (approved):

s19nal (approved):

s19nal (approved):

dplore:
@robshakir can you re-review/approve?  I fixed the community-member union type to point to the intended bgp-community-regex type which introduced the constaints on top of the posix extended regex type.  (The earlier approved commit was still referencing the 'plain' posix extended regex type)

robshakir (approved):

PR #1048, opened 2024-02-12
AFT model improvements (state-synced default change, description refinements, other as protocol/encap type)

nokia1adam:

Change Scope

Description changes:
* next-hop/mac-address: description now notes that absence of a value may indicate that the mac address will be learned by mechanisms such as ARP/ND and not by direct programming
* next-hop/origin-protocol: description now notes that in the context of a next-hop this refers to the protocol that resolves the next-hop
* next-hop-group/backup-next-hop-group: description now notes that this indicates availability of a backup NHG, not its active use in the datapath at the time when this state is retrieved
* next-hop-group/next-hop/weight: description now notes that this is the raw weight from higher level protocol interaction, and may not be equal to the weight used by the datapath hashing algorithm

Default changes:
afts/state-synced/state/ipv4-unicast: default is changed from FALSE to TRUE
afts/state-synced/state/ipv6-unicast: default is changed from FALSE to TRUE
Rationale: state-synced FALSE represents an abnormal condition, and should therefore be the exception case, not the normal case

Config changes:
INSTALL_PROTOCOL_TYPE: new identity OTHER to allow an implementation to specify that the entry it is reporting was installed by a protocol that is not defined by OpenConfig
encapsulation-header-type: new enum OTHER to allow an implementation to specify that the encapsulation format is not defined by OpenConfig

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -316,7 +316,7 @@ submodule openconfig-aft-common {
       description
         "The MAC address of the next-hop if resolved by the local
         network instance.
-        
+
         This leaf may have no value if the mac address is determined
         by mechanisms outside route programming -i.e. by ARP or IPv6
         neighbor discovery.";
@@ -647,12 +647,12 @@ submodule openconfig-aft-common {
         path "../../../next-hop-group/state/id";
       }
       description
-        "The backup next-hop-group that is available for the current group. 
+        "The backup next-hop-group that is available for the current group.
         When all entries within the next-hop group become unusable, the backup
         next-hop group is used if specified.
-        
+
         This leaf does not indicate whether the backup-next-hop-group is in
-        active use or not; it only indicates whether a backup NHG is 
+        active use or not; it only indicates whether a backup NHG is
         available";
     }
   }
@@ -679,7 +679,7 @@ submodule openconfig-aft-common {
         is balanced across the next-hops within the group in the
         proportion of weight/(sum of weights of the next-hops within
         the next-hop group).
-        
+
         This is the programmed weight as supplied by the higher level
         protocol or agent; it is not the necessarily the weight
         used in the datapath programming of the NHG.";
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,13 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.4.1";
+
+  revision "2024-02-12" {
+    description
+      "Change state-synced defaults to true.";
+    reference "2.4.1";
+  }
 
   revision "2023-09-26" {
     description
@@ -309,7 +315,11 @@ submodule openconfig-aft-common {
       type oc-yang:mac-address;
       description
         "The MAC address of the next-hop if resolved by the local
-        network instance.";
+        network instance.
+        
+        This leaf may have no value if the mac address is determined
+        by mechanisms outside route programming -i.e. by ARP or IPv6
+        neighbor discovery.";
     }
 
     leaf pop-top-label {
@@ -388,14 +398,15 @@ submodule openconfig-aft-common {
   grouping aft-common-install-protocol {
     description
       "Grouping for a common reference to the protocol which
-      installed an entry.";
+      installed an entry or that resolved a next-hop.";
 
     leaf origin-protocol {
       type identityref {
         base "oc-pol-types:INSTALL_PROTOCOL_TYPE";
       }
       description
-        "The protocol from which the AFT entry was learned.";
+        "The (pseudo) protocol that installed an entry or that
+        resolved a next-hop.";
     }
 
   }
@@ -636,9 +647,13 @@ submodule openconfig-aft-common {
         path "../../../next-hop-group/state/id";
       }
       description
-        "The backup next-hop-group for the current group. When all
-        entries within the next-hop group become unusable, the backup
-        next-hop group is used if specified.";
+        "The backup next-hop-group that is available for the current group. 
+        When all entries within the next-hop group become unusable, the backup
+        next-hop group is used if specified.
+        
+        This leaf does not indicate whether the backup-next-hop-group is in
+        active use or not; it only indicates whether a backup NHG is 
+        available";
     }
   }
 
@@ -663,7 +678,11 @@ submodule openconfig-aft-common {
         "The weight applied to the next-hop within the group. Traffic
         is balanced across the next-hops within the group in the
         proportion of weight/(sum of weights of the next-hops within
-        the next-hop group).";
+        the next-hop group).
+        
+        This is the programmed weight as supplied by the higher level
+        protocol or agent; it is not the necessarily the weight
+        used in the datapath programming of the NHG.";
     }
   }
 
--- a/release/models/aft/openconfig-aft-state-synced.yang
+++ b/release/models/aft/openconfig-aft-state-synced.yang
@@ -16,7 +16,13 @@ submodule openconfig-aft-state-synced {
     "Submodule containing definitions of groupings for the state
     synced signals corresponding to various abstract forwarding tables.";
 
-  oc-ext:openconfig-version "2.4.0";
+  oc-ext:openconfig-version "2.4.1";
+
+  revision "2024-02-12" {
+    description
+      "Change state-synced defaults to true.";
+    reference "2.4.1";
+  }
 
   revision "2023-09-26" {
     description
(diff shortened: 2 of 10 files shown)


release/models/aft/openconfig-aft-common.yang
```diff
@@ -663,7 +678,11 @@ submodule openconfig-aft-common {
         "The weight applied to the next-hop within the group. Traffic
         is balanced across the next-hops within the group in the
         proportion of weight/(sum of weights of the next-hops within
-        the next-hop group).";
+        the next-hop group).
+
+        This is the programmed weight as supplied by the higher level
+        protocol or agent; it is not the necessarily the weight
+        used in the datapath programming of the NHG.";
```
robshakir, line 685: It would be useful to clarify here why this might diverge -- i.e., if we are calculating how traffic is being split, I _would_ expect that this represents the weight that is used in the data plane -- otherwise the existing leaf is not correct.

Is the case that you are handling here the one where there is some quantisation further down the programming of the forwarding plane?

release/models/aft/openconfig-aft-state-synced.yang
```diff
@@ -49,7 +55,7 @@ submodule openconfig-aft-state-synced {
 
       leaf ipv4-unicast {
         type boolean;
-        default false;
+        default true;
```
robshakir, line 58: I'm not sure that we should change this default -- if the default is true in some implementation, then can't it just send a value of `true` here? Otherwise we're assuming that the happy case has happened before /any/ telemetry as such has been sent, which seems a bad idea for a health signal.

release/models/aft/openconfig-aft-state-synced.yang
```diff
@@ -60,7 +66,7 @@ submodule openconfig-aft-state-synced {
 
       leaf ipv6-unicast {
         type boolean;
-        default false;
+        default true;
```
robshakir, line 69: Same comment as for v4.

release/models/policy/openconfig-policy-types.yang
```diff
@@ -283,4 +289,13 @@ module openconfig-policy-types {
 
       10.244.136.79/32.";
   }
+
+  identity OTHER {
```
robshakir, line 293: I also don't think we should add this -- again, is there a reason not to just add the protocols that are not listed here?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -388,14 +398,15 @@ submodule openconfig-aft-common {
...
+      installed an entry or that resolved a next-hop.";
 
     leaf origin-protocol {
       type identityref {
         base "oc-pol-types:INSTALL_PROTOCOL_TYPE";
       }
       description
-        "The protocol from which the AFT entry was learned.";
+        "The (pseudo) protocol that installed an entry or that
+        resolved a next-hop.";
```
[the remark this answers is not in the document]
nokia1adam: The problem (lack of clarity) comes from the use of this leaf in the context of a next-hop. If the device learns a BGP route to next-hop 1.1.1.1, and the address 1.1.1.1 is resolved by an IS-IS route for 1.1.1.1/32, then what protocol was the next-hop 1.1.1.1 learned from? Arguments can be made to state BGP, and arguments can be made to state ISIS. The existing description doesn't make the answer clear.

I think your answer to this question is BGP, but this makes it problematic when a next-hop is shared by multiple protocols. If origin-protocol should be left empty in these cases then that is fine but I think all of this needs to be described.

release/models/aft/openconfig-aft-common.yang
nokia1adam: Yes, if you consider that weights programmed by gRIBI have a uint64 range and weights signaled in the BGP link bandwidth community have mantissa-exponent floating point range, it shouldn't be assumed that the full range of these weights are used by the datapath.

If this leaf is supposed to represent the quantized weight used by the datapath then this should be specified. I personally think there is value seeing both raw/signaled and normalized weights for each next-hop.

release/models/aft/openconfig-aft-state-synced.yang
nokia1adam: I think this sets a bad precedent. Suppose vendor X discovers an internal synchronization issue between actual state and telemetry state and they want to expose this issue (for good reasons) in state. Why should every other vendor have to report that they don't have this synchronization issue? With a properly chosen default, this can be avoided.

release/models/policy/openconfig-policy-types.yang
nokia1adam: I can recall a discussion we had many years ago on a related topic. :-)

I'm not sure it is realistic to extend the list of identities to be a superset of ALL the "install" protocol types across ALL vendors, especially when you consider that some vendors may have protocol types that are proprietary and not tied to standardized protocols. And even if you could come up with a list that is 99% complete now the OC model will always be in catch-up mode as new protocols get implemented.

If you are against 'other' then it should be documented that an implementation is free to provide no value for this leaf if the installed type is not expressible in OC.

dplore: I think we are ok in OC in general to rapidly update identities and enums in OC.   PR's can be merged pretty quickly for these kinds of changes (2 weeks or even less for obvious, documented identity definitions.  An example is OC Ethernet PMD types). This is suitable for protocols that are supported by 2 or more vendors.

For truly vendor proprietary protocols, we could prescribe a specific model structure for vendors to augment an OC model.  In this case, it could be as simple as creating a 'vendor-protocol-identity' leaf which refers to identity provided by a vendor in a separate, vendor specific yang module.    (I would not want to mix/modify a list of OC identities with vendor identities due to confusion and collisions between vendors).  Examples of this are the [pipeline-counters ](https://github.com/openconfig/public/blob/master/doc/vendor_counter_guide.md) model.

release/models/aft/openconfig-aft-state-synced.yang
dplore: It's completely legal to stream false, even when the schema default is false. I think this would cover your use case, yes @nokia1adam ?

release/models/aft/openconfig-aft-types.yang
```diff
@@ -89,6 +95,10 @@ module openconfig-aft-types {
         description
           "The encapsulation header is a VXLAN packet header";
       }
+      enum OTHER {
```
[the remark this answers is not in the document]
dplore: +1.  It's very easy to add more enums.  OC review and approvals for adding enums is quite fast.  CHeck the ethernet PMD as an example.  We have a lot of them and continue to add them.

release/models/aft/openconfig-aft-state-synced.yang
nokia1adam: Hi @dplore , I don't think this helps me.

In the implementation I am working with, the false value represents a problem that requires investigation. So I certainly hope to be not streaming false all the time. The default of false requires me to constantly stream the state of this leaf as true, which is consuming extra resources just to indicate a normal/good state.

release/models/aft/openconfig-aft-types.yang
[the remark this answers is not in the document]
nokia1adam: OK, I will add a few of the obvious missing ones and resubmit the PR.

release/models/policy/openconfig-policy-types.yang
nokia1adam: Hi @dplore , sure let me resubmit the PR with some obvious missing identities that I know are supported by 2+ vendors, and also add a vendor container that could be an attachment point for vendor-specific identities that come from augmentation.

release/models/aft/openconfig-aft-common.yang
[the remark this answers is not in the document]
robshakir: Agreed -- these clarifications seem worth making, but I'm not sure that the suggested edit makes it clearer. I'm happy to say that we should leave this blank in these cases, or give some examples here as to what we expect things to be populated with.

release/models/aft/openconfig-aft-common.yang
robshakir: ACK -- I would just specify in this description why the weights might diverge here. I'm open to adding something that gives us the signalled and normalised values.

release/models/aft/openconfig-aft-state-synced.yang
robshakir: I see the argument, but there are always going to be cases where such differences exist. In the overall volume of telemetry of AFT, does this one leaf cause us issues? Isn't it just sent once at the start of the session being set to true (with a `TARGET_DEFINED` or `ON_CHANGE` subscription)? If not, let's discuss with other implementations as to what the best choice here is.

(if we change the default to true, and some implementation says that until there's an explicit signal that it's false - then we just move the work around).

On this matter, from an operational perspective, assuming that things are healthy until they say unhealthy seems less preferable -- it seems like it's better to have some explicit signal that it's healthy.

release/models/policy/openconfig-policy-types.yang
robshakir: This is an identity -- it can be extended by vendors through augments for their proprietary protocols, you can simply use the `install-protocol-type` identity as the `base`.

dplore: Vendors could indeed add additional identities with an augment.  I suppose there is some remote chance that there could be a conflict where the same identity is used by two different vendors but has different meanings.  This seems unlikely, but might be very confusing if it were to occur.   Using a "vendor" container is slightly more complex, but would guarantee this won't happen.

PR #1056, opened 2024-02-23
Clarify input-power threshold

dplore:
### Change Scope

* Clarify that transceiver input-power thresholds apply to the physical channel and not the aggregate input power.

### DIFF
--- a/release/models/platform/openconfig-platform-transceiver.yang
+++ b/release/models/platform/openconfig-platform-transceiver.yang
@@ -66,7 +66,13 @@ module openconfig-platform-transceiver {
       specify a physical-channel within a TRANSCEIVER component
       (i.e. gray optic) that it is associated with.";
 
-  oc-ext:openconfig-version "0.13.0";
+  oc-ext:openconfig-version "0.14.0";
+
+revision "2023-08-30" {
+    description
+      "Clariy transceiver module threshold for input-power.";
+    reference "0.14.0";
+  }
 
 revision "2023-08-30" {
     description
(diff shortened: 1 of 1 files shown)


release/models/platform/openconfig-platform-transceiver.yang
```diff
@@ -451,9 +457,11 @@ revision "2023-08-30" {
...
     container thresholds {
       description
-        "Enclosing container for transceiver alarm thresholds.
-        Each threshold is compared to the instant value of the
-        measured parameter corresponding to the threshold";
+        "Enclosing container for transceiver alarm thresholds. Each threshold
+        is compared to the instant value of the measured parameter
+        corresponding to the threshold. For optical input power, the
+        threshold applies only to the physical-channel(s).  This threshold
+        does not apply to the aggregate transceiver input-power.";
```
ejbrever, line 464: should these messages be on each of the specific leaves instead? It is perhaps redundant, but seems more explicit and if we add something like aggregate power we won't have to remember to update the top level description?
https://github.com/openconfig/public/blob/5d24889a323a84fd077cbb83c2c7b5fb4896a6fe/release/models/platform/openconfig-platform-transceiver.yang#L794

Also, this should be for input and output.

ejbrever, line 464: Should we also clarify that the threshold defined applies to every physical channels?

dplore: done

dplore: Done

release/models/platform/openconfig-platform-transceiver.yang
```diff
@@ -66,7 +66,13 @@ module openconfig-platform-transceiver {
       specify a physical-channel within a TRANSCEIVER component
       (i.e. gray optic) that it is associated with.";
 
-  oc-ext:openconfig-version "0.13.0";
+  oc-ext:openconfig-version "0.14.0";
+
+revision "2023-08-30" {
+    description
+      "Clariy transceiver module threshold for input-power.";
```
ejbrever, line 73: nit, "Clarify"

ejbrever (approved):

PR #1060, opened 2024-03-01
add gnoi service identity to grpc services

LimeHat:
### Change Scope
- Add `gNOI` grpc service identity to complement the existing list of configurable grpc services (gNMI, gRIBI, P4RT, and soon-to-be-merged gNSI).

### DIFF
--- a/release/models/system/openconfig-system-grpc.yang
+++ b/release/models/system/openconfig-system-grpc.yang
@@ -22,7 +22,7 @@ module openconfig-system-grpc {
     to be included in the list.";
 
 
-  oc-ext:openconfig-version "1.0.1";
+  oc-ext:openconfig-version "1.1.0";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
--- a/release/models/system/openconfig-system-grpc.yang
+++ b/release/models/system/openconfig-system-grpc.yang
@@ -22,9 +22,15 @@ module openconfig-system-grpc {
     to be included in the list.";
 
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "1.0.1";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
+
+  revision "2024-03-01" {
+    description
+      "gNOI service identity.";
+    reference "1.0.1";
+  }
 
   revision "2022-04-19" {
     description
@@ -54,6 +60,12 @@ module openconfig-system-grpc {
     base GRPC_SERVICE;
     description
       "gNMI: gRPC Network Management Interface";
+  }
+
+  identity GNOI {
+    base GRPC_SERVICE;
+    description
+      "gNOI: gRPC Network Operations Interface";
   }
 
   grouping grpc-service-structural {
--- a/release/models/system/openconfig-system-grpc.yang
+++ b/release/models/system/openconfig-system-grpc.yang
@@ -29,7 +29,7 @@ module openconfig-system-grpc {
   revision "2024-03-01" {
     description
       "gNOI service identity.";
-    reference "1.0.1";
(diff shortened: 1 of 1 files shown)


release/models/system/openconfig-system-grpc.yang
```diff
@@ -22,10 +22,16 @@ module openconfig-system-grpc {
     to be included in the list.";
 
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "1.0.1";
```
dplore, line 25:
```suggestion
  oc-ext:openconfig-version "1.1.0";
```

release/models/system/openconfig-system-grpc.yang
```diff
@@ -22,10 +22,16 @@ module openconfig-system-grpc {
...
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
+  revision "2024-03-01" {
+    description
+      "gNOI service identity.";
+    reference "1.0.1";
```
dplore, line 32:
```suggestion
    reference "1.1.0";
```

release/models/system/openconfig-system-grpc.yang
```diff
@@ -56,6 +62,12 @@ module openconfig-system-grpc {
       "gNMI: gRPC Network Management Interface";
   }
 
+  identity GNOI {
```
earies, line 65: Am currently working on publishing a PR shortly that encompasses some redesign to grpc-servers in general and in that I would advise against these service identities existing directly in this module.

Much like we already have in `p4rt`, the service groupings can be split out into their own domains and provide a demarcation point thus we would have something such as `openconfig-gnoi.yang` going fwd.

All specific service config/state then resides in that new module along w/ augmentations into the grpc-server tree.

This also means the `GNMI` identity should be relocated and removed from this module

dplore: Ah, I see what you're saying.  Any yang models for gN*I protocols should add identities rather than asserting all the gN*I protocols in a single file here.

For example https://github.com/openconfig/public/blob/9826cae9e0aa1b81dbe5abcbce842adcc35f7921/release/models/gnsi/openconfig-gnsi.yang#L35-L39

LimeHat: I don't mind either way. 
Currently there's no separate model for either gnmi or gnoi, hence if the consesus is to refactor, identity will be the only thing in the initial version.

earies: precisely and yes - had a few other constructs in mind but can very well produce a model w/ a single identity as 0.1.0

It's really about clear demarcations and what the YANG language is capable of in 1.0 (and even 1.1) today - no way to "deviate" identities so it's an implementation detail that needs to be signaled by other means (and more complex code to account for such on model processing).  If we group services into buckets, an implementation can say support gNMI but not gNSI, etc... and it is defined on model support boundaries

dplore: @LimeHat this was reviewed at the March 19, 2024 operators meeting.  There is rough consensus that using a separate model is preferred.  The inspiration for this pattern was the ever growing network-instances model.  We should probably update our contributing guide to reflect this bias towards 'composability' by  augmenting models with separate files

LimeHat: Ack, will update shortly.

dplore: Hi @LimeHat are you interested in pursuing this further?

earies: This can be covered by https://github.com/openconfig/public/pull/1096 if we agree on the approach

dplore:
I will close this in favor of #1096 ,  please feel free to comment @LimeHat or move discussion there.

PR #1061, opened 2024-03-05
gnsi: align with latest certz version

brianneville:
The certz yang file was updated since 
the PR to move from openconfig/gnsi to 
openconfig/public, and this updated
version of certz had not been
included in the PR.

see the changes to gnsi-certz.yang at:
https://github.com/openconfig/gnsi/pull/156

This commit also sets the container
name for the certz counters to be just
"counters" to be backward-compatible 
with the existing gnsi-certz.yang.

The module name/prefix already 
mark these counters as certz-related,
so we don't need to duplicate this
information in the container name too.

### Change Scope

* changes to models:
```yang
    +--ro certz-counters
       +--ro access-rejects?       oc-yang:counter64
       +--ro last-access-reject?   oc-types:timeticks64
       +--ro access-accepts?       oc-yang:counter64
       +--ro last-access-accept?   oc-types:timeticks64
```       
is being changed to:
```yang
    +--ro counters
       +--ro connection-rejects?       oc-yang:counter64
       +--ro last-connection-reject?   oc-types:timeticks64
       +--ro connection-accepts?       oc-yang:counter64
       +--ro last-connection-accept?   oc-types:timeticks64
```
       
* this is not backwards-compatible with the current model at 
https://github.com/openconfig/public/blob/9826cae9e0aa1b81dbe5abcbce842adcc35f7921/release/models/gnsi/openconfig-gnsi-certz.yang (merged 2 days ago so presumably not in-use yet).
However, it is backwards compatible with the current model from the gNSI repo at 
https://github.com/openconfig/gnsi/blob/718ee9102f4676e11485ff151e9caa4fa74f57e1/certz/gnsi-certz.yang
which is in-use by vendors.
The current model from the gNSI repo provides the paths:
```yang
module: gnsi-certz

    +--ro certificate-version?                             version
    +--ro certificate-created-on?                          created-on
    +--ro ca-trust-bundle-version?                         version
    +--ro ca-trust-bundle-created-on?                      created-on
    +--ro certificate-revocation-list-bundle-version?      version
    +--ro certificate-revocation-list-bundle-created-on?   created-on
    +--ro authentication-policy-version?                   version
    +--ro authentication-policy-created-on?                created-on
    +--ro ssl-profile-id?                                  string
    +--ro counters
       +--ro connection-rejects?       oc-yang:counter64
       +--ro last-connection-reject?   oc-types:timeticks64
       +--ro connection-accepts?       oc-yang:counter64
       +--ro last-connection-accept?   oc-types:timeticks64
```

### Platform Implementations

 * Arista Certz support: https://www.arista.com/en/support/toi/eos-4-31-0f/18445-support-for-gnsi-grpc-network-security-interface
 * Cisco Certz support: https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5000/sysman/711x/configuration/guide/b-system-management-cg-ncs5000-711x/configuring-synce.pdf

### DIFF
--- a/release/models/gnsi/openconfig-gnsi-certz.yang
+++ b/release/models/gnsi/openconfig-gnsi-certz.yang
@@ -30,13 +30,11 @@ module openconfig-gnsi-certz {
     "This module provides a data model for the metadata of gRPC credentials
     installed on a networking device.";
 
-  oc-ext:openconfig-version "0.5.0";
+  oc-ext:openconfig-version "0.6.0";
 
-  revision 2024-03- {
+  revision 2024-03-05 {
     description
-      "align yang with the openconfig/gnsi version as of:
-      https://github.com/openconfig/gnsi/tree/e08ebfe537e1603015291898e6d5e9abe1562a0b
-      (update counter names)";
+      "rename access/reject counters";
     reference "0.6.0";
   }
 
--- a/release/models/gnsi/openconfig-gnsi-certz.yang
+++ b/release/models/gnsi/openconfig-gnsi-certz.yang
@@ -32,6 +32,14 @@ module openconfig-gnsi-certz {
 
   oc-ext:openconfig-version "0.5.0";
 
+  revision 2024-03- {
+    description
+      "align yang with the openconfig/gnsi version as of:
+      https://github.com/openconfig/gnsi/tree/e08ebfe537e1603015291898e6d5e9abe1562a0b
+      (update counter names)";
+    reference "0.6.0";
+  }
+
   revision 2024-02-13 {
     description
       "Major style updates and move to openconfig/public from openconfig/gnsi.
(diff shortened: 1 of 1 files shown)


release/models/gnsi/openconfig-gnsi-certz.yang
```diff
@@ -32,6 +32,14 @@ module openconfig-gnsi-certz {
 
   oc-ext:openconfig-version "0.5.0";
 
+  revision 2024-03- {
```
dplore, line 35:
```suggestion
  revision 2024-03-05 {
```

release/models/gnsi/openconfig-gnsi-certz.yang
```diff
@@ -32,6 +32,14 @@ module openconfig-gnsi-certz {
 
   oc-ext:openconfig-version "0.5.0";
```
dplore, line 33:
```suggestion
  oc-ext:openconfig-version "0.6.0";
```

release/models/gnsi/openconfig-gnsi-certz.yang
```diff
@@ -32,6 +32,14 @@ module openconfig-gnsi-certz {
...
 
+  revision 2024-03- {
+    description
+      "align yang with the openconfig/gnsi version as of:
+      https://github.com/openconfig/gnsi/tree/e08ebfe537e1603015291898e6d5e9abe1562a0b
+      (update counter names)";
```
dplore, line 39: I suggest copying the description of the revision(s) that introduced these same changes here.  This way we can think of the content of this module much more independently from the repository/storage system used.  

The distinction I'd like to make is content in these revision statements and the git log are relevant (IMO), but the  history of changes for openconfig/public repo should be hermetic/independent of other repositories)

release/models/gnsi/openconfig-gnsi-certz.yang
brianneville: addressed in https://github.com/openconfig/public/pull/1061/commits/c0ffe5e385876828242f0eb6de7d9f6508d55143

release/models/gnsi/openconfig-gnsi-certz.yang
brianneville: addressed in https://github.com/openconfig/public/pull/1061/commits/c0ffe5e385876828242f0eb6de7d9f6508d55143

That makes sense - I copied the description from the gNSI repo here: https://github.com/openconfig/gnsi/blob/718ee9102f4676e11485ff151e9caa4fa74f57e1/certz/gnsi-certz.yang#L30

dplore (approved):

PR #1062, opened 2024-03-05
Update component temperature thresholds

dplore:
### Change Scope

* Deprecate existing temperature threshold because it uses units (uint32) which are not compatible with temperature (decimal)
* Add high and low temperature thresholds using decimal64 celsius units
### Platform Implementations

 * Implementation A: Cisco IOS XR
 ```
=============================================================================================================
Location  TEMPERATURE                          Value     Crit    Major    Minor    Minor    Major    Crit
          Sensor                             (deg C)     (Lo)     (Lo)     (Lo)     (Hi)     (Hi)    (Hi)
-------------------------------------------------------------------------------------------------------------
0/RP0/CPU0 
          MB_JM01_L1_TEMP                        37      -10       -5        0      130      135      140
          MB_JM01_L2_TEMP                        36      -10       -5        0      130      135      140
          MB_JM11_L1_TEMP                        33      -10       -5        0      130      135      140
          MB_JM21_L1_TEMP                        36      -10       -5        0      130      135      140 
```

### DIFF
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,7 +65,14 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.24.0";
+  oc-ext:openconfig-version "0.25.0";
+
+  revision "2024-03-05" {
+    description
+      "Add high and low temperature thresholds. Deprecate existing 
+      temperature threshold.";
+    reference "0.25.0";
+  }
 
   revision "2023-11-28" {
     description
@@ -598,10 +605,6 @@ module openconfig-platform {
     description
       "Temperature alarm data for platform components";
 
-    // TODO(aashaikh): consider if these leaves could be in a
-    // reusable grouping (not temperature-specific); threshold
-    // may always need to be units specific.
-
     leaf alarm-status {
       type boolean;
       description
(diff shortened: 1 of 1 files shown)


dplore:
Reviewed in Mar 12, 2024 OC operators meeting.  @s19nal noted that the examples show 3 thresholds for high and low and we should allow all these instead of single high and low thresholds, otherwise the vendor has to choose which one to use.  I will update this PR to reflect all 6 thresholds

dplore:
Added severity for upper and lower.
Added tree diff view to the PR description.

nkitchen:
> Reviewed in Mar 12, 2024 OC operators meeting. @s19nal noted that the examples show 3 thresholds for high and low and we should allow all these instead of single high and low thresholds, otherwise the vendor has to choose which one to use. I will update this PR to reflect all 6 thresholds

How does the newest version of this PR allow 3 thresholds to be expressed?  It looks to me like it can express 2 thresholds, with corresponding severities.

Another thing I find confusing: The description of `alarm-severity-lower` is "The severity of the current low temperature alarm" -- what is the "current low temperature alarm", in contrast with "the current high temperature alarm" from the description of `alarm-severity-upper`? We only have one leaf `alarm-status`, so I would expect that there is only a single current alarm.

To me, the model used by the transceivers makes more sense: There is a list of thresholds, keyed by severity, and each entry in the list can have a lower and/or upper temperature threshold.

Issue #781, opened 2023-01-06
Tunnel/routed-vlan interface structures and design divergence implications


earies:
_Per OpenConfig community agenda for 2023-01-05_

Currently, `openconfig-interfaces` presents 2 design patterns from a
hierarchical perspective in relation to the existence of where L2 and L3
containers and attributes exist.

The primary design pattern is that for _most_ interfaces, there is a hierarchy
of `interface->subinterface` - L2 related attributes are generally covered at
the interface while L3 related at the subinterface levels.

```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw subinterfaces
           +--rw subinterface* [index]
              +--rw oc-ip:ipv4
                 +--rw oc-ip:addresses
                 |  +--rw oc-ip:address* [ip]
                 |     +--rw oc-ip:ip        -> ../config/ip
                 |     +--rw oc-ip:config
                 |     |  +--rw oc-ip:ip?              oc-inet:ipv4-address
                 |     |  +--rw oc-ip:prefix-length?   uint8
                 |     +--ro oc-ip:state
                 |     |  +--ro oc-ip:ip?              oc-inet:ipv4-address
                 |     |  +--ro oc-ip:prefix-length?   uint8
                 |     |  +--ro oc-ip:origin?          ip-address-origin
                 ...
```

It is also worth noting that various other models have a concept to reference
back to this concept breaking apart references to the interface + subinterface
pair.

e.g.

```
module: openconfig-network-instance
  +--rw network-instances
     +--rw network-instance* [name]
        +--rw interfaces
           +--rw interface* [id]
              +--rw id        -> ../config/id
              +--rw config
              |  +--rw id?                            string
              |  +--rw interface?                     -> /oc-if:interfaces/interface/name
              |  +--rw subinterface?                  -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
```

However for `routed-vlan` and `tunnel` interfaces, the design pattern changes
where there is no concept of "subinterfaces".  All attributes sit at the root
of an interface.

e.g. `routed-vlan` interface

```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw oc-vlan:routed-vlan
           +--rw oc-ip:ipv4
              +--rw oc-ip:addresses
              |  +--rw oc-ip:address* [ip]
              |     +--rw oc-ip:ip        -> ../config/ip
[...]

Issue #780, opened 2023-01-06
Practical usage of interface 'type' and the association to IETF -> IANA types


earies:
_Per OpenConfig community agenda for 2023-01-05_

Since the initial publish of the `openconfig-interfaces` modeling, the
following mandatory nodes have existed

```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw config
        |  +--rw type    identityref
        +--ro state
           +--ro type    identityref
```

Which is defined as the following 2-stage identityref back to `iana-if-type`

```
leaf type {
  type identityref {
    base ietf-if:interface-type;
  }
  mandatory true;
  description
    "The type of the interface.

    When an interface entry is created, a server MAY
    initialize the type leaf with a valid value, e.g., if it
    is possible to derive the type from the name of the
    interface.

    If a client tries to set the type of an interface to a
    value that can never be used by the system, e.g., if the
    type is not supported or if the type does not match the
    name of the interface, the server MUST reject the request.
    A NETCONF server MUST reply with an rpc-error with the
    error-tag 'invalid-value' in this case.";
  reference
    "RFC 2863: The Interfaces Group MIB - ifType";
}
```

For a number of years, there has been motivation to remove external
dependencies from OpenConfig modeling.  While on one aspect, leveraging others
work for reuse is good practice, however it also creates dependencies which can
prove difficult to control over time (ref: Redefinition of relevant IETF types
modules) especially if the intent of the 2 projects diverge.

The `iana-if-type` module version `2017-01-19` currently defines 285 possible
identities for use while OpenConfig related use-cases only have necessity for
around 5 true interface "types".  Deviations cannot convey which identities are
supported/not-supported thus each implementation needs to implement custom
guardrails to start that can only be signaled out-of-band or via error
handling to the consumer/client.

Aside from the vast list of potential types, implementations generally have not
supported specifying the "type" as a r/w object but rather the interface
nomenclature drives the _type_ followed by potential encapsulations where
applicable. (You cannot for instance dictate that an interface named `et-0/0/0`
can be of type `softwareLoopback`)

Since the model node is marked as `mandatory` means the payload _MUST_ include
this leaf when provisioning an interface (which in the case of physical
interfaces are often already provisioned at time of hardware initialization).
In most cases, what you will find is that this node does not translate to
anything meaningful within the implementation from a configuration perspective
(e.g. no-op).  Essentially these cases create what one would consider a bad
API.  You are giving the perception that you instructed the server to do
[...]

PR #1083, opened 2024-04-03
Add interfaces layer property to openconfig-interfaces model

mikewiebe:
### Change Scope
The proposal is to add an additional "layer" leaf to interface
This is to explicitly specify the interface layer.
```
interfaces
   +-- interface[name]
           +-- config
           +     +-- name
           +     +-- type
           |     +-- layer <<< new leaf
           +-- state
                 +-- layer
```
- The layer can by Layer2 or Layer3
- This change is backward compatible
   When the 'layer" is not specified, the behavior is the same as if the "layer" is not defined in the previous model.

pyang tree output 

```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw config
        |  +--rw name?            string
        |  +--rw type             identityref
        |  +--rw mtu?             uint16
        |  +--rw loopback-mode?   oc-opt-types:loopback-mode-type
        |  +--rw description?     string
        |  +--rw enabled?         boolean
        |  +--rw layer?           identityref
        +--ro state
        |  +--ro name?            string
        |  +--ro type             identityref
        |  +--ro mtu?             uint16
        |  +--ro loopback-mode?   oc-opt-types:loopback-mode-type
        |  +--ro description?     string
        |  +--ro enabled?         boolean
        |  +--ro layer?           identityref
        |  +--ro ifindex?         uint32
        |  +--ro admin-status     enumeration
        |  +--ro oper-status      enumeration
        |  +--ro last-change?     oc-types:timeticks64
        |  +--ro logical?         boolean
        |  +--ro management?      boolean
        |  +--ro cpu?             boolean
```

### Rationale

Currently in openconfig-interface, there is no “layer” information for an interface,
while L2 (switching) / L3 (routing) child configs are all mixed across the interface model.
- Example L2 config : switch-vlan, dot1x, etc
- Example L3 config : proxy-arp, ip addressed, dhcp, etc

Since the interface can mix and match both L2 and L3 config, the model user could have the impression that any given interface can operate concurrently in both the switching and routing mode. This is not true across vendors. Commonly an interface either operates in L2 or L3 mode, and it requires explicit configuration to toggle the interface layer.

The lack of the “layer” affects both the model user and the target device :
- Currently the model user would need to infer the interface layer.
  * Via the interface "name"
     The interface name may suggest its operating layer.
     For example, when the interface name is like vl12, Vlan12, lo100,
     the user can infer these interfaces are logical, and thus are used for routing.
     But such naming semantics is rather vendor specific

  * Via the interface "type"
     Some specific interface types could also suggest its operarting layer.
     For example, the type "iana:softwareLoopback" is a lookback, and thus deemed for routing.
     Though commonly, the interface type refers more to the underlying medium/technology.
     For example, the type "iana:ethernetCsmacd" only means the inteface is an ethernet link,
     it does not tell whether such ethernet interface serves for switching or routing.

- Oppositely, the target device also needs to infer the model user’s intent.
   Take the above ethernet interface as example. The interface may support either L2 or L3 functionality.  
   . If the user ever touches the L3-ish config, like ip (or dhcp, etc), then toggle the interface to L3
   . If the user ever touches the L2-ish config, like vlan (or dot1q, etc), then toggle the interface to L2.

These guess works are not straight-forward nor unreliable.
Thus the proposal is to add a new config leaf “interfaces/interface/config/layer” to eliminate such uncertainty/ambiguity.
Though to emphasize again, the new "layer" config is optional.  If the "layer" is not specified, the behavior remains the same as the current behavior.

### Platform Implementations

 * Cisco NXOS:
     * [Configure Layer2 Interfaces](https://www.cisco.com/c/en/us/td/docs/dcn/nx-os/nexus9000/104x/configuration/interfaces/cisco-nexus-9000-series-nx-os-interfaces-configuration-guide-release-104x/m_configuring_layer_2_interfaces_9x.html)
     * [Configure Layer 3 Interfaces](https://www.cisco.com/c/en/us/td/docs/dcn/nx-os/nexus9000/104x/configuration/interfaces/cisco-nexus-9000-series-nx-os-interfaces-configuration-guide-release-104x/m_configuring_layer_3_interfaces_9x.html)
 * Arista:
     * [Configure Layer2 Interfaces](https://www.arista.com/en/um-eos/eos-layer-2-configuration)
     * [Configure Layer3 Interfaces](https://www.arista.com/en/um-eos/eos-layer-3-configuration)

### DIFF
--- a/release/models/interfaces/openconfig-if-types.yang
+++ b/release/models/interfaces/openconfig-if-types.yang
@@ -1,0 +1,52 @@
+module openconfig-if-types {
+
+  yang-version "1";
+
+  // namespace
+  namespace "http://openconfig.net/yang/interfaces/types";
+
+  prefix "oc-if-types";
+
+  import openconfig-extensions { prefix "oc-ext"; }
+
+  // meta
+  organization "OpenConfig working group";
+
+  contact
+    "OpenConfig working group
+    www.openconfig.net";
+
+  description
+    "Types associated with a interface";
+
+  oc-ext:openconfig-version "3.7.1";
+
+  revision "2024-03-14" {
+    description
+      "Add layer types";
+    reference "3.7.1";
+  }
+
+  // OpenConfig specific extensions for module metadata.
+  oc-ext:catalog-organization "openconfig";
+  oc-ext:origin "openconfig";
+
+  // identity statements
+  identity interface_layer {
+    description
+      "A base identity which can be extended to indicate different
+      layers supported by an interface.";
+  }
+
+  identity L2 {
+    base interface_layer;
+    description
+      "Layer 2 Switching/Bridging interface";
+  }
+
+  identity L3 {
+    base interface_layer;
+    description
+      "Layer 3 Routing interface";
+  }
+}
--- a/release/models/interfaces/openconfig-if-types.yang
+++ b/release/models/interfaces/openconfig-if-types.yang
@@ -32,20 +32,20 @@ module openconfig-if-types {
   oc-ext:origin "openconfig";
 
   // identity statements
-  identity interface_layer {
+  identity INTERFACE_LAYER {
     description
       "A base identity which can be extended to indicate different
       layers supported by an interface.";
   }
 
   identity L2 {
-    base interface_layer;
+    base INTERFACE_LAYER;
     description
       "Layer 2 Switching/Bridging interface";
   }
 
   identity L3 {
-    base interface_layer;
+    base INTERFACE_LAYER;
     description
       "Layer 3 Routing interface";
   }
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -13,6 +13,7 @@ module openconfig-interfaces {
   import openconfig-types { prefix oc-types; }
   import openconfig-extensions { prefix oc-ext; }
   import openconfig-transport-types { prefix oc-opt-types; }
+  import openconfig-if-types { prefix oc-if-types; }
 
   // meta
   organization "OpenConfig working group";
@@ -51,7 +52,14 @@ module openconfig-interfaces {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "3.7.0";
+  oc-ext:openconfig-version "3.8.0";
+
+  revision "2024-04-03" {
+      description
+        "Add layer leaf to interface to indicate the layer at which the interface operates.";
+      reference
+        "3.8.0";
+  }
 
   revision "2023-11-06" {
       description
@@ -342,6 +350,23 @@ module openconfig-interfaces {
         datastore.";
       reference
         "RFC 2863: The Interfaces Group MIB - ifAlias";
+    }
+
+    leaf enabled {
+      type boolean;
+      default "true";
+      description
+        "This leaf contains the configured, desired state of the
+        interface.
+        Systems that implement the IF-MIB use the value of this
+        leaf in the 'running' datastore to set
+        IF-MIB.ifAdminStatus to 'up' or 'down' after an ifEntry
+        has been initialized, as described in RFC 2863.
+        Changes in this leaf in the 'running' datastore are
+        reflected in ifAdminStatus, but if ifAdminStatus is
+        changed over SNMP, this leaf is not affected.";
+      reference
+        "RFC 2863: The Interfaces Group MIB - ifAdminStatus";
     }
 
     leaf enabled {
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -358,10 +358,12 @@ module openconfig-interfaces {
       description
         "This leaf contains the configured, desired state of the
         interface.
+
         Systems that implement the IF-MIB use the value of this
         leaf in the 'running' datastore to set
         IF-MIB.ifAdminStatus to 'up' or 'down' after an ifEntry
         has been initialized, as described in RFC 2863.
+
         Changes in this leaf in the 'running' datastore are
         reflected in ifAdminStatus, but if ifAdminStatus is
         changed over SNMP, this leaf is not affected.";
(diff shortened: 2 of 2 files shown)


dplore:
Thanks Mike,  please provide some explicit examples where there is a problem without this leaf and using the new leaf solves the issue.   Consider that subinterfaces is where OC defines properties for logical and layer 3 interfaces.

mikewiebe:
> Thanks Mike, please provide some explicit examples where there is a problem without this leaf and using the new leaf solves the issue. Consider that subinterfaces is where OC defines properties for logical and layer 3 interfaces.

Here is the additional information:

**Problem Statement:**

Take an ethernet port ‘eth1/1’ as example.
This is an ethernet port which can run either in L2 or L3 layer based on the ‘switchport’ config.
The layer then dictates the valid child config under the interface, ie, L2 and L3 configs are mutual exclusive.

- When the interface is in L2 mode,
  it can only config the L2-ish config, like the below trunk config
```
      interface Ethernet1/3
        switchport
        switchport mode trunk
        switchport trunk native vlan 100
```

  The trunk config is equivalent to the below opeconfig path.
```
      interfaces/interface[name=eth1/3]/ethernet/switched-vlan/interface-mode  = TRUNK
                                                               native-vlan     = 100
```

- When the interface is in L3 mode.
  it can only config the L3-ish config, like the below ip config
```
      interface Ethernet1/3
        no switchport
        ip address 1.1.1.1/24
        ipv6 address 1::1/24
```

  The ip/ipv6 config is equivalent to the below opeconfig path.
```
      interfaces/interface[name=eth1/3]/subinterfaces/subinterface/ipv4/addresses/address/ip = 1.1.1.1
                                                                   ipv6/addresses/address/ip = 1::1
```


Now, for an openconfig user, the existing ‘name=eth1/3” and “type=ethernetCsmacd” could not differentiate the interface layer,so user cannot change the L2-ish / L3-ish freely.
If e1/3 is currently in the L2, then the user cannot config any L3-ish config via openconfig.
Vice-versa for e1/2 in L3.


There exist two current workarounds:

- Mixed model workaround
  The user needs to defer to the proprietary way to toggle e1/3, through vendor specific cli or device yang.
  Only then the user can proceed the change at the openconfig side.

- Inference workaround
  The target device needs to infer the user intent based on some heuristics like
  - if the request config the L2-ish path (ex interfaces/.../switched-vlan/...) then silently change the interface to L2.
  - if the request unconfig the L2-ish path, then revert the interface to L3.
 … etc
  However, this is pretty error-prone.
  As a single openconfig request can always mix all various L2 and L3 config.
  Netconf is even worse as it allows both config / unconfig to mix in a single payload.
  
These two workaround are far less than ideal for the usability perspective.
Thus we need the change.

LimeHat:
```
      interface Ethernet1/3
        no switchport
        ip address 1.1.1.1/24
        ipv6 address 1::1/24
```
Note that in OC, IP addresses always belong to subinterfaces (which is not the case in this cli example).

>If e1/3 is in the L2 layer, then sub-interfaces is not even supported.
In a specific implementation, perhaps.

In OC, subinterfaces can be used with L2 network instances.

I understand that the idea is to introduce "switchport/no switchport" equivalent to OC, but I'm not convinced it fits well into the overall architecture. There's a bit more basis for that for legacy-ish l2 switching that does not use subifs, but then there's already `interface-mode` that is related to this use case and covers this scenario to a certain extent. https://github.com/openconfig/public/blob/4f408783df925f4eb9d3bf7cd320c66cae4b6fb0/release/models/vlan/openconfig-vlan.yang#L177-L182

@robshakir any thoughts?

jorabada:
I agree with @LimeHat with all his comments. Another question would be: if an implementation needs this, can the interface `type` not be used? in the referred implementations it seems pointless to have `type ethernetCsmacd` and other types like `l2vlan` or `l3ipvlan` may provide the required distinction..

earies:
A few comments:

> Consider that subinterfaces is where OC defines properties for logical and layer 3 interfaces
> ...
> Note that in OC, IP addresses always belong to subinterfaces (which is not the case in this cli example).

Not entirely true - see: https://github.com/openconfig/public/issues/781

> Since this is not a submodule, we can start with 1.0.0

Any new model that has been introduced to date should have carried `0.1.0` and at some point of adoption/stabilization move to `1.0.0`

> in the referred implementations it seems pointless to have type ethernetCsmacd

Agreed - how many implementations do you have to dictate an ethernet interface to be ethernet?  Then validate to ensure the dictated type actually corresponds to the actual underlying physical hardware?  Or drop on the floor, do nothing with it but merely reflect it back...  For logical interfaces where you can dictate various encapsulations sure - also see: https://github.com/openconfig/public/issues/780

As an implementor (and cross referencing other implementations), this `type` has < 3% usage of the model identities defined so an implementation must go through an extra effort for true validations that fall outside of what can be expressed in the YANG language today in 1.0/1.1

> Today, without the “layer” property, the user needs to first use CLI or device YANG to change e1/3 to L3, then he/she can config sub-interfaces.

Not necessarily however this is also not a unique problem when translating between alien schemas and native constructs that mismatch in implementation behavior.  Translation logic could handle this but it comes at a cost - this is what happens when trying to normalize things that may vary greatly underneath - the 1:many and many:1 translation w/ additional checks then occurs which is derived off the intended payload.  Part of this is also that the model does not provide some of the guardrails in that a client can generate a schema compliant configuration that would ultimately fail on the device when applied.  This again is a common problem statement in all of this which also applies to the underlying native implementation as well.

> I understand that the idea is to introduce "switchport/no switchport" equivalent to OC, but I'm not convinced it fits well into the overall architecture.

And seems along the same lines as classical `ip routing` enablement...

We (Juniper) don't have this explicit concept but rather take paths based on "families" - essentially no precursor step necessary to dictate the mode

dplore:
Hi there, I am not seeing an operational use case for this change.  Can you provide some specific examples of configuration that is needed operationally and cannot be done without this change?  The two examples seem relatively straight forward for the NOS to understand if layer 2 switching or layer 3 routing is requested by the user:

```
      interfaces/interface[name=eth1/3]/ethernet/switched-vlan/interface-mode  = TRUNK
                                                               native-vlan     = 100
```

```
      interfaces/interface[name=eth1/3]/subinterfaces/subinterface/ipv4/addresses/address/ip = 1.1.1.1
                                                                   ipv6/addresses/address/ip = 1::1
```

The NOS should enforce that the interface in question cannot be configured with both of these snippets at the same time.

PR #1084, opened 2024-04-03
Add `ext-community-count` container and clarify that `community-count` does not include extended communities.

wenovus:
### Change Scope
* Added `ext-community-count` container.
* Major version bump to v8 for module `openconfig-bgp-policy`.

### Platform Implementations

 * JunOS: [link to documentation](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/policy-configuring-match-conditions-in-routing-policy-terms.html#:~:text=NOTE%3A%20The%20community%2Dcount%20attribute%20only%20works%20with%20standard%20communities.%20It%20does%20not%20work%20with%20extended%20communities)
 * Nokia CLI 7.156 in [this link](https://infocenter.nokia.com/public/7750SR212R1A/index.jsp?topic=%2Fcom.sr.classic%2Fhtml%2Fclassic_c_commands.html) includes all communities by default, but there is an option to select on one type of community, which is a feature that OpenConfig currently does not have.
 * Note Arista EOS supports this path per https://www.arista.com/en/support/toi/path-support?pn=EOS-4.32.1F. But there is no detailed explanation if it covers std and/or ext communities.

### Tree Diff
```diff
>        +--rw ext-community-count
>        |  +--rw config
>        |  |  +--rw operator?   identityref
>        |  |  +--rw value?      uint32
>        |  +--ro state
>        |     +--ro operator?   identityref
>        |     +--ro value?      uint32
```


dplore:
Hi @wenovus , do you have at least one additional reference you can find that indicates if external communities are counted or not?

wenovus:
> Hi @wenovus , do you have at least one additional reference you can find that indicates if external communities are counted or not?

I couldn't find anything for Cisco and Arista, but I found that Nokia's CLI by default actually includes all of them, but you can filter on one, a feature that OpenConfig currently does not have:

Jump to 7.156 in [this link](https://infocenter.nokia.com/public/7750SR212R1A/index.jsp?topic=%2Fcom.sr.classic%2Fhtml%2Fclassic_c_commands.html)

LimeHat:
Isn't configurable behavior a better choice in this case? `standard-community-count`, `extended-community-count`, etc.

Is the proposed clarification based solely on the default behavior of a single vendor?

wenovus:
> Isn't configurable behavior a better choice in this case? `standard-community-count`, `extended-community-count`, etc.
> 
> Is the proposed clarification based solely on the default behavior of a single vendor?

Thanks Sergey, indeed if this PR were to go ahead in its current form it would deviate from the CLI from Nokia. It would be good to get some inputs to see what way makes the most sense here.

I can see adding a bool for each of the community types, so `exclude-communities` and `include-ext-communities`. Note that for "communities" I'm proposing to include by default to minimize mis-use where no communities are included.

The issue is that OpenConfig currently doesn't have `large-communities`, so if we included large-communities by default, then it would amount to including communities that are not viewable by OpenConfig. To avoid this I think it makes sense to keep the default behaviour as communities only, but later on adding these bool flags to allow other communities to be included.

dplore:
Should we add ext-community-count as well?  Do you have a use case to count the extended communities @wenovus ?

wenovus:
> Should we add ext-community-count as well? Do you have a use case to count the extended communities @wenovus ?

I don't have one right now. Does it make sense to merge this PR with the intention of adding `ext-community-count` later on when the need arises?

dplore:
I think the key item we want to preserve and clarify is to support the operational use case in a clear and vendor neutral way. 
 The operational use case requires a count of standard communities as part of a bgp policy condition.  

My thought is updating the model to specify leafs for std and ext communities does this in a clear and vendor neutral way. 
 It is unambiguous about community type and vendor neutral in that multiple vendor implementation can adapt to the model, even if all vendors will not support both leafs.    

Note Arista EOS supports this path per https://www.arista.com/en/support/toi/path-support?pn=EOS-4.32.1F.  But there is no detailed explanation if it covers std and/or ext communities.

dplore:
Moving this to last call, to be merged on July 16, 2024.  

Please comment if this change is breaking in your currently shipping implementations or not.  If not, I would prefer to version this as a non-breaking change.

dplore (approved):

jhaas-pfrc:
Just a note that juniper has "community-count" that is only RFC 1997 communities, and has recently added "large-community-count".

PR #1085, opened 2024-04-04
Support gNPSI in OpenConfig

sallylsy:
### Change Scope
* Support Collectors for Generic Network Packet Streaming Interface(gNPSI)
* This is a new model which augment to system module and is backwards compatible.
```
module: openconfig-system
  +--rw system
     +--rw oc-sys-grpc:grpc-servers
        +--rw oc-sys-grpc:grpc-server* [name]
           +--ro oc-sys-grpc:connections
              +--ro oc-sys-grpc:connection* [address port]
                 +--ro oc-sys-grpc:address    -> ../state/address
                 +--ro oc-sys-grpc:port       -> ../state/port
                 +--ro oc-sys-grpc:state
                    +--ro oc-sys-grpc:address?    oc-inet:ip-address
                    +--ro oc-sys-grpc:port?       oc-inet:port-number
                    +--ro oc-sys-grpc:counters
                       +--ro oc-sys-grpc:bytes-sent?        oc-yang:counter64
                       +--ro oc-sys-grpc:packets-sent?      oc-yang:counter64
                       +--ro oc-sys-grpc:data-send-error?   oc-yang:counter64
```

### Current Use Case

* As a network engineer, I want to monitor the traffic with a gnpsi streaming connections to troubleshoot problems collecting data.  I want to observe the gNPSI server on the device  sending packets/bytes to clients over time.  In one implementation, the client (collector) creates a new gRPC connection (TCP stream) for each subscription.

* We would need to add support for per collector(client) stats since this information is essential for debugging purposes. Should there be a deviation on collector side, server stats can be used to compare with collector stats to culprit the issue. This is a similar path as sFlow per collector path

### Platform Implementations

    Augment grcp-server module
    Add support for bytes-sent, packets-sent, packets-error counters per client

Ref: https://github.com/openconfig/gnpsi


LimeHat:
Is the proposed model designed primarily for the dial-out case?

It appears that only dial-in is currently defined https://github.com/openconfig/gnpsi/blob/b27f152fa2ea9426c6306750b1e3d91d764c601e/proto/gnpsi/gnpsi.proto#L29-L34

dplore:
> Is the proposed model designed primarily for the dial-out case?
> 
> It appears that only dial-in is currently defined https://github.com/openconfig/gnpsi/blob/b27f152fa2ea9426c6306750b1e3d91d764c601e/proto/gnpsi/gnpsi.proto#L29-L34

Yes, it's my understanding that gNPSI is only intended for the dial-in scenario.  ie: where a collector opens the connection to a device.   It's not explicitly stated at https://github.com/openconfig/gnpsi/blob/main/README.md though. I'll open an issue there.

earies:
Few overall comments as this looks like it was refactored mostly to represent "state"

Much like the conversation in https://github.com/openconfig/public/pull/1060#discussion_r1513384531

* This is a gRPC service - I would propose we consolidate gRPC service related state into a common location enumerated per service - In this, we can also have per-client state/stats
* We need another GRPC_SERVICE identity defined here for GNPSI - this gRPC service enablement when relocated will handle the enable/disable of the service - not here and not with defaults
* Your port defaults is incorrect and not warranted here

I would propose dropping this PR as-is in favor of re-representing this as a gRPC service model

LimeHat:
Fully agree with the previous comment. 
Simply dropping the "configuration" part (which was done in the last commit `0a93b5`) does not transform this model from dial-out to dial-in.

dplore:
Hi @sallylsy ,  just helping out here.  Ebben and Sergey are recommending adding gnpsi to:
[/system/grpc-servers](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-grpc-servers) instead of `/sampling`.   

I think this makes sense as /system/grpc-servers is where we config other grpc services and report statistics on them.  If we were configuring how packet sampling is performed, then we would use /sampling

Regarding the port that is used for listening for gNPSI requests, that needs to be configurable.  We were not able to obtain an IANA reserved port specifically for gNPSI, so a default port should probably not be assigned.

sallylsy:
Thanks for the suggestions. I have added the gnpsi per collector counters under grpc-server. For the gnpsi config paths, we will use  existing paths in [/system/grpc-servers](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-grpc-servers)

earies:
I think the definition is still incorrect even after relocation - I will propose to hold on this until after the grpc-server structure is in for services - some reasoning behind this and the latest commit tree proposal

* You are rooting `collectors` directly under a grpc-server instance - This is not how gNPSI is designed and this is a generic term for a container at a location that will host a variety of services
* gNPSI is a request/response service - the term "collector" doesn't really apply here, there are no configured collectors as would be in dial-out services like NETFLOW/sFlow/Syslog, etc... - a client connects and receives a stream - address + port would be covered just by regular gRPC client state
* The same goes for bytes/packets counters - this can be a generic construct per "client" no matter the gRPC service

I will try to get around to at least getting the initial structure PR in by Monday

dplore:
> I think the definition is still incorrect even after relocation - I will propose to hold on this until after the grpc-server structure is in for services - some reasoning behind this and the latest commit tree proposal
> 
> * You are rooting `collectors` directly under a grpc-server instance - This is not how gNPSI is designed and this is a generic term for a container at a location that will host a variety of services
> * gNPSI is a request/response service - the term "collector" doesn't really apply here, there are no configured collectors as would be in dial-out services like NETFLOW/sFlow/Syslog, etc... - a client connects and receives a stream - address + port would be covered just by regular gRPC client state
> * The same goes for bytes/packets counters - this can be a generic construct per "client" no matter the gRPC service
> 
> I will try to get around to at least getting the initial structure PR in by Monday

A collector is a client that has opened a gRPC connection to the gNPSI server hosted on the device.  Its true, there are no collectors which are configured on the device.  The counters represent how much data the gNPSI server on the device has been sent to a particular client.   Perhaps we should use the more generic name `client` instead of `collector`?  Would that be more intuitive?

sallylsy:
I have refactor `collector` to `client`.

LimeHat:
>A collector is a client that has opened a gRPC connection to the gNPSI server hosted on the device. 

It will be a gRPC connection, on top of which a streaming gNPSI RPC will be opened. Things like client IP address and port numbers are not tied to gNPSI and should not be modelled as such; there should be a generic gRPC state `client` model that is extended with per-service containers.

sallylsy:
Based on the current discussion, we expended the generic gRPC state `clients` for each gRPC server. Proposed model is
```
module: openconfig-system
  +--rw system
     +--rw oc-sys-grpc:grpc-servers
     |  +--rw oc-sys-grpc:grpc-server* [name]
     |     +--rw oc-sys-grpc:gnpsi
     |     |  +--rw oc-sys-grpc:clients
     |     |     +--rw oc-sys-grpc:client* [address port]
     |     |        +--rw oc-sys-grpc:address    -> ../state/address
     |     |        +--rw oc-sys-grpc:port       -> ../state/port
     |     |        +--rw oc-sys-grpc:state
     |     |           +--rw oc-sys-grpc:address?    oc-inet:ip-address
     |     |           +--rw oc-sys-grpc:port?       oc-inet:port-number
     |     |           +--rw oc-sys-grpc:counters
     |     |              +--rw oc-sys-grpc:bytes-sent?      oc-yang:counter64
     |     |              +--rw oc-sys-grpc:packets-sent?    oc-yang:counter64
     |     |              +--rw oc-sys-grpc:packets-error?   oc-yang:counter64
     |     +--rw oc-sys-grpc:gnsi
     |     |  +--rw oc-sys-grpc:clients
     |     |     +--rw oc-sys-grpc:client* [address port]
     |     |        +--rw oc-sys-grpc:address    -> ../state/address
     |     |        +--rw oc-sys-grpc:port       -> ../state/port
     |     |        +--rw oc-sys-grpc:state
     |     |           +--rw oc-sys-grpc:address?    oc-inet:ip-address
     |     |           +--rw oc-sys-grpc:port?       oc-inet:port-number
     |     |           +--rw oc-sys-grpc:counters
     |     +--rw oc-sys-grpc:gnmi
     |     |  +--rw oc-sys-grpc:clients
     |     |     +--rw oc-sys-grpc:client* [address port]
     |     |        +--rw oc-sys-grpc:address    -> ../state/address
     |     |        +--rw oc-sys-grpc:port       -> ../state/port
     |     |        +--rw oc-sys-grpc:state
     |     |           +--rw oc-sys-grpc:address?    oc-inet:ip-address
     |     |           +--rw oc-sys-grpc:port?       oc-inet:port-number
     |     |           +--rw oc-sys-grpc:counters
     |     +--rw oc-sys-grpc:gnoi
     |     |  +--rw oc-sys-grpc:clients
     |     |     +--rw oc-sys-grpc:client* [address port]
     |     |        +--rw oc-sys-grpc:address    -> ../state/address
     |     |        +--rw oc-sys-grpc:port       -> ../state/port
     |     |        +--rw oc-sys-grpc:state
     |     |           +--rw oc-sys-grpc:address?    oc-inet:ip-address
     |     |           +--rw oc-sys-grpc:port?       oc-inet:port-number
     |     |           +--rw oc-sys-grpc:counters
```
The `counters` container will be customized for each gRPC server. Let me know if this makes more sense.

dplore:
This will be reviewed at Apr 23, 2024 OC operators meeting

LimeHat:
Hi @sallylsy , I don't see the latest changes reflect in yang?

The tree does look better. Perhaps we don't need to introduce `counters` to other services, at least for now (until there's a clear use case for those).

I would propose to change the definition of a `client`. In the output above, it is defined as a combination of IP and port, however, this does feel suboptimal with gRPC services.
- gRPC connection provides only a transport layer.
- There can be any number of independent clients multiplexed over a single gRPC connection, even for a single service (this is a common case with p4rt and gNMI/gNOI services); but even with gNPSI, in principle, nothing prevents you from opening three independent streams over the same gRPC connection (with the same ip/port combo).
- My proposal would be to introduce `uint id` field that will be the key for the client list (generated by the system). This will allow to uniquely identify each client in a multiplexed connection

sallylsy:
@LimeHat I have updated the yang file as you suggested. And the `counters` is only added for `gNPSI` for now. 

We still prefer to have the [address port] as key. One tcp client connection can use two protocols in this model.  And counters are unique to the protocol. We are not sure if refactoring to id actually adds value.

LimeHat:
> . One tcp client connection can use two protocols in this model. And counters are unique to the protocol. We are not sure if refactoring to id actually adds value.

Not sure If I follow. Can you elaborate on this?

The current description of the leafs specifies that the counters are maintained per client, e.g., "The total number of bytes sampled and sent to the client."

If you intent is to maintain some sort of global counter per-protocol, not per client, when this needs to be changed (along with the whole model once again)

Otherwise, it is simply not correct to model the client as a grpc connection. gRPC connection can exist standalone without any clients, or it can contain many clients. You cannot maintain per-client counters if you don't have unique identifiers for your clients.

LimeHat:
@earies 
>None of this is specific to gNPSI but rather can apply across all gRPC clients accessing various services. 
if you talk about the proposed counters, I would disagree.

There's a difference between the sampling stats and the general (per client or per connection) stats.

For example, the number of bytes sent to a client is different from the number of sampled bytes: gNPSI, in addition to sampled bytes, sends metadata (such as timestamp) to clients (even if we count only the payload). So I would argue this is service-specific data (which, traditionally, existed in the context of sflow/netflow).

LimeHat:
Also, speaking of the counters.

1. In #914, there was a proposal to add sampled bytes to the sflow model, but it didn't progress due to the lack of references.
2. `packets-error` seems to be completely new
    * it would be good to add some description of what exactly is expected to be counted here
    * Are there any references for such a metric? Perhaps in the context of sflow/ipfix, if gNPSI references are not available yet

sallylsy:
> Not sure If I follow. Can you elaborate on this?
> You cannot maintain per-client counters if you don't have unique identifiers for your clients.

Since we are proposing the per-client counters, we indeed need to make sure the client identifier is unique. My understanding is, the counters would have smaller granularity than clients. For example, a client could connect to gnpsi by both ipv4 and ipv6, on same <**port**>. The expected behavior should be collect statistics for ipv4+<**port**> and ipv6+<**port**> separately. If this is the goal, having unique identifier makes no different comparing having the combination of address+port. This behavior could provide more flexibility to client, enabling a client to get the statistics of a sum of both two protocol, or any of them.

earies:
> @earies
> 
> > None of this is specific to gNPSI but rather can apply across all gRPC clients accessing various services.
> 
> if you talk about the proposed counters, I would disagree.
> 
> There's a difference between the sampling stats and the general (per client or per connection) stats.
> 
> For example, the number of bytes sent to a client is different from the number of sampled bytes (even if we count only the payload): gNPSI, in addition to sampled bytes, sends metadata (such as timestamp) to clients. So I would argue this is service-specific data (which, traditionally, existed in the context of sflow/netflow).

Good point and agree.  It is true this then falls into the service specific bucket so not necessarily leverage a common "per-client" state/counters in general.

A few comments on this patch set and why I think we should align to https://github.com/openconfig/public/pull/1096

* This adds everything to the based grpc modeling for gNPSI.  gNPSI (and alot of gRPC services) are optional features that may or may not be for everyone and every use-case.  A design pattern that should be promoted is one of segmentation and augmentation.  I would even lobby that gNPSI, gNSI and potentially others when it comes to YANG data-modeling aspects should be collocated within that projects repo and augmenting into the OC tree from there
* The `port` definition for some reason indicates UDP port which is incorrect
* The client list keyed off `address` and `port` is indeed common and essentially needs to describe what a TCP connection (or future UDP/QUIC) entails to be unique.  There will be common connection state for a client (including streams) then there will be per service related.  Your proposal of a `uint id` is one method of abstracting this but you always have to learn this - a proactive query for "give me all the state for a client connected from IP address <x>" cannot be accomplished.
* As you will see in https://github.com/openconfig/public/pull/1096, I'd also like to start proposing the use of more and more "types" modules that do not contain subtrees and can be imported and used from a variety of modules.  Embedding anything gNPSI directly into core modules like this is an incorrect approach imo.  This also breaks the model used in P4RT and gRIBI currently
* As far as the service specific counters go, it's really "sampling" related but the description is combining it with what is sent to the client - we could end up w/ raw per client variants including all services, overhead, etc.. but for something like gNPSI, we can have a bucket of sampled (bytes/packets) _and_ whatever "error" is meant to be...

LimeHat:
@sallylsy 
I think we are looking at very different definitions of a client; and, perhaps, different definitions of gNPSI behavior in general.

> The expected behavior should be collect statistics for ipv4+<port> and ipv6+<port> separately. 

I don't see how that can happen, or what would be a reason for that.

One key point about any gRPC service is that the transport layer is abstracted. Unlike sflow, ipfix, and other existing dial-out sampling technologies, a client is not defined by a transport connection in the gRPC/gNPSI world.

If certain clients should receive only ipv4 samples, and other clients only ipv6 samples, or any other filtering is needed - it should become a part of the gNPSI service definition. The service definition is not tied to the transport layer. Any transport can be used to encapsulate any data. 

In my view, gNPSI client exists as long as there's an open stream of gNPSI service (`Subscribe` RPC) https://github.com/openconfig/gnpsi/blob/b27f152fa2ea9426c6306750b1e3d91d764c601e/proto/gnpsi/gnpsi.proto#L29 

Again, you can open any number of gNPSI streams over the same transport connection. You can have 1, 5, 10, 100 clients over the same transport connection (a single src/dst ip/port pair).

LimeHat:
Thanks, @earies; I haven't had a chance to review #1096, but I will get to it later.

>The client list keyed off address and port is indeed common and essentially needs to describe what a TCP connection (or future UDP/QUIC) entails to be unique. There will be common connection state for a client (including streams) then there will be per service related. Your proposal of a uint id is one method of abstracting this but you always have to learn this - a proactive query for "give me all the state for a client connected from IP address " cannot be accomplished.

I think we need to draw a clear line between a gRPC connection and a service client.
Indeed, you can model the connection as an IP/port combo. However, this method cannot be used to model svc clients; multiplexing (multiple clients over the same connection) immediately breaks that schema.
My proposal is not to abstract the connection under the id, my proposal is to abstract service clients.

sallylsy:
@LimeHat Thanks for the explanation. I agree that we want to differentiate the gRPC server connection (address+port) and a server client. I wonder what is your proposal to get the identifier per client. What information other than address and port (maybe we don't need them at all)?

sallylsy:
I have updated our use case in the PR description. As per

> I would propose to change the definition of a client. In the output above, it is defined as a combination of IP and port, however, this feels suboptimal with gRPC services.

Re-define the client would be a suboptimal step. Current definition also fulfill current use case for gNPSI. I would suggest to keep the current definition of client for now.

LimeHat:
>I wonder what is your proposal to get the identifier per client. What information other than address and port (maybe we don't need them at all) we need?

The system itself should identify the clients, uint is enough for this case.

>1. For each client, we return the aggregated statistic. Basically, we don't serve statistic per client, which can be done by the current proposal. 

In that case, 
1) you need to change the proposal to remove any per-client wording
2) together with the wording, you'll likely need to change the counters that you want to monitor
3) are we sure this is meaningful?

> 2. Server proto needs to provide information. Each client needs to provide unique identifier to the gRPC request. This is more complex issue because gRPC server needs to figure out a way to ensure the uniqueness of each client. The second idea will definitely bring more work other than the OC model change.

No, that's not always true. A NOS is perfectly capable of generating unique IDs for clients if we stick to the definition that a client is an RPC. 
This is nothing new; there are implementations that work like this. Take, for example, the p4rt protocol (I'm pretty sure @dplore has a reference for this exact scenario). While the p4rt client state does not have an OC model, each implementation needs to keep track of clients (StreamChannel RPCs) that have associated service-specific properties (e.g. election-id and role) and global properties (e.g. ip/port/username)

The id in proto is required only when you want to have client-controlled identification of the clients (which allows, for example, to associate multiple different RPCs with a single client).

LimeHat:
If you want to avoid doing per-RPC/per-client state, in my opinion, you have two viable options:

1) introduce global gNPSI counters outside of gRPC-servers context, perhaps to `/sampling` tree.
something similar is currently available with sflow (per interface sampled packets):
https://github.com/openconfig/public/blob/a2c7a09e619c56782b3f4495a065fd7c34ef1175/release/models/sampling/openconfig-sampling-sflow.yang#L74-L82

These counters are not associated with a client or a connection.

2) Introduce a generic, per-connection gRPC state (e.g., bytes sent) that is not specific to gNPSI sampling. That will be just plain gRPC connection monitoring data, not related to any service in particular.

sallylsy:
Hi @LimeHat, I think a generic, per-connection gRPC state is not something useful for our use case. The reason we want to have the per collector counter is:

>We would need to add support for per collector(client) stats since this information is essential for debugging purposes. Should there be a deviation on collector side, server stats can be used to compare with collector stats to culprit the issue. This is a similar path as sFlow per collector path

The sflow counters you pointed out is at per port level, which dose fulfill the user case we need as well. 

@dplore found there is specific client counter relevant for specific protos under `/system`. For example: https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-grpc-servers-grpc-server-acctz-counters-state-client-counters-idle-timeouts. From my point of view, we want to fulfill our user cases and follow the current pattern in OC model. I think we are in the good shape with the current proposal.

LimeHat:
>From my point of view, we want to fulfill our user case

Can you elaborate on the use cases?

>@dplore found there is specific client counter relevant for specific protos under /system.

These counters are not per client. They are global, include all clients, and are not associated with a specific client or a connection (just as I indicated in opt#1 in my last comment).

>  I think we are in the good shape with the current proposal.

I'm sorry but I cannot agree with that. Modeling a gNPSI client as a transport connection is incorrect for the reasons we already discussed.

dplore:
> > From my point of view, we want to fulfill our user case
> 
> Can you elaborate on the use cases?

See the description (first comment) for this PR for the latest use case description.

> These counters are not per client. They are global, include all clients (of the specific service), and are not associated with a specific client or a connection (just as I indicated in opt#1 in my last comment).
> 
> > I think we are in the good shape with the current proposal.
> 
> I'm sorry but I cannot agree with that. Modeling a gNPSI client as a transport connection is incorrect for the reasons we already discussed.

The use case is to monitor the gnpsi server to "collector" transport (packets).  We do not require  RPC level monitoring.  So counting all the packets a gnpsi server has sent on a tcp connection meets the use case and is the intent for this model.  
 This is considerably more important to track for gnpsi because we expect it can send large amounts of data for long periods of time.  (much greater than even gnmi).

LimeHat:
> So counting all the packets a gnpsi server has sent on a tcp connection meets the use case and is the intent for this model.

That is fine, but the model needs to be refactored to talk specifically about gRPC connections. The current proposal doesn't do that.  Valid options are:
1) per client state modeling, where each gNPSI client is an RPC
2) per connection state modeling, which is not directly tied to gNPSI
3) per service state modeling, which is tied to gNPSI but not to individual connections and clients

The current model mixes p. 1 and p. 2 and that's not a correct approach in my view.

dplore:
@sallylsy I suggest we use the word 'connection' in the model instead of client and also define in the yang description that the connection refers to the TCP layer transport.  This will help clarify that we are not talking about gRPC clients which are often associated with RPC level communications (for example in the [grpc core concepts](https://grpc.io/docs/what-is-grpc/core-concepts/) )

release/models/system/openconfig-system-grpc.yang
```diff
@@ -97,10 +110,103 @@ module openconfig-system-grpc {
...
+            description
+              "Operational data for gNPSI counters.";
+            uses gnpsi-counters-top;
+          }
+        }
+      }
+    }
+  }
+
+  grouping gnpsi-counters-top {
```
earies, line 167: As I've mentioned previously, suggest following the pattern in https://github.com/openconfig/public/pull/1096 for segmented model design.

gNPSI is a separate optional feature so it should be segmented as such and not included at all within the base gRPC modeling.  There can be many features that have gRPC services along w/ types that can be reused among other models

The intent of the restructuring PR is to provide commonality and reuse where appropriate along w/ well defined demarcations where one can support/not-support model augments vs. needing to produce deviations to a monolith model

release/models/gnpsi/openconfig-gnpsi.yang
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    leaf port {
+      type oc-inet:port-number;
+      description
+        "UDP port number for the gNPSI connection.";
+    }
+  }
+
+  grouping connections-top {
```
LimeHat, line 74: There are no gNPSI connections; there are only gRPC connections with gNPSI streams (clients) running on top of them. If, as @dplore suggested, the intention is to expose grpc/tcp state information, then all of this should be modeled as generic gRPC state outside of gNPSI model.

LimeHat (changes requested):

release/models/gnpsi/openconfig-gnpsi.yang
sallylsy: Revised. Connections are in generic gRPC state now :)

release/models/gnpsi/openconfig-gnpsi.yang
```diff
@@ -0,0 +1,96 @@
...
+        client.";
+    }
+
+    leaf packets-error {
+      type oc-yang:counter64;
+      description
+        "The total number of error packets failed to be sampled
+        and sent to the client.";
+    }
+  }
```
LimeHat, line 85: packet/byte counters should also be moved to the gRPC state (and the error counter should probably be removed) since they're not specific to gNPSI streams anymore

LimeHat (changes requested):

release/models/gnpsi/openconfig-gnpsi.yang
sallylsy: What about move packet/byte counters to gRPC state and leave error counter for gNPSI. gNPSI needs error counter

dplore: agreed, the `packets-error` leaf is gnpsi specific

LimeHat: Could you please elaborate on how this error counter is supposed to work?
What exactly is counted as an error?

sallylsy: Any error when writing into gRPC stream would be considered as a packets-error.

Ref: https://github.com/search?q=repo%3Aopenconfig%2Fgnpsi%20IncrementWriteErrorCount()&type=code

sallylsy: I have updated the PR as you suggested :)

release/models/system/openconfig-system-grpc.yang
sallylsy: PR is restructured per reference of  https://github.com/openconfig/public/pull/1096

release/models/gnpsi/openconfig-gnpsi.yang
```diff
@@ -0,0 +1,70 @@
...
+    description
+      "Top-level container of operational data for gNPSI counters.";
+
+    leaf packets-error {
+      type oc-yang:counter64;
+      description
+        "Error packets/datagram failed to be sent out by gNPSI
+        service for a specific client.";
+    }
+  }
```
dplore, line 57: Reading the [gnpsi server reference implementation](https://github.com/search?q=repo%3Aopenconfig%2Fgnpsi%20IncrementWriteErrorCount()&type=code), I suggest this name and description.  

```suggestion
leaf sample-send-error {
      type oc-yang:counter64;
      description
        "A count of errors the gnpsi server encountered when sending a packet sample to a grpc client.";
    }
  }
```

dplore (approved):

release/models/system/openconfig-system-grpc.yang
```diff
@@ -200,6 +208,92 @@ module openconfig-system-grpc {
...
+    leaf address {
+      type oc-inet:ip-address;
+      description
+        "IPv4/IPv6 address of the gRPC server connection.";
+    }
+
+    leaf port {
+      type oc-inet:port-number;
+      description
+        "UDP port number for the gRPC server connection.";
```
LimeHat, line 251: gRPC, in most cases, doesn't use udp

release/models/gnpsi/openconfig-gnpsi.yang
```diff
@@ -0,0 +1,70 @@
...
+
+  grouping gnpsi-counters-top {
+    description
+      "Top-level container of operational data for gNPSI counters.";
+    leaf flow-send-error {
+      type oc-yang:counter64;
+      description
+        "A count of errors the gnpsi server encountered when sending a
+        flow message to a grpc client.  In sflow this is an sflow datagram.
```
LimeHat, line 54: I'm still a bit puzzled by this.

If this is a generic error encountered when attempting to send a packet via a gRPC stream, why is it gNPSI specific?

Moreover, in the lemming reference that @sallylsy provided, the stream is closed immediately after encountering an error. Which, in many cases, will lead to the closure of the connection as well (and hence this data will not be available in telemetry).

Could you please elaborate on the purpose of this counter, and are there any real implementations of it?

LimeHat (changes requested):

release/models/gnpsi/openconfig-gnpsi.yang
dplore: I discussed with Sally and gnpsi team.  It does seem like the gnpsi error sending flow message may be generalized as a error sending any grpc data.  I think it's up to the server implementation to decide if a send() error should be retried or the connection is closed.  But incrementing a counter that an error on send() occurred seems applicable to any grpc service.  The latest commit updates this.

dplore (approved):
Thanks for adding the GNSPI identity back in.  LGTM

earies:
I have to say at this point that the OpenConfig set of projects needs to implement a proper governance model as we are seeing more merges without proper attention or agreement.  I've mentioned multiple times to collaborate along w/ a draft sitting in https://github.com/openconfig/public/pull/1096 and that embedding gNPSI directly into the core gRPC model is not desirable - detach and modularize and really consider the impacts if this is not done.  P4RT is already done this way.

If we walk back a bit to even the question of if gNPSI a good or feasible idea - that is debatable and may very well need to be proven out given the fact we are encapsulating something more efficient into something less efficient in various architectural and deployment scenarios.  That being the case calls for proper demarcation otherwise what you end up w/ this design pattern is forced compliance to a model-set for 1 set of features while needing to produce deviations if not all endpoint services are supported.

If this is detached as proposed in the above PR the the demarcations lay onto domains and modules of which can either ship or not ship as proper augments.

I suggest reconsidering this merge and instituting a proper governance model for everyones benefit - this also includes removing a divide of approvals towards true transparency.  I will add this specific item to the upcoming OC community call agenda.

earies:
Apologies for a bit of misread on the latest commits that ended up as the merge - I do see a slight shift here to a proper demarcation however we have a few additional issues looking closer

* This PR is no longer about gNPSI support (other than an identity/module introduction)  We now have an incorrect PR title and incorrect git commit log as far as intent/completion
* This PR is rather about gRPC connection state (more on that later)
* This new gNPSI identity is now stubbed out and unused as is the base identity.  The `services` leaf-list points to the old identity so at most this is prep work for follow-on commits from a gNPSI perspective.
* The container `connections` is flagged as `config false` but nested underneath a container with this flag thus redundant
* Regarding the leaf `data-send-error`, it seems this started off as a ref to gNPSI reference implementation code but to be clear, this is to now be generalized as any write error to a connected socket correct?  If the client disconnects as any result, this state is ephemeral in nature

On client connections and associated counters in general, I think we have a few items up for discussion here... The `grpc-server` modeling accommodates a list of servers that have services mapped to each server.  In _most_ situations, what we'll probably see is either 1 server hosting all services or a split of some services distributed to different servers with what should be different profiles (listening port, network-instance, etc.. esp. as some of these services have IANA registered ports).  The model does however also allow for the same service to be distributed across multiple servers which now raises how counters and state should be truly categorized.

An implementation can choose to front end such services by way of a reverse proxy fanning out to various backend services.  Now, one instance of a service could very well be reachable by > 1 grpc-server endpoints (on the proxy).  When we talk about connections keyed by `address` and `port`, this more of the raw TCP connection.  To be more specific, we are talking about _source_ address and port of the connecting client (leaf and descriptions should probably specify this) - Essentially then this state needs to be tracked on the listening endpoint on a specific port to pair up.

As we start to consider "per-service" state, we are going to have to consider the above scenario where 1 service could be distributed across multiple servers thus raising global vs. per-server state (where counters and state could very well then be identical across multiple server instances)

In either case, there are still unaddressed items and a mismatch of intent for the original PR

dplore:
Hi, I updated the PR title to better reflect the content.  Note the git log shows the old title, but the comments are hopefully a little more clear:

```
commit 3c810f3b1168e74de473d0d0592151cfbffd27b5 (HEAD -> master, origin/master, origin/HEAD)
Author: sallylsy <108024270+sallylsy@users.noreply.github.com>
Date:   Thu May 30 16:09:52 2024 -0700

    Support gNPSI in OpenConfig (#1085)
    
    * (M) release/models/gnpsi/openconfig-gnpsi-types.yang
    * (M) release/models/grpc/openconfig-grpc-types.yang
    
    * Add gRPC connections and statistics in OpenConfig
    * Add gNPSI as a gRPC service type
    
    ---------
    
    Co-authored-by: Darren Loher <dloher@google.com>
```

Regarding #1096  I do like the style of using separate files with augments to create models with some modularity.  It is a big change and the community will need to do some impact analysis on it.  I'll add comments there.

Regarding the client counters,  gnpsi is early in it's development and is the operational use case for grpc connection counters.  I follow what you are saying about having a reverse proxy multiplexing the connections for multiple grpc services, standalone services and a combination of both.  On the surface, it seems like this model could accommodate all the above scenarios.    FWIW, in the current target implementation, there is a separate grpc listener running gnpsi.   

We can iterate on the current model to accommodate more implementations.

nkitchen:
On a superficial read, the declaration of the `GNPSI` identity looks like it has an error:
```
  identity GNPSI {
    base oc-grpct:GRPC_SERVICE;
    description
      "gNMI: gRPC Network Management Interface";
  }
```
Should the description have referred to gNPSI instead of gNMI?

earies:
> On a superficial read, the declaration of the `GNPSI` identity looks like it has an error:
> ```
>   identity GNPSI {
>     base oc-grpct:GRPC_SERVICE;
>     description
>       "gNMI: gRPC Network Management Interface";
>   }
> ```
> Should the description have referred to gNPSI instead of gNMI?

That is correct as well as the identity is using an unused/unrefd base identity.

dplore:
> > On a superficial read, the declaration of the `GNPSI` identity looks like it has an error:
> > ```
> >   identity GNPSI {
> >     base oc-grpct:GRPC_SERVICE;
> >     description
> >       "gNMI: gRPC Network Management Interface";
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should the description have referred to gNPSI instead of gNMI?
> 
> That is correct as well as the identity is using an unused/unrefd base identity.

@sallylsy can you submit a fix for this?

sallylsy:
> Copied

Fix is in https://github.com/openconfig/public/pull/1145

Issue #1066, opened 2024-03-09
Add media-channel expected-power to openconfig-channel-monitor


kcroussore:
Currently under the openconfig channel-monitor model, there is a leaf for measured power:

https://github.com/openconfig/public/blame/815341544e58548526e7f9bec4e0ed2d615b69ac/release/models/optical-transport/openconfig-channel-monitor.yang#L163

Power is reported as integrated power between lower-frequency and upper-frequency.
This allows system vendors to implement optical channel monitors which report per-media-channel optical powers, based on configured media-channel frequency ranges.

To supplement this, it would be helpful to add an additional leaf, "expected-power", which reports the optical power expected for a media-channel at the same system measurement point.
This would enable operators to easily identify discrepancies between measured and expected optical powers within the same OpenConfig model (channel-monitor).

A proposed implementation would look like:

leaf expected-power {
      type decimal64 {
        fraction-digits 2;
      }
      units dBm;
      description
        "Expected optical power over the specified spectrum";
    }
  }

In this proposal, both "power" and "expected-power" are "state" data (i.e. "expected-power" is not configured under the channel-monitor model).

System vendors can easily calculate expected powers based on wavelength-router port-spectrum-power-profiles, for example, defined at WSS input and output ports and optical amplifier line out ports.

dplore:
Can you comment on the reason that `expected-power` should be configured on the device?  My initial reaction is the expected power would be stored offline, in an NMS system which was measuring the actual values, doing the comparison and so on.  Another question might be, is this value expected to used as an alarm threshold?

kcroussore:
@dplore the intent here is to provide a leaf under /channel-monitor.../ where the system can report both "power" and "expected-power" at the same measurement point.
I slightly updated the original comment to clarify that the intention is to add "state" data, not configuration.

The general use case is something like:
1) Operators configure a wavelength-router port-spectrum-power-profile (PSPP), e.g. at a WSS input port. A set of media-channels is also configured on the same port.
2) System vendors already internally calculate the expected power per media-channel based on the defined PSPP (this is needed for optical channel control).
3) System vendors already stream "power" for individual media-channels, under /channel-monitor.../power/.
4) This proposal would allow system vendors to add "expected-power" under the same telemetry path.
5) Operators then have a direct comparison of "power" and "expected-power" without having to do the calculation in the operators own SDN controller.

So this should simplify the use case where operators want both expected and measured power at a system measurement point (which is useful for troubleshooting).

[...]

PR #1087, opened 2024-04-04
Add media-channel expected-power to openconfig-channel-monitor

kcroussore:
Reference: https://github.com/openconfig/public/issues/1066

### DIFF
--- a/release/models/optical-transport/openconfig-channel-monitor.yang
+++ b/release/models/optical-transport/openconfig-channel-monitor.yang
@@ -26,7 +26,7 @@ module openconfig-channel-monitor {
     channel monitor (OCM) for optical transport line system
     elements such as wavelength routers (ROADMs) and amplifiers.";
 
-  oc-ext:openconfig-version "0.4.1";
+  oc-ext:openconfig-version "0.5.0";
 
   revision "2024-04-04" {
     description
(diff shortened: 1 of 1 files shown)


release/models/optical-transport/openconfig-channel-monitor.yang
```diff
@@ -26,7 +26,13 @@ module openconfig-channel-monitor {
     channel monitor (OCM) for optical transport line system
     elements such as wavelength routers (ROADMs) and amplifiers.";
 
-  oc-ext:openconfig-version "0.4.0";
+  oc-ext:openconfig-version "0.4.1";
```
dplore, line 29: Since this is adding a leaf and is backwards compatible, the minor revision should be updated (per https://semver.org/ versioning standards followed by OC)

```suggestion
  oc-ext:openconfig-version "0.5.0";
```

release/models/optical-transport/openconfig-channel-monitor.yang
```diff
@@ -26,7 +26,13 @@ module openconfig-channel-monitor {
...
+
+  revision "2024-04-04" {
+    description
+      "Add new leaf media-channel target-power.";
+    reference "0.4.1";
```
dplore, line 34:
```suggestion
    reference "0.5.0";
```

dplore (changes requested):

release/models/optical-transport/openconfig-channel-monitor.yang
kcroussore: Done, thanks for the pointer to the versioning standard.

dplore (approved):

Issue #1067, opened 2024-03-09
OpenConfig Support for Configurable Media-Channel Compensation Range and ASE recovery hysteresis


kcroussore:
This proposal would add 3 configurable parameters to the existing [openconfig-wavelength-router](https://github.com/openconfig/public/blob/master/release/models/optical-transport/openconfig-wavelength-router.yang) model (listed below with proposed names):
A leaf:  max-undershoot-compensation
A leaf: max-overshoot-compensation
A leaf: ase-recovery-hysteresis

The new parameters would be added in the following path:
“wavelength-router/media-channels/channel[index]/config/”, similar to the existing parameter “attenuation-control-mode”.

The purpose of the new parameters is to enable the operator to determine the details of how media-channel power compensation is performed.
Specifically, the operator can prescribe the amount of channel power overshoot and undershoot that should be actively compensated.
Additionally, operators can prescribe the hysteresis value used to determine if a channel should be added initially, or reverted after being switched to ASE due to low power.

Under the current model, these implementation details must be specified using vendor native models (in addition to configuration which can be set directly through OpenConfig).
The proposed model update will transfer ownership of media-channel optical control to OpenConfig directly.
The new parameters would be optional, allowing existing operator implementations which rely partially on native vendor models to be preserved.

Below is a summary of model additions that will be proposed.
These parameters will be applied on a per-media-channel basis (i.e. they are set on every media-channel and can be different across media-channels).

 // identity statements
  identity ATTENUATION_CONTROL_RANGE {
    description
      "The definition for different types of modes to specify attenuation control range settings on a media channel.";
  }

  identity CONTROL_RANGE_FULL {
    base ATTENUATION_CONTROL_RANGE;
    description
      "Media channel power compensation range is not constrained, and limited by available WSS dynamic range. Operator does not explicitly set the undershoot and overshoot compensation values.";
  }

  identity CONTROL_RANGE_LIMITED {
    base ATTENUATION_CONTROL_RANGE;
    description
      "Media channel undershoot and overshoot compensation ranges are defined by max-undershoot-compensation and max-overshoot-compensation.";
  }

leaf attenuation-control-range {
      type identityref {
        base ATTENUATION_CONTROL_RANGE;
      }
      description
        "Sets the attenuation control range for the media channel. The control range defines the amount of media channel undershoot and overshoot compensation that will be applied when performing power targeting to meet the defined output port spectrum power profile";
    }
  }

New optical control parameters:

"max-undershoot-compensation"
[...]

PR #1090, opened 2024-04-08
Add OpenConfig Support for Configurable Media-Channel Compensation Range and ASE recovery offset #1067

kcroussore:
### Change Scope

* Reference: [#1067 ](https://github.com/openconfig/public/issues/1067)

### DIFF
--- a/release/models/optical-transport/openconfig-wavelength-router.yang
+++ b/release/models/optical-transport/openconfig-wavelength-router.yang
@@ -591,8 +591,8 @@ module openconfig-wavelength-router {
       description
         "Defines the optical power offset value relative to the
         ASE injection threshold at which point ASE should be removed
-        and the media-channel should be injected. Requires 
-        ase-injection-delta or ase-injection-threshold to be 
+        and the media-channel should be injected. Requires
+        ase-injection-delta or ase-injection-threshold to be
         configured";
     }
 
--- a/release/models/optical-transport/openconfig-wavelength-router.yang
+++ b/release/models/optical-transport/openconfig-wavelength-router.yang
@@ -622,7 +622,7 @@ module openconfig-wavelength-router {
       type decimal64 {
         fraction-digits 2;
       }
-      when "../attenuation-control-range = ‘CONTROL_RANGE_LIMITED’";
+      when "../attenuation-control-range = 'CONTROL_RANGE_LIMITED'";
       units dB;
       description
         "Defines the maximum allowable WSS attenuation adjustment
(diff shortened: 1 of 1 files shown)


release/models/optical-transport/openconfig-wavelength-router.yang
```diff
@@ -554,6 +583,20 @@ module openconfig-wavelength-router {
...
+        fraction-digits 2;
+      }
+      units dB;
+      description
+        "Defines the optical power offset value relative to the
+        ASE injection threshold at which point ASE should be removed
+        and the media-channel should be restored.
+        Also specifies the optical power offset relative to
+        injection threshold in dB above which a media channel will
+        be initially added to the system";
```
dplore, line 597: Perhaps just an academic comment, but is this typically called hysteresis or a threshold in the optical industry?  I'm more used to seeing hysteresis used for 'debounce' values and/or when there is a time delay involved.

dplore (commented):
LGTM from a yang point of view.  @ejbrever can you review for optical content?

release/models/optical-transport/openconfig-wavelength-router.yang
```diff
@@ -554,6 +583,20 @@ module openconfig-wavelength-router {
...
+    leaf ase-recovery-hysteresis {
+      type decimal64 {
+        fraction-digits 2;
+      }
+      units dB;
+      description
+        "Defines the optical power offset value relative to the
+        ASE injection threshold at which point ASE should be removed
+        and the media-channel should be restored.
+        Also specifies the optical power offset relative to
```
ejbrever, line 595: I wasn't sure I understood this second sentence as well. It seems the "recovery" hysteresis would control the initial addition of the media channel as well? In that sense is recovery the right name here? I was looking back at the doc when we were discussing this and it looks like originally you were thinking of calling this "ase-injection-hysteresis" although I agree in changing it as I'm not clear that is the right name here either.

Apologies, I could definitely be misunderstanding, but I think this is about when ASE should be removed and the media channel added. Would something like "ase-removal-hysteresis" be more explicit? Or flipping things around since we already have "ase-injection-threshold" would "media-channel-injection-hysteresis" fit well?

ejbrever: To Darren's point below, maybe instead of hysteresis "offset" could be an alternative?

kcroussore: @dplore @ejbrever thanks for the review and comments. I tried to clarify this and liked the idea of calling this "media-channel-injection-hysteresis" instead, but it made sense to me to remove hysteresis and use "offset" instead.
Please take a look at the updated name and description and we can adjust further if needed.

ejbrever: Looks good, thanks.

ejbrever (approved):
LGTM

kcroussore:
@dplore I believe I got my formatting mistakes cleaned up here - can you take a look and if everything is good, add review?

dplore (approved):
LGTM.

dplore:
Last call for comments on this change.  We will review on April 16, 2024 OC Operators meeting before merging.

dplore:
Reviewed today in OC Operators call without objection.

PR #1091, opened 2024-04-09
Add storage counters related to errors

dplore:
### Change Scope

* Add `/components/component/storage/state/counters/` as a container to represent storage device counters
* Add a few select leaves into this container based on common/often available [SMART telemetry](https://en.wikipedia.org/wiki/Self-Monitoring,_Analysis_and_Reporting_Technology) which are focused on detecting storage device errors.
* Since this is adding leafs, this change is backwards compatible

### Operational use case
While rare, in a large population of devices storage errors have led to a device becoming unhealthy, unable to accept software updates or unable to boot due to non-volatile media (flash, ssd media) errors.  This is a counter to be able to measure the accumulation of storage areas as an statistic for storage component health.

Note, `/components/component/healthz/state/status` is also a useful data point, but as a boolean only value, it is very coarse.  Storage counters can be used to predict a storage device will fail in the future.  

### Tree View
```diff
module: openconfig-platform
  +--rw components
     +--rw component* [name]

[... snip ...]

         +--rw storage
         |  +--rw config
         |  +--ro state
+        |     +--ro oc-storage:counters
+        |        +--ro oc-storage:soft-read-error-rate?                  oc-yang:counter64
+        |        +--ro oc-storage:reallocated-sectors?                   oc-yang:counter64
+        |        +--ro oc-storage:end-to-end-error?                      oc-yang:counter64
+        |        +--ro oc-storage:offline-uncorrectable-sectors-count?   oc-yang:counter64
+        |        +--ro oc-storage:life-left?                             uint8
+        |        +--ro oc-storage:percentage-used?                       uint8
```

### Platform Implementations

* Linux  smartmontools - https://www.smartmontools.org/
* Cisco [IOS XR implements syslog messages](https://www.cisco.com/en/US/docs/ios_xr_sw/error/messages/ios-xr-sem-guide.html) for `MEDIA` errors
* Arista EOS exposes [storage related errors via dmesg and console](https://arista.my.site.com/AristaCommunity/s/article/Troubleshooting-Switch-Failure#Comm_Kna_ka02I000000brT9QAI_22)


sulrich:
## pyang tree output

```
   +--ro mount-points
     |  +--ro mount-point* [name]
     |     +--ro name     -> ../state/name
     |     +--ro state
     |        +--ro name?                string
     |        +--ro storage-component?   -> /oc-platform:components/component/name
     |        +--ro size?                uint64
     |        +--ro available?           uint64
     |        +--ro utilized?            uint64
     |        +--ro counters
     |           +--ro io-errors?   uint64
```

release/models/system/openconfig-system.yang
```diff
@@ -372,6 +378,24 @@ module openconfig-system {
...
+    }
+
+  }
+
+  grouping system-mount-point-error-counters {
+    description
+      "Counters for system memory errors.";
+
+    leaf io-errors {
+      type uint64;
```
earies, line 395: Likely rather `oc-yang:counter64`

earies (commented):
Could you elaborate precisely what something like this error counter would map to for an underlying OS (e.g. Linux) implementation?

This also is attempting to categorize per mount point vs. device (block)

release/models/system/openconfig-system.yang
dplore: Thanks, fixing this and moving these under /components/component/storage/state/ with descriptions for each leaf

dplore:
Addressed comments.  This is now ready for review.

release/models/platform/openconfig-platform.yang
```diff
@@ -65,7 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.24.0";
+  oc-ext:openconfig-version "0.25.0";
```
earies, line 68: There are no edits to `openconfig-platform` or its submodule `openconfig-platform-common` and `openconfig-platform-storage` is entirely new and augments into the tree so I'm curious why versions are being touched on any modules other than the new module?

release/models/platform/openconfig-platform-storage.yang
```diff
@@ -0,0 +1,75 @@
...
+  oc-ext:catalog-organization "openconfig";
+  oc-ext:origin "openconfig";
+
+  // identity statements
+  // typedef statements
+  // grouping statements
+  grouping storage-state {
+    description
+      "Storage component state.  These counters are derived from the
+       linux kernel block layer statistics in /sys/block/<dev>/stat.";
```
earies, line 52: Thx for updating this.  A few comments:

* Confirming that since this is the first use of `/components/component/storage` that it is now expected that each storage device (at least block dev) such as those enumerated w/ `lsblk` be represented as a component of type `STORAGE` w/ a respective parent of the physical component housing the storage device (The relationship of containers to respective types is not drawn out in the modeling but rather an implication of when such containers should be in use depending on the type of component)
* This and various other models will often assume and imply an underlying Linux OS which while in most cases is true, is not guaranteed but it would be best to then blanket such implementation assumptions with Linux being the mandatory underlying OS

release/models/platform/openconfig-platform-storage.yang
```diff
@@ -0,0 +1,75 @@
...
+    reference
+      "https://www.kernel.org/doc/Documentation/block/stat.txt";
+
+    leaf io-errors {
```
earies, line 56: This augments directly into `/components/component/storage/state` but it is a counter value.  The design of other OC modeling would nest this under a `counters` container

release/models/platform/openconfig-platform-storage.yang
```diff
@@ -0,0 +1,75 @@
...
+      type oc-yang:counter64;
+      description
+        "This values increments when an I/O request completes with a
+         failure.  This value corresponds to 'discard I/Os' om the linux
```
earies, line 60: s/om/on/

release/models/platform/openconfig-platform.yang
dplore: Fixed

release/models/platform/openconfig-platform-storage.yang
dplore: We'll introduce a [featureprofiles test ](https://github.com/openconfig/featureprofiles) which will reference this path and may provide additional expectations on related leaves, such as `parent`.  

We don't always have detailed expectations for all possible related leaves.  At the moment, we don't have an operational use case to model the component tree for storage devices, so I expect this to remain undefined.

Regarding the linux reference, I'll loosen the description slightly to state that linux is a reference example and should be used if applicable.  Other implementations count I/O operations which cannot be completed due to an issue with the storage device.

release/models/platform/openconfig-platform-storage.yang
dplore: Thanks for this.  I also prefer using the counters container as most of OC does this.

release/models/platform/openconfig-platform.yang
dplore: Actually, our CI check catches this as a module revision must be in sync with a submodule (openconfig-platform-common.yang ) needing to be in sync 

Ref:
https://gist.github.com/OpenConfigBot/15bc064de0245e83f65bad75c3870f35

dplore:
This was reviewed in Apr 9, 2024 OC Operators meeting without objection.  Addressed latest comments and now placing this on last-call for merge on May 21, 2024

LimeHat:
>       "This values increments when an I/O request completes with a
>         failure.  This value corresponds to 'discard I/Os' on the linux
>         kernel block layer statistics.

Is there a reference that can confirm that statement?

My understanding is that "discard i/o" is just another type of i/o operation, which is often used with SSD drives (see also [fstrim man](https://man7.org/linux/man-pages/man8/fstrim.8.html)); and not an error.

dplore:
> > ```
> >   "This values increments when an I/O request completes with a
> >     failure.  This value corresponds to 'discard I/Os' on the linux
> >     kernel block layer statistics.
> > ```
> 
> Is there a reference that can confirm that statement?
> 
> My understanding is that "discard i/o" is just another type of i/o operation, which is often used with SSD drives (see also [fstrim man](https://man7.org/linux/man-pages/man8/fstrim.8.html)); and not an error.

Here's the reference I found: https://www.kernel.org/doc/Documentation/block/stat.txt

LimeHat:
I checked that link, yes, but there's no indication that discard i/o is related to errors in any way.

If anything, it confirms my understanding, since they describe the discard operations in the same way as read/write e.g.:

```
read sectors, write sectors, discard_sectors
============================================
These values count the number of sectors read from, written to, or
discarded from this block device.  The "sectors" in question are the
standard UNIX 512-byte sectors, not any device- or filesystem-specific
block size.  The counters are incremented when the I/O completes.
```

LimeHat:
Another ref: blkdiscard
https://man7.org/linux/man-pages/man8/blkdiscard.8.html

earies:
For the driving use-case (trying to understand when storage media is having issues), maybe its best to narrow this in via SMART data vs. what is exposed in sysfs

@dplore - maybe best to align w/ what precisely is being monitored in your compute environment for such case?

LimeHat:
I agree with the suggestion to use SMART.

release/models/platform/openconfig-platform.yang
earies: I think these 2 modules can be dropped out of this patch set - the master version for both is 0.27.0 and neither are being touched as part of this PR

dplore:
> For the driving use-case (trying to understand when storage media is having issues), maybe its best to narrow this in via SMART data vs. what is exposed in sysfs
> 
> @dplore - maybe best to align w/ what precisely is being monitored in your compute environment for such case?

Ironically the current description is aligned with at least one use case for what is monitored in one of our network environments.  I do agree that SMART is a better data set to base this on and will refactor for that.

dplore:
Updated this PR to use a select few SMART counters.  I appreciate any feedback on this  approach.

earies:
Overall I think your latest patch is a better approach but it should be noted that this is up to the drive manufacturer as far as implementation which is going to vary within platforms of the same vendor and across vendors.

For example, just grabbing one variant of SSD we ship today, 2 of the 4 attributes listed here are supported.  This structure should be noted that these leaf nodes should be supported if the underlying hardware supports, otherwise optional/excluded.

k-iakhontov (approved):

dplore:
This was reviewed in the OC Community meeting on Sep 12, 2024 without objection.  Setting last-call to Sep 19 to allow a little more time for public review.

@s19nal do you have any comments?  Can you approve?

s19nal (approved):
LGTM - all comments addressed

PR #1092, opened 2024-04-10
Add interface name type with regex pattern constraint

dplore:
### Change Scope

* This proposes a typedef for a fully qualified interface name with a specific pattern of the name followed by a subinterface index.  
* It's imagined that this would be used for interface-refs where it is necessary to point to a fully specified interfaces and subinterface.

See also https://github.com/openconfig/public/pull/573 for content on naming of subinterfaces.

### DIFF
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -206,8 +206,8 @@ module openconfig-interfaces {
 
   typedef interface-full-id {
     type string {
-      pattern '(.*)\.([0-9]{*})';
-      oc-ext:posix-pattern '^(.*)\.([0-9]{*})$';
+      pattern '(.*)\.([0-9]*)';
+      oc-ext:posix-pattern '^(.*)\.([0-9]*)$';
     }
     description
       "This is a string which identifies an interface + subinterface
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -206,8 +206,8 @@ module openconfig-interfaces {
 
   typedef interface-full-id {
     type string {
-      pattern '(.*)\.([0-9]*)';
-      oc-ext:posix-pattern '^(.*)\.([0-9]*)$';
+      pattern '(.+)\.([0-9]+)';
+      oc-ext:posix-pattern '^(.+)\.([0-9]+)$';
     }
     description
       "This is a string which identifies an interface + subinterface
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -51,7 +51,14 @@ module openconfig-interfaces {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "3.7.0";
+  oc-ext:openconfig-version "3.8.0";
+
+  revision "2024-04-10" {
+      description
+        "Add typedef interface-full-id.";
+      reference
+        "3.8.0";
+  }
 
   revision "2023-11-06" {
       description
@@ -195,6 +202,16 @@ module openconfig-interfaces {
       name a interface reference.  The id can be arbitrary but a
       useful convention is to use a combination of base interface
       name and subinterface index.";
+  }
+
+  typedef interface-full-id {
+    type string {
+      pattern '(.*)\.([0-9]{*})';
+      oc-ext:posix-pattern '^(.*)\.([0-9]{*})$';
+    }
+    description
+      "This is a string which identifies an interface + subinterface
+      index.";
   }
 
   // grouping statements
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -206,8 +206,8 @@ module openconfig-interfaces {
 
   typedef interface-full-id {
     type string {
-      pattern '(.+)\.([0-9]+)';
-      oc-ext:posix-pattern '^(.+)\.([0-9]+)$';
+      pattern '([^.]+)(\.[0-9]+)?';
+      oc-ext:posix-pattern '^([^.]+)(\.[0-9]+)?$';
     }
     description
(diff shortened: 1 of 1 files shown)


release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -197,6 +204,16 @@ module openconfig-interfaces {
       name and subinterface index.";
   }
 
+  typedef interface-full-id {
+    type string {
+      pattern '(.*)\.([0-9]*)';
+      oc-ext:posix-pattern '^(.*)\.([0-9]*)$';
```
rolandphung, line 210: Should this be `(.+)\.([0-9]+)` to avoid `.` as a valid interface id?

dplore: Yes! Done.

LimeHat:
>It's imagined that this would be used for interface-refs where it is necessary to point to a fully specified interfaces and subinterface.

I think interface-refs already have a strict definition. Am I missing something?
https://github.com/openconfig/public/blob/0ae6953d5d0ee8156f48088ff7a1bb37ae0ce179/release/models/interfaces/openconfig-interfaces.yang#L181-L207

release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -197,6 +204,16 @@ module openconfig-interfaces {
       name and subinterface index.";
   }
 
+  typedef interface-full-id {
+    type string {
+      pattern '(.+)\.([0-9]+)';
```
earies, line 209: I'm not sure we can impose a restriction via pattern here unless we always can assume a subinterface is in existance - also see https://github.com/openconfig/public/issues/781 on the issue w/ tunnel/routed-vlan interfaces

A description statement may be the only way to cover that this is meant to either be an exact correlation to an interface name (should no subinterfaces exist) or `interface.subinterface`

If we make the regex cover anything post the `.` is optional then there is no use for the regex.

I'd vote for covering this in a very descriptive description stmt rather

dplore:
I've updated the regex and description to show the suffix representing subinterface as optional.

This was discussed in the 2024-04-16 OC operators meeting.  

It is observed that interface names are a very long standing area with a loose specification by design to allow for varied implementations, especially when it comes to media, hardware hierarchy and human readable names.    There is some concern that we should understand what operational problems we are making better and weigh this against the cost of any change which is widespread in the models.  Comments about the problem we are solving, particularly in a network operational context  are needed to move this forward.

release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -197,6 +204,18 @@ module openconfig-interfaces {
       name and subinterface index.";
   }
 
+  typedef interface-full-id {
+    type string {
+      pattern '([^.]+)(\.[0-9]+)?';
```
earies, line 209: For the most part I think we are probably ok on common "system-defined" physical/logical/virtual interface naming across network vendor implementations but this excludes any interface name that may carry a `.` unless the interface has no subinterface and encodes an integer post a single `.` (which is also legal linux netdev naming)

```
$ cat interfaces.txt 
et-0/0/0
et-0/0/0.0
et-0/0/0.1024
et-0/0/0:0.0
et-0/0/0:0.1024
ae0
ae0.0
ae0_0
TenGigabitEthernet0/0/0
TenGigabitEthernet0/0/0.0
mgmt0
mgmt0.0
Foo Bar Car
Foo Bar Car.1024
Foo Bar Car.1024.10
Foo/Bar/Car
Foo/Bar/Car:0
Foo/Bar/Car:0.100
Foo.Bar.Car/Jar
Foo.Bar.Car/Jar.100
Foo.Bar.Car/Jar.100
10/1/12
foo.1.2.3
foo.bar:0
foo.bar:0.0

$ cat interfaces.txt  | grep -vE "^([^.]+)(\.[0-9]+)?$" 
Foo Bar Car.1024.10
Foo.Bar.Car/Jar
Foo.Bar.Car/Jar.100
Foo.Bar.Car/Jar.100
foo.1.2.3
foo.bar:0
foo.bar:0.0
```

Maybe less of an issue for system-defined names but user-defined names (think virtual tunnels, etc..) now put a constraint on what the user can name an interface

dplore: To entertain this idea a bit more, a robust way to be very permissive on interface names could be to replace the single string `interface-id` with a  strongly typed model representation with two leafs:
1. interface-name type string 
2. Optional, sub-interface-index type uint32

Thinking a bit about how to model this in yang, we could add/replace each leaf which uses the interface-id type with the above two leaves.  Today `typedef interface-id` is [present in 17 leafs](https://github.com/search?q=repo%3Aopenconfig%2Fpublic%20%22type%20oc-if%3Ainterface-id%22&type=code) which at a glance all relate some configuration to an interface.  

I am not sure it's worth the churn in the model to make a change like this and would like to hear what the issues are.

LimeHat: Is there a problem with subinterface references that was not solved in #802 ?

earies: Am a bit confused - it seems what you are proposing is not much different than the `interface-ref-common` grouping:

https://github.com/openconfig/public/blob/master/release/models/interfaces/openconfig-interfaces.yang#L209-L235

But the problem trying to be solved here was the fact that we have a single leaf node key used in a list _somewhere_ that is either completely user-defined (made up objects that can differ all over each use) _or_ we use a typedef with some set of rules behind it to drive consistency in the list key naming per implementation.

Those "rules" are either gated by a pattern (which is likely too restrictive or too loose) or a very descriptive description stmt on an YANG unbounded string.

Now, this could be combined into the above grouping so that a common grouping of a list and child nodes (the ultimate distinct leafrefs) can be used and reapplied but this is primarily about ensuring that a key used to reference back to an interface (or subinterface combo) is consistent across all usage in each domain.

It's not about consistency across implementations but a set of rules within the same implementation at least.

cfyo:
Do you have examples of unqualified interface names that you are trying to avoid by doing this?

dplore:
I'll close this PR as I don't have an example of implementations of interface leaf references which are causing a problem.  If the community identifies this, we can reopen.

PR #1094, opened 2024-04-12
Add FAN_TRAY component type.

SydneyCaulfeild:
### Change Scope

  * Adding a component type fan tray components. These components currently have a type of FRU which is not specific enough to classify them. These components can be found on devices from different vendors such as Cisco and Juniper.
#### FAN_TRAY
 * Platform Implementation A:
   * For example on a CISCOXR, there are 4 Fan Tray components (`state/type` is "openconfig-platform-types:FRU").
   * This type is too general to determine what the component is. The component's name (with "FT") or `state/description` (`openconfig/components/component/0/FT0/state/description, Cisco 8808 Fan Tray`) in theory could be parsed, but this is not safe. A more specific type is needed.

 * Platform Implementation B:
   * Similarly on a Juniper, there are 2 Fan Tray components (`state/type` is "openconfig-platform-types:FRU").
   * These don't have a description, but they have `openconfig/components/component/Fan Tray 0/state/name, Fan Tray 0`.


s19nal:
where there is a FAN_TRAY, would there also be individual FAN components (with the FAN_TRAY as a parent)?  In which case it seems useful to specify both separately.  If not, could the tray assembly simply be assigned as type FAN? But isn't the point of FRU (field replaceable unit) is that it is deemed a single component?
I still think this change is a useful addition, despite the above :)

SydneyCaulfeild:
Hi @s19nal, we do have devices where there is a FAN_TRAY that contains multiple fan components. For example, I see a model where a FAN_TRAY FRU component has 8 fans. That being said, in this model the individual fans in the FAN_TRAY are not FRUs (they cannot be replaced by themselves, instead the entire FAN_TRAY has to be replaced). So these fans do not have part/serial numbers or a type.

Although I haven't found an example where a FAN_TRAY FRU component has FAN FRU child components, there are FAN_TRAY FRU components with individual, physical fans so I think this new type will avoid confusion!

s19nal:
@SydneyCaulfeild  agreed - more specific and less confusion I like :)  looks good to me!

dplore:
Reviewed in OC Operators meeting on April 23rd.  There were some comments about FANs and FANTRAY's where both are modeled as type "FAN", and using removable leaf to determine which component is replaceable.     But this seems confusing to have two different components with the same TYPE, especially where one is a parent of the other.

I've set this to last-call and merging on May 7,2024.

LimeHat:
If the implementation does not expose fans as standalone components, and only models the fan tray as an atomic FRU, what would be the correct component type?

SydneyCaulfeild:
We could update the definition for this new FAN_TRAY type to state that if the component is an atomic FRU and has more than one fan in it, it is a FAN_TRAY. Otherwise, regardless of if it is an FRU or not, each individual fan is type FAN. Does that make sense?

`If an implementation does not expose fans as standalone components, and only models the fan tray as an atomic FRU, what would be the correct component type?` -> this means it would be a FAN_TRAY because it houses several fans (even though these individual fans are not modeled).

I.e., possible scenarios are:

- FRU of type FAN_TRAY where the individual "child" fans are not modeled.
- FRU of type FAN_TRAY with FRU FANs.
- FRU FAN that exists on its own (is not in a tray).

SydneyCaulfeild:
Another reason I think it is worth distinguishing between a fan tray and a fan is that there are other component types with child fans, and it would help to clarify that these are individual fans are not fan trays. For example, in Cisco 8808 devices, there are power tray components and as you go down the subcomponents you get to a FAN (ex. `0/PT0 ` power tray -> `0/PT0-Power Module Slot 1` -> `0/PT0-PM1` power supply -> `0/PT0-PM1-Fan 0 ` FAN).

dplore:
We reviewed in OC operators Apr 30, 2024 meeting without objection.

@SydneyCaulfeild looks like a merge conflict has come up that needs to be resolved and this will be ready to merge.

SydneyCaulfeild:
@dplore thanks for letting me know! I have resolved the merge conflict.

rgwilton:
FWIW, I think that modelling FAN_TRAYS and FANs as separate components makes sense.

dplore:
@oscargdd provided feedback in the May 7, 2024 OC Operators meeting that the model also solves for their use cases.

dplore (approved):

PR #1095, opened 2024-04-13
Add component install-position, install-component and deprecate location

### LINKED ISSUE #1076: Semantics and format of Component Location
In order to get some consistency for the leaf  `/components/component/state/location`, I propose that we constrain it's meaning to be relevant to the parent component.  For example, a port's location would be a reference to the position in a CHASSIS (fixed-form factor) or LINECARD (modular form factor).       A value "5" for a PORT location  would for example mean the 5th port on the parent LINECARD.   

Today we see inconsistencies where some devices concatenate the linecard slot (CHASSIS position) and and linecard port (LINECARD position) together with some separator (ie: `1/5` or `1/5-port`) while others use only the parent location (ie: `5`)

https://github.com/openconfig/public/blob/6fd1f5d24d35b693864aa3218315b2acbeae9055/release/models/platform/openconfig-platform.yang#L404-L412

dplore:
Fixes #1076

### Change Scope

* Deprecate the `location` leaf as it's current usage varies across implementations.
* Introduce the `install-position`, `install-component` which more precisely define where a component is installed into a target index or `install-position` of a target `install-component`.  

The operational use case for`install-position` is: as a hardware replacement technician I want to receive an indication of the location of a replaceable hardware component within some network device physical package (chassis, shelf, linecard, etc).  An example subset of the replaceable components in scope could include a chassis, fan, fabric, powersupply , linecard module or transceiver module. The value of the position indication should exactly match some physical label or display on the hardware or if that is not feasible, it should exactly match documentation identifying the available positions on the hardware component(s).

### Tree view
```diff
*** /Users/dloher/old-tree.txt  Thu May  2 15:29:21 2024
--- new-tree.txt        Thu May  2 15:29:08 2024
***************
--- 8670,8678 ----
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw name                           -> ../config/name
        +--rw config
        |  +--rw name?   string
        +--ro state          |  +--ro name?                           string
          |  +--ro type?                           union
          |  +--ro id?                             string
!         |  x--ro location?                       string
>         |  +--ro install-position?               string
>         |  +--ro install-component?              string
          |  +--ro description?                    string
          |  +--ro mfg-name?                       string
          |  +--ro mfg-date?                       oc-yang:date

[... snip ...]

        +--rw oc-linecard:linecard
           +--rw oc-linecard:config
           |  +--rw oc-linecard:power-admin-state?   oc-platform-types:component-power-type
           +--ro oc-linecard:state
           |  +--ro oc-linecard:power-admin-state?   oc-platform-types:component-power-type
!          |  x--ro oc-linecard:slot-id?             string

```
### Platform Implementations

 * IOS XR
 ```
openconfig-platform-types:PORT
    /components/component[name="Port0/1/0/12"]/state/parent = "0/1/CPU0-NPU1"
    /components/component[name="Port0/1/0/12"]/state/location = "0/1/CPU0"

    New values:
    /components/component[name="Port0/1/0/12"]/state/install-position = "12"
    /components/component[name="Port0/1/0/12"]/state/install-component = "0/0/CPU0"
```
(IOSXR on the 8800 series names linecards including a CPU identifier)

* JunOS
```
openconfig-platform-types:PORT components:
/components/component[name="FPC0:PIC0:PORT1"]/state/location = "FPC0:PIC0:PORT1"

New value:
/components/component[name="FPC0:PIC0:PORT1"]/state/install-position = "1"
/components/component[name="FPC0:PIC0:PORT1"]/state/install-component = "FPC0:PIC0"
```

* EOS
```
openconfig-platform-types:PORT components:
/components/component[name="Ethernet4/1-Port"]/state/location = "4/1"

New value:
/components/component[name="Ethernet4/1-Port"]/state/install-position = "1"
/components/component[name="FPC0:PIC0:PORT1"]/state/install-component = "Linecard4"

```

### CODE REVIEW

**Comment by LimeHat** [comment]
This is an interesting example. 
Does an ASIC (`NPU1`) represent a helpful parent position for this use case?

**Comment by dplore** [comment]
> This is an interesting example. Does an ASIC (`NPU1`) represent a helpful parent-position value for this use case?

No, the port to NPU does not map to the `parent-position` use case.    

It is perhaps and example of why location may not be needed.  The port to NPU component relationship is good to represent using the `parent` leaf.  `location` is redundant.

**Comment by LimeHat** [comment]
>It is perhaps and example of why location may not be needed. The port to NPU component relationship is good to represent using the parent leaf. location is redundant.

Perhaps, but it also shows that the parent-position is not applicable in some cases. In fact, the location leaf in this scenario would be more useful, since the location doesn't have to be directly related to the parent.

**Comment by dplore** [comment]
> > It is perhaps and example of why location may not be needed. The port to NPU component relationship is good to represent using the parent leaf. location is redundant.
> 
> Perhaps, but it also shows that the parent-position is not applicable in some cases. In fact, the location leaf in this scenario would be more useful, since the location doesn't have to be directly related to the parent.
> 
> Just want to make sure you're ok with this (some components may not have a `parent-position`, or have a parent-position that isn't useful)

I expanded on the description for parent-position.

File: release/models/interfaces/openconfig-interfaces.yang
Comment (s19nal) [inline release/models/interfaces/openconfig-interfaces.yang:1]: @dplore  this change doesn't seem to be related to this PR - should it be part of something separate?

Comment (dplore) [reply release/models/interfaces/openconfig-interfaces.yang:1]: Opps, I accidently added commits from that other PR here.  Fixed.

File: release/models/platform/openconfig-platform.yang [hunk 1]
```diff
@@ -403,12 +409,37 @@ module openconfig-platform {
...
+        that some devices may use letters or alphanumerics.  The
+        position name should be the same name used to physically
+        identify the position in documentation or printed on the
+        device.
+
+        Some components with a parent may not have a parent-position.
+
+        For component types that have an explicit slot-id attribute,
+        such as linecards, the system should populate the more
+        specific slot-id.";
```
LimeHat, line 442: The part about "the more specific slot-id" is not entirely clear to me.

If I have a flat linecard numbering schema, for example, `linecard1` to `linecard9`, both slot-id and parent-position will indicate a single number [1,9].

dplore: This part of the description (carried over from the `location` leaf) is advising that component type `LINECARD` should use [/components/component/linecard/state/slot-id](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-linecard-state-slot-id) instead of parent-position.   

However now that you point it out, I see parent-position and slot-id are redundant.  We could expand the scope of this change to deprecate slot-id as well since it has the same meaning.  WDYT?

LimeHat: It seems reasonable for simple use cases like in the example above.

One scenario that we need to think about is when more complex numbering schemas are in use: some platforms have parent linecards with sub-slots (that contain Ethernet interfaces).  
In that case, something like `linecardX/Y` is typically used to identify the subslots, and there may be a divergence (e.g. a subslot can be modeled as a linecard with slot-id `x/y`, but the parent location will be `y`). Perhaps this text was added to describe such cases, and can be rephrased to better reflect that.

rgwilton:
@dplore, for clarity, in your example for IOS XR, should the parent-position not be '12' rather than 'NPU1'?

dplore:
> @dplore, for clarity, in your example for IOS XR, should the parent-position not be '12' rather than 'NPU1'?

Yes, thanks for this.  I updated the example in the PR description to reflect this.  

In that particular example, the parent of the PORT is an NPU.  The operational use case we are solving for requires a mapping from the PORT to the position in the LINECARD (or CHASSIS in a fixed form factor device).    Looking at this physically, both relationships exist; the port is wired to an NPU and NPU also occupies a position in the LINECARD/CHASSIS.    So the hierarchy is `PORT -> INTEGRATED_CIRCUIT -> LINECARD -> CHASSIS` or `PORT -> INTEGRATED_CIRCUIT -> CHASSIS` or .  We see this often.  There could be additional additional ancestors in the tree between PORT and LINECARD/CHASSIS.  

@SydneyCaulfeild and I discussed and had some ideas on how to resolve this programmatically.   I'll let her comment.

LimeHat:
Based on the updated example for IOS XR, it seems that `parent-position` should be `ancestor-position` instead?

According to the conversation above, my understanding was that the parent-position would not be applicable to a port when the parent is not a linecard module.

SydneyCaulfeild:
I would also like to point out that some components already apply this logic of "use the location of a removable ancestor" in their current `state/location` path, so I don't think that this constraint is necessarily new.

For example, in a Cisco 8808, a component of type POWER_SUPPLY is the grandchild of a power tray. Example names: `0/PT0-PM1` power supply -> `0/PT0-Power Module Slot 1` intermediate component -> `0/PT0` power tray. The `state/location` of the power supply right now is `0/PT0` which is the location of the grandparent, not the parent.

release/models/platform/openconfig-platform.yang
```diff
@@ -403,12 +409,51 @@ module openconfig-platform {
...
+      status deprecated;
       description
         "System-supplied description of the location of the
         component within the system. This could be a bay position,
         slot number, socket location, etc. For component types that
         have an explicit slot-id attribute, such as linecards, the
-        system should populate the more specific slot-id.";
+        system should populate the more specific slot-id.
+
+        This leaf is deprecated and replaced by parent-position.";
```
SydneyCaulfeild, line 420: We should update this to also say `install-position` instead of `parent-position`. Same applies to the version description in release/models/platform/openconfig-platform-common.yang.

release/models/platform/openconfig-platform.yang
```diff
@@ -403,12 +409,51 @@ module openconfig-platform {
...
+        specific slot-id.";
+    }
+
+    leaf install-component {
+      type string;
+      description
+        "A component may have zero or more intermediate component
+        ancestors which do not represent the install-position.  This
+        leaf represents the ancestor component which contains the
+        'install-position'.  For example, in the component tree
```
SydneyCaulfeild, line 453: Could we end this example by saying that `install-position` would be 2 and `install-component` would be "CPU1" (or whatever other line card name is used)?

We should also probably state that this leaf will hold the name of the ancestor (not the type or anything else).

release/models/platform/openconfig-platform.yang
```diff
@@ -403,12 +409,51 @@ module openconfig-platform {
...
+        parent component name should be present in the 'parent' leaf.
+
+        Typically the install-position is a number, but it is observed
+        that some devices may use letters or alphanumerics.  The
+        position name should be the same name used to physically
+        identify the position in documentation or printed on the
+        device.
+
+        Any component which is removable is expected to have
+        an install-position.  The ancestor component
```
SydneyCaulfeild, line 440: Not sure if this sentence "The ancestor component" is incomplete or was meant to be removed:)

dplore: opps, forgot to save before pushing to the repo.  Fixed.

release/models/platform/openconfig-platform.yang
dplore: done

release/models/platform/openconfig-platform-common.yang
```diff
@@ -20,7 +20,13 @@ submodule openconfig-platform-common {
     "This modules contains common groupings that are used in multiple
     components within the platform module.";
 
-  oc-ext:openconfig-version "0.24.0";
+  oc-ext:openconfig-version "0.25.0";
+
+   revision "2024-04-12" {
+    description
+      "Add parent-position and deprecate location.";
```
SydneyCaulfeild, line 27: "Add install-position and install-component and deprecate location."

release/models/platform/openconfig-platform.yang
dplore: I think the added `install-component` leaf will provide a clear and direct mapping to the component where the insertion/install occurs.  The implementation can choose to model the "subslot" as a component whose parent is the LINECARD and has children of type PORT.

dplore:
This was re-reviewed in the April 30 2024 OC operators meeting without objection.  

@oscargdd wanted to review as his organization was running into similar challenges using the `location` leaf.  @oscargdd please review and if you approve.

dplore:
Setting this to last call for May 14th to give Oscar time to review.

dplore:
Refreshed the PR description.  @SydneyCaulfeild please review if it meets your expectations.

SydneyCaulfeild:
@dplore the PR meets my expectations. Thanks for adding those examples!

release/models/platform/openconfig-platform.yang
```diff
@@ -411,12 +417,58 @@ module openconfig-platform {
...
+        identify the position in documentation or printed on the
+        device.
+
+        Any component which is removable is expected to have
+        an install-position and an install-component which points to
+        an ancestor component where the connection occurs.
+
+        For component types that have an explicit slot-id attribute,
+        such as LINECARD, the system should populate the more
+        specific slot-id.";
```
LimeHat, line 454: Likely this part can be removed

LimeHat (changes requested):

release/models/platform/openconfig-platform.yang
dplore: Thanks for the comment.  I think this is still a relevant and useful hint that install-position should not be used for LINECARD.  

I do think we could open a separate issue to address if we should change `slot-id`.  While slot-id is perhaps redundant to install-position, it does not seem to be in conflict.  For this PR, I want to keep the scope to addressing the 'location' leaf.

dplore: This made me think we should deprecate slot-id as it is now redundant to install-position and install-component.  I have proposed this for comment.

s19nal (approved):
looks like all the open comments have been addressed Darren - happy to approve this one.

release/models/platform/openconfig-platform.yang
```diff
@@ -411,12 +418,59 @@ module openconfig-platform {
...
+        an ancestor component where the connection occurs.
+
+        For component types that have an explicit slot-id attribute,
+        such as LINECARD, the system should populate slot-id,
+        install-position and install-component.  This will facilitate a
+        transition to deprecate slot-id.";
+    }
+
+    leaf install-component {
+      type string;
```
earies, line 460: If this is always a reference to another component then it seems this should rather be a leafref to ensure the relationship (It's just a different form of parent/child relationship mainly from a physical vs. logical aspect)

SydneyCaulfeild: I also think this can be a leafref!

SydneyCaulfeild (approved):

release/models/platform/openconfig-platform.yang
dplore: Done

release/models/platform/openconfig-platform.yang
```diff
@@ -411,12 +418,63 @@ module openconfig-platform {
...
-        system should populate the more specific slot-id.";
+        system should populate the more specific slot-id.
+
+        This leaf is deprecated and replaced by install-position and
+        install-component.";
+    }
+
+    leaf install-position {
+      type leafref {
+        path "../name";
```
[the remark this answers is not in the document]
dplore: Thank you for catching this.  Yes, it should've remained as a string. I will update it now.

PR #1096, opened 2024-04-24
Refactor gRPC service modeling

earies:
* (A) release/models/gnmi/openconfig-gnmi.yang
  * (A) release/models/gnmi/openconfig-gnmi-types.yang
  * (A) release/models/gnoi/openconfig-gnoi.yang
  * (A) release/models/gnoi/openconfig-gnoi-types.yang
  * (A) release/models/gnpsi/openconfig-gnpsi.yang
  * (A) release/models/gnpsi/openconfig-gnpsi-types.yang
  * (M) release/models/gribi/openconfig-gribi.yang
  * (A) release/models/gribi/openconfig-gribi-types.yang
  * (A) release/models/p4rt/openconfig-p4rt-types.yang
  * (M) release/models/gnsi/openconfig-gnsi.yang
  * (M) release/models/gnsi/openconfig-gnsi-acctz.yang
  * (M) release/models/gnsi/openconfig-gnsi-authz.yang
  * (M) release/models/gnsi/openconfig-gnsi-certz.yang
  * (M) release/models/gnsi/openconfig-gnsi-credentialz.yang
  * (M) release/models/gnsi/openconfig-gnsi-pathz.yang
  * (A) release/models/gnsi/openconfig-gnsi-types.yang
  * (A) release/models/grpc/openconfig-grpc-services-common.yang
  * (A) release/models/grpc/openconfig-grpc-types.yang
    - Refactor gRPC service modeling into modular augment approach that
      is determined by "allowed services" per endpoint
    - Relocate service identities to per-service types modules
    - Added structure of per-service containers to contain per-service
      config/state

### Change Scope

**NOTE: WORK IN PROGRESS, SOLICITING EARLY FEEDBACK**

* Initial refactor of gRPC service modeling
  - Addition of per-service categories
  - Detach common types into types modules for reuse and expansion
  - Addition of common gRPC constructs that each service must implement to
    satisfy the consistent API
  - Modular anchor point to add additional services

Intent is to restructure first to modular approach so that each gRPC service
can have consistent modeling w/o scattering constructs in various subtrees.

### Platform Implementations

N/A: Refactor for purpose of new implementation sustainability/growth

### DIFF
--- a/release/models/gnmi/openconfig-gnmi-types.yang
+++ b/release/models/gnmi/openconfig-gnmi-types.yang
@@ -1,0 +1,43 @@
+module openconfig-gnmi-types {
+  yang-version "1";
+  namespace "http://openconfig.net/yang/gnmi/types";
+  prefix "oc-gnmit";
+
+  import openconfig-extensions {
+    prefix oc-ext;
+  }
+
+  import openconfig-grpc-types {
+    prefix oc-grpct;
+  }
+
+  organization
+    "OpenConfig working group";
+  contact
+    "www.openconfig.net";
+
+  description
+    "This module adds configuration and operational state relating to gRPC
+    services running on a network device. The GRPC_SERVICE identity is used
+    to create an extensible list of services that can be instantiated, with
+    a base set defined in this module. New services can extend the identity
+    to be included in the list.";
+
+
+  oc-ext:openconfig-version "0.1.0";
+  oc-ext:catalog-organization "openconfig";
+  oc-ext:origin "openconfig";
+
+  revision 2024-04-23 {
+    description
+      "Initial revision.";
+    reference "0.1.0";
+  }
+
+  identity GNMI {
+    base oc-grpct:GRPC_SERVICE;
+    description
+      "gNMI: gRPC Network Management Interface";
+  }
+
+}
--- a/release/models/gnsi/openconfig-gnsi-acctz.yang
+++ b/release/models/gnsi/openconfig-gnsi-acctz.yang
@@ -3,22 +3,30 @@ module openconfig-gnsi-acctz {
   namespace "https://github.com/openconfig/yang/gnsi/acctz";
   prefix oc-gnsi-acctz;
 
-  import openconfig-extensions { prefix oc-ext; }
+  import openconfig-extensions {
+    prefix oc-ext;
+  }
+
   import openconfig-system {
     prefix oc-sys;
   }
+
   import openconfig-system-grpc {
     prefix oc-sys-grpc;
   }
+
   import openconfig-types {
     prefix oc-types;
   }
+
   import openconfig-yang-types {
     prefix oc-yang;
   }
+
   import openconfig-gnsi {
     prefix oc-gnsi;
   }
+
   organization
     "OpenConfig Working Group";
 
@@ -205,41 +213,44 @@ module openconfig-gnsi-acctz {
     }
   }
 
-  grouping grpc-server-acctz-counters {
+  grouping acctz-counters {
     description
       "A collection of counters from the gNSI.acctz module.";
 
-    container acctz-counters {
+    container counters {
       config false;
       description
         "A collection of counters from the gNSI.acctz module
         for acctz clients and sources.";
 
-      container state {
+      leaf last-cleared {
+        type oc-types:timeticks64;
         description
-          "Operational state relating to acctz-counters.";
-
-        leaf counters-last-cleared {
-          type oc-types:timeticks64;
-          description
-            "The last time that the counters were cleared (reset to
-            zero). This value is reported as nanoseconds since epoch
-            (January 1st, 1970 00:00:00 GMT).";
-        }
-
-        uses client-counters;
+          "The last time that the counters were cleared (reset to
+          zero). This value is reported as nanoseconds since epoch
+          (January 1st, 1970 00:00:00 GMT).";
       }
 
+      uses client-counters;
       uses source-counters;
     }
   }
 
-  // Augments section.
-  augment "/oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server" {
-    when "config[contains(services, 'oc-gnsi:GNSI')]/enable = 'true'";
+  grouping acctz-base {
     description
-      "Counters collected by the gNSI.acctz module.";
+      "";
 
-    uses grpc-server-acctz-counters;
+    container acctz {
+      description
+        "";
+
+      uses acctz-counters;
+    }
+  }
+
+  augment "/oc-sys:system/oc-sys-grpc:grpc-servers" +
+          "/oc-sys-grpc:grpc-server/oc-sys-grpc:services/oc-gnsi:gnsi" {
+
+    uses acctz-base;
   }
 }
--- a/release/models/gnsi/openconfig-gnsi-credentialz.yang
+++ b/release/models/gnsi/openconfig-gnsi-credentialz.yang
@@ -3,22 +3,30 @@ module openconfig-gnsi-credentialz {
   namespace "https://github.com/openconfig/yang/gnsi/credentialz";
   prefix oc-gnsi-credz;
 
-  import openconfig-extensions { prefix oc-ext; }
+  import openconfig-extensions {
+    prefix oc-ext;
+  }
+
   import openconfig-system {
     prefix oc-sys;
   }
+
   import openconfig-types {
     prefix oc-types;
   }
+
   import openconfig-yang-types {
     prefix oc-yang;
   }
+
   import openconfig-system-grpc {
     prefix oc-sys-grpc;
   }
+
   import openconfig-gnsi {
     prefix oc-gnsi;
   }
+
   organization
     "OpenConfig Working Group";
 
@@ -40,6 +48,12 @@ module openconfig-gnsi-credentialz {
       /system/aaa/authentication/users/user/state/password-hashed";
 
   oc-ext:openconfig-version "0.6.0";
+
+  revision 2024-04-23 {
+    description
+      "";
+    reference "";
+  }
 
   revision 2024-02-13 {
     description
@@ -256,17 +270,18 @@ module openconfig-gnsi-credentialz {
     }
   }
 
-  // Augments section.
-
-  augment "/oc-sys:system" {
-    when "oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server" +
-        "/oc-sys-grpc:config[contains(oc-sys-grpc:services, 'oc-gnsi:GNSI')]" +
-        "/oc-sys-grpc:enable = 'true'";
+  grouping credentialz-base {
     description
-      "Console credentials freshness data.";
+      "";
 
-    uses console-config-state;
+    container credentialz {
+      description
+        "";
+
+      uses console-config-state;
+    }
   }
+
   augment "/oc-sys:system/oc-sys:ssh-server/oc-sys:state" {
     when "../../oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server" +
         "/oc-sys-grpc:config[contains(oc-sys-grpc:services, 'oc-gnsi:GNSI')]" +
@@ -298,5 +313,11 @@ module openconfig-gnsi-credentialz {
 
     uses user-console-credentials-version;
     uses user-ssh-credentials-version;
+  }
+
+  augment "/oc-sys:system/oc-sys-grpc:grpc-servers" +
+          "/oc-sys-grpc:grpc-server/oc-sys-grpc:services/oc-gnsi:gnsi" {
+  
+    uses credentialz-base;
   }
 }
--- a/release/models/gnsi/openconfig-gnsi-pathz.yang
+++ b/release/models/gnsi/openconfig-gnsi-pathz.yang
@@ -3,19 +3,26 @@ module openconfig-gnsi-pathz {
   namespace "https://github.com/openconfig/yang/gnsi/pathz";
   prefix oc-gnsi-pathz;
 
-  import openconfig-extensions { prefix oc-ext; }
+  import openconfig-extensions {
+    prefix oc-ext;
+  }
+
   import openconfig-system {
     prefix oc-sys;
   }
+
   import openconfig-system-grpc {
     prefix oc-sys-grpc;
   }
+
   import openconfig-types {
     prefix oc-types;
   }
+
   import openconfig-yang-types {
     prefix oc-yang;
   }
+
   import openconfig-gnsi {
     prefix oc-gnsi;
   }
@@ -33,6 +40,12 @@ module openconfig-gnsi-pathz {
     device.";
 
   oc-ext:openconfig-version "0.3.0";
+
+  revision 2024-04-23 {
+    description
+      "";
+    reference "";
+  }
 
   revision 2024-02-13 {
     description
@@ -295,37 +308,23 @@ module openconfig-gnsi-pathz {
     }
   }
 
-  // Augments section.
-
-  augment "/oc-sys:system" {
-    when "oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server" +
-        "/oc-sys-grpc:config[contains(oc-sys-grpc:services, 'oc-gnsi:GNSI')]" +
-        "/oc-sys-grpc:enable = 'true'";
+  grouping pathz-base {
     description
-      "Collection of OpenConfig-path-based authorization policies that
-      have been installed on the device using the gNSI OpenConfig-path-
-      based authorization policy management service.
-      Each policy listed here is identified by its status (either ACTIVE
-          or SANDBOX) and has its version and creation date/time listed.";
+      "";
 
-    uses system-gnmi-pathz-policies;
+    container pathz {
+      description
+        "";
+
+      uses grpc-server-gnmi-pathz-policy-state;
+      uses gnmi-pathz-policy-success-failure-counters;
+      uses system-gnmi-pathz-policies;
+    }
   }
-  augment "/oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server" +
-          "/oc-sys-grpc:state" {
-    when "../config[contains(services, 'oc-gnsi:GNSI')]/enable = 'true'";
-    description
-      "A gNMI server OpenConfig-path-based authorization policy freshness
-      information.";
 
-    uses grpc-server-gnmi-pathz-policy-state;
-  }
-  augment "/oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server" {
-    when "oc-sys-grpc:config[contains(oc-sys-grpc:services, 'oc-gnsi:GNSI')]" +
-        "/oc-sys-grpc:enable = 'true'";
-    description
-      "A gNMI server OpenConfig-path-based authorization policy
-      success/failure counters.";
+  augment "/oc-sys:system/oc-sys-grpc:grpc-servers" +
+          "/oc-sys-grpc:grpc-server/oc-sys-grpc:services/oc-gnsi:gnsi" {
 
-    uses gnmi-pathz-policy-success-failure-counters;
+    uses pathz-base;
   }
 }
--- a/release/models/grpc/openconfig-grpc-services-common.yang
+++ b/release/models/grpc/openconfig-grpc-services-common.yang
@@ -1,0 +1,57 @@
+module openconfig-grpc-services-common {
+  yang-version "1";
+  namespace "http://openconfig.net/yang/grpc/common";
+  prefix "oc-grpc-common";
+
+  import openconfig-extensions { prefix oc-ext; }
+
+  organization
+    "OpenConfig working group";
+  contact
+    "www.openconfig.net";
+
+  description
+    "";
+
+  oc-ext:openconfig-version "0.1.0";
+  oc-ext:catalog-organization "openconfig";
+  oc-ext:origin "openconfig";
+
+  revision 2024-03-10 {
+    description
+      "Initial revision.";
+    reference "0.1.0";
+  }
+
+  grouping common-config {
+    description
+      "";
+
+    leaf enable {
+      type boolean;
+      description
+        "";
+    }
+  }
+
+  grouping common-services {
+    description
+      "";
+
+    container config {
+      description
+        "";
+
+      uses common-config;
+    }
+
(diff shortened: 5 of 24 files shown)


earies:
Initial refactored tree of `/system/grpc-servers`

_NOTE: Various existing services would be in need of refactoring for consistency and alignment of config/state_

```
module: openconfig-system
  +--rw system
     +--rw oc-sys-grpc:grpc-servers
        +--rw oc-sys-grpc:grpc-server* [name]
           +--rw oc-sys-grpc:name        -> ../config/name
           +--rw oc-sys-grpc:config
           |  +--rw oc-sys-grpc:name?                      string
           |  +--rw oc-sys-grpc:services*                  identityref
           |  +--rw oc-sys-grpc:enable?                    boolean
           |  +--rw oc-sys-grpc:port?                      oc-inet:port-number
           |  +--rw oc-sys-grpc:transport-security?        boolean
           |  +--rw oc-sys-grpc:certificate-id?            string
           |  +--rw oc-sys-grpc:metadata-authentication?   boolean
           |  +--rw oc-sys-grpc:listen-addresses*          union
           |  +--rw oc-sys-grpc:network-instance?          oc-ni:network-instance-ref
           +--ro oc-sys-grpc:state
           |  +--ro oc-sys-grpc:name?                      string
           |  +--ro oc-sys-grpc:services*                  identityref
           |  +--ro oc-sys-grpc:enable?                    boolean
           |  +--ro oc-sys-grpc:port?                      oc-inet:port-number
           |  +--ro oc-sys-grpc:transport-security?        boolean
           |  +--ro oc-sys-grpc:certificate-id?            string
           |  +--ro oc-sys-grpc:metadata-authentication?   boolean
           |  +--ro oc-sys-grpc:listen-addresses*          union
           |  +--ro oc-sys-grpc:network-instance?          oc-ni:network-instance-ref
           +--rw oc-sys-grpc:services
              +--rw oc-gnmi:gnmi
              |  +--rw oc-gnmi:config
              |  |  +--rw oc-gnmi:enable?   boolean
              |  +--ro oc-gnmi:state
              |     +--ro oc-gnmi:enable?   boolean
              +--rw oc-gnoi:gnoi
              |  +--rw oc-gnoi:config
              |  |  +--rw oc-gnoi:enable?   boolean
              |  +--ro oc-gnoi:state
              |     +--ro oc-gnoi:enable?   boolean
              +--rw oc-p4rt:p4rt
              |  +--rw oc-p4rt:config
              |  |  +--rw oc-p4rt:enable?   boolean
              |  +--ro oc-p4rt:state
              |     +--ro oc-p4rt:enable?   boolean
              +--rw oc-gribi:gribi
              |  +--rw oc-gribi:config
              |  |  +--rw oc-gribi:enable?   boolean
              |  +--ro oc-gribi:state
              |     +--ro oc-gribi:enable?   boolean
              +--rw oc-gnsi:gnsi
              |  +--rw oc-gnsi:config
              |  |  +--rw oc-gnsi:enable?   boolean
              |  +--ro oc-gnsi:state
              |  |  +--ro oc-gnsi:enable?   boolean
              |  +--rw oc-gnsi-acctz:acctz
              |  |  +--ro oc-gnsi-acctz:counters
              |  |     +--ro oc-gnsi-acctz:last-cleared?      oc-types:timeticks64
              |  |     +--ro oc-gnsi-acctz:client-counters
              |  |     |  +--ro oc-gnsi-acctz:history-istruncated?   oc-yang:counter64
              |  |     |  +--ro oc-gnsi-acctz:idle-timeouts?         oc-yang:counter64
              |  |     |  +--ro oc-gnsi-acctz:record-requests?       oc-yang:counter64
              |  |     |  +--ro oc-gnsi-acctz:record-responses?      oc-yang:counter64
              |  |     +--ro oc-gnsi-acctz:source-counters
              |  |        +--ro oc-gnsi-acctz:source-records* [service type]
              |  |           +--ro oc-gnsi-acctz:service    -> ../state/service
              |  |           +--ro oc-gnsi-acctz:type       -> ../state/type
              |  |           +--ro oc-gnsi-acctz:state
              |  |              +--ro oc-gnsi-acctz:service?   service-request
              |  |              +--ro oc-gnsi-acctz:type?      service-type
              |  |              +--ro oc-gnsi-acctz:records?   oc-yang:counter64
              |  +--rw oc-gnsi-authz:authz
              |  |  +--ro oc-gnsi-authz:authz-policy-counters
              |  |     +--ro oc-gnsi-authz:rpcs
              |  |        +--ro oc-gnsi-authz:rpc* [name]
              |  |           +--ro oc-gnsi-authz:name     -> ../state/name
              |  |           +--ro oc-gnsi-authz:state
              |  |              +--ro oc-gnsi-authz:name?                 string
              |  |              +--ro oc-gnsi-authz:access-rejects?       oc-yang:counter64
              |  |              +--ro oc-gnsi-authz:last-access-reject?   oc-types:timeticks64
              |  |              +--ro oc-gnsi-authz:access-accepts?       oc-yang:counter64
              |  |              +--ro oc-gnsi-authz:last-access-accept?   oc-types:timeticks64
              |  +--rw oc-gnsi-certz:certz
              |  |  +--rw oc-gnsi-certz:certificate-version?                             version
              |  |  +--rw oc-gnsi-certz:certificate-created-on?                          created-on
              |  |  +--rw oc-gnsi-certz:ca-trust-bundle-version?                         version
              |  |  +--rw oc-gnsi-certz:ca-trust-bundle-created-on?                      created-on
              |  |  +--rw oc-gnsi-certz:certificate-revocation-list-bundle-version?      version
              |  |  +--rw oc-gnsi-certz:certificate-revocation-list-bundle-created-on?   created-on
              |  |  +--rw oc-gnsi-certz:authentication-policy-version?                   version
              |  |  +--rw oc-gnsi-certz:authentication-policy-created-on?                created-on
              |  |  +--rw oc-gnsi-certz:ssl-profile-id?                                  string
              |  |  +--ro oc-gnsi-certz:counters
              |  |     +--ro oc-gnsi-certz:connection-rejects?       oc-yang:counter64
              |  |     +--ro oc-gnsi-certz:last-connection-reject?   oc-types:timeticks64
              |  |     +--ro oc-gnsi-certz:connection-accepts?       oc-yang:counter64
              |  |     +--ro oc-gnsi-certz:last-connection-accept?   oc-types:timeticks64
              |  +--rw oc-gnsi-credz:credentialz
              |  |  +--rw oc-gnsi-credz:console
              |  |     +--rw oc-gnsi-credz:config
              |  |     +--ro oc-gnsi-credz:state
              |  |        +--ro oc-gnsi-credz:counters
              |  |           +--ro oc-gnsi-credz:access-rejects?       oc-yang:counter64
              |  |           +--ro oc-gnsi-credz:last-access-reject?   oc-types:timeticks64
              |  |           +--ro oc-gnsi-credz:access-accepts?       oc-yang:counter64
              |  |           +--ro oc-gnsi-credz:last-access-accept?   oc-types:timeticks64
              |  +--rw oc-gnsi-pathz:pathz
              |     +--rw oc-gnsi-pathz:gnmi-pathz-policy-version?      version
              |     +--rw oc-gnsi-pathz:gnmi-pathz-policy-created-on?   created-on
              |     +--ro oc-gnsi-pathz:gnmi-pathz-policy-counters
              |     |  +--ro oc-gnsi-pathz:paths
              |     |     +--ro oc-gnsi-pathz:path* [name]
              |     |        +--ro oc-gnsi-pathz:name     -> ../state/name
              |     |        +--ro oc-gnsi-pathz:state
              |     |           +--ro oc-gnsi-pathz:name?     string
              |     |           +--ro oc-gnsi-pathz:reads
              |     |           |  +--ro oc-gnsi-pathz:access-rejects?       oc-yang:counter64
              |     |           |  +--ro oc-gnsi-pathz:last-access-reject?   oc-types:timeticks64
              |     |           |  +--ro oc-gnsi-pathz:access-accepts?       oc-yang:counter64
              |     |           |  +--ro oc-gnsi-pathz:last-access-accept?   oc-types:timeticks64
              |     |           +--ro oc-gnsi-pathz:writes
              |     |              +--ro oc-gnsi-pathz:access-rejects?       oc-yang:counter64
              |     |              +--ro oc-gnsi-pathz:last-access-reject?   oc-types:timeticks64
              |     |              +--ro oc-gnsi-pathz:access-accepts?       oc-yang:counter64
              |     |              +--ro oc-gnsi-pathz:last-access-accept?   oc-types:timeticks64
              |     +--ro oc-gnsi-pathz:gnmi-pathz-policies
              |        +--ro oc-gnsi-pathz:policies
              |           +--ro oc-gnsi-pathz:policy* [instance]
              |              +--ro oc-gnsi-pathz:instance    -> ../state/instance
              |              +--ro oc-gnsi-pathz:state
              |                 +--ro oc-gnsi-pathz:instance?     enumeration
              |                 +--ro oc-gnsi-pathz:version?      version
              |                 +--ro oc-gnsi-pathz:created-on?   created-on
              +--rw oc-gnpsi:gnpsi
                 +--rw oc-gnpsi:config
                 |  +--rw oc-gnpsi:enable?   boolean
                 +--ro oc-gnpsi:state
                    +--ro oc-gnpsi:enable?   boolean
```

earies:
> Looks to me there are some redundancy. For example, how would you interpret the difference in these two paths:
> 
> system/grpc-servers/grpc-server[name=gnmi]/config/enable
> 
> system/grpc-servers/grpc-server[name=gnmi]/services/gnmi/config/enable

The first is the endpoint/proxy enablement/disablement (e.g. an entire listening server/port)

The latter is granular control per service enablement (an abstract aggregation of services infact) - this could cascade to deregistration of a backend service to the proxy/endpoint.  This hierarchy also sets it up for very granular per service/API/message construct config/state within.

Also, there is the leaf-list of `services` directly under the root of the grpc-server instance under `config/services` which if this lands would complement per-service characteristics.

The leaf-list then turns into "allowed services on the endpoint" where we could default to all being enabled should they be defined.  The per-service models augment nodes to the tree _only_ if that service is allowed on the endpoint.  The existence of a leaf-list member determines if allowed is all going fwd.

release/models/gnmi/openconfig-gnmi-types.yang
dplore, line 1: Should the grpc server types be in release/models/grpc/openconfig-grpc-types.yang ?  This way we have a simple list of well known, OC model adopted types?  

Types can be defined even if the implementation doesn't support them.  Errors can be 'grpc server xyz not supported' versus 'not found'.  

A single file would only be usable if the implementation chooses to include the protocol specific files only  when implemented.    The alternative that seems to be proposed here is the implementation only includes the <protocol>-type model files when supported.    I sort of prefer a single file with all the types as it's easy to see the list of all defined OC gRPC protocols in one model file.

release/models/gnsi/openconfig-gnsi-credentialz.yang
```diff
@@ -256,17 +270,18 @@ module openconfig-gnsi-credentialz {
...
+
+    container credentialz {
+      description
+        "";
 
-    uses console-config-state;
+      uses console-config-state;
+    }
   }
+
```
dplore, line 284: the augments below should be refactor to follow the style you recommended.  Perhaps you didn't do this yet because you were waiting for comments on the recommended style.

release/models/gnsi/openconfig-gnsi-pathz.yang
dplore, line 1: several more changes needed to this file to follow the `<protocol>/state/counters/<counter_name>` pattern.   Currently modeled as

```
     |     +--ro oc-gnsi-pathz:gnmi-pathz-policy-counters
     |        +--ro oc-gnsi-pathz:paths
     |           +--ro oc-gnsi-pathz:path* [name]
     |              +--ro oc-gnsi-pathz:name     -> ../state/name
     |              +--ro oc-gnsi-pathz:state
     |                 +--ro oc-gnsi-pathz:name?     string
     |                 +--ro oc-gnsi-pathz:reads
     |                 |  +--ro oc-gnsi-pathz:access-rejects?       oc-yang:counter64
     |                 |  +--ro oc-gnsi-pathz:last-access-reject?   oc-types:timeticks64
     |                 |  +--ro oc-gnsi-pathz:access-accepts?       oc-yang:counter64
     |                 |  +--ro oc-gnsi-pathz:last-access-accept?   oc-types:timeticks64
     |                 +--ro oc-gnsi-pathz:writes
     |                    +--ro oc-gnsi-pathz:access-rejects?       oc-yang:counter64
     |                    +--ro oc-gnsi-pathz:last-access-reject?   oc-types:timeticks64
     |                    +--ro oc-gnsi-pathz:access-accepts?       oc-yang:counter64
     |                    +--ro oc-gnsi-pathz:last-access-accept?   oc-types:timeticks64
     +--rw oc-gnsi-credz:console
```

release/models/gnsi/openconfig-gnsi.yang
```diff
@@ -1,40 +1,62 @@
...
+      description
+        "";
+
+      uses oc-grpc-common:common-services;
+    }
+  }
+
+  augment "/oc-sys:system/oc-sys-grpc:grpc-servers" +
+          "/oc-sys-grpc:grpc-server/oc-sys-grpc:services" {
+    when "../oc-sys-grpc:config/oc-sys-grpc:services = 'oc-gnsit:GNMI'";
```
dplore, line 58: if keeping the augment and when clause, this one should be changed to be gnsi specific

release/models/gnsi/openconfig-gnsi-acctz.yang
```diff
@@ -205,41 +213,44 @@ module openconfig-gnsi-acctz {
...
+    container acctz {
+      description
+        "";
+
+      uses acctz-counters;
+    }
+  }
+
+  augment "/oc-sys:system/oc-sys-grpc:grpc-servers" +
+          "/oc-sys-grpc:grpc-server/oc-sys-grpc:services/oc-gnsi:gnsi" {
```
dplore, line 253: Do you want a when statement here? 

Here and throughout the protocol model files, is it simpler to not include any when statement and have the implementation choose whether or not to include the model based on support of the protocol?

release/models/grpc/openconfig-grpc-services-common.yang
```diff
@@ -0,0 +1,57 @@
...
+    description
+      "Initial revision.";
+    reference "0.1.0";
+  }
+
+  grouping common-config {
+    description
+      "";
+
+    leaf enable {
```
LimeHat, line 30: Is there a need for the `enable` leaf for each service?

earies: Possibly - my point was rather first to draw out the modular skeleton hierarchies to demonstrate what per service/RPC config/state could expand upon for comment.

Currently, the bucketization of a `GRPC_SERVICE` today is done mostly on a "project" basis (e.g. gNOI) and each project can encapsulate >= 1 gRPC "services" and each can comprise of >= 1 RPCs -> message structs -> fields and so on.  The behaviors of some are more well defined than others (some may carry a specification, some vast comments, some few comments)

So as we break this down, there could be a necessity for granularity at each of these layers.  Enable/Disable of a service currently could be accomplished by removing the entry from the leaf-list so there would have to be some discussion on what this means, if it is of value, etc...

But at this layer could be various other branches to represent the above mentioned breakdown where state or config override behaviors, etc... could exist

This common model is meant to be set of nodes that would apply to every gRPC service (enable was an easy example is all....)

release/models/gnsi/openconfig-gnsi-acctz.yang
earies: My thoughts are yes, a when statement should be used here

It is more of a signal that "this subtree should be expected and exists _if_ the service is enabled on the server endpoint"

The model inclusion/exclusion is the hard demarcation where the service is not supported whatsoever

release/models/grpc/openconfig-grpc-services-common.yang
LimeHat: > Enable/Disable of a service currently could be accomplished by removing the entry from the leaf-list so there would have to be some discussion on what this means, if it is of value, etc...
Yep, that's pretty much my point. The existing leaf-list already accomplishes that, and the explicit enable appears to be redundant. 

Now I see what you mean, thanks.

release/models/gnsi/openconfig-gnsi.yang
earies: Yep good catch - rushed out replication ...  will readjust in next patch

release/models/gnsi/openconfig-gnsi-credentialz.yang
earies: Agreed - also there are a number of issues within the gNSI modeling that have been discovered recently that will have to aggregate into a separate discussion/PR

release/models/gnmi/openconfig-gnmi-types.yang
earies: That's a good suggestion esp. for the case you describe (to be able to signal "not supported" vs. "i have no clue what this is")

My initial thoughts were that alot of gRPC API based projects (and that may only be a portion of the intent of the project) are marshaled off on their own repositories, may exist within the OpenConfig umbrella, may not (P4RT or proprietary native services) and that development can be contained and modularized as multi-repo vs. drawing a relationship to a monorepo.

With that in mind, I was thinking it is possibly a good thing to keep even YANG schemas segmented as such (as gNSI was initially until a few mo. back) but have an intent to augment an OpenConfig subtree from "outside" creating a very clear demarcation.  Project is supported or not....

But since there has already been a design pattern of bringing things under a common umbrella, this already signifies that a given project/service is "OC adopted" then (I think there are alot of rough edges to be molded here for every project to be consumed/used by everyone in this case)

I guess one major downside of this approach is from a configuration aspect...

The node that carries the identityref would enumerate _all_ as possible by default.  The YANG language currently does not allow a method to elegantly "remove" an identity so it falls back to an out-of-band signal and internal implementation detail to have to say well yes.... gNPSI might be in the types model so it gives the perception of support but gNPSI is not supported, etc....

release/models/gnsi/openconfig-gnsi-pathz.yang
earies: Agreed - gNSI modeling is in need of a bit of an overhaul and dissection (will follow this up separately since these are already merged)

dplore:
Hi @earies the OC Operators took another look at this.  This does look like a good work in progress that we should incorporate into OpenConfig release 6.x.   (5.x will be released in the coming days).  

This OC 6.x release can include these breaking changes and refactoring of the gnsi related containers/counters.

Would you like to freshen this PR for a new review?

brianneville:
Hi,

Since 2023, Arista has built out support for large portions of the current gNSI YANG models. 
This change will be backwards-incompatible and breaking for customers who rely on those YANG models.
The benefits of restructuring all these models would have to significantly outweigh the cost of introducing such a breaking change.

>N/A: Refactor for purpose of new implementation sustainability/growth

Are there concrete future plans for these YANG models which necessitates such a refactor? 
Complete restructuring of this many models should not be done lightly and if there is nothing that immediately necessitates such a refactor, it feels harder to justify. 
While the gNSI YANG models are new to the openconfig/public repo, the models themselves are fairly mature (they just existed in the gNSI repo until they were relocated and then removed in https://github.com/openconfig/gnsi/pull/166)

release/models/grpc/openconfig-grpc-services-common.yang
brianneville: >So as we break this down, there could be a necessity for granularity at each of these layers. 

I think this granularity would be beneficial - network operators may not want to run _all_ gNSI services on their system. 
For example a network operator may wish to use gNSI Acctz for accounting, but may have no interest in starting gNSI Pathz and managing a gNSI Pathz policy for the system (managing a policy would be necessary when Pathz is enabled as Pathz assumes an implicit deny for all users on all gNMI paths unless a policy has rules specifying otherwise).

This granularity would also make more sense given the `GRPC_SERVICE` identity, which is being overloaded with mulitple meanings here (i.e. it can either refer to a particular _service_ like gNMI, or a _collection of services_ like gNSI or gNOI).

release/models/gnsi/openconfig-gnsi-acctz.yang
```diff
@@ -205,41 +213,44 @@ module openconfig-gnsi-acctz {
...
+      "";
+
+    container acctz {
+      description
+        "";
+
+      uses acctz-counters;
+    }
+  }
+
```
adityasingh-anet, line 250: Just a heads up if this change moves forward, the gnsi-acctz yang model has been updated with a revision that removes the `idle-timeouts` and `counters-last-cleared` leaves.

Additionally, the source-counters and client-counters are now on different augmented paths:
https://github.com/openconfig/public/pull/1195

earies:
> Looks to me there are some redundancy. For example, how would you interpret the difference in these two paths:
> 
> system/grpc-servers/grpc-server[name=gnmi]/config/enable
> 
> system/grpc-servers/grpc-server[name=gnmi]/services/gnmi/config/enable

The motivation behind above is to have per-service granular control.  The prior is for the entire endpoint/reverse-proxy.  The 2 can be complementary

earies:
Apologies for the delay as other items took precedence - Just to provide an update as I'll be looping back to this PR shortly, addressing previous comments along w/ some possible refactoring, I wanted to solicit commentary around a few items that have come up since (also in relation to some concerns above)

1. Breaking changes and massive restructuring

I think we all understand the pain when something is supported, shipped and then changed however the unfortunate side effects of the projects under this umbrella is future growth and/or operationalization/usefulness is often not realized until post publish (and some implementation of some kind).   If we look at some of the gNSI modeling today as an example (since there is intent to relocate/restructure as part of this), I would question the ultimate usefulness to secops for instance for pathz counters (This in itself needs reconsideration and restructuring - an enumeration per "path" is incorrect and misaligned)

In either case I agree, such changes should not be taken lightly and have proper justification.

2. Proper anchor points

As I've thought about this restructuring a bit more, I think there is definite need to have per-server and per-service classification (and believe folks are agreeing so far) _but_ many of the services that fall within have intent that may not only be system-wide (and not classified per server) but also not unique to gRPC.  Today, these are gRPC services to provision and also potentially reap state but state can also exist in YANG modeled subtrees.  Each one of these from an implementation perspective could also expose native services or other NBI to interact with the ultimate "feature" at the end of the day so some of the backend state should likely sit outside of each classified "grpc-server"

Examples: policy hits may not be tracked per server but rather global thus duplicating data under each server hosting the service

Items such as per-service config knobs, per-client stats, RPC counters, etc... are all warranted under server->service subtrees but the ultimate feature itself may still be warranted to sit outside of the gRPC server modeling

Thoughts/Comments?

PR #1099, opened 2024-04-24
jakesnyder-1097 adding INACTIVE to client_state #1098

jsnyder81:
### Change Scope

* Added state to indicate that client is no longer active in the 802.11 state machine, but still present in client data.
### Platform Implementations

 * Implementation A: Aruba purges client data after a period of time, but retains it in the table to reduce impact of adding it back in for quick events such as roaming, but issues arise because there is no state to map this to.  Need a way to reconcile this to ensure interoperability between vendor specific implementations.

Unsure if other vendors implement this, this is related to how things work under the hood.


mike-albano (approved):
LGTM

PR #1100, opened 2024-04-26
Add backup-active to AFT NHG state

dplore:
### Change Scope

* Adds `/network-instances/network-instance/afts/next-hop-groups/next-hop-group/state/backup-active`
* This leaf indicates if the backup next hop group is in use.

```diff
*** /Users/dloher/old-tree.txt  Thu Apr 25 17:59:29 2024
--- new-tree.txt        Thu Apr 25 17:59:48 2024
***************
*** 3396,3401 ****
--- 3396,3402 ----
module: openconfig-network-instance
     +--rw network-instances
          +--rw network-instance* [name]        +--ro afts
          |  +--ro next-hop-groups
          |  |  +--ro next-hop-group* [id]
          |  |     +--ro id             -> ../state/id
          |  |     +--ro state
          |  |     |  +--ro id?                      uint64
          |  |     |  +--ro next-hop-group-name?     string
          |  |     |  +--ro programmed-id?           uint64
          |  |     |  +--ro color?                   uint64
          |  |     |  +--ro backup-next-hop-group?   -> ../../../next-hop-group/state/id
+         |  |     |  +--ro backup-active?           boolean
          |  |     +--ro next-hops
          |  |     |  +--ro next-hop* [index]
          |  |     |     +--ro index    -> ../state/index
```

### Platform Implementations

This is a new feature which is not currently supported by any known implementations.  It is intended to represent a feature request.


robshakir:
Generally LGTM -- is there any commentary as to the supportability in different implementations?

dplore:
This is added to the [May 2, 2024 OC community agenda](https://docs.google.com/document/d/11sXUF2zujhbPs_bt-RFJhh2DcIbEvnOpd9GypE-u3cc/edit).

nflath (approved):

aaronmillisor:
This functionality can be supported in IOS XR.

dplore:
Setting to last call.  I would like to hear a few more comments from other implementors about the feasibility of this leaf.

dplore:
Some implementors have asked for additional time to review.  Setting last call to June 20, 2024

kgurunath:
Below are the queries regarding the necessity, intented scope of this leaf and other alternative approaches.

**a. Necessity of the Leaf:** 

        Alternative_1:
	IMO, the SDN controller listens to opeconfig's afts and interface sensor paths. By co-relating these two sensor values, the controller can infer the time instance at which, all primary legs of the Next Hop Group (NHG) are down. 
	
        Alternative_2: 
	Examining the ‘packets-forwarded’ leaf within the ‘next-hops’ container is another approach. By monitoring this parameter, the controller could infer whether traffic forwarding occurs through primary-legs or backup-leg. 

Both of the above approaches might provide an alternative way to achieve the desired functionality without introducing a new leaf. Can they be explored?

**b. Intended Scope and Transitionary State:**

	Regarding the leaf’s intended scope, it’s essential to clarify whether it applies exclusively to gRIBI programmed routes?
	In general, when dynamic routes are learned through protocols, the control plane typically initiates global repairs. Therefore, understanding how this leaf fits into the overall control plane behavior during such transitions is crucial.

Answers to the above queries will indeed enhance our understanding of the pull request’s intent and use-case.

robshakir:
@kgurunath: What happens if the NHG viability is not solely to do with the interface status? e.g., recursive resolution to a route that doesn't exist, ARP entry missing etc. It seems like there's potentially more complex logic than solely what is being suggested, right?

dplore:
We've had confirmation from multiple implementors that this leaf is feasible to implement, so we will merge this.

Regarding the question of the install-protocol, currently there is not a constraint on this.  The expectation is that if a backup-next-hop-group is present, then `backup-active` should be supported.     Of course, not all implementations and not all hardware can fully support every feature/path defined in OC.  

That said, our particular use case is in fact for gRIBI programmed AFT entries.  Other operators may have different use cases.

dplore:
@robshakir for your approval before merging

robshakir (approved):

PR #1101, opened 2024-04-28
Add bridge-id to interfaces

ayelrod:
### Change Scope

/interfaces/interface/config/bridge-id

* Introduce a new leaf to specify an interface's bridge membership.
* Bridge interfaces are modeled as interfaces themselves, meaning the bridge-id is a leafref to an interface of type [bridge](https://github.com/openconfig/yang/blob/65fb9ff7590595be1ddeafef8fbaac37cb4c0671/standard/ietf/RFC/iana-if-type.yang#L1176).

### Platform Implementations

- [Linux Bridge](https://man7.org/linux/man-pages/man8/brctl.8.html) - set up, maintain, and inspect the ethernet bridge configuration in the Linux kernel.

The intent is to bridge two or more interfaces on the same device using a Linux ethernet bridge. Network traffic coming in on any of these ports will be forwarded to the other ports transparently.

### DIFF
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -51,7 +51,14 @@ module openconfig-interfaces {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "3.7.1";
+  oc-ext:openconfig-version "3.8.0";
+
+  revision "2024-04-28" {
+      description
+        "Add bridge-id to interfaces.";
+      reference
+        "3.8.0";
+  }
 
   revision "2024-04-04" {
       description
@@ -451,6 +458,15 @@ module openconfig-interfaces {
         "Sets the loopback type on the interface. Setting the
         mode to something besides NONE activates the loopback in
         the specified mode.";
+    }
+
+    leaf bridge-id {
+      type leafref {
+        path "/oc-if:interfaces/oc-if:interface/oc-if:name";
+      }
+      description
+        "Specify the logical bridge interface to which
+        this interface belongs.";
     }
 
     uses interface-common-config;


earies:
Am a bit confused by your intention and placement here.  It seems what you are looking to do is define the L3 interface for a given L2 segment/VLAN or "bridge-domain" is that correct?

The only constructs that exist today that could correlate to a "bridge-domain" is a network-instance classification (where VLANs and members are defined/bound) inside a special instance type.  The L3 interface for a bridge is defined as a `routed-vlan` interface (`/interfaces/interface/routed-vlan`) in which the interface (e.g. irb, vlan, svi) itself references the L2 segment (VLAN ID)

```
  augment /oc-if:interfaces/oc-if:interface:
    +--rw routed-vlan
       +--rw config
       |  +--rw vlan?   union
       +--ro state
          +--ro vlan?   union
```

As this patch indicates, you merely have a leafref inside an interface back to an interface which does not suffice so would suggest expanding on what it is you are trying to convey.

Also the above construct appears in need of addressing as it is too loose in the referencing of what a vlan ID entails

```
  grouping vlan-routed-config {
    description
      "Configuration data for routed vlans (SVI, IRB, etc.)";

    leaf vlan {
      type union {
        // TODO: in YANG 1.1, unions support leafref types which
        // should be used here to reference a configured VLAN by
        // id or name
        type uint16;
        type string;
      }
      description
        "References the VLAN for which this IP interface
        provides routing services -- similar to a switch virtual
        interface (SVI), or integrated routing and bridging interface
        (IRB) in some implementations.";
    }

  }
```

ayelrod:
Thanks for helping review @earies!

I may have had some incorrect links in the implementations section, I removed them. The intent is to model a linux bridge where two interfaces on the same device are bridged and all traffic that comes in on one interface also goes to the other. This one additional leaf should be enough to accomplish this. The bridge is an interface itself (named "br0" for example) and two interfaces will have a bridge-id pointing to that bridge (bridge-id: "br0" in this case).

Let me know what you think about this.

LimeHat:
>The intent is to model a linux bridge where two interfaces on the same device are bridged

That can be modeled as an L2 network instance with two interfaces associated with it.

E.g.
https://github.com/openconfig/public/blob/4d4aa4b7dd94acfae84e3f567b534116e6bd5ca4/release/models/network-instance/openconfig-network-instance-types.yang#L158-L162

See also https://github.com/openconfig/public/blob/master/doc/network_instance_examples.md


> all traffic that comes in on one interface also goes to the other.  

that's not how bridges work, generally speaking (unless you disable mac-learning or all your traffic is BUM)

rgwilton:
I also agree that this should be modelled as an L2VSI network instance.  Specifically, I don't think that it is correct to try and model a bridge as an interface.  Looking at the Linux documentation, it would also seem to better fit an L2VSI construct.

earies:
+1 to @rgwilton + @LimeHat 's suggestions here on proper modeling for what you are looking to achieve from existing OC constructs

When we blend classical router/switch behaviors with other worlds, I think we'll see alot more of these types of things where constructs that have similar meaning will be implemented differently.  In Linux, whether using std linux bridging, OVS, etc... will ultimately translate down into the creation of a kernel netdev construct - aka. "interface" - This construct by nature can then inherit other attributes such as also becoming an IRB/SVI L3 interface or just a standard "bridge" - in either case I view this as a somewhat collapsed mechanism of defining a bridge construct and attaching L2 ports along w/ an L3 "gateway" where in this case the "gateway" and the bridge could be named one in the same.

dplore:
Agreed the [network-instance examples documentation](https://github.com/openconfig/public/blob/master/doc/network_instance_examples.md)  defines the canonical method for bridge creation to be network-instance.    Apologies to @ayelrod as I had suggested the interface approach before discovering that documentation.  Thanks to the community for the feedback!

ayelrod:
Thanks everyone for the feedback. Closing this PR.

PR #1036, opened 2024-01-22
Publish PTP Yang model

sallylsy:
### Change Scope

* Support PTP (Precision Time Protocol) in OpenConfig. PTP model implementation is based on IEEE1588
* This change is backwards compatible

### Model

```
module: openconfig-ptp
  +--rw ptp
     +--rw instances
     |  +--rw instance* [id]
     |     +--rw id                 -> ../config/id
     |     +--rw config
     |     |  +--rw id?   uint32
     |     +--ro state
     |     |  +--ro id?                        uint32
     |     |  +--ro current-dataset-state
     |     |  |  +--ro steps-removed?        uint16
     |     |  |  +--ro offset-from-master?   oc-ptp-types:time-interval-type
     |     |  |  +--ro mean-path-delay?      oc-ptp-types:time-interval-type
     |     |  +--ro time-properties-dataset
     |     |     +--ro current-utc-offset-valid?   boolean
     |     |     +--ro current-utc-offset?         int16
     |     |     +--ro leap59?                     boolean
     |     |     +--ro leap61?                     boolean
     |     |     +--ro time-traceable?             boolean
     |     |     +--ro frequency-traceable?        boolean
     |     |     +--ro ptp-timescale?              boolean
     |     |     +--ro time-source?                uint8
     |     +--rw default-dataset
     |     |  +--rw config
     |     |  |  +--rw priority1?           uint8
     |     |  |  +--rw priority2?           uint8
     |     |  |  +--rw domain-number?       uint8
     |     |  |  +--rw slave-only?          boolean
     |     |  |  +--rw clock-type?          oc-ptp-types:clock-type-enumeration
     |     |  |  +--rw sdoid?               uint16
     |     |  |  +--rw network-transport?   oc-ptp-types:network-transport-enumeration
     |     |  |  +--rw unicast-multicast?   oc-ptp-types:unicast-multicast-enumeration
     |     |  |  +--rw domain-profile?      oc-ptp-types:domain-profile-enumeration
     |     |  |  +--rw udp6-scope?          uint8
     |     |  |  +--rw master-ip*           inet:ip-address
     |     |  +--ro state
     |     |     +--ro priority1?           uint8
     |     |     +--ro priority2?           uint8
     |     |     +--ro domain-number?       uint8
     |     |     +--ro slave-only?          boolean
     |     |     +--ro clock-type?          oc-ptp-types:clock-type-enumeration
     |     |     +--ro sdoid?               uint16
[...]

Issue #1039, opened 2024-01-26
Comments on proposed PTP model PR #1036


proberts2022:
Re; PR #1036 

I am new to this process of adding to the openconfig model but I hope review comments are welcomed.  I have many years working with PTP and when I look at this model that has been defined for PTP in openconfig, I see differences here compared to the datasets of the IEEE1588 standard itself and also with the draft YANG model as defined by the IEEE1588 working group.  

I think the model defined within openconfig should align with the datasets of the standard and avoid as much as possible additional attributes that may be implementation specific.

Here is a list of the items I saw that are different in this model compared to the standard:

Firstly, I am confused as to why the parent-dataset is defined as a list when an individual PTP clock instance can only have one single parent clock at any given time.  Within the IEEE1588 standard there is only one instance of a parentDS.  I think this level of list hierarchy should be removed.

Secondly, while the model does seem to include support for PTP over IP, it doesn’t include any support for the unicast discovery and unicast message negotiation.  Is the intention that only static IP is to be supported for now?

default-dataset:config:domain-profile 
-	this is defining the profile in use by the clock.  I agree this is necessary and the method to set this is not well defined in the standard itself.

default-dataset:config:network-transport 
-	This seems to be indicating that the PTP clock shall only operate with one transport protocol.  This is not a restriction required by the standard itself.  In theory, a clock may use PTP over UDP/IPv4 communication with some external PTP clocks while using PTP over UDP/IPv6 with other PTP clocks.  The transport layer protocol should be defined for each port in the port-dataset and I believe this is via the configurable underlying-interface.  I think this leaf should be removed

default-dataset:config:unicast-multicast 
port-dataset:config:unicast-multicast
-	I support the concept of this configurable parameter at the port level, but I don’t agree with a global setting for the entire clock.  And what are the procedures when a port setting conflicts with the global clock setting.  I think the default-dataset leaf should be removed.

default-dataset:config:udp6-scope
-	This is not in the standard but does seem useful as is to apply to all UDP/IPv6 multicast messages generated out of this PTP clock, if that transport mode is used.  Maybe this should be under the port so it the model allows for it to differ on a given IP communication path

default-dataset:config:master-ip [list]
-	It is not clear to me what this list is defining.  Is this a list of IP used by PTP clocks external to this clock and if so how is this list to be used?  Is this the acceptable master table but using IP addresses?  Is it addresses t which this clock should send Signallign messages with Unicast Negotiation TLVs?

[...]

PR #1103, opened 2024-05-02
Support PTP Yang model based on IEEE standard

sallylsy:
### Change Scope

Original PR: https://github.com/openconfig/public/pull/1036

This model is based on [ieee1588-ptp-ms.yang](https://github.com/YangModels/yang/blob/287f9e307c3b04282c6de1e75ab8afea6225f11a/standard/ieee/draft/1588/ieee1588-ptp-ms.yang)

Operational and configurational data are added based on the definition of IEEE standard.


RodneyCummings:
FYI: The official YANG modules for IEEE Std 1588e-2024 are now available at:
https://github.com/YangModels/yang/tree/main/standard/ieee/published/1588

The only change from the draft P1588e module used for this PR is change of "1588e-XXXX" to "1588e-2024" in the initial description (no changes to nodes).

proberts2022:
There now seems to be Issue #1039 , Pull request #1036 and this new Pull Request #1103 
Please advise on where we are to add comments

RodneyCummings:
Hi Peter,

I'm adding Doug for feedback on management of unicast and mixed unicast/multicast. 1588-2019 has data sets (i.e., info model) for unicast, and the YANG in 1588e modeled those data sets 1-1, and OpenConfig's YANG was in turn based on that 1588e YANG (as I understand). Doug was involved in the 1588-2019 data sets for unicast. I wasn't directly involved there.

Doug: My impression is that the OpenConfig YANG for 1588 is adding more data sets members for unicast. We might want to reflect that over to 1588 someday.

Rodney

dplore:
@sallylsy would you like to move this forward?

sallylsy:
Yes, sorry for the late reply.

unicast-multicast
The default-ds level indicates how we get the clock. And the port-ds level indicates how we send the clock down the stream - using unicast or multicast

default-ds:config:udp6-scope.
We inherited this leaf from SONIC. https://github.com/project-arlo/sonic-mgmt-framework/blob/562cd84ff3fec9ca705c7df621742f2daa61ce71/models/yang/sonic/sonic-ptp.yang#L477
Thanks for the explanation. From the description in SONIC, this will be used as the second byte of the primary address. My understanding is, this would be apply to all the IPv6 under the clock instance. We do not have an use case now, my suggestion would be 1. we leave this leaf at the current level, or 2. we drop this leaf from the model for now

master-ip
I double checked the IPs populated for this field. This is a list of host ip and associated with the design between ToR and host. The field may not apply generically, I would drop this leaf.

sallylsy:
I reconsidered udp6-scope. This is defined in SONIC yang model. We did not find a strong motivation from IEEE or use case for this leaf existing in OC model. Dropped this leaf

sallylsy:
IEEE copyright information is updated in the description

sallylsy:
Hi @dplore , I fixed the revision and lint errors.

dplore (approved):

PR #1105, opened 2024-05-03
Use reference of servers in /system/aaa/server-groups. Move the actual server list to top level of /system/aaa.

WenyiChengANET:
(M) system/openconfig-aaa-radius.yang
(M) system/openconfig-aaa-tacacs.yang
(M) system/openconfig-aaa.yang

**Change Scope:**

Add port, network-instance, type to aaa server as extra keys.
Deprecate port/authen-port from aaa-tacacs and aaa-radius config. Move it to server config as one of the keys.
Deprecate the server list in the “server-groups” container and move “servers” up to the top level for aaa.
Add a list of server ref matching the top level server list in the “server-groups” container.
This is a slow change which keeps the original nodes and marks them as deprecated.

**Platform Implementations:**

Arista AAA configuration:
Servers with the same host/port can be configured in different VRFs(network-instance).
One server can be added to multiple server groups.

```
(config)#radius-server host 1.2.3.4 vrf mgmt-vrf auth-port 123
RADIUS host 1.2.3.4 with auth-port 123 and acct-port 1813 created in vrf mgmt-vrf
(config)#radius-server host 1.2.3.4 vrf mgmt-vrf-2 auth-port 123
RADIUS host 1.2.3.4 with auth-port 123 and acct-port 1813 created in vrf mgmt-vrf-2
(config)#aaa group server radius rad-grp-1
(config-sg-radius-rad-grp-1)#server 1.2.3.4 vrf mgmt-vrf auth-port 123
(config-sg-radius-rad-grp-1)#server 1.2.3.4 vrf mgmt-vrf-2 auth-port 123
(config-sg-radius-rad-grp-1)#exit
(config)#aaa group server radius rad-grp-2
(config-sg-radius-rad-grp-2)#server 1.2.3.4 vrf mgmt-vrf auth-port 123
(config-sg-radius-rad-grp-2)#exit
(config)#show radius
…
RADIUS server-group: rad-grp-1
     0: 1.2.3.4, authentication port 123, accounting port 1813 (VRF mgmt-vrf)
     1: 1.2.3.4, authentication port 123, accounting port 1813 (VRF mgmt-vrf-2)

RADIUS server-group: rad-grp-2
     0: 1.2.3.4, authentication port 123, accounting port 1813 (VRF mgmt-vrf)
…
```

Arista [documentation](https://www.arista.com/en/um-eos/eos-user-security#xx1347513:~:text=root%20of%20trust%3E-,Server%20Groups,-A%20server%20group)

**Pyang output for the new model**:
***please note that the old nodes has been removed from this output for easy review. In the actual change they are marked as deprecated.**

```
module: openconfig-system
  +--rw system
     +--rw aaa
        +--rw config
        +--ro state
        ...<unchanged part>

        +--rw servers
        |  +--rw server* [address port network-instance type]
        |     +--rw address             -> ../config/address
        |     +--rw port                -> ../config/port
        |     +--rw network-instance    -> ../config/network-instance
        |     +--rw type                -> ../config/type
        |     +--rw config
        |     |  +--rw name?               string
        |     |  +--rw address?            oc-inet:ip-address
        |     |  +--rw port?               oc-inet:port-number
        |     |  +--rw network-instance?   oc-ni:network-instance-ref
        |     |  +--rw type?               identityref
        |     |  +--rw timeout?            uint16
        |     +--ro state
        |     |  +--ro name?                  string
        |     |  +--ro address?               oc-inet:ip-address
        |     |  +--ro port?                  oc-inet:port-number
        |     |  +--ro network-instance?      oc-ni:network-instance-ref
        |     |  +--ro type?                  identityref
        |     |  +--ro timeout?               uint16
        |     |  +--ro connection-opens?      oc-yang:counter64
        |     |  +--ro connection-closes?     oc-yang:counter64
        |     |  +--ro connection-aborts?     oc-yang:counter64
        |     |  +--ro connection-failures?   oc-yang:counter64
        |     |  +--ro connection-timeouts?   oc-yang:counter64
        |     |  +--ro messages-sent?         oc-yang:counter64
        |     |  +--ro messages-received?     oc-yang:counter64
        |     |  +--ro errors-received?       oc-yang:counter64
        |     +--rw tacacs
        |     |  +--rw config
        |     |  |  x--rw port?                oc-inet:port-number
        |     |  |  +--rw secret-key?          oc-types:routing-password
        |     |  |  +--rw secret-key-hashed?   oc-aaa-types:crypt-password-type
        |     |  |  +--rw source-address?      oc-inet:ip-address
        |     |  +--ro state
        |     |     x--ro port?                oc-inet:port-number
        |     |     +--ro secret-key?          oc-types:routing-password
        |     |     +--ro secret-key-hashed?   oc-aaa-types:crypt-password-type
        |     |     +--ro source-address?      oc-inet:ip-address
        |     +--rw radius
        |        +--rw config
        |        |  x--rw auth-port?             oc-inet:port-number
        |        |  +--rw acct-port?             oc-inet:port-number
        |        |  +--rw secret-key?            oc-types:routing-password
        |        |  +--rw secret-key-hashed?     oc-aaa-types:crypt-password-type
        |        |  +--rw source-address?        oc-inet:ip-address
        |        |  +--rw retransmit-attempts?   uint8
        |        +--ro state
        |           x--ro auth-port?             oc-inet:port-number
        |           +--ro acct-port?             oc-inet:port-number
        |           +--ro secret-key?            oc-types:routing-password
        |           +--ro secret-key-hashed?     oc-aaa-types:crypt-password-type
        |           +--ro source-address?        oc-inet:ip-address
        |           +--ro retransmit-attempts?   uint8
        |           +--ro counters
        |              +--ro retried-access-requests?   oc-yang:counter64
        |              +--ro access-accepts?            oc-yang:counter64
        |              +--ro access-rejects?            oc-yang:counter64
        |              +--ro timeout-access-requests?   oc-yang:counter64
        +--rw server-groups
           +--rw server-group* [name]
              +--rw name             -> ../config/name
              +--rw config
              |  +--rw name?   string
              |  +--rw type?   identityref
              +--ro state
              |  +--ro name?   string
              |  +--ro type?   identityref
              +--rw group-members
                 +--rw group-member* [address port network-instance type]
                    +--rw address             -> ../config/address
                    +--rw port                -> ../config/port
                    +--rw network-instance    -> ../config/network-instance
                    +--rw type                -> ../config/type
                    +--rw config
                    |  +--rw address?            -> ../../../../../../servers/server/address
                    |  +--rw port?               -> deref(../address)/../port
                    |  +--rw network-instance?   -> deref(../port)/../network-instance
                    |  +--rw type?               -> deref(../network-instance)/../type
                    +--ro state
                       +--ro address?            -> ../../../../../../servers/server/address
                       +--ro port?               -> deref(../address)/../port
                       +--ro network-instance?   -> deref(../port)/../network-instance
                       +--ro type?               -> deref(../network-instance)/../type
```

### DIFF
--- a/release/models/system/openconfig-aaa.yang
+++ b/release/models/system/openconfig-aaa.yang
@@ -12,6 +12,7 @@ module openconfig-aaa {
   import openconfig-inet-types { prefix oc-inet; }
   import openconfig-yang-types { prefix oc-yang; }
   import openconfig-aaa-types { prefix oc-aaa-types; }
+  import openconfig-network-instance { prefix "oc-ni"; }
 
   include openconfig-aaa-tacacs;
   include openconfig-aaa-radius;
@@ -100,6 +101,7 @@ module openconfig-aaa {
   grouping aaa-servergroup-common-config {
     description
       "Configuration data for AAA server groups";
+    status deprecated;
 
     leaf name {
       type string;
@@ -120,6 +122,7 @@ module openconfig-aaa {
   grouping aaa-servergroup-common-state {
     description
       "Operational state data for AAA server groups";
+    status deprecated;
 
     //TODO: add list of group members as opstate
   }
@@ -127,6 +130,7 @@ module openconfig-aaa {
   grouping aaa-servergroup-common-top {
     description
       "Top-level grouping for AAA server groups";
+    status deprecated;
 
     container server-groups {
       description
@@ -151,7 +155,9 @@ module openconfig-aaa {
           description
             "Configuration data for each server group";
 
-          uses aaa-servergroup-common-config;
+          uses aaa-servergroup-common-config {
+            status deprecated;
+          }
         }
 
         container state {
@@ -160,11 +166,17 @@ module openconfig-aaa {
           description
             "Operational state data for each server group";
 
-          uses aaa-servergroup-common-config;
-          uses aaa-servergroup-common-state;
+          uses aaa-servergroup-common-config {
+            status deprecated;
+          }
+          uses aaa-servergroup-common-state {
+            status deprecated;
+          }
         }
 
-        uses aaa-server-top;
+        uses aaa-server-top {
+          status deprecated;
+        }
       }
     }
   }
@@ -172,6 +184,7 @@ module openconfig-aaa {
   grouping aaa-server-config {
     description
       "Common configuration data for AAA servers";
+    status deprecated;
 
     leaf name {
       type string;
@@ -193,6 +206,62 @@ module openconfig-aaa {
         server";
     }
   }
+
+  grouping aaa-server-config-data {
+    description
+      "Common configuration data for AAA servers";
+
+    leaf name {
+      type string;
+      description
+        "Name assigned to the server";
+    }
+
+
+    leaf address {
+      type oc-inet:ip-address;
+      description "Address of the authentication server";
+    }
+
+    leaf port {
+      type oc-inet:port-number;
+      description "Port of the authentication server";
+    }
+
+    leaf network-instance {
+      type oc-ni:network-instance-ref;
+      description "Network-instance of the authentication server";
+    }
+
+    leaf type {
+      type identityref {
+        base oc-aaa-types:AAA_SERVER_TYPE;
+      }
+      description
+        "AAA server type";
+    }
+
+    leaf timeout {
+      type uint16;
+      units seconds;
+      description
+        "Set the timeout in seconds on responses from the AAA
+        server";
+    }
+
+    list server-group {
+      key "name";
+      description
+        "List of the server groups this server belongs to";
+
+      leaf name {
+        type string;
+        description
+          "Server group name";
+      }
+    }
+  }
+
 
   grouping aaa-server-state {
     description
@@ -251,9 +320,83 @@ module openconfig-aaa {
 
   }
 
+  grouping aaa-server-list-top {
+    description
+      "Top-level grouping for list of AAA servers";
+
+    container servers {
+      description
+        "Enclosing container the list of servers";
+
+      list server {
+        key "address port network-instance type";
+        description
+          "List of AAA servers";
+
+        leaf address {
+          type leafref {
+            path "../config/address";
+          }
+          description
+            "Reference to the configured address of the AAA server";
+        }
+
+        leaf port {
+          type leafref {
+            path "../config/port";
+          }
+          description
+            "Reference to the configured port of the AAA server";
+        }
+
+        leaf network-instance {
+          type leafref {
+            path "../config/network-instance";
+          }
+          description
+            "Reference to the configured network-instance of the AAA server";
+        }
+
+        leaf type {
+          type leafref {
+            path "../config/type";
+          }
+          description
+            "AAA server type";
+        }
+
+        container config {
+          description
+            "Configuration data ";
+
+          uses aaa-server-config-data;
+        }
+
+        container state {
+          config false;
+
+          description
+            "Operational state data ";
+
+          uses aaa-server-config-data;
+          uses aaa-server-state;
+        }
+
+        uses aaa-tacacs-server-top {
+          when "config/type = 'oc-aaa:TACACS'";
+        }
+
+        uses aaa-radius-server-top {
+          when "config/type = 'oc-aaa:RADIUS'";
+        }
+      }
+    }
+  }
+
   grouping aaa-server-top {
     description
       "Top-level grouping for list of AAA servers";
+    status deprecated;
 
     container servers {
       description
@@ -276,7 +419,9 @@ module openconfig-aaa {
           description
             "Configuration data ";
 
-          uses aaa-server-config;
+          uses aaa-server-config {
+            status deprecated;
+          }
         }
 
         container state {
@@ -285,7 +430,9 @@ module openconfig-aaa {
           description
             "Operational state data ";
 
-          uses aaa-server-config;
+          uses aaa-server-config {
+            status deprecated;
+          }
           uses aaa-server-state;
         }
 
@@ -834,7 +981,10 @@ module openconfig-aaa {
       uses aaa-authentication-top;
       uses aaa-authorization-top;
       uses aaa-accounting-top;
-      uses aaa-servergroup-common-top;
+      uses aaa-servergroup-common-top {
+        status deprecated;
+      }
+      uses aaa-server-list-top;
 
     }
   }
--- a/release/models/system/openconfig-aaa.yang
+++ b/release/models/system/openconfig-aaa.yang
@@ -116,34 +116,6 @@ module openconfig-aaa {
         "AAA server type -- all servers in the group must be of this
         type";
     }
-
-    list servers {
-      key "address port network-instance type";
-
-      leaf address {
-        type leafref {
-          path "/system/aaa/servers/server/address";
-        }
-      }
-
-      leaf port {
-        type leafref {
-          path "deref(../address)/../port";
-        }
-      }
-
-      leaf network-instance {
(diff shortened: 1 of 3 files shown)


WenyiChengANET:
It looks like that some checks are complaining about the deref I use in path. Is deref not support in path? Pyang doesn't give me any error on that.

If it's not supported, any suggestion on how to use leafref on a node with multiple keys and make sure all keys are from the same element? Below is the code using deref in the PR:
```
  grouping aaa-servergroup-member-config {
    description
      "Common configuration data for server group member";

    leaf address {
      type leafref {
        path "../../../../../../servers/server/address";
      }
      description
        "Reference to the configured address of the server group member";
    }

    leaf port {
      type leafref {
        path "deref(../address)/../port";
      }
      description
        "Reference to the configured port of the server group member";
    }

    leaf network-instance {
      type leafref {
        path "deref(../port)/../network-instance";
      }
      description
        "Reference to the configured network-instance of the server group member";
    }

    leaf type {
      type leafref {
        path "deref(../network-instance)/../type";
      }
      description
        "Reference to the configured type of the server group member";
    }
  }
  ```

**Comment by earies** [comment]
> It looks like that some checks are complaining about the deref I use in path. Is deref not support in path? Pyang doesn't give me any error on that.
> 

`deref()` is a valid YANG 1.1 XPath function - OpenConfig models are 1.0 based

`deref()` is also not a permitted path-arg in 1.1

https://datatracker.ietf.org/doc/html/rfc7950#section-9.9.2
https://datatracker.ietf.org/doc/html/rfc7950#page-205

**Comment by AI-Complete** [comment]
Wendyi who did you request review this commit?

**Comment by WenyiChengANET** [comment]
Removed deref from the path. Use name as the single key for server and as the leafref in server group member list.

File: release/models/system/openconfig-aaa.yang [hunk 1]
```diff
@@ -194,6 +266,49 @@ module openconfig-aaa {
...
+      type oc-inet:port-number;
+      description "Port of the authentication server";
+    }
+
+    leaf network-instance {
+      type oc-ni:network-instance-ref;
+      description "Network-instance of the authentication server";
+    }
+
+    leaf type {
```
dplore, line 294: This aaa server type overlaps/conflicts with the type defined at the server-group level.  Do we need this defined per server?

Also, I observe that there is only one base type, so maybe we don't really even need type at all?

WenyiChengANET: Thanks for reviewing Darren.
`This aaa server type overlaps/conflicts with the type defined at the server-group level. Do we need this defined per server?`
An AAA server can be used either directly by itself or within a server group, so we need this type defined per server as well. Vendors can decide what to do if they see a conflict between the type of the group and the type of the server.

`Also, I observe that there is only one base type, so maybe we don't really even need type at all?`
`leaf type` here is an identityref which has existing identity RADIUS and TACACS added to it from their own files:
https://github.com/openconfig/public/blob/1d1edd12ad5e8dc191b1d01fe53d55f72bf893d8/release/models/system/openconfig-aaa-radius.yang#L74
https://github.com/openconfig/public/blob/1d1edd12ad5e8dc191b1d01fe53d55f72bf893d8/release/models/system/openconfig-aaa-tacacs.yang#L73
so it seems to have two types with it. Am I using this wrong?

WenyiChengANET:
> Placeholder note for myself, this change looks like it is needed for supporting aaa servers that have the same IP address but are in different network instances, is that right @WenyiChengANET ?

This is correct. It also moves the `servers` list out from `server-groups` and puts it on the top level, so that the same server can be added to multiple groups without having duplicated copies.

dplore:
This change is needed for configuring AAA servers in network-instances.  Although there is some variability regarding configuring individual servers vs. always requiring a server group, there are two instances allowing this, so I am approving.

Quoting more implementation references:

 [Arista EOS documentation](https://www.arista.com/en/um-eos/eos-user-security#xx1347513:~:text=root%20of%20trust%3E-,Server%20Groups,-A%20server%20group) shows server and server group configuration.

[Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/security/b-system-security-cr-cisco8000/authentication-authorization-and-accounting-commands.html#wp3457061824) supports single aaa server and grouped aaa server configuration.

[JunOS] appears to only support aaa configuration in a group

[SRLinux](https://documentation.nokia.com/srlinux/22-3/SR_Linux_Book_Files/Data_Model_Reference/srl_nokia-system_0.html#d4689) appears to only support aaa configuration in a group

dplore (approved):

dplore:
Setting last call to Nov 14, 2024

jsterne:
Should this one maybe get a bit more visibility and discussion? I don't think it has come up on the monthly vendor meeting. Although it isn't NBC right now (since the old model is still available but deprecated) it will be a breaking change later.

The references above do show implementations with servers defined outside of server-groups, but all those implementations seem to use address as the server key today.  

Do any OpenConfig models actually point to an individual server rather than a server-group? Is there a plan to do so? I can see how some implementations may support that but if OpenConfig models don't refer to an individual server then is there a need to make that change in the model?

Adding the NI leafref does seem like a needed addition though.

dplore:
Revised last call date 2024-Dec-20

LimeHat:
> Deprecate the server list in the “server-groups” container and move “servers” up to the top level for aaa.

>An AAA server can be used either directly by itself or within a server group, so we need this type defined per server as well. 

1. OpenConfig today does not define the usage of a single server by itself. Nor it should, in my opinion.  
Accounting and authorization method reference server groups (which makes sense) or an identity. Is there an operational use case for singular, group-independent server usage?

https://github.com/openconfig/public/blob/f01a86de7e54f5c6f54f6b74c408a08014e6d4a3/release/models/system/openconfig-aaa.yang#L464-L481

https://github.com/openconfig/public/blob/f01a86de7e54f5c6f54f6b74c408a08014e6d4a3/release/models/system/openconfig-aaa.yang#L613-L630


>Vendors can decide what to do if they see a conflict between the type of the group and the type of the server.
2. Leaving such things up to "vendors to decide" is not great since the purpose of OC is to define a consistent behavior across all implementations.

3. This seems like a major change (which is breaking from the operational point of view, even thought not breaking in yang terms as noted by Jason) that is done with the primary purpose of aligning the OC model with the Arista's implementation.

rgwilton:
I've checked with the IOS XR AAA team, and we allow a server to be configured outside a server group and then referenced (but as Jason indicates it is keyed by IP address).

I think that the change that is being proposed here is more impactful than it needs to be.  I.e., I suspect in some (perhaps many) cases, defining the servers under a server-group is logically the right place to put them.  So, I don't think that we should be deprecating those existing leaves.

It seems that the core desire is to allow servers being listed outside server-groups, so that they can be shared across multiple groups.  This seems reasonable, but I would perhaps renamed the top level list from servers to shared-servers, and rename group-members to something like shared-server-members.

dplore:
Based on feedback from the OC operators and OC community meetings, there is consensus that we need to add network-instance to the aaa models.  There is not consensus to refactor the model to allow aaa servers to be defined outside of a aaa group.  

@WenyiChengANET to move this PR partially forward, can you refactor this one or open a new PR which only adds `network-instance`.

dplore:
Hi @WenyiChengANET we would still like to see the change for just the network-instance and move the grouping changes to a separate PR.  We are not seeing requests from operators or vendors outside of Arista to change the model regarding the groupings.

dplore:
Hi @WenyiChengANET we are waiting to refactor this to only include the network-instance addition.

dplore (changes requested):
Please refactor this PR (or create a new one) which only adds the network-instance changes

PR #1109, opened 2024-05-08
Updates to the ES model to allow the configuration of an ES-import route target

abamberger-arista:
### Change Scope

The existing openconfig ethernet segment model has no way to configure the ES-import route target (defined in RFC7432 Section 7.6) for a type 0 operator configured ESI.  This proposal adds a new field, “es-import-route-target”, to the existing ES config model, to allow the explicit configuration of an ES-import route target.

New tree state after proposed change (additions in **bold**):
<pre>
module: openconfig-ethernet-segments
+--rw ethernet-segments
   +--rw ethernet-segment* [name]
      +--rw config
      <b>|  +--rw es-import-route-target?   oc-yang-types:mac-address</b>
      |  ...
      +--ro state
         <b>+--ro es-import-route-target?   oc-yang-types:mac-address</b>
	   ...
</pre>

New Yang Paths:
* ethernet-segments/ethernet-segment/config/es-import-route-target
* ethernet-segments/ethernet-segment/state/es-import-route-target

#### Arista EOS:
https://www.arista.com/en/um-eos/eos-vxlan-configuration#topic_ckc_dh4_ynb
https://www.arista.com/en/support/toi/eos-4-22-0f/14236-evpn-vxlan-all-active-multihoming
<pre>
interface Port-Channel1
   evpn ethernet-segment
      identifier 00aa:bbbb:cccc:dddd:eeee
      route-target import 12:23:34:45:56:67
</pre>

### DIFF
--- a/release/models/ethernet-segments/openconfig-ethernet-segments.yang
+++ b/release/models/ethernet-segments/openconfig-ethernet-segments.yang
@@ -26,9 +26,9 @@ module openconfig-ethernet-segments {
 
   oc-ext:openconfig-version "0.2.0";
 
-  revision "2024-05-08" {
+  revision "2024-06-10" {
     description
-      "Added ability to allow configuring es-import-route-target for an ES";
+      "Added ability to allow configuring es-import-route-target for a type-0 ES";
     reference "0.2.0";
   }
 
@@ -207,6 +207,10 @@ grouping evpn-ethernet-segment-df-config {
     }
 
     leaf es-import-route-target {
+      when "../esi = 'oc-evpn-types:TYPE_0_OPERATOR_CONFIGURED'" {
+        description
+          "For Type-0 ESIs, manually specify an ES-import route target";
+      }
       type oc-yang-types:mac-address;
       description
         "Received ES routes with an ES-Import extended community set to this value
--- a/release/models/ethernet-segments/openconfig-ethernet-segments.yang
+++ b/release/models/ethernet-segments/openconfig-ethernet-segments.yang
@@ -210,7 +210,9 @@ grouping evpn-ethernet-segment-df-config {
       type oc-yang-types:mac-address;
       description
         "Received ES routes with an ES-Import extended community set to this value
-        are imported into this ethernet segment";
+        are imported into this ethernet segment.  When configured by a user, the
+        es-import-route-target will be used, otherwise the system automatically
+        derives the value used by following the procedures in RFC 7432 Section 7.6.";
       reference
         "RFC 7432: BGP MPLS-Based Ethernet VPN Section 7.6";
     }
--- a/release/models/ethernet-segments/openconfig-ethernet-segments.yang
+++ b/release/models/ethernet-segments/openconfig-ethernet-segments.yang
@@ -26,9 +26,9 @@ module openconfig-ethernet-segments {
 
   oc-ext:openconfig-version "0.2.0";
 
-  revision "2024-06-10" {
+  revision "2024-09-06" {
     description
-      "Added ability to allow configuring es-import-route-target for a type-0 ES";
+      "Added ability to configure an explicit es-import-route-target for an ES";
     reference "0.2.0";
   }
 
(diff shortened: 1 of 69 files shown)


jorabada:
The ES import route target is autoderived in case of ES types 1, 2 and 3, and may be autoderived in the rest of the cases too. That is what other implementations do as well. So I don't think this change is necessary...

abamberger-arista:
RFC7432 only specifies how the ES-import RT can be auto-derived from ES types 1, 2, and 3.  In particular, for type 0 ESs, since they're operator configured and completely arbitrary, and the RFC doesn't specify any way to do auto-derivation for that type, I think this is required at least to be able to properly support type 0 ESs in a general way.

abamberger-arista:
I updated the PR to add a when clause to the new "es-import-route-target" leaf to make it only apply when the ESI type is Type-0, which should make this a bit cleaner to use.  As mentioned earlier, types 1, 2, and 3 are defined by the RFC to be auto-derived from the ESI, so don't require an explicitly specified es-import-rt.  Types 4 and 5 aren't specified one way or the other, but the when clause can be updated to include these in the future if necessary.

dplore:
@jorabada can you please comment?

jorabada:
> I updated the PR to add a when clause to the new "es-import-route-target" leaf to make it only apply when the ESI type is Type-0, which should make this a bit cleaner to use. As mentioned earlier, types 1, 2, and 3 are defined by the RFC to be auto-derived from the ESI, so don't require an explicitly specified es-import-rt. Types 4 and 5 aren't specified one way or the other, but the when clause can be updated to include these in the future if necessary.

I don't think I agree with this. The es-import route-target can be autoderived for all types. rfc7432bis states that is autoderived for types 1..3, and MAY be autoderived for the other types. If you do this for type 0, I don't see why you wouldn't do it for the other types. Also, I suppose implementations do autoderivation as a bare minimum, and explicit configuration is supported only in some implementations out there.

abamberger-arista:
Would it make more sense to change the "es-import-route-target" leaf to be a union of either an explicit value, or an enum with an AUTO member?  That would then allow the user to be more explicit, both when configuring AUTO, and when configuring an explicit es-import-rt.  The explicit configuration portion of the union could be made conditional on types 0, 4, and 5, to adhere to the new draft RFC.

jorabada:
Thanks @abamberger-arista - as I was saying, I don't think this change is necessary at all. If there is really push from people to have it, your original suggestion is simpler: When configured explicitly, the es-import-route-target will be used. Otherwise, the autoderived one is used.

abamberger-arista:
@jorabada I guess I'm not seeing a good reason not to add it.  It's something explicitly defined in the RFC, it's backwards compatible, and it's optional.  I can revert back to the original proposal (which I agree is simpler, although it would technically allow users to push an explicit es-import-rt with ESI types that per the RFC are always supposed to be auto-generated, that is, types 1-3).  Are there any objections to getting this in in its original form (a single additional optional es-import-rt leaf)?

jorabada:
> @jorabada I guess I'm not seeing a good reason not to add it. It's something explicitly defined in the RFC, it's backwards compatible, and it's optional. I can revert back to the original proposal (which I agree is simpler, although it would technically allow users to push an explicit es-import-rt with ESI types that per the RFC are always supposed to be auto-generated, that is, types 1-3). Are there any objections to getting this in in its original form (a single additional optional es-import-rt leaf)?

As I said, I don't think it is necessary, but I won't object if you add it (in its original form)... thanks.

abamberger-arista:
Thanks for taking a look again, I've restored it to its original form.  I understand this may be a niche use case in practice, but since I don't really see much of a downside in adding it, and it will allow implementations that do support it to be a bit more full featured when using OC for configuration.  @dplore are all concerns resolved from your end on this?

release/models/ethernet-segments/openconfig-ethernet-segments.yang
```diff
@@ -199,6 +206,15 @@ grouping evpn-ethernet-segment-df-config {
         "RFC 7432: BGP MPLS-Based Ethernet VPN";
     }
 
+    leaf es-import-route-target {
+      type oc-yang-types:mac-address;
+      description
+        "Received ES routes with an ES-Import extended community set to this value
+        are imported into this ethernet segment";
```
dplore, line 213: Can you add text regarding auto derived and what happens if configured?  ie:

"When configured by a user, the es-import-route-target will be used, otherwise the system automatically derives the value used."
Also, end the sentence with a period please.  Thanks!

dplore: Setting to last call for comments.  THis will merge Oct 21, 2024

abamberger-arista: Done

dplore (approved):

PR #1110, opened 2024-05-13
Correct description of QSFP56_DD to be generic.

ejbrever:
This is a correction to https://github.com/openconfig/public/pull/564

The initial description referenced 200 and 400G which causes confusion as 800G is becoming possible.


ejbrever:
@dplore would you mind taking a look please? This is just a description update. Thanks.

dplore (approved):

rgwilton:
Would it be better to use different identifies for the higher bandwidth ones, e.g. keep QSFP56_DD to a max of 400G?

Looking at http://www.qsfp-dd.com/wp-content/uploads/2023/09/QSFP-DD-Hardware-Rev7.0.pdf seems to suggest perhaps calling the others QSFP-DD800 and QSFP-DD1600.

ejbrever:
The thought process was that the form factor is the same as before and that the rate is additionally defined elsewhere. To represent 800G you can set (QSFP56_DD + PROT_800GE). So it seemed redundant to have form factor types that represent a form factor plus the rate. There is also a maintenance cost in the model if we have to add new form factors to represent every rate. Please let us know your thoughts though.

rgwilton:
I was thinking that using different types would help mean indicate that plugging in a 800G optics into a 400G slot isn't going to work.

It is also worth noting the previous definitions also define maximum rates:
```
  identity QSFP28_DD {
    base TRANSCEIVER_FORM_FACTOR_TYPE;
    description
      "QSFP-DD with electrical interfaces consisting of 8 lanes that operate at up to
      25 Gbps with NRZ modulation";
    reference "http://qsfp-dd.com";
  }
  identity QSFP56 {
    base TRANSCEIVER_FORM_FACTOR_TYPE;
    description
      "QSFP pluggable optic with support for up to 4x56G physical
      channels";
  }
```

This proposed change seems to be treating QSPF56_DD differently from the previous ones?

ejbrever:
I think revisiting some of those other descriptions is worth while if those form factors offer support for multiple rates. I haven't asked around though if that is the case though.

For RMAs, I am not sure operators would want the form factor to play a role in deciding what optic to replace it with. I think the part number would be the preferred data to determine that.

In the previous discussions, the form factor was the same though with QSFP56_DD Type1 and Type2. Then we started talking about Type3 for 800G and it was noted that the form factor is also the same. There may be some different heat sink extrusions, but the form factor was same. There's a bit more discussion in https://github.com/openconfig/public/pull/564 as well which may be of interest.

dplore:
> I think revisiting some of those other descriptions is worth while if those form factors offer support for multiple rates. I haven't asked around though if that is the case though.
> 
> For RMAs, I am not sure operators would want the form factor to play a role in deciding what optic to replace it with. I think the part number would be the preferred data to determine that.
> 
> In the previous discussions, the form factor was the same though with QSFP56_DD Type1 and Type2. Then we started talking about Type3 for 800G and it was noted that the form factor is also the same. There may be some different heat sink extrusions, but the form factor was same. There's a bit more discussion in #564 as well which may be of interest.


In #564 we asserted the PMD should be used to differentiate link speed.  The form factor identity describes the physical package of the transceiver interface, including the number of electrical lanes.    Happy to have more comments on this if it is creating issues.

Ethernet PMDs: [/components/component/transceiver/state/ethernet-pmd](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-transceiver-state-ethernet-pmd)

Transceiver form factors: [/components/component/transceiver/state/form-factor](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-transceiver-state-form-factor)

dplore:
Last call for comments.  This will merge on June 3, 2024.

PR #1114, opened 2024-05-25
enabling multiple protocol instances in

rszarecki:
Enable support for source and desination instance of respective protocol.

change /network-instances/network-instance/tables/ so it become read-only and make tables per protocol instance.
  +--rw network-instances
        +--rw tables
           +--rw table* [protocol instance-name address-family]
              +--rw protocol          -> ../state/protocol
              +--rw address-family    -> ../state/address-family
              +--rw instance-name     -> ../state/instance-name
              +--ro state
                 +--ro protocol?         -> ../../../../protocols/protocol/config/identifier
                 +--ro instance-name?    -> ../../../../protocols/protocol/config/name
                 +--ro address-family?   identityref
I still do not see much value, but ...
2. Extent table-connections to include dst-instance-name and src-instance-name. While doing so define config leafs of keys references to .../protocols/protocol/config rather then to .../tables/table/config as it is today

  +--rw network-instances
        +--rw table-connections
           +--rw table-connection* [src-protocol src-name dst-protocol dst-name address-family]
              +--rw src-protocol         -> ../config/src-protocol
              +--rw src-instance-name?   -> ../config/src-instance-name
              +--rw dst-protocol?        -> ../config/dst-protocol
              +--rw dst-instance-name?   -> ../config/dst-instance-name
              +--rw address-family?      -> ../config/address-family
              +--rw config
              |  +--rw src-protocol?                 -> ../../../../protocols/protocol/config/identifier
              |  +--rw src-instance-name?            -> ../../../../protocols/protocol/config/name
              |  +--rw address-family?               -> ../../../../protocols/protocol[identifier=current()/../src-protocol]/config/address-family
              |  +--rw dst-protocol?                 -> ../../../../protocols/protocol/config/identifier
              |  [...]
              +--ro state
                 +--ro src-protocol?                 -> ../../../../protocols/protocol/config/identifier
                 +--ro src-instance-name?            -> ../../../../protocols/protocol/config/name
                 +--ro address-family?               -> ../../../../protocols/protocol[identifier=current()/../src-protocol]/config/address-family
                 +--ro dst-protocol?                 -> ../../../../protocols/protocol/config/identifier
                 +--ro dst-instance-name?            -> ../../../../protocols/protocol/config/name
                 [...]
I personally do not like string "table-connections". Network engineers would rather understand "redistribition", "route exporting". But "table-connections" sounds odd and ambigous with forwarding table chaining. I keep "table-connections" because this make change non-breaking.

### Cisco XR
```
router bgp 12345 instance FOO
   net 00.0000.0000.12a5.00
   address-family ipv4 unicast
    metric-style wide
    redistribute isis BAR level 2
!
router isis BAR
   ...
```

### DIFF
--- a/release/models/network-instance/openconfig-network-instance.yang
+++ b/release/models/network-instance/openconfig-network-instance.yang
@@ -481,7 +481,7 @@ module openconfig-network-instance {
             between tables";
 
           list table-connection {
-            key "src-protocol dst-protocol address-family";
+            key "src-protocol src-name dst-protocol dst-name address-family";
 
             description
               "A list of connections between pairs of routing or
@@ -495,6 +495,7 @@ module openconfig-network-instance {
 
             reference
               "Route Redistribution in OpenConfig Network Instance:
+              // [TODO:rewrite to represent changes]
               https://github.com/openconfig/public/blob/master/doc/network_instance_redistribution.md#interconnection-of-protocol-ribs";
 
             leaf src-protocol {
@@ -502,9 +503,21 @@ module openconfig-network-instance {
                 path "../config/src-protocol";
               }
               description
-                "The name of the protocol associated with the table
+                "The protocol associated with the table
                 which should be utilised as the source of forwarding
                 or routing information";
+            }
+
+            leaf src-instance-name {
+              type leafref {
+                path "../config/src-instance-name";
+              }
+              description
+                "The name of the instance of protocol associated with the table
+                which should be utilised as the source of forwarding
+                or routing information.
+                If not specified and only one instance of given protocol is
+                configured, implementation may default to this instance.";
             }
 
             leaf dst-protocol {
@@ -512,8 +525,21 @@ module openconfig-network-instance {
                 path "../config/dst-protocol";
               }
               description
-                "The table to which routing entries should be
-                exported";
+                "The protocol associated with the table
+                which should be utilised as the destination of forwarding
+                or routing information";
+            }
+
+            leaf dst-instance-name {
+              type leafref {
+                path "../config/dst-instance-name";
+              }
+              description
+                "The name of the instance of protocol associated with the table
+                which should be utilised as the destination of forwarding
+                or routing information.
+                If not specified and only one instance of given protocol is
+                configured, implementation may default to this instance.";
             }
 
             leaf address-family {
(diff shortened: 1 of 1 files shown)


dplore:
Instead of refactoring tables and  table-connections, how about deprecating both of those completely and introducing a new tree
`/network-instances/network-instance/protocol-redistribution`

It could look like that you have proposed.  I would say since we are using src-instance-name, we no longer need the protocol leaf.  A protocols/protocol/config/name will only be one protocol.  (no longer a table concept with multiple protocols possible).   ie:

```
  +--rw network-instances
        +--rw protocol-redistributions
           +--rw protocol-redistribution* [src-instance-name  dst-instance-name address-family]
              +--rw src-instance-name?   -> ../config/src-instance-name
              +--rw dst-instance-name?   -> ../config/dst-instance-name
              +--rw address-family?      -> ../config/address-family
              +--rw config
              |  +--rw src-instance-name?            -> ../../../../protocols/protocol/config/name
              |  +--rw dst-instance-name?            -> ../../../../protocols/protocol/config/name
              |  +--rw address-family?               -> ../../../../protocols/protocol[identifier=current()/../src-protocol]/config/address-family
              |  [...]
              +--ro state
                 +--ro src-instance-name?            -> ../../../../protocols/protocol/config/name
                 +--ro dst-instance-name?            -> ../../../../protocols/protocol/config/name
                 +--ro address-family?               -> ../../../../protocols/protocol[identifier=current()/../src-protocol]/config/address-family
                 [...]

```

WDYT?

rszarecki:
@dplore 
1. from my point of view, we can deprecate `/network-instances/network-instance/tables/ ` completly.
2. we need both `src/dst-protocol` and `src/dst-instance-name`. This is because instance name is free-text-string and can be same for multiple protocols. For example: ( src-protocol=BGP, src-instance-name="default", dst-protocol=ISIS, src-instance-name="default").
3. "table-connections" --> "protocol-redistributions" (or route-redistributions); This would be breaking change. I hope if we keep table-connections AND allow inertpetate lack of instance-name as "the only one existing" (see description), we may be backward compatible.

nokia1adam:
> @dplore
> 
> 1. from my point of view, we can deprecate `/network-instances/network-instance/tables/ ` completly.


+1 on deprecating /network-instances/network-instance/tables.

As configuration objects they serve no purpose other than anchoring the endpoints of a table connection, and surely there are simpler ways to accomplish this, especially in light of trying to redistribute routes of a particular protocol instance.

If the configuration of these tables is supposed to be added automatically by the system (and protected from delete), what happens when there is a replace-from-root type operation?

If the tables were state only I would have less objection, but that breaks their utility as a leaf-ref target and takes us back to square one of just using a simpler way to define a table connection.

rszarecki:
In this proposal, table-connections KEYs are leaf-ref to '/protocol/protocols/' directly. Not to fables as it was previously.

rszarecki:
I refactored it quite heavly, following off-line discussion. Please see PR description.

dplore:
This does support the operational use case to redistribute routes between two network instances, which the current model does not (at least, not as far I can determine).    So I am in favor of making this change.

I'd like to ask the community for more feedback.  Adding this to the November 2024 openconfig community meeting.

rszarecki:
> This does support the operational use case to redistribute routes between two network instances, which the current model does not (at least, not as far I can determine).    So I am in favor of making this change.
> 
> 
> 
> I'd like to ask the community for more feedback.  Adding this to the November 2024 openconfig community meeting.
> 
> 
> 
> 

Correction: This does support the operational use case to redistribute routes between two protocol instances (of potentially the same protocol like BGP) within the same network-instance

dplore:
This was presented at the Nov 7, 2024 OC Community meeting to create more visibility.  We'd like to see comments from the community on this proposal.

PR #1116, opened 2024-05-29
Clarify /network-instances/network-instance/tables 

dplore:
### Change Scope

* Explicitly state that users cannot create or delete `/network-instances/network-instance/tables/table` nodes.
* This change is a minor clarification and backwards compatible.

### DIFF
--- a/release/models/network-instance/openconfig-network-instance.yang
+++ b/release/models/network-instance/openconfig-network-instance.yang
@@ -643,9 +643,8 @@ module openconfig-network-instance {
 	            removal of the table should not require additional or
 	            explicit configurations.
 
-              Users are not expected to create or delete tables.
-              Instead a user may configure table-connections which
-              reference these tables.";
+              Users cannot create or delete tables.  Instead a user may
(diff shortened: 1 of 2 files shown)


release/models/network-instance/openconfig-network-instance.yang
```diff
@@ -633,7 +641,11 @@ module openconfig-network-instance {
               address families enabled, the protocol=BGP,
               address-family=IPv4 table is created by the system. The
 	            removal of the table should not require additional or
-	            explicit configurations";
+	            explicit configurations.
+
+              Users are not expected to create or delete tables.
```
chrisy, line 646: nit: Perhaps pedantic but since we’re clarifying; to me saying users are not expected to do a thing implies that they still able to do that thing but it is not required.

chrisy (commented):
One minor but otherwise lgtm

release/models/network-instance/openconfig-network-instance.yang
dplore: That's fair.

PR #1117, opened 2024-05-30
fix install-position

dplore:
Fix the type for install-position to a string.  The leafref type assigned in #1095 was unintended.

### DIFF
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -20,7 +20,13 @@ submodule openconfig-platform-common {
     "This modules contains common groupings that are used in multiple
     components within the platform module.";
 
-  oc-ext:openconfig-version "0.26.0";
+  oc-ext:openconfig-version "0.27.0";
+
+  revision "2024-05-29" {
+    description
+      "Change install-position from leaf-ref to string.";
+    reference "0.27.0";
+  }
 
    revision "2024-04-12" {
     description
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,7 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-oc-ext:openconfig-version "0.26.0";
+  oc-ext:openconfig-version "0.27.0";
+
+  revision "2024-05-29" {
+    description
+      "Change install-position from leaf-ref to string.";
+    reference "0.27.0";
+  }
 
   revision "2024-04-12" {
     description
@@ -431,9 +437,7 @@ oc-ext:openconfig-version "0.26.0";
     }
 
     leaf install-position {
-      type leafref {
-        path "../name";
-      }
+      type string;
       description
         "System-supplied index to a position where this component is
         installed.  The position may be referred in device documenation


dplore:
@SydneyCaulfeild

SydneyCaulfeild (approved):

robshakir (approved):
This looks OK to revert to me. 

A question: given that this is a backwards incompatible change, what's the policy on:

a) versioning of this change? (should this model, being implemented already not be a 0.x module?)
b) whether there is a need for a backwards incompatibility marker given that we are changing the type from leafref to string to just string? (the encoding should be the same, the validation rules are now just looser).

dplore:
> This looks OK to revert to me.
> 
> A question: given that this is a backwards incompatible change, what's the policy on:
> 
> a) versioning of this change? (should this model, being implemented already not be a 0.x module?) b) whether there is a need for a backwards incompatibility marker given that we are changing the type from leafref to string to just string? (the encoding should be the same, the validation rules are now just looser).

In this case, the change is to a new leaf that was introduced about 24 hours earlier.   Between the current (as of the PR) [OC release v4.1.0](https://github.com/openconfig/public/releases/tag/v4.1.0) and the next release (tentatively v4.2.0), this will appear as the addition of a leaf and a non-breaking change.  The leafref to string change will not be observed between these two OC public releases.   So I believe we should not mark this change as breaking.

Regarding whether the platform model should be a v0 or v1+ is a separate discussion I think.  For those who might not be familiar, the [OC release policy](https://github.com/openconfig/public/blob/9eeeb2019ac61d501feae5127eb31bc4db15a048/doc/releases.md?plain=1#L59-L61) states  v0.x models are not subject to have breaking changes declared.  In practice, v0.x models are only promoted to v1.x when the model is expected to no longer have breaking changes made on a frequent basis.  

To be very strict, this is a breaking change, albeit as Rob said, only for a more loose validation rule.  If there was some implementation of the leafref node, a new client might attempt to use a string which does not match any component name which is now legal in the string node, yet the server would deny it.

earies:
> > This looks OK to revert to me.
> > A question: given that this is a backwards incompatible change, what's the policy on:
> > a) versioning of this change? (should this model, being implemented already not be a 0.x module?) b) whether there is a need for a backwards incompatibility marker given that we are changing the type from leafref to string to just string? (the encoding should be the same, the validation rules are now just looser).
> 
> In this case, the change is to a new leaf that was introduced about 24 hours earlier. Between the current (as of the PR) [OC release v4.1.0](https://github.com/openconfig/public/releases/tag/v4.1.0) and the next release (tentatively v4.2.0), this will appear as the addition of a leaf and a non-breaking change. The leafref to string change will not be observed between these two OC public releases. So I believe we should not mark this change as breaking.
> 
> Regarding whether the platform model should be a v0 or v1+ is a separate discussion I think. For those who might not be familiar, the [OC release policy](https://github.com/openconfig/public/blob/9eeeb2019ac61d501feae5127eb31bc4db15a048/doc/releases.md?plain=1#L59-L61) states v0.x models are not subject to have breaking changes declared. In practice, v0.x models are only promoted to v1.x when the model is expected to no longer have breaking changes made on a frequent basis.
> 
> To be very strict, this is a breaking change, albeit as Rob said, only for a more loose validation rule. If there was some implementation of the leafref node, a new client might attempt to use a string which does not match any component name which is now legal in the string node, yet the server using the old model expecting a leafref to a component name would deny it.

What we ran into here is the very common micro changes between releases which inside most development projects means you do not increment a version/revision.  They are aggregated at time of the next major release.

Otherwise much like internal development, you would end up w/ a potentially enormous changelog that could indicate alot of back-and-forth corrections, etc.. much like this one.  Not everything is caught in every code review but very often caught shortly after.

IMO, what I would propose is to never recommend master/main as stable and treat as experimental (I think we all realize that) - an item like this then I would treat as just another commit that fixed the previous from 24h earlier, no increment in version # but maybe rewrite the revision-date is all.

Aside from that - what is a true breaking change vs. not can quickly become subjective depending on the POV.  If we follow IETF guidelines for YANG backwards compatibility rules - yes this is backwards compatible.  If we look at it from how values are encoded on the wire, it might be the same but codegen, validation, etc... can all change.  If we were between releases then this would have to be a backwards incompatible change and signaled/recorded as such but in this case this is just top of tree churn where the rules should probably be relaxed but not forgotten (as buffering them up for release will also pose issues and misses imo)

robshakir:
I think the issue with the version changes is that we cut a module version change/revision change for every micro change of the file -- hence we rather do get this massive changelog in the modules. I'm OK with saying that for these kind of "oh ****, we missed that this got broken" that we don't declare a backwards incompatibility based on the amount of time that the breakage was available for.

We should - however - have something in the CI that caught this. @dplore -- can you look at what happened here? Did we get a report that said that there was a backwards incompatibility issue and miss this, or was there a CI gap?

earies:
Another potential solution (debatable w/ various pros/cons) would be to marshal all merges into a development branch that comes together as a single merge to master/main upon each release cut

master/main is always representing the last official version - as of today, alignment with `4.1.0`

A development branch is then cut from 4.1.0 tagged something like "next" - all active development requests PRs against "next" - allows multiple iterations and consolidation then at release time, determination is made for all changes to be cut into the next release (whatever the version determined may be - 4.2.0, 5.0.0, etc...)

CI logic and validation thus changes for this workflow and could potentially be different for what goes -> next vs. -> main/master

PR #1120, opened 2024-05-31
Add gratuitous-arp-accepted and unsolicited-na-accepted

cagdasalagoz-arista:
### Change Scope

Two new boolean leaf values will be added to paths below,

- `interfaces/interface/subinterfaces/subinterface/ipv4/config/`
- `interfaces/interface/subinterfaces/subinterface/ipv6/config/`.

**A. Yang tree for IPv4**
```
/oc-if:interfaces/oc-if:interface/oc-if:subinterfaces/oc-if:subinterface:
    +--rw ipv4
       +--rw config
       |  +--rw enabled?                     boolean
       |  +--rw mtu?                         uint16
       |  +--rw dhcp-client?                 boolean
       |  +--rw gratuitous-arp-accepted?     boolean       <<<new
       +--ro state
          +--ro enabled?                     boolean
          +--ro mtu?                         uint16
          +--ro dhcp-client?                 boolean
          +--ro gratuitous-arp-accepted?     boolean       <<<new
          +--ro counters
```

**B. Yang tree for IPv6**
```
/oc-if:interfaces/oc-if:interface/oc-if:subinterfaces/oc-if:subinterface:
    +--rw ipv6
       +--rw config
       |  +--rw enabled?                     boolean
       |  +--rw mtu?                         uint32
       |  +--rw dup-addr-detect-transmits?   uint32
       |  +--rw dhcp-client?                 boolean
       |  +--rw unsolicited-na-accepted?     boolean    <<<new
       +--ro state
          +--ro enabled?                     boolean
          +--ro mtu?                         uint32
          +--ro dup-addr-detect-transmits?   uint32
          +--ro dhcp-client?                 boolean
          +--ro unsolicited-na-accepted?     boolean    <<<new
          +--ro counters
```
As the name suggests these will be used to toggle if the device will accept (and process) or reject the gratuitous arp updates.

### Platform Implementations

**A. Implementation for IPv4**

**Arista**

These commands configure interface ethernet 2/1 to accept gratuitous ARP request packets.
```
switch(config)# interface ethernet 2/1
switch(config-if-Et2/1)# arp gratuitous accept
switch(config-if-Et2/1)#
```

**B. Implementation for IPv6**

**Arista**
```
switch(config-if-Et1/1)#ipv6 nd na unsolicited accept
   
switch#show ipv6 interface et1/1
. . .
  ND unsolicited NAs are accepted
```

### DIFF
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -44,7 +44,7 @@ module openconfig-if-ip {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "3.5.2";
+  oc-ext:openconfig-version "3.6.0";
 
   revision "2024-05-28" {
     description
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -620,14 +620,14 @@ revision "2023-06-30" {
         "RFC 4862: IPv6 Stateless Address Autoconfiguration";
     }
 
-    leaf unsolicited-na-accepted {
+    leaf learn-unsolicited {
       type boolean;
-      default true;
       description
-        "When set to true unsolicited neighbor advertisements
-        will be accepted and the ARP table will be updated.";
+        "When set to true neighbors will be learned from unsolicited
+        neighbor advertisements.";
       reference
-        "RFC 4861: Neighbor Discovery for IP version 6 (IPv6)";
+        "RFC 9131: Routers Creating Cache Entries upon
+        Receiving Unsolicited Neighbor Advertisements";
     }
 
     uses ip-common-global-config;
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -488,10 +488,8 @@ revision "2023-06-30" {
 
    leaf gratuitous-arp-accepted {
      type boolean;
-     default false;
      description
-        "When set to true, gratuitous neighbor advertisements
-        will be accepted.";
+        "When set to true, gratuitous ARPs will be accepted.";
      reference
       "RFC 826: An Ethernet Address Resolution Protocol";
     }
@@ -623,13 +621,12 @@ revision "2023-06-30" {
 
     leaf unsolicited-na-accepted {
       type boolean;
-      default false;
+      default true;
       description
         "When set to true unsolicited neighbor advertisements
         will be accepted.";
       reference
-        "RFC 9131: Gratuitous Neighbor Discovery: Creating
-        Neighbor Cache Entries on First-Hop Routers";
+        "RFC 4861: Neighbor Discovery for IP version 6 (IPv6)";
     }
 
     uses ip-common-global-config;
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -637,8 +637,8 @@ revision "2023-06-30" {
       }
       default "NONE";
       description
-        "When set to true neighbors will be learned from unsolicited
-        neighbor advertisements.";
+        "Sets if neighbors should be learned from unsolicited neighbor
+        advertisements for global or link local addresses or both.";
       reference
         "RFC 9131: Routers Creating Cache Entries upon
         Receiving Unsolicited Neighbor Advertisements";
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -49,7 +49,7 @@ module openconfig-if-ip {
   revision "2024-05-28" {
     description
       "Add gratuitous-arp-accepted for IPv4 and unsolicited-na-accepted for IPv6";
-    reference "3.5.2";
+    reference "3.6.0";
   }
 
   revision "2024-03-13" {
@@ -489,7 +489,8 @@ revision "2023-06-30" {
    leaf gratuitous-arp-accepted {
      type boolean;
      description
-        "When set to true, gratuitous ARPs will be accepted.";
+        "When set to true, gratuitous ARPs will be accepted and
+        the ARP table will be updated";
      reference
       "RFC 826: An Ethernet Address Resolution Protocol";
     }
@@ -624,7 +625,7 @@ revision "2023-06-30" {
       default true;
       description
         "When set to true unsolicited neighbor advertisements
-        will be accepted.";
+        will be accepted and the ARP table will be updated";
       reference
         "RFC 4861: Neighbor Discovery for IP version 6 (IPv6)";
     }
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -48,7 +48,7 @@ module openconfig-if-ip {
 
   revision "2024-05-28" {
     description
-      "Add gratuitous-arp-accepted for IPv4 and unsolicited-na-accepted for IPv6";
+      "Add gratuitous-arp-accepted for IPv4 and unsolicited-na-accepted for IPv6.";
     reference "3.6.0";
   }
 
@@ -490,7 +490,7 @@ revision "2023-06-30" {
      type boolean;
      description
         "When set to true, gratuitous ARPs will be accepted and
-        the ARP table will be updated";
+        the ARP table will be updated.";
      reference
       "RFC 826: An Ethernet Address Resolution Protocol";
     }
@@ -625,7 +625,7 @@ revision "2023-06-30" {
       default true;
       description
         "When set to true unsolicited neighbor advertisements
-        will be accepted and the ARP table will be updated";
+        will be accepted and the ARP table will be updated.";
       reference
         "RFC 4861: Neighbor Discovery for IP version 6 (IPv6)";
     }
--- a/release/models/interfaces/openconfig-if-ip.yang
+++ b/release/models/interfaces/openconfig-if-ip.yang
@@ -44,7 +44,13 @@ module openconfig-if-ip {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "3.5.1";
+  oc-ext:openconfig-version "3.5.2";
+
+  revision "2024-05-28" {
+    description
+      "Add gratuitous-arp-accepted for IPv4 and unsolicited-na-accepted for IPv6";
+    reference "3.5.2";
+  }
 
   revision "2024-03-13" {
     description
@@ -480,6 +486,16 @@ revision "2023-06-30" {
        "RFC 791: Internet Protocol";
     }
 
+   leaf gratuitous-arp-accepted {
+     type boolean;
+     default false;
+     description
+        "When set to true, gratuitous neighbor advertisements
+        will be accepted.";
+     reference
+      "RFC 826: An Ethernet Address Resolution Protocol";
+    }
+
     uses ip-common-global-config;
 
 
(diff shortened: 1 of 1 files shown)


dplore:
Additional references 

[Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/ip-addresses/b-ip-addresses-cr-8k/arp-commands.html#wp8703948390) - `no arp gratuitous ignore`

[JunOS](https://www.juniper.net/documentation/us/en/software/junos/multicast-l2/topics/task/interfaces-configuring-gratuitous-arp.html)  - set [gratuitous-arp-reply](https://www.juniper.net/documentation/en_US/junos/topics/reference/configuration-statement/gratuitous-arp-reply-edit-interfaces.html)

[Nokia SR Linux](https://documentation.nokia.com/cgi-bin/dbaccessfilename.cgi/3HE17570AAAA01_V1_SR%20Linux%20R21.6%20%20Advanced%20Solutions%20Guide.pdf) - `learn-unsolicited true`

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -480,6 +486,16 @@ revision "2023-06-30" {
        "RFC 791: Internet Protocol";
     }
 
+   leaf gratuitous-arp-accepted {
+     type boolean;
+     default false;
```
LimeHat, line 491: I would suggest removing the default, unless you have a strong argument for the `false` value.

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -480,6 +486,16 @@ revision "2023-06-30" {
...
+     description
+        "When set to true, gratuitous neighbor advertisements
```
LimeHat, line 493: gratuitous ARPs (NA/ND terminology is specific to v6)

LimeHat (changes requested):

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -605,6 +621,17 @@ revision "2023-06-30" {
         "RFC 4862: IPv6 Stateless Address Autoconfiguration";
     }
 
+    leaf unsolicited-na-accepted {
+      type boolean;
```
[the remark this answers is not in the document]
cagdasalagoz-arista: Thank you for your feedback Sergey,
I've updated the default value and the referance.

release/models/interfaces/openconfig-if-ip.yang
cagdasalagoz-arista: Done

release/models/interfaces/openconfig-if-ip.yang
cagdasalagoz-arista: Done

dplore (commented):
This looks good to me.  Small comment on updating the versioning and I think this is ready to go.

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -480,6 +486,16 @@ revision "2023-06-30" {
        "RFC 791: Internet Protocol";
     }
 
+   leaf gratuitous-arp-accepted {
+     type boolean;
+     default false;
+     description
+        "When set to true, gratuitous neighbor advertisements
+        will be accepted.";
```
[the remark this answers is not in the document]
cagdasalagoz-arista: fixed

dplore:
Last call for comments.  This will merge on Oct 8, 2024

LimeHat:
Can we please extend this last call? The latest changes escaped my attention, and I have a few concerns about the ipv6 part of this PR. Also, the arista doc reference is still behind the paywall, which makes it harder to follow for the broader public.

- according to the arista documentation, the config toggle is used to enable rfc9131-based behavior for processing unsolicited NAs (unfortunately, author gave me a different answer in the conversation earlier); which makes the current leaf description inaccurate. (I'll follow up with a more detailed comment)
- also, in the current form arista seems to be the only ipv6 reference. Is there a second one?

I have no objection to merging the ipv4 leaf (if the author and the community decides to split this into 2 parts)

dplore:
I found this reference in RFC9131 which is relevant to the new leaf `/interfaces/interface/subinterfaces/subinterface/ipv6/config/unsolicited-na-accepted`

For IPv6 accept unsolicited Neighbor Advertisements (related to RFC9131 Section 4.2), see
https://datatracker.ietf.org/doc/html/rfc9131#name-routers-creating-cache-entr

Seems to make sense to me to add this as well to be up to date with RFC9131?

cagdasalagoz-arista:
Correct, the feature is about accepting and processing unsolicited-na requests.
I'll update the `unsolicited-na-accepted` leaf as :
```
    leaf unsolicited-na-accepted {
      type boolean;
      default true;
      description
        "When set to true unsolicited neighbor advertisements
        will be accepted and the ARP table will be updated.";
      reference
        "RFC 9131: Routers Creating Cache Entries upon
        Receiving Unsolicited Neighbor Advertisements";
    }
```
Is this correct? @LimeHat

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -605,6 +620,16 @@ revision "2023-06-30" {
         "RFC 4862: IPv6 Stateless Address Autoconfiguration";
     }
 
+    leaf unsolicited-na-accepted {
+      type boolean;
+      default true;
```
LimeHat, line 625: since this leaf is supposed to enable 9131 behavior only, please remove the default

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -605,6 +620,16 @@ revision "2023-06-30" {
...
+      description
+        "When set to true unsolicited neighbor advertisements
+        will be accepted and the ARP table will be updated.";
```
LimeHat, line 628: Please remove:
* any references to ARP (it doesn't exist in ipv6).
* any references to `accepted` (without specifying the conditions) - this is not what RFC9131 describes.

Unsolicited NA should be accepted with or without 9131 behavior, and will be processed in either case if the entry already exists in cache (this is why I initially asked which behavior you are trying to describe and based on your response asked for the default value).

RFC9131 changes the behavior only for the specific case when the entry doesn't exist yet, quoting the rfc:

> This document updates [[RFC4861](https://www.rfc-editor.org/rfc/rfc9131.html#RFC4861)] so that routers create a new Neighbor Cache entry upon receiving an unsolicited Neighbor Advertisement for an address that does not already have a Neighbor Cache entry. These changes do not modify the router behavior specified in [[RFC4861](https://www.rfc-editor.org/rfc/rfc9131.html#RFC4861)] for the scenario when the corresponding Neighbor Cache entry already exists.

The description should be modified accordingly

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -605,6 +620,16 @@ revision "2023-06-30" {
...
+      reference
+        "RFC 4861: Neighbor Discovery for IP version 6 (IPv6)";
```
LimeHat, line 630: update the reference

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -605,6 +620,16 @@ revision "2023-06-30" {
         "RFC 4862: IPv6 Stateless Address Autoconfiguration";
     }
 
+    leaf unsolicited-na-accepted {
+      type boolean;
```
LimeHat, line 624: I would question whether this should be boolean.

@cagdasalagoz-arista's implementation does it this way; 
in srlinux we have an implementation that offers greater flexibility and uses [a enum instead](https://github.com/nokia/srlinux-yang-models/blob/6dd7b5a91433966dc9e4b763389a1d090355771b/srlinux-yang-models/srl_nokia/models/interfaces/srl_nokia-interfaces-nbr.yang#L342-L360).

Obviously from our point of view the enum is a better option; but considering that there are only 2 implementations I'm not sure what is the procedure to decide / break the tie here. @dplore any thoughts?

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -605,6 +620,16 @@ revision "2023-06-30" {
         "RFC 4862: IPv6 Stateless Address Autoconfiguration";
     }
 
+    leaf unsolicited-na-accepted {
```
LimeHat, line 623: Minor, but I'd propose to consider other names (I left a note in another comment below, `accepted` is not a completely  accurate way to represent the meaning)

LimeHat (changes requested):

LimeHat:
Hi Darren @dplore ,
>Seems to make sense to me to add this as well to be up to date with RFC9131?

It does make sense, but the initial description of the leaf was not accurate and the first conversation in the comments didn't help to clear it up, unfortunately. :)

Another question is what type of the leaf this should be, boolean or enum (please see my comment above).

release/models/interfaces/openconfig-if-ip.yang
cagdasalagoz-arista: I've updated to `learn-unsolicited`

release/models/interfaces/openconfig-if-ip.yang
cagdasalagoz-arista: Referance is updated to 
```
        RFC 9131: Routers Creating Cache Entries upon
        Receiving Unsolicited Neighbor Advertisements
```

release/models/interfaces/openconfig-if-ip.yang
cagdasalagoz-arista: description is updated to:
```
        When set to true neighbors will be learned from unsolicited
        neighbor advertisements.
```

release/models/interfaces/openconfig-if-ip.yang
cagdasalagoz-arista: Removed the default value

release/models/interfaces/openconfig-if-ip.yang
dplore: I do like the enum in this case.  It does not seem overly complex.  Vendors may only support a subset of the options.

@cagdasalagoz-arista can you pelase update similar to the following?   I checked the arista doc and it looks like it also implements "none" as the default, so I think we can leave that in.

```
      leaf learn-unsolicited {
        type enumeration {
          enum none {
            value 0;
          }
          enum global {
            value 1;
          }
          enum link-local {
            value 2;
          }
          enum both {
            value 3;
          }
        }
        default "none";
        description
          "Sets if neighbors should be learned from unsolicited neighbor advertisements for global or link local addresses or both.";
      }

```

release/models/interfaces/openconfig-if-ip.yang
dplore: @cagdasalagoz-arista please the text to use "Neighbor Cache" instead of ARP, since this is IPv6.

Please also update the description so it is clear if you are intending
1. To enable/disable RFC4861 behavior 
or
2. To enable/disable RFC9131 behavior

The key difference between the two is whether or not the router already has a Neighbor Cache entry.  If there is already an entry, RFC4861 says it can be updated.  If there is not one, the cache won't be updated.

RFC9131 says do RFC4861 AND also update the cache is there is no entry.

As written, it is not clear in this yang node which behavior is being configured.

cagdasalagoz-arista: Hello @dplore ,
I think you accidentally commented on the older version of the commit. The latest version is like this:
```
    leaf learn-unsolicited {
      type enumeration {
        enum NONE {
          value 0;
        }
        enum GLOBAL {
          value 1;
        }
        enum LINK_LOCAL {
          value 2;
        }
        enum BOTH {
          value 3;
        }
      }
      default "NONE";
      description
        "Sets if neighbors should be learned from unsolicited neighbor
        advertisements for global or link local addresses or both.";
      reference
        "RFC 9131: Routers Creating Cache Entries upon
        Receiving Unsolicited Neighbor Advertisements";
    }
```
Reference contains `RCF9131`

dplore:
Additional implementations:

JunOS [ipv6 neighbor discovery guide](https://www.juniper.net/documentation/us/en/software/junos/neighbor-discovery/neighbor-discovery.pdf)

Cisco IOS XR - [ipv6 nd na glean]( https://www.cisco.com/c/en/us/td/docs/routers/asr903/software/guide/ip/16-12-1/b-ip6b-xe-16-12-asr900/m_ip6-neighb-disc-xe.html#GUID-DAD0F8A4-E598-4813-992A-DB70857E8FD2)

dplore:
@LimeHat any additional comments?

dplore:
Reset last call to Nov 15, 2024

release/models/interfaces/openconfig-if-ip.yang
dplore: Thanks!

dplore (approved):

PR #1122, opened 2024-06-03
Configuration to control ACL set/entry counter allocation

nokia1adam:

Change Scope

At present, the allocation of system resources needed to provide counters for packets matching ACL entries cannot be controlled through configuration. This PR aims to fill this gap by introducing the following new paths:
acl/acl-sets/acl-set/config/counter
acl/acl-sets/acl-set/state/counter
acl/acl-sets/acl-set/acl-entries/acl-entry/config/counter
acl/acl-sets/acl-set/acl-entries/acl-entry/state/counter

The permitted values are: NONE, INTERFACE_ONLY, AGGREGATE_ONLY and INTERFACE_AGGREGATE. These are the same values as supported by the existing ACL_COUNTER_CAPABILITY except now NONE has been added as well, in order to allow the operator to disable stats collection for specific ACL sets or ACL entries in order to conserve limited resources.

The description for /acl/state/counter-capability has been amended to recommend that no value should be returned in state if the support of counters is not uniformly configured for all ACL sets and all ACL entries.

Implementations

Many vendors support some form of configuration control for ACL counters.
Arista: https://www.arista.com/en/um-eos/eos-acls-and-route-maps
Nokia: https://documentation.nokia.com/srlinux/24-3/books/acl-policy-based-routing/access-control-lists.html

### DIFF
--- a/release/models/acl/openconfig-acl.yang
+++ b/release/models/acl/openconfig-acl.yang
@@ -34,7 +34,14 @@ module openconfig-acl {
     packets should be handled. Entries have a type that indicates
     the type of match criteria, e.g., MAC layer, IPv4, IPv6, etc.";
 
-  oc-ext:openconfig-version "1.3.3";
+  oc-ext:openconfig-version "1.3.4";
+
+  revision "2024-06-03" {
+    description
+      "Add configuration control for statistics collection per
+      ACL-set and per ACL entry.";
+    reference "1.3.4";
+  }
 
   revision "2023-02-06" {
     description
@@ -233,6 +240,12 @@ module openconfig-acl {
       and reported per ACL entry.";
   }
 
+  identity NONE {
+    base ACL_COUNTER_CAPABILITY;
+    description
+      "ACL counters are not incremented.";
+  }
+
   // grouping statements
 
   // input interface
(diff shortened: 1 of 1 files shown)


dplore:
OC does not have a way to enable/disable ACL counters so this sounds like a good gap to close.  

EOS support looks like `counters per-entry` which enables/disables per acl-set and are reported as AGGREGATE_ONLY if enabled.    Enable is at the acl-set level.

My interpretation of the SRLinux doc is it supports enable/disable of `statistics-per-entry` at an ACL set level, but can report statistics at a per  interface level.   Does this map to INTERFACE_ONLY?  Is that correct @nokia1adam

nokia1adam:
@dplore, the SRLinux implementation has two different controls: one to control whether the ACL is shared by multiple interfaces or not and another to control whether stats are enabled or disabled for all entries of the acl-set. So these two controls together will determine whether a particular acl-set's entries have AGGREGATE or INTERFACE capability.

release/models/acl/openconfig-acl.yang
```diff
@@ -527,6 +550,15 @@ module openconfig-acl {
         "Description, or comment, for the ACL set";
     }
 
+    leaf counter {
```
dplore, line 553: this leaf is not needed, as the grouping "access-list-entries-config" is used in both config and state containers.

dplore (commented):
Please also propose some description of what should happen if the counter leaf is not configured.  (I would guess the default value is implementation defined)

dplore:
@nokia1adam let me know if you'd like to continue this PR

dplore:
@nokia1adam let me know if you'd like to continue this PR

I will close as stale for now

PR #437, opened 2021-06-29
Add initial support for EVPN VXLAN Overlay Endpoint Oper Data

mikewiebe:
**Summary of changes:**

* Adds support for EVPN VXLAN Overlay Endpoint Oper Data
    * State of peers learned over the endpoint
    * State of VNI's learned over a peer
* Addresses whitespace issues in `openconfig-evpn.yang`.

PR #1125, opened 2024-06-08
Updates to network instance models for VLAN-VNI and VRF-VNI mappings

abamberger-arista:
### Change Scope

This is part 2 of splitting the original PR #1108 into 3 separate PRs. This one targets the changes to add the ability to configure local VLAN-to-VNI and VRF-to-VNI mappings.

The existing endpoint-vnis subtree for VXLAN endpoints is read-only, and seems intended for reporting state “learned on the local VXLAN Tunnel End Point from remote VTEPs in the default network instance” (see https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html).

To support the configuration of local VLAN-to-VNI and VRF-to-VNI mappings on a VTEP, this change proposes the addition of a “local-endpoint-vnis” subtree, which will be congruent with the existing “endpoint-vnis” subtree, but be read/write, and allow the configuration of local VLAN-to-VNI and VRF-to-VNI mappings.

It will specifically contain support for configuring the VNI, VNI type (L2 or L3), and either the bridge-domain (for L2 VNIs) or l3-vrf-name (for L3 VNIs).

Because this is a new subtree being added, this change is fully backwards compatible.

New tree state after proposed change (additions in <b>bold</b>):

<pre>
+--rw network-instances
      +--rw connection-points
         +--rw connection-point* [connection-point-id]
            +--rw endpoints
               +--rw endpoint* [endpoint-id]
                  +--rw vxlan<b>
                     +--rw local-endpoint-vnis
                        +--rw local-endpoint-vni* [vni]
                           +--rw vni       -> ../config/vni
                              +--rw config
                              |  +--rw vni?             oc-evpn-types:evi-id
                              |  +--rw vni-type?        enumeration
                              |  +--rw bridge-domain?   uint32
                              |  +--rw l3-vrf-name?     string
                              +--ro state
                                 +--ro vni?             oc-evpn-types:evi-id
                                 +--ro vni-type?        enumeration
                                 +--ro bridge-domain?   uint32
                                 +--ro l3-vrf-name?     string</b>
</pre>

New Yang Paths:

* network-instances/network-instance/connection-points/connection-point/endpoints/endpoint/vxlan/local-endpoint-vnis/…

#### Arista EOS

VLAN-to-VNI and VRF-to-VNI mapping:
https://www.arista.com/en/um-eos/eos-vxlan-configuration#xx1152323

<pre>
interface Vxlan1
   vxlan udp-port 4789
   vxlan vlan 200 vni 658120
   vxlan vlan 100 vni 100
   vxlan vrf test vni 12345
</pre>

### DIFF
--- a/release/models/network-instance/openconfig-evpn.yang
+++ b/release/models/network-instance/openconfig-evpn.yang
@@ -40,7 +40,14 @@ module openconfig-evpn {
     domains, this is not currently supported and requires an extension
     of the model.";
 
-  oc-ext:openconfig-version "0.8.0";
+  oc-ext:openconfig-version "0.9.0";
+
+  revision "2024-06-07" {
+   description
+      "Add new vni-list leaf list to EVI config to support VLAN-aware-bundle
+      MACVRFs.  Make existing vni leaf conditional on MACVRF type being VLAN-based";
+   reference   "0.9.0";
+  }
 
   revision "2024-04-03" {
    description
@@ -645,10 +652,25 @@ module openconfig-evpn {
       Using Ethernet VPN";
 
     leaf vni {
+      when "../../../config/service-type = 'oc-evpn-types:VLAN_BASED' or
+            ../../../config/service-type = 'oc-evpn-types:VLAN_BUNDLE'" {
+        description
+          "For VLAN-based and VLAN-bundle EVIs, use a single VNI";
+      }
       type oc-evpn-types:vni-id;
       description
         "Virtual Network Identifier (VNI) associated to the EVI. This VNI is used for
         ingress and egress in the VXLAN domain.";
+    }
+
+    leaf-list vni-list {
+      when "../../../config/service-type = 'oc-evpn-types:VLAN_AWARE'" {
+        description
+          "For VLAN-aware-bundle EVIs, use a list of VNIs";
+      }
+      type oc-evpn-types:vni-id;
+      description
+        "List of VNIs participating in a VLAN-aware-bundle EVI";
     }
 
     leaf overlay-endpoint-network-instance {
--- a/release/models/network-instance/openconfig-evpn.yang
+++ b/release/models/network-instance/openconfig-evpn.yang
@@ -40,7 +40,14 @@ module openconfig-evpn {
     domains, this is not currently supported and requires an extension
     of the model.";
 
-  oc-ext:openconfig-version "0.9.0";
+  oc-ext:openconfig-version "0.10.0";
+
+  revision "2024-06-07" {
+   description
+      "Add new local-endpoint-vnis config container to Vxlan connection point to allow
+      configuration of local VNI-to-VLAN and VNI-to-VRF maps";
+   reference   "0.10.0";
+  }
 
   revision "2024-06-07" {
    description
@@ -1015,11 +1022,45 @@ module openconfig-evpn {
           config false;
           description
             "Container for state parameters related to this L2VNI or L3VNI";
+          uses evpn-endpoint-vni-config;
           uses evpn-endpoint-vni-state;
         }
 
         uses ipv4-top;
         uses ipv6-top;
+      }
+    }
+
+    container local-endpoint-vnis {
+      description
+        "Top level container for local configuration related to Layer 2 virtual
+        network identifiers (L2VNIs) and Layer 3 virtual network identifiers
+        (L3VNIs) in the default network instance";
+
+      list local-endpoint-vni {
+        key "vni";
+        description "List of L2VNIs and L3VNIs configured on the local VTEP";
+
+        leaf vni {
+          type leafref {
+            path '../config/vni';
+          }
+          description "L2VNI or L3VNI Identifier";
+        }
+
+        container config {
+          description
+            "Container for configuration parameters related to this local L2VNI or
+            L3VNI";
+          uses evpn-endpoint-vni-config;
+        }
+
+        container state {
+          config false;
+          description
+            "Container for state parameters related to this local L2VNI or L3VNI";
+          uses evpn-endpoint-vni-config;
+        }
       }
     }
   }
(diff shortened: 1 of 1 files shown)


release/models/network-instance/openconfig-evpn.yang
```diff
@@ -1144,16 +1207,52 @@ module openconfig-evpn {
...
+        L2VNI in the VXLAN fabric";
+    }
+
+    leaf l3-vrf-name {
+      type string;
+      description
+        "This refects the configured VRF instance that maps to this L3VNI
+        that is used for routing between subnets in the VXLAN fabric";
+    }
+  }
```
dplore, line 1249: Following precedent in OC for referring to a network instance:

```suggestion
    leaf network-instance {
          type oc-ni:network-instance-ref;
      description
        "This refects the configured VRF instance that maps to this L3VNI
        that is used for routing between subnets in the VXLAN fabric";
    }
  }
```

release/models/network-instance/openconfig-evpn.yang
```diff
@@ -1144,16 +1207,52 @@ module openconfig-evpn {
...
+      }
+      description "The type of virtual network identfier";
+    }
+
+    leaf bridge-domain {
+      type uint32;
+      description
+        "This reflects the configured VLAN or Bridge Domain that maps to this
+        L2VNI in the VXLAN fabric";
+    }
```
dplore, line 1241: In OC, if the the network-instance is of type `L2VSI`, then that also indicates the bridge domain.  So I think this leaf is not needed.

E.g.
https://github.com/openconfig/public/blob/4d4aa4b7dd94acfae84e3f567b534116e6bd5ca4/release/models/network-instance/openconfig-network-instance-types.yang#L158-L162

Ref: https://github.com/openconfig/public/blob/master/doc/network_instance_examples.md

release/models/network-instance/openconfig-evpn.yang
```diff
@@ -1144,16 +1207,52 @@ module openconfig-evpn {
...
+            broadcast domain";
+        }
+        enum L3 {
+          description
+            "This is a Layer 3 service virtual network identifier (L3VNI)
+            or VRF VNI that is used for communication between subnets";
+        }
+      }
+      description "The type of virtual network identfier";
+    }
```
dplore, line 1234: Is this redundant to the network-instance type?

dplore (changes requested):

release/models/network-instance/openconfig-evpn.yang
abamberger-arista: The intent here was to match the existing fields in the `endpoint-vnis` state container, both to match the existing model, and also so the common parts could be factored out into a common grouping.  If we wanted to make this different, we'd either have to have a difference in structure/data type between `endpoint-vnis` and `local-endpoint-vnis`, or we'd have to change `endpoint-vnis` in a non-backwards compatible way.

release/models/network-instance/openconfig-evpn.yang
abamberger-arista: Similarly to `l3-vrf-name`, this is intended to be symmetric with the existing `endpoint-vnis` container, for similar reasons.

release/models/network-instance/openconfig-evpn.yang
abamberger-arista: This is already part of the existing public model, it's just been refactored to pull it into a common location that can be used both by the existing `endpoint-vnis`, and the proposed `local-endpoint-vnis`

release/models/network-instance/openconfig-evpn.yang
dplore: I see, ok.

release/models/network-instance/openconfig-evpn.yang
dplore: I see, ok.   I think vlan is likely a better name, as we don't use 'bridge' often in OC.  Wouldn't this leaf refer directly to an OC configured vlan. (ie: /interfaces/interface/subinterfaces/subinterface/vlan/config/vlan-id) ?

dplore:
@mikewiebe, @limehat,  @earies  can you give a review as well?

release/models/network-instance/openconfig-evpn.yang
abamberger-arista: In practice yes, I'm assuming the reason it's this way in the first place is to make the distinction between a VLAN as a bridging domain, and a VLAN as a specific dot1q tag (where multiple tags could refer to the same bridging domain, based on other config)

abamberger-arista:
Just as an overall comment, since it's relevant to a few of the more specific comments here, this PR doesn't add any new terminal leaves, it's mostly just factoring out some parts of the existing `endpoint-vnis` container so they can be re-used in a new `local-endpoint-vnis` container.  The contents that have been factored out (`vni`, `vni-type`, `bridge-domain`, and `l3-vrf-name`) are existing leaves already in the public model in the `endpoint-vnis` container.

release/models/network-instance/openconfig-evpn.yang
dplore: Ok, resolving this as it's expanding on the state leaf already merged in to OC (https://github.com/openconfig/public/pull/437) and the desire to not create a breaking change over the choice of name for a leaf.

release/models/network-instance/openconfig-evpn.yang
```diff
@@ -40,7 +40,21 @@ module openconfig-evpn {
...
+      configuration of local VNI-to-VLAN and VNI-to-VRF maps";
+   reference   "0.10.0";
+  }
+
+  revision "2024-06-07" {
+   description
+      "Add new vni-list leaf list to EVI config to support VLAN-aware-bundle
+      MACVRFs.  Make existing vni leaf conditional on MACVRF type being VLAN-based";
+   reference   "0.9.0";
+  }
```
dplore, line 57: Do want to merge these in the order for #1125, #1126, #1127?  If yes,  the revision statements can reflect that.

(ie: this PR #1125 should have just version 0.10.0.  Note that 0.9.0 conflicts with  recently merged version https://github.com/openconfig/public/blob/6e28490719bd8a3966ce9a78c82c3b7a5a249412/release/models/network-instance/openconfig-evpn.yang#L45-L50)

dplore (commented):
LGTM, but please fix the revision statement as appropriate.  I will move this to last call, to be merged on Aug 28, 2024)

release/models/network-instance/openconfig-evpn.yang
dplore: got it

release/models/network-instance/openconfig-evpn.yang
abamberger-arista: Should be fixed now

abamberger-arista:
I went ahead and fixed the merge conflict with the just-merged previous PR

mikewiebe:
@dplore Looks like I missed the notification on this one.  I see the change is backwards compatible since it's adding a new tree so overall I think the change is ok.

PR #1126, opened 2024-06-08
Updates to network-instance EVPN models for EVPN redist support

abamberger-arista:
### Change Scope

This is part 3 of splitting the original PR #1108 into 3 separate PRs. This one targets changes required to configure redistribution of routes into EVPN

MACVRFs and IPVRFs can be configured to redistribute different types of MACs and IP Prefixes to their peers. For instance, a MACVRF might be configured to redistribute routes for dynamic MACs learned in the associated local VLANs, but not static MACs configured in those same VLANs.

To support the ability to configure these distribution settings, this change proposes the addition of a “redistribute” leaf list to the EVI subtree, along with a new identity type for types of redistributions to support. This initial proposal has 3 types of redistributions defined, REDISTRIBUTE_LEARNED and REDISTRIBUTE_STATIC, for redistributing dynamically learned and statically configured MACs, respectively, into a MACVRF, and REDISTRIBUTE_CONNECTED, for redistributing connected IP routes into an IPVRF. Additional types of redistribution could be added in later changes.

New tree state after proposed change (additions in <b>bold</b>):

<pre>
+--rw network-instances
      +--rw evpn
         +--rw evpn-instances
            +--rw evpn-instance* [evi]
               +--rw evi                     -> ../config/evi
                  +--rw config
                  |  +--rw evi?                    string
                  |  +--rw encapsulation-type?     identityref
                  |  +--rw service-type?           identityref
                  |  +--rw multicast-group?        oc-inet:ipaddress
                  |  +--rw multicast-mask?         oc-inet:ipaddress
                  |  +--rw replication-mode?       enumeration
                  |  +--rw route-distinguisher?    union
                  |  +--rw control-word-enabled?   boolean
                  <b>|  +--rw redistribute*           oc-evpn-types:redistribute-type</b>
                  +--ro state
                  |  +--ro evi?                    string
                  |  +--ro encapsulation-type?     identityref
                  |  +--ro service-type?           identityref
                  |  +--ro multicast-group?        oc-inet:ip-address
                  |  +--ro multicast-mask?         oc-inet:ip-address
                  |  +--ro replication-mode?       enumeration
                  |  +--ro route-distinguisher?    union
                  |  +--ro control-word-enabled?   boolean
                  <b>|  +--ro redistribute*           oc-evpn-types:redistribute-type</b>
</pre>

New Yang Paths:

* network-instances/network-instance/evpn/evpn-instances/evpn-instance/config/redistribute
* network-instances/network-instance/evpn/evpn-instances/evpn-instance/status/redistribute

#### Arista EOS

Redistribution config:
https://www.arista.com/en/um-eos/eos-sample-configurations#xx1247650

<pre>
router bgp 65002
   vlan 10
      rd 1.1.1.11:1010
      route-target both 1010:1010
      redistribute learned
   !
   vlan 11
      rd 1.1.1.11:1011
      route-target both 1011:1011
      redistribute learned
      redistribute static
   !
   vrf red
   ...
      redistribute connected
</pre>

### DIFF
--- a/release/models/network-instance/openconfig-evpn.yang
+++ b/release/models/network-instance/openconfig-evpn.yang
@@ -40,7 +40,14 @@ module openconfig-evpn {
     domains, this is not currently supported and requires an extension
     of the model.";
 
-  oc-ext:openconfig-version "0.8.0";
+  oc-ext:openconfig-version "0.9.0";
+
+  revision "2024-06-07" {
+   description
+      "Add new vni-list leaf list to EVI config to support VLAN-aware-bundle
+      MACVRFs.  Make existing vni leaf conditional on MACVRF type being VLAN-based";
+   reference   "0.9.0";
+  }
 
   revision "2024-04-03" {
    description
@@ -645,10 +652,25 @@ module openconfig-evpn {
       Using Ethernet VPN";
 
     leaf vni {
+      when "../../../config/service-type = 'oc-evpn-types:VLAN_BASED' or
+            ../../../config/service-type = 'oc-evpn-types:VLAN_BUNDLE'" {
+        description
+          "For VLAN-based and VLAN-bundle EVIs, use a single VNI";
+      }
       type oc-evpn-types:vni-id;
       description
         "Virtual Network Identifier (VNI) associated to the EVI. This VNI is used for
         ingress and egress in the VXLAN domain.";
+    }
+
+    leaf-list vni-list {
+      when "../../../config/service-type = 'oc-evpn-types:VLAN_AWARE'" {
+        description
+          "For VLAN-aware-bundle EVIs, use a list of VNIs";
+      }
+      type oc-evpn-types:vni-id;
+      description
+        "List of VNIs participating in a VLAN-aware-bundle EVI";
     }
 
     leaf overlay-endpoint-network-instance {
--- a/release/models/network-instance/openconfig-evpn.yang
+++ b/release/models/network-instance/openconfig-evpn.yang
@@ -40,7 +40,14 @@ module openconfig-evpn {
     domains, this is not currently supported and requires an extension
     of the model.";
 
-  oc-ext:openconfig-version "0.10.0";
+  oc-ext:openconfig-version "0.11.0";
+
+  revision "2024-06-07" {
+   description
+     "Add redistribute type leaf list to EVI config to support configuration of
+      various types of route distribution (for both L2 and L3).";
+   reference   "0.11.0";
+  }
 
   revision "2024-06-07" {
    description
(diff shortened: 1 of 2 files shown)


release/models/network-instance/openconfig-evpn.yang
```diff
@@ -645,12 +674,27 @@ module openconfig-evpn {
...
 
+    leaf-list vni-list {
+      when "../../../config/service-type = 'oc-evpn-types:VLAN_AWARE'" {
+        description
+          "For VLAN-aware-bundle EVIs, use a list of VNIs";
+      }
+      type oc-evpn-types:vni-id;
+      description
+        "List of VNIs participating in a VLAN-aware-bundle EVI";
+    }
```
dplore, line 696: Add reference  "RFC 7432: BGP MPLS-Based Ethernet VPN section-6.3";

release/models/network-instance/openconfig-evpn.yang
```diff
@@ -645,12 +674,27 @@ module openconfig-evpn {
       Using Ethernet VPN";
 
     leaf vni {
+      when "../../../config/service-type = 'oc-evpn-types:VLAN_BASED' or
+            ../../../config/service-type = 'oc-evpn-types:VLAN_BUNDLE'" {
+        description
+          "For VLAN-based and VLAN-bundle EVIs, use a single VNI";
+      }
```
dplore, line 681: Add reference "RFC 7432: BGP MPLS-Based Ethernet VPN section-6";

dplore:
Hi @abamberger-arista,  the state tree doesn't match this PR?  Where is `vni-list`?

release/models/network-instance/openconfig-evpn.yang
abamberger-arista: This was part of one of the previous PRs that has already merged

release/models/network-instance/openconfig-evpn.yang
abamberger-arista: Likewise, one of the previous PRs that has already merged

abamberger-arista:
> Hi @abamberger-arista, the state tree doesn't match this PR? Where is `vni-list`?

That was part of the previous PR (#1125)

LimeHat:
I have the same comment as in the original PR: I don't believe the redistribution is a fundamentally correct approach to control mac/type-2 route distribution. It is an implementation of a specific vendor's CLI, but it doesn't mean it should be adopted in OC as it is.

IMO, the baseline EVPN implies that locally learned mac addresses _are_ advertised to neighbors (that's the core rfc7432/rfc8365 behavior). And to my knowledge, this is what most implementations do. A typical L2 EVPN instance will not be properly functioning without the type-2 routes.

While it is true that there are certain use cases where an operator may wish to limit type-2 advertisements, that should not be the default behavior, and the use of redistribution to control that is highly questionable.  
Some implementations (I'm aware of 2 vendors) offer a single bool knob to disable mac-advertisements/type-2 routes for the limited number of use cases where it may be needed.

Is there a second reference that has static/dynamic granularity?

Last but not least, can you elaborate on this part?

```
REDISTRIBUTE_CONNECTED, for redistributing connected IP routes into an IPVRF. Additional types of redistribution could be added in later changes.
```

If we are talking about the regular redistribution, then it is already defined by OpenConfig standards. If this is something specific to EVPN, please provide more details as to what, exactly, this feature is supposed to do.

PR #1131, opened 2024-06-14
Add BGP cumulative link-bandwidth

dplore:
### Change Scope

* Extend configuration of BGP global and peer-groups to permit cumulative link bandwidth and transitive behavior as per [draft-ietf-bess-ebgp-dmz](https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/).   
* Also add link-bandwidth parameters to BGP neighbor configuration and state.  (Previously these were only defined at the global and peer-group levels)

* This change only adds leafs and is backwards compatible.

* Example flattened paths:
```
Prefix: /network-instances/network-instance/protocols/protocol/bgp/

global/afi-safis/afi-safi/use-multiple-paths/ebgp/link-bandwidth-ext-community/config/cumulative=true
global/afi-safis/afi-safi/use-multiple-paths/ebgp/link-bandwidth-ext-community/config/non-transitive=true
global/afi-safis/afi-safi/use-multiple-paths/ebgp/link-bandwidth-ext-community/config/equal=false
```

### Platform Implementations

On [Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/bgp/24xx/configuration/guide/b-bgp-cg-8k-24xx/implementing-bgp.html#concept_j3g_jvv_njb), there is no special configuration needed other than `link-bandwidth-ext-community/config/enabled ` .  The OS implicitly sums up the link-bandwidth of ECMP bgp paths.

[Arista EOS](https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp#xx1418621) expects an 'aggregate' configuration option on top of enabling link-bandwidth to be added on a per neighbor basis.  They further support 'divide' and 'equal' options.

[JunOS](https://community.juniper.net/blogs/moshiko-nayman/2024/05/13/bgp-link-bandwidth-with-junos) expects configuration of a BGP policy statement to enable 'aggregate-bandwidth' and additional options for transitive/non-transitive and a 'divide-equal' option.    There is also a configuration item, set per bgp neighbor or peer to allow link-bandwidth community to be sent via eBGP (ie: allow sending  BGP link-bandwidth which was defined as non-transitive in [draft-ietf-idr-link-bandwidth](https://datatracker.ietf.org/doc/draft-ietf-idr-link-bandwidth/), but then updated by [draft-ietf-bess-ebgp-dmz](https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/) to be transitive.

### Tree view
This common stanza of changes is added at each level of BGP configuration:
bgp/global
bgp/peer-group
bgp/neighbor
bgp/global/afi-safis
bgp/peer-group/afi-safis
bgp/neighbor/afi-safis

```diff
diff -u ~/old-tree.txt ~/new-tree.txt
--- /Users/dloher/old-tree.txt  2024-06-13 17:48:44
+++ /Users/dloher/new-tree.txt  2024-06-13 17:48:23
@@ -3593,9 +3593,15 @@
         |     |  |  |  +--rw ebgp
         |     |  |  |  |  +--rw link-bandwidth-ext-community
         |     |  |  |  |  |  +--rw config
-        |     |  |  |  |  |  |  +--rw enabled?   boolean
+        |     |  |  |  |  |  |  +--rw enabled?          boolean
+        |     |  |  |  |  |  |  +--rw cumulative?       boolean
+        |     |  |  |  |  |  |  +--rw non-transitive?   boolean
+        |     |  |  |  |  |  |  +--rw equal?            boolean
         |     |  |  |  |  |  +--ro state
-        |     |  |  |  |  |     +--ro enabled?   boolean
+        |     |  |  |  |  |     +--ro enabled?          boolean
+        |     |  |  |  |  |     +--ro cumulative?       boolean
+        |     |  |  |  |  |     +--ro non-transitive?   boolean
+        |     |  |  |  |  |     +--ro equal?            boolean
         |     |  |  |  |  +--rw config
         |     |  |  |  |  |  +--rw allow-multiple-as?   boolean
         |     |  |  |  |  |  +--rw maximum-paths?       uint32
@@ -3605,9 +3611,15 @@
         |     |  |  |  +--rw ibgp
         |     |  |  |     +--rw link-bandwidth-ext-community
         |     |  |  |     |  +--rw config
-        |     |  |  |     |  |  +--rw enabled?   boolean
+        |     |  |  |     |  |  +--rw enabled?          boolean
+        |     |  |  |     |  |  +--rw cumulative?       boolean
+        |     |  |  |     |  |  +--rw non-transitive?   boolean
+        |     |  |  |     |  |  +--rw equal?            boolean
         |     |  |  |     |  +--ro state
-        |     |  |  |     |     +--ro enabled?   boolean
+        |     |  |  |     |     +--ro enabled?          boolean
+        |     |  |  |     |     +--ro cumulative?       boolean
+        |     |  |  |     |     +--ro non-transitive?   boolean
+        |     |  |  |     |     +--ro equal?            boolean
         |     |  |  |     +--rw config
         |     |  |  |     |  +--rw maximum-paths?   uint32
         |     |  |  |     +--ro state
```

### DIFF
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -21,12 +21,18 @@ submodule openconfig-bgp-common-structure {
     "This sub-module contains groupings that are common across multiple BGP
     contexts and provide structure around other primitive groupings.";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-06-13" {
+    description
+      "Clarification to last-prefix-limit-exceeded in description";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
+    reference "9.7.1";
   }
 
   revision "2023-12-28" {
--- a/release/models/bgp/openconfig-bgp-common.yang
+++ b/release/models/bgp/openconfig-bgp-common.yang
@@ -24,19 +24,25 @@ submodule openconfig-bgp-common {
     may be application to a subset of global, peer-group or neighbor
     contexts.";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-06-13" {
+    description
+      "Clarification to last-prefix-limit-exceeded in description";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
+    reference "9.7.1";
   }
 
   revision "2023-12-28" {
     description
       "Add support for controling use of link-bandwidth extended
       community for BGP multipath.";
-          reference "9.7.0";
+    reference "9.7.0";
   }
 
   revision "2023-11-02" {
@@ -698,17 +704,52 @@ submodule openconfig-bgp-common {
     description
       "Parameters controlling usage of  of DMZ Link-Bandwidth
       extended community in pultipath RIB/FIB formation";
+
     leaf enabled {
       type boolean;
       description
-      "When set to TRUE, BGP multiplepath shall distributed traffic
-      load among contributing routes proportionally to value of
-      Local Administrator subfield of link-bandwidth extended
-      community [draft-ietf-idr-link-bandwidth-07].
-      This leaf has no effect if BGP multi-path is disabled or
-      if maximum-path attribute of BGP multi-path value is set
-      to 1";
+        "When set to TRUE, BGP multiplepath shall distributed traffic
+        load among contributing routes proportionally to value of
+        Local Administrator subfield of link-bandwidth extended
+        community [draft-ietf-idr-link-bandwidth-07].
+        This leaf has no effect if BGP multi-path is disabled or
+        if maximum-path attribute of BGP multi-path value is set
+        to 1";
     }
+
+    leaf cumulative {
+      type boolean;
+      when "../enabled = 'TRUE'";
+      description
+        "When set to TRUE, and sending of link-bandwidth community values
+        is enabled, the system will sum the bgp link-bandwidth community
+        values for equal cost paths as defined in draft-ietf-bess-ebgp-dmz.";
+      reference
+        "https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/";
+    }
+
+    leaf non-transitive {
+      type boolean;
+      when "../enabled = 'TRUE'";
+      description
+        "When set to TRUE and sending of link-bandwidth community values is
+        enabled, the system will send the link-bandwidth community to eBGP
+        peers.";
+      reference
+        "https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/";
+    }
+
+    leaf equal {
+      type boolean;
+      when "../enabled = 'TRUE'";
+      description
+        "When set to TRUE, and sending of link-bandwidth community values
+        is enabled, the system divide the summed bandwidth equally across
+        available paths as defined in draft-ietf-bess-ebgp-dmz.";
+      reference
+        "https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/";
+    }
+
   }
 
   grouping bgp-common-use-multiple-paths-ebgp-config {
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -24,12 +24,18 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-06-13" {
+    description
+      "Clarification to last-prefix-limit-exceeded in description";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
+    reference "9.7.1";
   }
 
   revision "2023-12-28" {
@@ -38,6 +44,7 @@ submodule openconfig-bgp-common-multiprotocol {
       community for BGP multipath.";
     reference "9.7.0";
   }
+
   revision "2023-11-02" {
     description
       "Fix revision '2023-03-31': send-community-type was added to the
--- a/release/models/bgp/openconfig-bgp-global.yang
+++ b/release/models/bgp/openconfig-bgp-global.yang
@@ -27,12 +27,18 @@ submodule openconfig-bgp-global {
     "This sub-module contains groupings that are specific to the
     global context of the OpenConfig BGP module";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-06-13" {
+    description
+      "Clarification to last-prefix-limit-exceeded in description";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
+    reference "9.7.1";
   }
 
   revision "2023-12-28" {
--- a/release/models/bgp/openconfig-bgp-peer-group.yang
+++ b/release/models/bgp/openconfig-bgp-peer-group.yang
@@ -25,12 +25,18 @@ submodule openconfig-bgp-peer-group {
     "This sub-module contains groupings that are specific to the
     peer-group context of the OpenConfig BGP module.";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-06-13" {
+    description
+      "Clarification to last-prefix-limit-exceeded in description";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
+    reference "9.7.1";
   }
 
   revision "2023-12-28" {
--- a/release/models/bgp/openconfig-bgp.yang
+++ b/release/models/bgp/openconfig-bgp.yang
@@ -68,19 +68,25 @@ module openconfig-bgp {
     whereas leaf not present inherits its value from the leaf present
     at the next higher level in the hierarchy.";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-06-13" {
+    description
+      "Clarification to last-prefix-limit-exceeded in description";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
+    reference "9.7.1";
   }
 
   revision "2023-12-28" {
     description
       "Add support for controling use of link-bandwidth extended
       community for BGP multipath.";
-          reference "9.7.0";
+    reference "9.7.0";
   }
 
   revision "2023-11-02" {


rszarecki:
The [draft-ietf-bess-ebgp-dmz](https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/) does 2 things:
1. allow for propagation link-bandwidth ext-community attribute form BGP Local-RIB to eBGP session despite link-bandwidth ext-community is non-transitiv. This is orthogonal to BGP multipathing configuration on given system.
2. it also alow to apply different aggregation algorithms to link-bandwidth in case when multiple path with  link-bandwidth ext-community exist in Local-RIB (that is BGP multipath is enabled)
Note: The original I-D [draft-ietf-idr-link-bandwidth-07 page 3, last sentence](https://datatracker.ietf.org/doc/html/draft-ietf-idr-link-bandwidth-07#section-3) explicitly allow initialization and transmission of link-bandwidth community on eBGP session (similiarly to other non-transitive BGP attribute like MED). It just do not allow **propagation** of link-bandwidth community for Local-RIB to eBGP session. 


The `global/afi-safis/afi-safi/use-multiple-paths/...` hierarchy is wrong place to control if and how link-bandwidth is propagated send over eBGP because:
- draft-bess do not requires multipathing to be enabled (see 1 above)
- `global/afi-safis/afi-safi/use-multiple-paths/...` is meant to provide constrains to proces of polulation fo Local-RIB and FIB's NHG. The `use-multiple-paths/ebgp/link-bandwidth-ext-community/config/enabled` control if FIB ECMP weights should be derived form link-bandwidth values or not. In later case basic ECMP will be programmed in dataplane cons=umig less TCAM/SRAM resources, but link-bandwidth will remain in Local-RIB and will be propagated to iBGP peers (and eBGP under draft-bess if enabled).
- provided that `.../afi-safis/afi-safi/use-multiple-paths/config/enabled` is set to "FALSE", and new leaf `non-transitive` is set to "TRUE"; what shall be system behaviour?  
- IMHO the knobs controlling [draft-ietf-bess-ebgp-dmz](https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/) shall be direct attributes of {neighbor|peer-group|global}/afi-safis/afi-safi attribute, similar to `send-community-type` (which is unfortunetly enum). My suggestion is:
  - define 2 new container dedicated for link-badwidth community (it is special case anyway) under `bgp/global/afi-safis/afi-safi/config/' - `tx-link-bandwidth` with leafs: `enabled`, `ebgp`default FALSE, `cummulative`, `average/equal` AND  `rx-link-bandwidth` with leafs: `enabled` and default TRUE. if `tx-link-bandwidth/enabled` not specified: TRUE on iBGP and FALSE on eBGP.

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -698,17 +704,52 @@ submodule openconfig-bgp-common {
     description
       "Parameters controlling usage of  of DMZ Link-Bandwidth
```
jhaas-pfrc, line 705: Perhaps cleanup of of in this PR?

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -698,17 +704,52 @@ submodule openconfig-bgp-common {
...
+        "When set to TRUE, BGP multiplepath shall distributed traffic
+        load among contributing routes proportionally to value of
+        Local Administrator subfield of link-bandwidth extended
+        community [draft-ietf-idr-link-bandwidth-07].
+        This leaf has no effect if BGP multi-path is disabled or
+        if maximum-path attribute of BGP multi-path value is set
+        to 1";
+    }
+
+    leaf cumulative {
```
jhaas-pfrc, line 720: I would recommend "send-cumulative" to represent the role in how this is transmitted.
Note that in junos, inconsistent configuration of this feature at the neighbor level can result in peergroup splits since it's "hidden policy".

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -698,17 +704,52 @@ submodule openconfig-bgp-common {
...
+      type boolean;
+      when "../enabled = 'TRUE'";
```
jhaas-pfrc, line 722: The "enabled" above is "do load balancing locally".  If the device is not locally performing ECMP but is a pinch-point for routes and thus sending the aggregated link-bw, advertising the aggregate may still be appropriate.  Perhaps consider removing this constraint.

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -698,17 +704,52 @@ submodule openconfig-bgp-common {
...
+      description
+        "When set to TRUE, and sending of link-bandwidth community values
+        is enabled, the system will sum the bgp link-bandwidth community
+        values for equal cost paths as defined in draft-ietf-bess-ebgp-dmz.";
+      reference
+        "https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/";
```
jhaas-pfrc, line 728: A general comment about references vs. IETF I-Ds: Before an RFC, drafts may change substantively from one version to another.  Consider always adding draft version to the reference.

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -698,17 +704,52 @@ submodule openconfig-bgp-common {
...
+    }
+
+    leaf non-transitive {
```
jhaas-pfrc, line 731: I'll add a general comment to the PR, but expect an update to the base link-bw I-D where both transitive and non-transitive forms are dealt with.  This flag will be inconsistent with upcoming procedure.

The desired behavior is "send non-transitive link-bw extended communities over an ebgp session", and the name needs more context for that.  

Perhaps "send-non-transitive-via-ebgp" ?

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -698,17 +704,52 @@ submodule openconfig-bgp-common {
...
+      when "../enabled = 'TRUE'";
+      description
+        "When set to TRUE and sending of link-bandwidth community values is
+        enabled, the system will send the link-bandwidth community to eBGP
+        peers.";
+      reference
+        "https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/";
+    }
+
+    leaf equal {
```
jhaas-pfrc, line 742: I'd suggest including the word "divide" in the name in some form.

jhaas-pfrc:
> [JunOS](https://community.juniper.net/blogs/moshiko-nayman/2024/05/13/bgp-link-bandwidth-with-junos) expects configuration of a BGP policy statement to enable 'aggregate-bandwidth' and additional options for transitive/non-transitive and a 'divide-equal' option. There is also a configuration item, set per bgp neighbor or peer to allow link-bandwidth community to be sent via eBGP (ie: allow sending BGP link-bandwidth which was defined as non-transitive in [draft-ietf-idr-link-bandwidth](https://datatracker.ietf.org/doc/draft-ietf-idr-link-bandwidth/), but then updated by [draft-ietf-bess-ebgp-dmz](https://datatracker.ietf.org/doc/draft-ietf-bess-ebgp-dmz/) to be transitive.

The details here are slightly off.  Juniper's implementation of link-bw uses a transitive extended community code point.  (Embarrassing, because Juniper is the one that published the base spec that uses non-transitive in the document.)  This causes interop issues between Juniper and implementations that don't understand the non-draft-compliant transitive format.

The link-bw draft is pending an upcoming update in IETF that will address both the transitive and non-transitive cases.  It is intended to contain interop procedures.  Indirectly, it also addresses some points covered in the DMZ draft.

release/models/bgp/openconfig-bgp-common.yang
rszarecki: average? it is average is not?

release/models/bgp/openconfig-bgp-common.yang
rszarecki: As I commented, IMHO this is anchored in wrong place. The "non-transitive-send-over-eBGP" is orthogonal to multipath.

release/models/bgp/openconfig-bgp-common.yang
jhaas-pfrc: "average" is effectively what it is.  That said, there's some desire to have symmetry in the cited draft.  I'd suggest either picking that as a name and working with the authors to incorporate it in their description, or stick to something in the draft like divide.

Note that Juniper has chosen divide-equal in its policy language.

release/models/bgp/openconfig-bgp-common.yang
jhaas-pfrc: Very broadly speaking, as I've discussed internally with Juniper developers, there's some ugly ambiguity in RFC 4360 with regard to non-transitive extended communities.  The ambiguity comes from unclear semantics for propagation vs. "origination".

- If the community is unmolested locally, is learned from another iBGP speaker, then it should be dropped at the AS boundary.
- If the community is attached as part of export procedures to an eBGP speaker, the community is effectively "originated" at that boundary.  A valid semantic would be that AS1 pushes the community into AS2 with the expectation that it doesn't leave AS2.

This second sense isn't in the RFC, but overlaps the point here for the DMZ draft.  

In those senses, I agree that this is a generic problem.  There's some motivation to do a RFC 4360-bis to clarify the point, but that work hasn't happened.

Meanwhile, implementations have inconsistent knobs for dealing with this scenario.  Right now, the only compelling use case (non-transitive ext-comms are rare) is this DMZ case.  Thus, having a knob here is "fine".  Dealing with the broader discussion is likely outside the OC use case until implemented by enough vendors.

release/models/bgp/openconfig-bgp-common.yang
dplore: Done

release/models/bgp/openconfig-bgp-common.yang
dplore: Added `non-transitive` boolean leaf to explicitly indicate if link-bandwidth communities from iBGP peers will be sent to eBGP peers.  WDYT?

release/models/bgp/openconfig-bgp-common.yang
dplore: Thank you!  Added revision `05`

rszarecki:
I will reiterate it again:

The 'global/afi-safis/afi-safi/use-multiple-paths/...' hierarchy is **wrong place** to control if and how link-bandwidth is propagated send over eBGP because:

- draft-bess do not requires multipathing to be enabled (see 1 above)
global/afi-safis/afi-safi/use-multiple-paths/... is meant to provide constrains to proces of polulation fo Local-RIB and FIB's NHG. The use-multiple-paths/ebgp/link-bandwidth-ext-community/config/enabled control if FIB ECMP weights should be derived form link-bandwidth values or not. In later case basic ECMP will be programmed in dataplane cons=umig less TCAM/SRAM resources, but link-bandwidth will remain in Local-RIB and will be propagated to iBGP peers (and eBGP under draft-bess if enabled).
provided that .../afi-safis/afi-safi/use-multiple-paths/config/enabled is set to "FALSE", and new leaf non-transitive is set to "TRUE"; what shall be system behaviour?
- IMHO the knobs controlling draft-ietf-bess-ebgp-dmz shall be direct attributes of {neighbor|peer-group|global}/afi-safis/afi-safi attribute, similar to send-community-type (which is unfortunetly enum). My suggestion is:
define 2 new container dedicated for link-badwidth community (it is special case anyway) under bgp/global/afi-safis/afi-safi/config/' - tx-link-bandwidthwith leafs:enabled, ebgpdefault FALSE, cummulative, average/equalAND rx-link-bandwidthwith leafs:enabledand default TRUE. iftx-link-bandwidth/enabled` not specified: TRUE on iBGP and FALSE on eBGP.

release/models/bgp/openconfig-bgp-common.yang
dplore: I elaborated on the description.  I would rather keep the leaf name somewhat short.  With the updated description do you think the leaf name `non-transitive-ebgp` is good enough?

release/models/bgp/openconfig-bgp-common.yang
dplore: I'll use divide since that is in the draft.

dplore:
> ```
> Prefix: /network-instances/network-instance/protocols/protocol/bgp/
> 
> bgp/global/afi-safis/afi-safi/link-bw/config/send-cumulative=true
> bgp/global/afi-safis/afi-safi/link-bw/config/non-transitive-ebgp=true
> bgp/global/afi-safis/afi-safi/link-bw/config/divide=false
> ```

Too fast on the comments!  I agree and have moved it.  Pushing commit here  now. :)

```
Prefix: /network-instances/network-instance/protocols/protocol/bgp/

bgp/global/afi-safis/afi-safi/link-bw/config/send-cumulative=true
bgp/global/afi-safis/afi-safi/link-bw/config/non-transitive-ebgp=true
bgp/global/afi-safis/afi-safi/link-bw/config/divide=false
```

dplore:
Reviewed by OC operators on Aug 20, 2024 without objection.

dplore:
@rszarecki and @jhaas-pfrc thank you for the earlier review, this is ready for your review again.

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -696,21 +702,57 @@ submodule openconfig-bgp-common {
 
   grouping bgp-common-use-multiple-paths-link-bandwidth-config {
     description
-      "Parameters controlling usage of  of DMZ Link-Bandwidth
-      extended community in pultipath RIB/FIB formation";
+      "Parameters controlling usage of DMZ Link-Bandwidth
+      extended community in multipath RIB/FIB formation";
```
robshakir, line 706: I don't think this needs to say anything about `multipath`.

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -696,21 +702,57 @@ submodule openconfig-bgp-common {
...
-      This leaf has no effect if BGP multi-path is disabled or
-      if maximum-path attribute of BGP multi-path value is set
-      to 1";
+        "When set to TRUE, BGP multiplepath shall distributed traffic
+        load among contributing routes proportionally to value of
+        Local Administrator subfield of link-bandwidth extended
+        community [draft-ietf-idr-link-bandwidth-07].
+        This leaf has no effect if BGP multi-path is disabled or
+        if maximum-path attribute of BGP multi-path value is set
+        to 1";
```
robshakir, line 717: This is a bit of a head-scratching sentence -- perhaps it should say:

"When set to true, and BGP is configured to use multiple paths, traffic is distributed using the proportions set by the Local Administrator subfield of the Link Bandwidth extended community [draft...]. This leaf has no effect if BGP multi-path is disabled, or if the maximum number of paths is set to 1."

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -696,21 +702,57 @@ submodule openconfig-bgp-common {
...
+        is enabled, the system will sum the bgp link-bandwidth community
+        values for equal cost paths as defined in
+        draft-ietf-bess-ebgp-dmz-05.";
+    }
+
+    leaf non-transitive-ebgp {
+      type boolean;
+      when "../send-cumulative = 'TRUE'";
+      description
+        "When set to TRUE and sending of link-bandwidth community values is
```
robshakir, line 740: Is this the correct description? Is this really sending a community that is marked as non-transitive and then making it transitive? My understanding was that this is for the _originator_ of the community to mark it transitive? @rszarecki

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -696,21 +702,57 @@ submodule openconfig-bgp-common {
...
+        community learned from iBGP peers to eBGP peers.";
+    }
+
+    leaf divide {
+      type boolean;
+      when "../send-cumulative = 'TRUE'";
+      description
+        "When set to TRUE, and sending of link-bandwidth community values
+        is enabled, the system divides the summed bandwidth equally across
+        available paths as defined in draft-ietf-bess-ebgp-dmz-05.";
```
robshakir, line 751: I'm not sure what this means. When we "divide the summed bandwidth equally across available paths" -- isn't that just doing ECMP -- and thus it's the same as just not setting `send-cumulative`?

robshakir (changes requested):

release/models/bgp/openconfig-bgp-common.yang
jhaas-pfrc: The IETF draft is currently up to date.  The behavior is "the extended community is non-transitive, but it's 'originated' by throwing it the next AS by the ASBR".

There's motivation to clarify this scenario in the base extended community RFC and that work will likely get taken up in IDR this year.

jhaas-pfrc:
> @rszarecki and @jhaas-pfrc thank you for the earlier review, this is ready for your review again.

I think in terms of the bess dmz and send non-transitive external being dealt with the current diff is fine.  I suggest paying attention to upcoming last calls for draft-ietf-idr-link-bandwidth likely happening in the next month or so.  I'll also be coordinating with the IETF bess chairs for their drafts for last call considerations.

With regard to the completeness of the work, the link-bandwidth extended community pattern probably needs some work.  Right now I think vendor implementations are only starting to rendezvous in terms of consistency with regard to what link-bw we use.  For example, Juniper added as part of this OC and the related IETF work the transitive version of the community.  But similarly, other vendors have added non-transitive support.

This means there's now one note of ambiguity for the type.  If it's the transitive form, document that.  There's probably related work to create a type for the non-transitive form as well.  That probably could happen in a separate pull request when the agenda for the OC implementers for link-bw have alignment.

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -319,4 +325,26 @@ submodule openconfig-bgp-common-structure {
     }
   }
 
+  grouping bgp-common-structure-link-bw {
+    description
+      "Structural grouping used to include link bandwidth configuration and
+      state for BGP neighbors, peer groups and global.";
+
+    container link-bw {
```
earies, line 333: Nit: use full name as is referenced everywhere else

```suggestion
    container link-bandwidth {
```

PR #1137, opened 2024-06-20
Addition of BFD sessions subtree under interface in openconfig-bfd.yang

Pull-eckermann:
Creation of a BFD single-hop session subtree at interface level, so that BFD sessions can be created besides routing protocols, as stated in [RFC-9127](https://datatracker.ietf.org/doc/rfc9127/).

### Change Scope

* Creation of BFD sessions subtree at interface level, witch allows the configuration of single-hop BFD sessions under the specified interface with well specified local-address and remote-address , outside the multiple BFD clients scope. Multiple BFD sessions can be created in the same interface.
### Platform Implementations

 * IPInfusion OcNOS implementation: OcNOS BFD yang model [ipi-bfd-interface.yang (https://github.com/IPInfusion/OcNOS/blob/OcNOS-SP-6.4.2/yang-files/ipi/bfd/ipi-bfd-interface.yang). Please refer to the bfd-interface-sessions-top grouping. Also, CLI configuration example can be faund in the [BFD Configuration Guide](https://docs.ipinfusion.com/service-provider/index.html#page/OcNOS-SP/BFD-Config/BaseBFDConfig.html#ww1051773)
 * RFC 9127 implementation: The [RFC 9127](https://datatracker.ietf.org/doc/rfc9127/) specifies at section 2.6 the yang.tree of the subtree for single-hop BFD sessions configuration that allows the creation of the sessions attached to a interface, represented as a list key. This was a inspiration for this pull-request. Also the actual OpenCOnfig model does not allow configuration of BFD sessions outside the multiple BFD clients scope, but this RFC states that it should be possible.

### DIFF
--- a/release/models/bfd/openconfig-bfd.yang
+++ b/release/models/bfd/openconfig-bfd.yang
@@ -508,6 +508,100 @@ module openconfig-bfd {
     uses bfd-session-microbfd-common;
   }
 
+  grouping bfd-interface-session-config {
+    description
+      "Configuration parameters for a BFD session on an interface.";
+
+    leaf local-address {
+      type oc-inet:ip-address;
+      description
+        "The local IP address used by the system for the single-hope BFD session
+        specified.";
+    }
+
+    leaf remote-address {
+      type oc-inet:ip-address;
+      description
+        "The remote IP destination that should be used by the system for
+        the single-hope BFD session specified.";
+    }
+
+    leaf enabled {
+      type boolean;
+      description
+        "When this leaf is set to true then the single-hope BFD session is enabled
+        - if it is set to false, it is administratively disabled.";
+    }
+
+    leaf demand-mode {
+      type boolean;
+      description
+        "When this leaf is set to true then the single-hope BFD session 
+        is operating in demand-mode.";
+    }
+  }
+
+  grouping bfd-interface-session-state {
+    // placeholder
+    description
+      "Operational state parameters relating to the single-hope BFD session 
+      running on the interface.";
+  }
+
+  grouping bfd-interface-sessions-structural {
+    description
+      "Structural grouping for BFD sessions configuration on interface
+      and state parameters.";
+
+    container sessions {
+      description
+        "Parameters relating to BFD sessions associated with the interface.";
+
+      list session {
+        key "local-address";
+        key "remote-address";
+
+        description
+          "This list contains configuration and state parameters
+          relating to interface single-hope session.";
+        reference
+          "RFC5881 - Bidirectional Forwarding Detection
+          (BFD) for IPv4 and IPv6 (Single Hop).";
+
+        leaf local-address {
+          type leafref {
+            path "../config/local-address";
+          }
+          description
+            "A reference to the local-address for the BFD single-hop
+            session";
+        }
+        leaf remote-address {
+          type leafref {
+            path "../config/remote-address";
+          }
+          description
+            "A reference to the remote-address for the BFD single-hop
+            session";
+        }
+
+        container config {
+          description
+            "Configuration parameters for the BFD single-hope session.";
+          uses bfd-interface-session-config;
+        }
+
+        container state {
+          config false;
+          description
+            "Operational state parameters for the BFD single-hope session.";
+          uses bfd-interface-session-config;
+          uses bfd-interface-session-state;
+        }
+      }
+    }
+  }
+
   grouping bfd-interface-microbfd-structural {
     description
       "Structural grouping for micro-bfd configuration and state
@@ -650,6 +744,7 @@ module openconfig-bfd {
 
           uses oc-if:interface-ref;
 
+          uses bfd-interface-sessions-structural;
           uses bfd-interface-microbfd-structural;
           uses bfd-interface-peer-structural;
         }


release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +508,100 @@ module openconfig-bfd {
...
+      type boolean;
+      description
+        "When this leaf is set to true then the single-hope BFD session is enabled
+        - if it is set to false, it is administratively disabled.";
+    }
+
+    leaf demand-mode {
+      type boolean;
+      description
+        "When this leaf is set to true then the single-hope BFD session
```
robshakir, line 539:
```suggestion
        "When this leaf is set to true then the single-hop BFD session 
```

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +508,100 @@ module openconfig-bfd {
...
+      description
+        "The local IP address used by the system for the single-hope BFD session
+        specified.";
+    }
+
+    leaf remote-address {
+      type oc-inet:ip-address;
+      description
+        "The remote IP destination that should be used by the system for
+        the single-hope BFD session specified.";
```
robshakir:
```suggestion
        the single-hop BFD session specified.";
```

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +508,100 @@ module openconfig-bfd {
...
   }
 
+  grouping bfd-interface-session-config {
+    description
+      "Configuration parameters for a BFD session on an interface.";
+
+    leaf local-address {
+      type oc-inet:ip-address;
+      description
+        "The local IP address used by the system for the single-hope BFD session
```
robshakir:
```suggestion
        "The local IP address used by the system for the single-hop BFD session
```

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +508,100 @@ module openconfig-bfd {
...
+  }
+
+  grouping bfd-interface-sessions-structural {
+    description
+      "Structural grouping for BFD sessions configuration on interface
+      and state parameters.";
+
+    container sessions {
+      description
+        "Parameters relating to BFD sessions associated with the interface.";
```
robshakir, line 558: Can we add something here that describes that these are statically configured single-hop sessions? I presume the use case here is where we don't have a protocol that ties to that BFD session (so there's no `enable` leaf). Is there anything implied by this being created?

release/models/bfd/openconfig-bfd.yang
Pull-eckermann: Typo fixed in the new commit

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +508,100 @@ module openconfig-bfd {
...
+
+      list session {
+        key "local-address";
+        key "remote-address";
```
robshakir, line 562: This should be `key "local-address remote-address"`. I don't think this syntax is valid. 

>   The "key" statement, which MUST be present if the list represents
>   configuration and MAY be present otherwise, takes as an argument a
>   string that specifies a space-separated list of one or more leaf
>   identifiers of this list.

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +508,100 @@ module openconfig-bfd {
...
+        container config {
+          description
+            "Configuration parameters for the BFD single-hope session.";
+          uses bfd-interface-session-config;
+        }
+
+        container state {
+          config false;
+          description
+            "Operational state parameters for the BFD single-hope session.";
```
robshakir:
```suggestion
            "Operational state parameters for the BFD single-hop session.";
```

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +508,100 @@ module openconfig-bfd {
...
+            path "../config/remote-address";
+          }
+          description
+            "A reference to the remote-address for the BFD single-hop
+            session";
+        }
+
+        container config {
+          description
+            "Configuration parameters for the BFD single-hope session.";
```
robshakir:
```suggestion
            "Configuration parameters for the BFD single-hop session.";
```

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +508,100 @@ module openconfig-bfd {
...
+      description
+        "Parameters relating to BFD sessions associated with the interface.";
+
+      list session {
+        key "local-address";
+        key "remote-address";
+
+        description
+          "This list contains configuration and state parameters
+          relating to interface single-hope session.";
```
robshakir:
```suggestion
          relating to interface single-hop session.";
```

robshakir (commented):
Thanks for the well-formed submission. Generally, looks reasonable to me, assuming I understand the context (question below).

There's one implementation listed here -- are there others that allow static non-protocol-related single hop sessions to be configured?

We should probably clarify what happens if both the per-protocol BFD enabled leaves are used as well as this configuration.

release/models/bfd/openconfig-bfd.yang
Pull-eckermann: This was fixed in the last commit

release/models/bfd/openconfig-bfd.yang
Pull-eckermann: This was fixed in the last commit

release/models/bfd/openconfig-bfd.yang
Pull-eckermann: This was fixed in the last commit

release/models/bfd/openconfig-bfd.yang
Pull-eckermann: This was fixed in the last commit

release/models/bfd/openconfig-bfd.yang
Pull-eckermann: This was fixed in the last commit

release/models/bfd/openconfig-bfd.yang
Pull-eckermann: Done, fixed in the last commit. Thanks for pointing this.

release/models/bfd/openconfig-bfd.yang
Pull-eckermann: Description of sessions container changed to specify that these are statically configured single-hop sessions.
Also changed the name of enable leaf to admin-down, to be more specific. When set to true, the session will be administratively down.

Pull-eckermann:
> Thanks for the well-formed submission. Generally, looks reasonable to me, assuming I understand the context (question below).
> 
> There's one implementation listed here -- are there others that allow static non-protocol-related single hop sessions to be configured?
> 
> We should probably clarify what happens if both the per-protocol BFD enabled leaves are used as well as this configuration.

Q1: yes, after some search I've found a CLI configuration example for Huawei system at this [link](https://www.h3c.com/en/d_202205/1604255_294551_0.htm) in section Configuring Single-Hop BFD.
Also, in the [huawei-bfd.yang](https://github.com/Huawei/yang/blob/1a069483f2ba7213aa4858f799220734ac3eb0f2/network-router/8.22.0/ne8000-x/huawei-bfd.yang) file we can see that this type of configuration is allowed in the static and auto-generated scenarios. THis can be find in "choice multicast-or-unicast-ip" for the unicast case, or in the "choice auto-session-link-type" for ip case. This configuration allows the creation of a static single-hop specifying the interface, source and destination address.

Q2: Probably system checks for two concurrent sessions if source and dest addresses are the same and if so, prioritize the session that was firstly established.

release/models/bfd/openconfig-bfd.yang
```diff
@@ -26,7 +26,14 @@ module openconfig-bfd {
     "An OpenConfig model of Bi-Directional Forwarding Detection (BFD)
     configuration and operational state.";
 
-  oc-ext:openconfig-version "0.3.0";
+  oc-ext:openconfig-version "0.3.1";
```
robshakir, line 29: Since we're adding a feature here, let's make this be a minor release - i.e., this should be `0.4.0`.

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +515,99 @@ module openconfig-bfd {
...
+    }
+
+    leaf remote-address {
+      type oc-inet:ip-address;
+      description
+        "The remote IP destination that should be used by the system for
+        the single-hop BFD session specified.";
+    }
+
+    leaf admin-down {
```
robshakir, line 536: We've generally used `enabled` as this leaf's name -- and have zero called `admin-down`, so let's go for some consistency here and use `enabled`.

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +515,99 @@ module openconfig-bfd {
...
+            "A reference to the local-address for the BFD single-hop
+            session";
+        }
+        leaf remote-address {
+          type leafref {
+            path "../config/remote-address";
+          }
+          description
+            "A reference to the remote-address for the BFD single-hop
+            session";
```
robshakir, line 591:
```suggestion
            session.";
```

release/models/bfd/openconfig-bfd.yang
```diff
@@ -508,6 +515,99 @@ module openconfig-bfd {
...
+          "RFC5881 - Bidirectional Forwarding Detection
+          (BFD) for IPv4 and IPv6 (Single Hop).";
+
+        leaf local-address {
+          type leafref {
+            path "../config/local-address";
+          }
+          description
+            "A reference to the local-address for the BFD single-hop
+            session";
```
robshakir, line 583:
```suggestion
            session.";
```

robshakir (commented):
Couple of small changes, and then LGTM.

Leaving to @dplore to ensure last-call process.

release/models/bfd/openconfig-bfd.yang
earies: This was also raised some time ago whether or not this should have been corrected for those that veered but current stats look like:

```
$ find . -name "*.yang" | xargs grep -E "leaf enable(d?) {" | awk '{print $3}' | sort | uniq -c
     15 enable
     51 enabled
```

LimeHat:
@Pull-eckermann could you please elaborate on the original problem statement?

>Creation of a BFD single-hop session subtree at interface level, so that BFD sessions can be created besides routing protocols, as stated in [RFC-9127](https://datatracker.ietf.org/doc/rfc9127/).

>Creation of BFD sessions subtree at interface level, witch allows the configuration of single-hop BFD sessions under the specified interface with well specified local-address and remote-address ,

I would disagree with that interpretation of the RFC9127 and BFD protocol behavior for a number of reasons.

Firstly, the protocol behavior is defined in RFCs 5880-5882.

RFC5880, section 3.1:
```
Addressing and Session Establishment

   A BFD session is established based on the needs of the application
   that will be making use of it.  It is up to the application to
   determine the need for BFD, and the addresses to use -- there is no
   discovery mechanism in BFD. 
```

RFC5881 tells us that the session must be associated with a protocol:

```
   In this application, there will be only a single BFD session between
   two systems over a given interface (logical or physical) for a
   particular protocol.  The BFD session must be bound to this
   interface.  As such, both sides of a session MUST take the "Active"
   role (sending initial BFD Control packets with a zero value of Your
   Discriminator), and any BFD packet from the remote machine with a
   zero value of Your Discriminator MUST be associated with the session
   bound to the remote system, interface, and protocol.
```

RFC5882 tells us that the protocol is expected to bootstrap the session:
```
 The service interface with BFD is straightforward.  The application
   supplies session parameters (neighbor address, time parameters,
   protocol options), and BFD provides the session state, of which the
   most interesting transitions are to and from the Up state.  The
   application is expected to bootstrap the BFD session, as BFD has no
   discovery mechanism.
   ```

In addition,

2. RFC9127 talks about the configuration of source & destination addresses primarily in the context of multihop and microbfd sessions (see sections 2.1.1-2.1.3)
3. Section 2.1 of RFC9127 states that these configuration parameters are designed so that BFD clients (such as routing protocols) can program the BFD module via the API call using the model. This is not typically used in OC.

Also, probably you want to use RFC9314 instead of 9127 as a reference.

To summarize, while technically the rfc9127 yang model does not prevent the creation of the config as you state, I don't believe it is designed for the purpose you stated.
Is there any other normative reference that allows the creation of independent BFD sessions that are not tied to any protocol or function the way you describe it?

To me, it seems that you're trying to "standardize" a non-standard implementation that requires a fixed provisioning of the destination IP via this pull request.  

I disagree with the proposal in its current form.

release/models/bfd/openconfig-bfd.yang
dplore: @Pull-eckermann please update version to 0.4.0.  Thanks!

release/models/bfd/openconfig-bfd.yang
dplore: @Pull-eckermann please rename this leaf to `enabled`

dplore:
@Pull-eckermann it seems @LimeHat has a point regarding the intent of the RFC.  However, OC doesn't need to strictly follow this.  If there is operational need and implementation precedent for configuration of BFD session IP addresses, (and not just built-in, device protocol based discovery auto session IP address assignment), then this could be added to openconfig.

I see one reference to [ipInfusion configuration of BFD session ip addresses](https://docs.ipinfusion.com/service-provider/index.html#page/OcNOS-SP/BFD-Config/BaseBFDConfig.html#ww1051773) .  Can you find at least one more implementation that supports this kind of configuration?

Issue #1113, opened 2024-05-15
`deprecated` statement under `uses` statement


rolandphung:
It seems that when deprecating `uses` statement in order to restructure the model, the pyang output is not able to represent this at all. This causes restructuring of groupings to be much harder to notice, especially when trying to maintain backwards compatibility. The output of pyang would contain all of the paths and nothing will be indicated as deprecated (`x--rw <node>` vs `+--rw <node>`)

 Doing a bit of research, the pyang devs interpreted the RFC for deprecated to not be recursively applied for containers (See https://github.com/mbj4668/pyang/issues/577). Looking at both RFC6020 and RFC7950 there isn't a defined behavior for the `uses` statement.

What would be the guidance for OpenConfig wrt. adding `status deprecated;` to `uses` statements?

The PR #1105 is one of such changes that are affected by this.

For example:
```
  grouping foo-config {
    leaf a {
      type uint32;
    }

    leaf b {
      type string;
    }
  }

  grouping foo {
    container config {
      uses foo-config;
    }

    container state {
      uses foo-config;
    }
  }

  grouping top {
    container new {
      uses foo;
    }

    container old {
      uses foo {
         status deprecated;
      }
    }
  }

  uses top;
```

```
module: openconfig-test
  +--rw new
  |  +--rw config
  |  |  +--rw a?   uint32
  |  |  +--rw b?   string
  |  +--rw state
  |     +--rw a?   uint32
  |     +--rw b?   string
  +--rw old
     +--rw config
     |  +--rw a?   uint32
     |  +--rw b?   string
     +--rw state
        +--rw a?   uint32
        +--rw b?   string
```

dplore:
The [Yang RFC 6020 Section 7.12.1](https://datatracker.ietf.org/doc/html/rfc6020#section-7.12.1) says `status` is a valid substatement for the `uses` statement.  So it seems what you want to do is supposed to be allowed by RFC.

However, my intuition is that the status field should only be used on nodes and types that appear in the tree (ie: leaf, container, typedef etc) and not to yang language like groupings or uses.  Would this have the effect you're looking for?

earies:
Personally I disagree with the statements in https://github.com/mbj4668/pyang/issues/577

There is no verbiage in YANG 1.0 or 1.1 stating inheritance rules which then I assume could be interpreted as there are none thus means every node must be explicitly marked.

https://datatracker.ietf.org/doc/html/rfc6020#section-7.19.2

[...]

PR #1138, opened 2024-06-23
Adding support for configuration of ECMP of next hops for a static LSP

Shashank-arista:
### Change Scope

* (M) release/models/mpls/openconfig-mpls-static.yang
* (M) release/models/mpls/openconfig-mpls.yang

Currently, model supports configuration of only a single next hop for a given LSP name. We are adding support to configure multiple next hops by adding a new container `next-hops` to support backward compatibility. The incoming-label and metric will be used from existing config itself as its common to all next-hops.

```  +--rw network-instances
        +--rw mpls
           +--rw lsps
              +--rw static-lsps
                 +--rw static-lsp* [name]
                    +--rw name       -> ../config/name
                    +--rw config
                    |  +--rw name?   string
                    +--ro state
                    |  +--ro name?   string
                    +--rw ingress
                    |  +--rw config
                    |  |  +--rw incoming-label?   oc-mplst:mpls-label
                    |  |  x--rw next-hop?         inet:ip-address
                    |  |  x--rw push-label?       oc-mplst:mpls-label
                    |  |  +--rw interface?        -> /oc-if:interfaces/interface/name
                    |  |  +--rw subinterface?     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                    |  |  +--rw metric?           uint8
                    |  +--ro state
                    |  |  +--ro incoming-label?   oc-mplst:mpls-label
                    |  |  x--ro next-hop?         inet:ip-address
                    |  |  x--ro push-label?       oc-mplst:mpls-label
                    |  |  +--ro interface?        -> /oc-if:interfaces/interface/name
                    |  |  +--ro subinterface?     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                    |  |  +--ro metric?           uint8
                    |  +--rw lsp-next-hops                                                                                                                                   [36/9557]
                    |     +--rw lsp-next-hop* [index]
                    |        +--rw index     -> ../config/index
                    |        +--rw config
                    |        |  +--rw index?          uint32
                    |        |  +--rw ip-address?     inet:ip-address
                    |        |  +--rw push-label?     oc-mplst:mpls-label
                    |        |  +--rw interface?      -> /oc-if:interfaces/interface/name
                    |        |  +--rw subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                    |        +--ro state
                    |           +--ro index?          uint32
                    |           +--ro ip-address?     inet:ip-address
                    |           +--ro push-label?     oc-mplst:mpls-label
                    |           +--ro interface?      -> /oc-if:interfaces/interface/name
                    |           +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                    +--rw transit
                    |  +--rw config
                    |  |  +--rw incoming-label?   oc-mplst:mpls-label
                    |  |  x--rw next-hop?         inet:ip-address
                    |  |  x--rw push-label?       oc-mplst:mpls-label
                    |  |  +--rw interface?        -> /oc-if:interfaces/interface/name
                    |  |  +--rw subinterface?     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                    |  |  +--rw metric?           uint8
                    |  +--ro state
                    |  |  +--ro incoming-label?   oc-mplst:mpls-label
                    |  |  x--ro next-hop?         inet:ip-address
                    |  |  x--ro push-label?       oc-mplst:mpls-label
                    |  |  +--ro interface?        -> /oc-if:interfaces/interface/name
                    |  |  +--ro subinterface?     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                    |  |  +--ro metric?           uint8
                    |  +--rw lsp-next-hops
                    |     +--rw lsp-next-hop* [index]
                    |        +--rw index     -> ../config/index
                    |        +--rw config
                    |        |  +--rw index?          uint32
                    |        |  +--rw ip-address?     inet:ip-address
                    |        |  +--rw push-label?     oc-mplst:mpls-label
                    |        |  +--rw interface?      -> /oc-if:interfaces/interface/name
                    |        |  +--rw subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                    |        +--ro state
                    |           +--ro index?          uint32
                    |           +--ro ip-address?     inet:ip-address
                    |           +--ro push-label?     oc-mplst:mpls-label
                    |           +--ro interface?      -> /oc-if:interfaces/interface/name
                    |           +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                    +--rw egress
                       +--rw config
                       |  +--rw incoming-label?   oc-mplst:mpls-label
                       |  x--rw next-hop?         inet:ip-address
                       |  x--rw push-label?       oc-mplst:mpls-label
                       |  +--rw interface?        -> /oc-if:interfaces/interface/name
                       |  +--rw subinterface?     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                       |  +--rw metric?           uint8
                       +--ro state
                       |  +--ro incoming-label?   oc-mplst:mpls-label
                       |  x--ro next-hop?         inet:ip-address
                       |  x--ro push-label?       oc-mplst:mpls-label
                       |  +--ro interface?        -> /oc-if:interfaces/interface/name
                       |  +--ro subinterface?     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                       |  +--ro metric?           uint8
                       +--rw lsp-next-hops
                          +--rw lsp-next-hop* [index]
                             +--rw index     -> ../config/index
                             +--rw config
                             |  +--rw index?          uint32
                             |  +--rw ip-address?     inet:ip-address
                             |  +--rw push-label?     oc-mplst:mpls-label
                             |  +--rw interface?      -> /oc-if:interfaces/interface/name
                             |  +--rw subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                             +--ro state
                                +--ro index?          uint32
                                +--ro ip-address?     inet:ip-address
                                +--ro push-label?     oc-mplst:mpls-label
                                +--ro interface?      -> /oc-if:interfaces/interface/name
                                +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
```

### Platform Implementations

 *  Arista: https://www.arista.com/en/um-eos/eos-mpls-commands#xx1141474
    CLI configuration: mpls static top-label top_tag[ bgp peer [peer IP]] [DEST_INTF] NEXTHOP_ADDR ACTION [PRIORITY]

### DIFF
--- a/release/models/mpls/openconfig-mpls-igp.yang
+++ b/release/models/mpls/openconfig-mpls-igp.yang
@@ -21,7 +21,16 @@ submodule openconfig-mpls-igp {
     "Configuration generic configuration parameters for IGP-congruent
     LSPs";
 
-  oc-ext:openconfig-version "3.5.0";
+  oc-ext:openconfig-version "3.6.0";
+
+  revision "2024-06-19" {
+    description
+      "Added support with backward compatibility to configure ECMP
+      of next-hops for the same LSP name. Also marked attributes
+      push-label, next-hop and interface as deprecated under egress,
+      transit and ingress config";
+    reference "3.6.0";
+  }
 
   revision "2023-12-14" {
     description
--- a/release/models/mpls/openconfig-mpls-static.yang
+++ b/release/models/mpls/openconfig-mpls-static.yang
@@ -24,7 +24,16 @@ submodule openconfig-mpls-static {
     "Defines static LSP configuration";
 
 
-  oc-ext:openconfig-version "3.5.0";
+  oc-ext:openconfig-version "3.6.0";
+
+  revision "2024-06-19" {
+    description
+      "Added support with backward compatibility to configure ECMP
+      of next-hops for the same LSP name. Also marked attributes
+      push-label, next-hop and interface as deprecated under egress,
+      transit and ingress config";
+    reference "3.6.0";
+  }
 
   revision "2023-12-14" {
     description
@@ -156,36 +165,126 @@ submodule openconfig-mpls-static {
 
   // grouping statements
 
-  grouping static-lsp-common-config {
+  grouping static-lsp-nexthop-index {
     description
-      "common definitions for static LSPs";
+      "Next hop index definition for static LSPs";
 
-    leaf next-hop {
+    leaf index {
+      type uint32;
+      description
+        "An user-specified identifier utilised to uniquely reference
+        the next-hop entry in the next-hop list. The value of this
+        index has no semantic meaning other than for referencing
+        the entry.";
+    }
+  }
+
+  grouping static-lsp-nexthop-common-config {
+    description
+      "Common definition of next hop configuration for static LSPs";
+
+    leaf ip-address {
       type inet:ip-address;
       description
-        "next hop IP address for the LSP";
-    }
-
-    leaf incoming-label {
-      type oc-mplst:mpls-label;
-      description
-        "label value on the incoming packet";
+        "Next hop IP address for the LSP";
     }
 
     leaf push-label {
       type oc-mplst:mpls-label;
       description
-        "label value to push at the current hop for the
+        "Label value to push at the current hop for the
         LSP";
     }
 
     // interface-ref
     uses oc-if:interface-ref-common;
+  }
+
+  grouping static-lsp-common-config {
+    description
+      "Common definitions for static LSPs";
+
+    leaf incoming-label {
+      type oc-mplst:mpls-label;
+      description
+        "Label value on the incoming packet";
+    }
+
+    leaf next-hop {
+      type inet:ip-address;
+      status deprecated;
+      description
+        "Next hop IP address for the LSP";
+    }
+
+    leaf push-label {
+      type oc-mplst:mpls-label;
+      status deprecated;
+      description
+        "Label value to push at the current hop for the
+        LSP";
+    }
+
+    // interface-ref
+    uses oc-if:interface-ref-common {
+      status deprecated;
+    }
 
     leaf metric {
       type uint8;
       description
-        "Specifies metric value used for the MPLS route";
+        "Specifies metric value used for the MPLS route.";
+    }
+  }
+
+  grouping static-lsp-nexthops-common {
+    description
+      "MPLS Static LSP next-hops configuration";
+
+    container lsp-next-hops {
+      description
+        "Configuration and state parameters relating to the
+        next-hops that are to be utilised for the MPLS static
+        route being specified";
+
+      list lsp-next-hop {
+        key "index";
+
+        description
+          "A list of next-hops to be utilised for the MPLS
+          static route being specified.";
+
+        leaf index {
+          type leafref {
+            path "../config/index";
+          }
+          description
+            "A reference to the index of the current next-hop.
+            The index is intended to be a user-specified value
+            which can be used to reference the next-hop in
+            question, without any other semantics being
+            assigned to it.";
+        }
+
+        container config {
+          description
+            "Configuration parameters relating to the next-hop
+            entry";
+
+          uses static-lsp-nexthop-index;
+          uses static-lsp-nexthop-common-config;
+        }
+
+        container state {
+          config false;
+          description
+            "Operational state parameters relating to the
+            next-hop entry";
+
+          uses static-lsp-nexthop-index;
+          uses static-lsp-nexthop-common-config;
+        }
+      }
     }
   }
 
@@ -227,6 +326,8 @@ submodule openconfig-mpls-static {
         uses static-lsp-ingress-config;
         uses static-lsp-ingress-state;
       }
+
+      uses static-lsp-nexthops-common;
     }
   }
 
@@ -268,6 +369,8 @@ submodule openconfig-mpls-static {
         uses static-lsp-transit-config;
         uses static-lsp-transit-state;
       }
+
+      uses static-lsp-nexthops-common;
     }
   }
 
@@ -309,6 +412,8 @@ submodule openconfig-mpls-static {
         uses static-lsp-egress-config;
         uses static-lsp-egress-state;
       }
+
+      uses static-lsp-nexthops-common;
     }
   }
 
--- a/release/models/mpls/openconfig-mpls-te.yang
+++ b/release/models/mpls/openconfig-mpls-te.yang
@@ -30,7 +30,16 @@ submodule openconfig-mpls-te {
     signaling protocol or mechanism (see related submodules for
     signaling protocol-specific configuration).";
 
-  oc-ext:openconfig-version "3.5.0";
+  oc-ext:openconfig-version "3.6.0";
+
+  revision "2024-06-19" {
+    description
+      "Added support with backward compatibility to configure ECMP
+      of next-hops for the same LSP name. Also marked attributes
+      push-label, next-hop and interface as deprecated under egress,
+      transit and ingress config";
+    reference "3.6.0";
+  }
 
   revision "2023-12-14" {
     description
--- a/release/models/mpls/openconfig-mpls.yang
+++ b/release/models/mpls/openconfig-mpls.yang
@@ -70,7 +70,16 @@ module openconfig-mpls {
                +------+      |ROUTING|      +-----+
                              +-------+
     ";
-  oc-ext:openconfig-version "3.5.0";
+  oc-ext:openconfig-version "3.6.0";
+
+  revision "2024-06-19" {
+    description
+      "Added support with backward compatibility to configure ECMP
+      of next-hops for the same LSP name. Also marked attributes
+      push-label, next-hop and interface as deprecated under egress,
+      transit and ingress config";
+    reference "3.6.0";
+  }
 
   revision "2023-12-14" {
     description


LimeHat:
I would propose to align this with the [AFT model](https://github.com/openconfig/public/blob/c00868ed96e8e48993e26d8fba20f093722c0e39/release/models/aft/openconfig-aft.yang) and introduce the concept of next-hop-groups.  
cc @robshakir, any thoughts on this?

Also, I'm not entirely clear on the `support backward compatibility` thing. The tree in the description doesn't look like a backward-compatible change to me. Could you elaborate on that?

Shashank-arista:
> I would propose to align this with the [AFT model](https://github.com/openconfig/public/blob/c00868ed96e8e48993e26d8fba20f093722c0e39/release/models/aft/openconfig-aft.yang) and introduce the concept of next-hop-groups. cc @robshakir, any thoughts on this?
> 
> Also, I'm not entirely clear on the `support backward compatibility` thing. The tree in the description doesn't look like a backward-compatible change to me. Could you elaborate on that?

By backward compatibility, I meant the leaves model under (egress | transit | ingress)/config to configure a single next-hop are not being removed to change into a list. Instead, we are introducing a new container for the list of next-hops.

LimeHat:
Ah, I see what you mean.

I don't think it makes too much sense to keep the two (redundant/conflicting) methods in the long term, and the usual procedure of marking the old leafs as `deprecated` should be used.

Shashank-arista:
> Ah, I see what you mean.
> 
> I don't think it makes too much sense to keep the two (redundant/conflicting) methods in the long term, and the usual procedure of marking the old leafs as `deprecated` should be used.

Sure thanks, will update that. Also regarding next-hops, I have tried to use here similar to [Static routes](https://github.com/openconfig/public/blob/master/release/models/local-routing/openconfig-local-routing.yang#L301)

LimeHat:
I'd be happy if we added the concept of NHGs to static-routes too! That makes a lot of sense.

I think you guys support it as well? 
https://www.arista.com/en/um-eos/eos-nexthop-groups

Shashank-arista:
> I'd be happy if we added the concept of NHGs to static-routes too! That makes a lot of sense.
> 
> I think you guys support it as well? https://www.arista.com/en/um-eos/eos-nexthop-groups

Yes, next-hop-groups is like an addition/enhancement to next-hops, right? We can still configure a list of next-hops and use them directly without configuring next-hop groups right?

In EOS, we can create a new next-hop each time we run a command: 
mpls static top-label top_tag [ bgp peer [peer IP]] [DEST_INTF] NEXTHOP_ADDR ACTION [PRIORITY]

Here, we need not use next-hop group

robshakir:
w.r.t `next-hop-groups` -- the general benefit of these is if we end up with lots of sharing. I don't see that we necessarily must have the same approach between AFT and the data model used for config (for example, we do not have this for static routes).

robshakir:
For reference:

* Cisco XR -- allows multiple `path` statements [docs](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/mpls/79x/b-mpls-cg-cisco8000-79x/implementing-mpls-static-labeling.html)
* JUNOS -- allows one next-hop, multiple LSPs possible to the same destination using different "name" [docs](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/static-label-switched-path-edit-protocols-mpls.html) - this is similar to the current model.
* SROS -- allows only one next-hop (docs)[https://infocenter.nokia.com/public/7750SR217R1A/index.jsp?topic=%2Fcom.nokia.MPLS_Guide_21.7.R1%2Fconfiguring_a_s-ai9emdyo7e.html]

@LimeHat @earies @rgwilton -- do you have thoughts here?

LimeHat:
> the general benefit of these is if we end up with lots of sharing.
yes, and I think we should allow that.

> I don't see that we necessarily must have the same approach between AFT and the data model used for config 

Not a must, but a uniform approach would be highly beneficial from an implementation point of view. Otherwise, we need to support different ways of "config" (e.g., for gRIBI vs. CLI), which has to be normalized behind the scenes multiple times/in different ways (e.g., for hw programming, for AFT outputs, etc.).

> (for example, we do not have this for static routes).
I'd love to see it for static routes as well. :-)

earies:
> @LimeHat @earies @rgwilton -- do you have thoughts here?

Will revert back shortly - quick passthrough appears to not align well to existing support within our implementation as proposed

Shashank-arista:
> w.r.t `next-hop-groups` -- the general benefit of these is if we end up with lots of sharing. I don't see that we necessarily must have the same approach between AFT and the data model used for config (for example, we do not have this for static routes).

Also in Arista implementation we don't have option to use next-hop group to configure transit labels and its optional for egress labels (https://www.arista.com/en/um-eos/eos-mpls-commands#xx1141474)

dplore:
To be reviewed on July 30,2024 OC Operators meeting

dplore:
Should this also deprecate the following leafs?  This way it's clear the preferred way to configure any/all LSP's is via the next-hops/next-hop subtree.

```
                    |  |  +--rw interface?        -> /oc-if:interfaces/interface/name
                    |  |  +--rw subinterface?     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                    |  |  +--rw metric?           uint8
```

robshakir:
`ygot` is failing here because after path compression we have duplicate names -- i.e., `next-hops/next-hop` and `state/next-hop` are duplicate names in this generated code. Can we select different naming?

dplore:
> next-hops

To avoid a breaking change we can't rename the existing, to be deprecated leaf so we'll need the newly introduced container and list to use something like `lsp-next-hops/-lsp-next-hop`

Shashank-arista:
> > Should this also deprecate the following leafs? This way it's clear the preferred way to configure any/all LSP's is via the next-hops/next-hop subtree.
> > ```
> >                     |  |  +--rw interface?        -> /oc-if:interfaces/interface/name
> >                     |  |  +--rw subinterface?     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
> >                     |  |  +--rw metric?           uint8
> > ```
> 
> Actually, `interface` and `subinterface` are marked as deprecated in the changes but they are not getting displayed in pyang due to some issue which is discussed here: #1113
> 
> Regarding `metric` deprecation, `incoming-label` and `metric ` together is considered as a key for lsp and is same for a route. So we have kept both them under the `config` as above except others.

Shashank-arista:
> Reviewed with OC Operators on July 30,2024. The following names were proposed. We'll leave this open a couple more days and then should choose one:
> 
> Ideas to resolve conflict due to ygot compression:
> 
> * lsp-next-hops/lsp-next-hop   # most explicit, uses canonical terms, stutters
> * lsp-nhs/lsp-nh                       # pretty cryptic still
> * nhs/nh                                   # too cryptic
> * next-nodes/next-node          # succinct, doesn’t stutter, node is not the canonical term

I would be updating the model using the first option: lsp-next-hops/lsp-next-hop

dplore:
> > Reviewed with OC Operators on July 30,2024. The following names were proposed. We'll leave this open a couple more days and then should choose one:
> > Ideas to resolve conflict due to ygot compression:
> > 
> > * lsp-next-hops/lsp-next-hop   # most explicit, uses canonical terms, stutters
> > * lsp-nhs/lsp-nh                       # pretty cryptic still
> > * nhs/nh                                   # too cryptic
> > * next-nodes/next-node          # succinct, doesn’t stutter, node is not the canonical term
> 
> I would be updating the model using the first option: lsp-next-hops/lsp-next-hop

I approve, please do submit the change.

Shashank-arista:
> > > Reviewed with OC Operators on July 30,2024. The following names were proposed. We'll leave this open a couple more days and then should choose one:
> > > Ideas to resolve conflict due to ygot compression:
> > > 
> > > * lsp-next-hops/lsp-next-hop   # most explicit, uses canonical terms, stutters
> > > * lsp-nhs/lsp-nh                       # pretty cryptic still
> > > * nhs/nh                                   # too cryptic
> > > * next-nodes/next-node          # succinct, doesn’t stutter, node is not the canonical term
> > 
> > 
> > I would be updating the model using the first option: lsp-next-hops/lsp-next-hop
> 
> I approve, please do submit the change.

Thanks, I have updated the PR with the change.

dplore:
Regarding operational use case, Google and others on the Operator group voiced support that this is a real use case in use.

Regarding multiple vendor implementations,  it seems there are three, although they are achieved in different ways.   Cisco and Arista seem to both support configuring multiple next-hops for a static LSP:

* Cisco has a path command which can be repeated in this [example for MPLS over GRE](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/mpls/79x/b-mpls-cg-cisco8000-79x/implementing-mpls-static-labeling.html#concept_fkh_mhz_t4b).  This looks very similar to the proposed OC model.
```
Router(config)# mpls static
Router(config-mpls-static)# lsp v4-encap-payload-1
Router(config-mpls-static-lsp)# in-label 10001 allocate per-prefix 111.0.0.1/32
Router(config-mpls-static-lsp)# forward
Router(config-mpls-static-lsp-fwd))# path 1 nexthop tunnel-ip1 out-label 20001
Router(config-mpls-static-lsp-fwd))# path 2 nexthop tunnel-ip2 out-label 20001
Router(config-mpls-static-lsp-fwd))# path 3 nexthop tunnel-ip3 out-label 20001
Router(config-mpls-static-lsp-fwd))# path 4 nexthop tunnel-ip4 out-label 20001
Router(config-mpls-static-lsp-fwd)# commit
```

* Arista [static MPLS LSP configuration](https://www.arista.com/en/um-eos/eos-mpls-commands#xx1141474) 
@Shashank-arista , is repeating this CLI configuration the method to define multiple next hops in EOS?  It was not clear to me from the docs.
`mpls static top-label top_tag[ bgp peer [peer IP]] [DEST_INTF] NEXTHOP_ADDR ACTION [PRIORITY]`

* As @robshakir  pointed out, Juniper appears to support this in concept, but the implementation is via multiple LSPs to the same destination using different "name" [docs](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/static-label-switched-path-edit-protocols-mpls.html)

@LimeHat I think adding next-hop-group is also an option.  That could be via a separate PR.  As Rob pointed out, the precedent with IP static routes uses only a nexthop.  So I think this PR follows that precedent which is ok.  Internally a NOS is free to construct an AFT using groups to deliver this configured functionality.

dplore:
Last call for comment Aug 19th, 2024

Shashank-arista:
> Regarding operational use case, Google and others on the Operator group voiced support that this is a real use case in use.
> 
> Regarding multiple vendor implementations, it seems there are three, although they are achieved in different ways. Cisco and Arista seem to both support configuring multiple next-hops for a static LSP:
> 
> * Cisco has a path command which can be repeated in this [example for MPLS over GRE](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/mpls/79x/b-mpls-cg-cisco8000-79x/implementing-mpls-static-labeling.html#concept_fkh_mhz_t4b).  This looks very similar to the proposed OC model.
> 
> ```
> Router(config)# mpls static
> Router(config-mpls-static)# lsp v4-encap-payload-1
> Router(config-mpls-static-lsp)# in-label 10001 allocate per-prefix 111.0.0.1/32
> Router(config-mpls-static-lsp)# forward
> Router(config-mpls-static-lsp-fwd))# path 1 nexthop tunnel-ip1 out-label 20001
> Router(config-mpls-static-lsp-fwd))# path 2 nexthop tunnel-ip2 out-label 20001
> Router(config-mpls-static-lsp-fwd))# path 3 nexthop tunnel-ip3 out-label 20001
> Router(config-mpls-static-lsp-fwd))# path 4 nexthop tunnel-ip4 out-label 20001
> Router(config-mpls-static-lsp-fwd)# commit
> ```
> 
> * Arista [static MPLS LSP configuration](https://www.arista.com/en/um-eos/eos-mpls-commands#xx1141474)
>   @Shashank-arista , is repeating this CLI configuration the method to define multiple next hops in EOS?  It was not clear to me from the docs.
>   `mpls static top-label top_tag[ bgp peer [peer IP]] [DEST_INTF] NEXTHOP_ADDR ACTION [PRIORITY]`
> * As @robshakir  pointed out, Juniper appears to support this in concept, but the implementation is via multiple LSPs to the same destination using different "name" [docs](https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/static-label-switched-path-edit-protocols-mpls.html)
> 
> @LimeHat I think adding next-hop-group is also an option. That could be via a separate PR. As Rob pointed out, the precedent with IP static routes uses only a nexthop. So I think this PR follows that precedent which is ok. Internally a NOS is free to construct an AFT using groups to deliver this configured functionality.

@dplore, yes, repeating the CLI configuration is the method to define multiple next hops in EOS.

dplore:
Updated the [implementation references](https://github.com/openconfig/public/pull/1138#issuecomment-2285021373), including adding a reference to SRLinux.

dplore (approved):

Issue #1082, opened 2024-03-27
Support for ACL with netmask set with non-contiguous bits


Pull-eckermann:
In the current OpenConfig, configuration of source-address and destination-address leaves in the /oc-acl:acl/acl-sets/acl-set/acl-entries/acl-entry/ipv4/config (or ipv6) xpath are only supported with netmasks that are left-contiguous, cause this leaves are defined with type oc-inet:ipv4-prefix or oc-inet:ipv6-prefix, which only allows CIDR mask format.

The contiguous mask is applicable when assigning an IP address to an interface, or while adding routes. However, it does not necessarily need to be contiguos for ACLs. ACL's should be capable of filtering based on any kinds of masks. This way multiple non consecutive ranges of networks can be covered in one shot.

Ipinfusiuon OcNOS implements this in the yang [ipi-acl-types.yang](https://github.com/IPInfusion/OcNOS/blob/OcNOS-SP-6.4.2/yang-files/ipi/acl/ipi-acl-types.yang) with the typedef acl_any_ipv4_src_addr_t (line 244), as a union.

Please let me know if this is possible and if OpenConfig was designed to work in this way.

PR #1143, opened 2024-06-28
Acl contribution

Pull-eckermann:
Added support for ACL entry netmask set with non-contiguous bits

This pull resquest is being created in the context of issue [1082](https://github.com/openconfig/public/issues/1082)

In the current OpenConfig, configuration of source-address and destination-address leaves in the /oc-acl:acl/acl-sets/acl-set/acl-entries/acl-entry/ipv4/config (or ipv6) xpath are only supported with netmasks that are left-contiguous, cause this leaves are defined with type oc-inet:ipv4-prefix or oc-inet:ipv6-prefix, which only allows CIDR mask format.

The contiguous mask is applicable when assigning an IP address to an interface, or while adding routes. However, it does not necessarily need to be contiguos for ACLs. ACL's should be capable of filtering based on any kinds of masks. This way multiple non consecutive ranges of networks can be covered in one shot.

### Change Scope

* Changed type of source-address and destination-address leaves (ipv4 and ipv6) to oc-inet:ipv4/6-address. Also created new leaves to represent the netmask for each address, allowing CIDR format or wildcard bits.
* Type of source and destination address were changed but this change is backwards compatible.

### Platform Implementations

* For CISCO, documentatin of this type of configuration can be found [here](https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/sec_data_acl/configuration/15-mt/sec-data-acl-15-mt-book.pdf) in "Wildcard Mask for Addresses in an Access List" section.
* JUNO OS from Junyper also have support to this, docs can be found [here](https://www.juniper.net/documentation/us/en/software/junos/security-policies/topics/topic-map/security-policy-configuration.html) in "Understanding Wildcard Addresses" section.
* FortiOS from Fortinet also have support to this and an example is found in this [page](https://docs.fortinet.com/document/fortigate/7.4.3/administration-guide/838461/wildcard-addressing).
* Huawei documentation about this can be found [here](https://support.huawei.com/enterprise/en/doc/EDOC1100086647#EN-US_TOPIC_0173009747) in Table 1-4.

### DIFF
--- a/release/models/acl/openconfig-packet-match-types.yang
+++ b/release/models/acl/openconfig-packet-match-types.yang
@@ -371,4 +371,28 @@ module openconfig-packet-match-types {
       decimal notation, or using a type defined by the
       ETHERTYPE identity";
   }
+
+  typedef ipv4-prefix-mask-type {
+    type union {
+      type uint8 {
+        range 0..32;
+      }
+      type oc-inet:ipv4-address;
+    }
+    description
+      "The ipv4-prefix mask to be used in acl-entries. Can be in CIDR
+      formmat or wildcard bits.";
+  }
+
+  typedef ipv6-prefix-mask-type {
+    type union {
+      type uint8 {
+        range 0..128;
+      }
+      type oc-inet:ipv6-address;
+    }
+    description
+      "The ipv6-prefix mask to be used in acl-entries. Can be in CIDR
+      formmat or wildcard bits.";
+  }
 }
--- a/release/models/acl/openconfig-packet-match.yang
+++ b/release/models/acl/openconfig-packet-match.yang
@@ -308,9 +308,15 @@ module openconfig-packet-match {
       for IPv4";
 
     leaf source-address {
-      type oc-inet:ipv4-prefix;
+      type oc-inet:ipv4-address;
       description
-        "Source IPv4 address prefix.";
+        "Source IPv4 address.";
+    }
+
+    leaf source-address-mask {
+      type oc-pkt-match-types:ipv4-prefix-mask-type;
+      description
+        "Source IPv4 address prefix mask.";
     }
 
     leaf source-address-prefix-set {
@@ -324,9 +330,15 @@ module openconfig-packet-match {
     }
 
     leaf destination-address {
-      type oc-inet:ipv4-prefix;
+      type oc-inet:ipv4-address;
       description
-        "Destination IPv4 address prefix.";
+        "Destination IPv4 address.";
+    }
+
+    leaf destination-address-mask {
+      type oc-pkt-match-types:ipv4-prefix-mask-type;
+      description
+        "Destination IPv4 address prefix mask.";
     }
 
     leaf destination-address-prefix-set {
@@ -378,9 +390,15 @@ module openconfig-packet-match {
       "Configuration data for IPv6 match fields";
 
     leaf source-address {
-      type oc-inet:ipv6-prefix;
+      type oc-inet:ipv6-address;
       description
-        "Source IPv6 address prefix.";
+        "Source IPv6 address.";
+    }
+
+    leaf source-address-mask {
+      type oc-pkt-match-types:ipv6-prefix-mask-type;
+      description
+        "Source IPv6 address prefix mask.";
     }
 
     leaf source-address-prefix-set {
@@ -400,9 +418,15 @@ module openconfig-packet-match {
     }
 
     leaf destination-address {
-      type oc-inet:ipv6-prefix;
+      type oc-inet:ipv6-address;
       description
-        "Destination IPv6 address prefix.";
+        "Destination IPv6 address.";
+    }
+
+    leaf destination-address-mask {
+      type oc-pkt-match-types:ipv6-prefix-mask-type;
+      description
+        "Destination IPv6 address prefix mask.";
     }
 
     leaf destination-address-prefix-set {


LimeHat:
I'd prefer to see new leafs added instead of changing the existing `source-address`/`destination-address`.

Changing the type is a major, breaking change and, in my opinion, is not warranted in this case.

release/models/acl/openconfig-packet-match-types.yang
```diff
@@ -371,4 +371,28 @@ module openconfig-packet-match-types {
...
+  typedef ipv4-prefix-mask-type {
+    type union {
+      type uint8 {
+        range 0..32;
+      }
+      type oc-inet:ipv4-address;
+    }
+    description
+      "The ipv4-prefix mask to be used in acl-entries. Can be in CIDR
+      formmat or wildcard bits.";
```
robshakir, line 384: (nit) `s/formmat/format/`

Since this a `typedef` that can be used anywhere, we should probably not say where this is used. If it's single usage, prefer just using the above `type` in that place.

robshakir, line 384: It's not actually clear to me what this `typedef` means -- is it solely the mask?

release/models/acl/openconfig-packet-match-types.yang
```diff
@@ -371,4 +371,28 @@ module openconfig-packet-match-types {
...
+  typedef ipv6-prefix-mask-type {
+    type union {
+      type uint8 {
+        range 0..128;
+      }
+      type oc-inet:ipv6-address;
+    }
+    description
+      "The ipv6-prefix mask to be used in acl-entries. Can be in CIDR
+      formmat or wildcard bits.";
```
robshakir, line 396: Same comments as above.

release/models/acl/openconfig-packet-match.yang
```diff
@@ -324,9 +330,15 @@ module openconfig-packet-match {
     }
 
     leaf destination-address {
-      type oc-inet:ipv4-prefix;
+      type oc-inet:ipv4-address;
```
robshakir, line 333: Yeah, let's not do this. This is a breaking change that will cause all consumers to need to change immediately.

If we really think that this prefix specification is not needed, then we should mark this as `deprecated` and then define a new leaf. Otherwise, we should consider adding new leaves for the specific case that is of concern.

It seems like there is an ask for being able to match a prefix against some bitmask. I'd consider this somewhat of a special case and say that we should probably have something like a `masked-source-address` and `source-address-mask` leaf (and corresponding destination based value).

I'd suggest we _only_ use these leaves in the case of non-wildcard matches that cannot be expressed by a CIDR mask.

robshakir (changes requested):
Thanks for the PR and the research to show support on different platforms -- one note, the fact that platforms use two leaves to express address and mask does not imply that they support wildcard matches, so I'd prefer to keep these as two separate configurable entities.

+1 to comments in the thread from @limehat about how we implement the change -- see comments below too.

release/models/acl/openconfig-packet-match-types.yang
Pull-eckermann: Deleting this changes as this typedef is not needed anymore

release/models/acl/openconfig-packet-match-types.yang
Pull-eckermann: Deleting this changes as this typedef is not needed anymore

Pull-eckermann:
@LimeHat @robshakir Breaking change removed and implemented the sugested changes. Two new leaves were added to the specific case that is of concern.

release/models/acl/openconfig-packet-match.yang
```diff
@@ -313,6 +313,18 @@ module openconfig-packet-match {
         "Source IPv4 address prefix.";
     }
 
+    leaf masked-source-address {
+      type oc-inet:ipv4-address;
+      description
+        "Source IPv4 address used with a mask.";
```
LimeHat, line 319: I would recommend adding more detailed descriptions to these leafs, namely:
1) indicate that `masked-source-address` and `source-mask` are always used together, but mutually exclusive with `source-address`
2) indicate that `source-address` is the preferred method of configuring ACLs as suggested by @robshakir

release/models/acl/openconfig-packet-match.yang
```diff
@@ -313,6 +313,18 @@ module openconfig-packet-match {
...
+    }
+
+    leaf destination-source-mask {
```
LimeHat, line 322: destination or source?

release/models/acl/openconfig-packet-match.yang
```diff
@@ -313,6 +313,18 @@ module openconfig-packet-match {
...
+      type oc-inet:ipv4-address;
+      description
+        "Source IPv4 address mask.";
```
LimeHat, line 325: Please specify that this is a wildcard mask. Maybe even better to add wildcard to the name of the leaf.

release/models/acl/openconfig-packet-match.yang
```diff
@@ -329,6 +341,18 @@ module openconfig-packet-match {
...
+    leaf masked-destination-address {
+      type oc-inet:ipv4-address;
+      description
+        "Destination IPv4 address used with a mask.";
+    }
+
+    leaf destination-address-mask {
+      type oc-inet:ipv4-address;
+      description
+        "Destination IPv4 address mask.";
```
LimeHat, line 353: Please specify that this is a wildcard mask. Maybe even better to add `wildcard` to the name of the leaf.

LimeHat (changes requested):

release/models/acl/openconfig-packet-match.yang
Pull-eckermann: Done in the last commit.

release/models/acl/openconfig-packet-match.yang
Pull-eckermann: Done in the last commit.

Pull-eckermann:
@dplore can you /gcbrun again for this PR?

release/models/acl/openconfig-packet-match.yang
```diff
@@ -313,6 +319,21 @@ module openconfig-packet-match {
         "Source IPv4 address prefix.";
     }
 
+    leaf masked-source-address {
+      type oc-inet:ipv4-address;
+      description
+        "Source IPv4 address, to be used with source-wildcard-mask.
+        This configuration is used in cases of matches that cannot 
+        be expressed by a CIDR mask..";
+    }
```
dplore, line 328: Here and throughout I suggest more explicit language:

```suggestion
        "Source IPv4 address, to be used with source-wildcard-mask.
        This node must only be used if a masked address is required.
        If this node is populated, ../source-address must not be used.";
    }
```

release/models/acl/openconfig-packet-match.yang
dplore: Thanks for the update.  I suggested more explicit text

release/models/acl/openconfig-packet-match.yang
Pull-eckermann: Done

release/models/acl/openconfig-packet-match.yang
```diff
@@ -383,6 +419,21 @@ module openconfig-packet-match {
...
+    leaf masked-source-address {
+      type oc-inet:ipv6-address;
+      description
+        "Source IPv6 address, to be used with source-wildcard-mask.
+        This node must only be used if a masked address is required.
+        If this node is populated, ../source-address must not be used.";
+    }
+
+    leaf source-wildcard-mask {
+      type oc-inet:ipv4-address;
```
sahul-arista, line 431: here the type is supposed to be oc-inet:ipv6-address ?

PR #1145, opened 2024-07-01
Update description for GNPSI enum

sallylsy:
GNPSI description fix as per https://github.com/openconfig/public/pull/1085#issuecomment-2151142298


release/models/gnpsi/openconfig-gnpsi-types.yang
```diff
@@ -33,6 +33,6 @@ module openconfig-gnpsi-types {
   identity GNPSI {
     base oc-grpct:GRPC_SERVICE;
```
earies, line 34: As it stands today, this identity is not used whatsoever due to the base identity not having an identityref

The transition for the current `services` leaf-list is a backwards incompatible change when this does occur

Issue #1146, opened 2024-07-10
Attenuator mode - system controlled


ejbrever:
@kcroussore

Today in the optical attenuator model, modes are supported for constant power and constant attenuation. In some platforms, Nokia OLS for example, the system can derive the appropriate targets and adjust as necessary. An example of this use is taking into account a fiber span loss measurement which can be used to derive the optimal optical attenuation. As the the span loss will vary over time, the system would adjust as needed.

I'd like to propose adding a new attenuation mode of SYSTEM_CONTROLLED for this use case.

kcroussore:
I think this makes sense and will prove useful:
CONSTANT_POWER = system is respecting a fixed, externally defined power target
CONSTANT_ATTENUATION = system is respecting a fixed defined attenation regardless of input, output power
SYSTEM_CONTROLED = system is determining the proper power target and setting attenuation to meet that target

ggrammel:
@ejbrever in the example https://github.com/openconfig/public/issues/1146#issue-2401891193 you mention OLS and "system" and I wonder about two points: which entity is actually controlling the span loss compensation and how (in general terms) this is supposed to operate in an OC context? 
My understanding is: to figure out the span loss, a control logic needs to know the TX power of the upstream U-AMP and the Rx power at the downstream D-AMP so that the U-AMP can adjust the Tx-power according to what the D-AMP is measuring as rx-power. IOW if the controller does not set the power or attenuation, then SYSTEM_CONTROLLED would require an undefined communication protocol running between U-AMP and D-AMP. 
The other point is about which entity ("system") is supposed to perform the calculation of "appropriate targets". The proposal seems to suggest it is not the OC-controller but perhaps the AMP itself is setting its own targets. In that case, would computing those targets require (initial) configuration by an OC-controller and would it require additional input such as e.g. Rx-power at U-AMP to get the job done?

ejbrever:
1. There is a protocol running between the near/far end amps, it is the optical supervisory channel.
2. Correct, the device itself (and not the OC controller) would calculate its own targets, but there isn't a need for additional OC driven configs. If that becomes necessary in the future, I think we can decide how to add support for it.

dplore:
I think it's ok to add the state leafs for this in https://github.com/openconfig/public/pull/1147 without the config in the same pull request if you don't have the need to read/write such config.  (Guess it might be enabled by default on some devices?)

PR #1147, opened 2024-07-10
Attenuator mode - system controlled

ejbrever:
Some more details in: https://github.com/openconfig/public/issues/1146

### DIFF
--- a/release/models/optical-transport/openconfig-optical-attenuator.yang
+++ b/release/models/optical-transport/openconfig-optical-attenuator.yang
@@ -143,7 +143,8 @@ module openconfig-optical-attenuator {
       description
         "If the system-derived-target-output-power is equal to or greater than
         'max-output-power + max-output-power-threshold', a device alarm will
-        be raised. This leaf is only relevant when in SYSTEM_CONTROLLED mode.";
+        be raised within /system/alarms. This leaf is only relevant when
+        in SYSTEM_CONTROLLED mode.";
     }
   }
 
--- a/release/models/optical-transport/openconfig-optical-attenuator.yang
+++ b/release/models/optical-transport/openconfig-optical-attenuator.yang
@@ -186,7 +186,7 @@ module openconfig-optical-attenuator {
     leaf system-derived-target-output-power {
       type decimal64 {
         fraction-digits 2;
-        }
+      }
       units dBm;
       description
         "The target output power as determined by the device.
--- a/release/models/optical-transport/openconfig-optical-attenuator.yang
+++ b/release/models/optical-transport/openconfig-optical-attenuator.yang
@@ -24,7 +24,13 @@ module openconfig-optical-attenuator {
     for variable optical attenuators, deployed as part of a transport
     line system.";
 
-  oc-ext:openconfig-version "0.1.0";
+  oc-ext:openconfig-version "0.2.0";
+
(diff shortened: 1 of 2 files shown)


ejbrever:
@kcroussore

release/models/optical-transport/openconfig-optical-attenuator.yang
```diff
@@ -186,7 +186,7 @@ module openconfig-optical-attenuator {
     leaf system-derived-target-output-power {
```
kcroussore, line 186: just curious - would the system derived target belong in "state", or under "config' similar to optical-amplifier target-gain and target-output-power?

ejbrever: Since only the system can define it, I put it under state. So a user could view what the system is doing, but in this mode I don't think they should be able to configure it. Please let me know your thoughts though.

release/models/optical-transport/openconfig-optical-attenuator.yang
```diff
@@ -112,6 +124,27 @@ module openconfig-optical-attenuator {
...
+
+    leaf max-output-power-threshold {
+      type decimal64 {
+        fraction-digits 2;
+      }
+      units dB;
+      description
+        "If the system-derived-target-output-power is equal to or greater than
+        'max-output-power + max-output-power-threshold', a device alarm will
+        be raised. This leaf is only relevant when in SYSTEM_CONTROLLED mode.";
```
dplore, line 146: Naive question here on optical alarms and OC, but which alarm do we expect to be raised?  Do we define it and refer to the appropriate alarm in OpenConfig?

release/models/optical-transport/openconfig-optical-attenuator.yang
dplore: State only makes sense to me.

release/models/optical-transport/openconfig-optical-attenuator.yang
ejbrever: No problem. All of the optical vendors today use the generic alarms list in OC where the "text" leaf is just a vendor chosen strings. We haven't really gone down the path of trying to standardize and define every alarm. When we introduced the alarms list we discussed how that would be ideal, but given there can be 1000 alarms on a platform, it just hasn't been something there's been much progress to date on.

dplore: ok, maybe it makes sense then to just refer to the oc tree for alarms in the description.  (to emphasize we expect the alarm to be set in [/system/alarms](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-alarms)

release/models/optical-transport/openconfig-optical-attenuator.yang
kcroussore: lgtm as well

kcroussore (approved):
lgtm

release/models/optical-transport/openconfig-optical-attenuator.yang
ejbrever: Makes sense and added, thanks.

dplore (approved):

dplore:
Set last call for comments; this will merge on Aug 8, 2024.

PR #1148, opened 2024-07-15
New set-as-path BGP action, and support last AS for for setting and updating AS path.

xuqma:
### Change Scope

* Add set-as-path BGP action to replacing the entire AS path attribute with the following new path:
  * `/routing-policy/policy-definitions/policy-definition/statements/statement/actions/bgp-actions/set-as-path` (a list of ASNs or last-as)
* Add oc-bgp-pol:bgp-as-mode-type to specify last AS in set-as-path.
  * This type is only used for BGP policies, so I decided to put it in `oc-bgp-pol` instead of `oc-bgp-types`.

New paths after the model update:

```
To be updated.
```

### Platform Implementations

 * Arista Implmentation: [RCF as_path.replace](https://www.arista.com/en/support/toi/eos-4-27-2f/15101-routing-control-functions-attribute-and-builtin-function-reference#as-path-replace) and [Route Map set as path all replacement](https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1313923).


release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -1290,6 +1319,184 @@ module openconfig-bgp-policy {
...
+          "oc-bgp-pol:bgp-defined-sets/" +
+          "oc-bgp-pol:as-path-sets/" +
+          "oc-bgp-pol:as-path-set/" +
+          "oc-bgp-pol:as-path-set-name";
+      }
+      description
+        "References a list of defined AS path sets by name";
+    }
+
+    // TODO(xqm): do we need to keep this deperecated leaf?
```
xuqma, line 1397: @dplore Could you please help to confirm if we should keep this leaf which is deprecated in `set-community` and `set-ext-community`?

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -1290,6 +1319,184 @@ module openconfig-bgp-policy {
...
+      }
+    }
+  }
+
+  grouping set-as-path-action-config {
+    description
+      "Configuration data for the set-as-path action";
+
+    // TODO(xqm): should we rename this to set-bgp-defined-set-action-common?
+    uses set-community-action-common;
```
xuqma, line 1459: @dplore I am reusing the `INLINE/REFERENCE` method defined in `set-community-action-common`. Ideally we should rename this so that it's not specific to communities. However, it might introduce some breaking change IIUC. Please advise. Thanks!

xuqma:
Removed the extra comma and rebased against master.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -896,9 +916,18 @@ module openconfig-bgp-policy {
...
+        the local AS number will be used for prepending.";
+    }
+
+    leaf mode {
+      type bgp-as-mode-type;
+      description
+        "Use a special AS number, e.g. last AS, to prepend to the
+        AS path. If neither leaf nor asn leaf is specified but
+        repeat-n is set, then the local AS number will be used for
+        prepending.";
```
dplore, line 930: Since there is only one enum and no clear use cases for expanding this, maybe it makes sense to sense to just make this leaf `use-last-as`

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -229,6 +236,19 @@ module openconfig-bgp-policy {
       the IGP cost using the enum (predefined value).";
   }
 
+  typedef bgp-as-mode-type {
```
dplore, line 239: Although only used in policy, I do think this should be moved to release/models/bgp/openconfig-bgp-types.yang

dplore, line 239: We use the term 'builtin' in other parts of OC to refer to special values or logic that is defined by the device. 

```suggestion
  typedef bgp-as-builtin-type {
```

release/models/bgp/openconfig-bgp-policy.yang
dplore: Thanks for considering, but yes, no need to add a reference to deprecated leafs.    This can be removed.

release/models/bgp/openconfig-bgp-policy.yang
dplore: IMO, we should be verbose and create a new grouping for aspath INLINE/REFERENCE.

release/models/bgp/openconfig-bgp-policy.yang
xuqma: Removed `as-path-set-ref`. Thanks for the confirmation.

release/models/bgp/openconfig-bgp-policy.yang
xuqma: Created `set-as-path-action-config` and new option type `bgp-set-as-path-option-type`.

release/models/bgp/openconfig-bgp-policy.yang
xuqma: Updated the leaf to `use-last-as` as a boolean.

release/models/bgp/openconfig-bgp-policy.yang
xuqma: Moved `bgp-as-builtin-type ` to `openconfig-bgp-types.yang`. Thanks for the suggestion.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -163,6 +170,29 @@ module openconfig-bgp-policy {
 
   // typedef statements
 
+  typedef bgp-set-as-path-option-type {
+    type enumeration {
+      // TODO(xqm): maybe use prepend instead?
+      enum ADD {
```
xuqma, line 176: I am not so sure about certain behaviors after thinking about this a little more. Unlike communities, as-path is ordered and therefore could be a bit tricky.

1. I wonder what would the expected behavior for `add as-path "1 2 3"`. Do we prepend "1 2 3" to the AS path? How about `delete as-path "1 2 3"? Do we delete 1) every AS1, AS2 and AS3 from the path 2) all "1 2 3" sub-paths from the as-path?
2. Using as-path-set reference as input might complicate things further. For instance, an as-path-set can have multiple members such as `["1 2", "3 4 5"]`. After `replace as-path-set`, will the AS path become `"1 2 3 4 5"`? If there are regex in as-path-set, maybe it works for delete, but it's more like invalid input for add/replace.

Would like to have your thoughts to better define the semantics in these cases. Thanks Darren!

dplore: Regarding precedence, I am aware of as path prepending 'n' times, adding the "last-asn" to the list and replacing the originating asn with one's own asn (as-override) .   

I'm not aware of existing implementations that allow arbitrary add/remove operations from an asn path.  

Regarding "editing" an as-path set, practically in  gNMI (although OC isn't gnmi only, netconf is also used),  there is no 'list editing / insert' operation.  To change an 'ordered-by-user' list in gnmi, the whole list must be replaced.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -28,7 +28,14 @@ module openconfig-bgp-policy {
     It augments the base routing-policy module with BGP-specific
     options for conditions and actions.";
 
-  oc-ext:openconfig-version "8.0.0";
+  oc-ext:openconfig-version "8.2.0";
```
xuqma, line 31: Split the change for last-as support to https://github.com/openconfig/public/pull/1221 as discussed offline, which takes version `8.1.0`. This change is expected to merge after that PR, so we use version `8.2.0` after that.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -1322,6 +1362,191 @@ module openconfig-bgp-policy {
     }
   }
 
+  grouping set-as-path-inline-config {
+    description
+      "Configuration data for inline specification of
+      set-as-path action";
+
+    leaf-list as-paths {
+      type union {
```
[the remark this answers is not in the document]
dplore: Ok, we can declare OC which achieves an operational use case that multiple vendors only deliver by different syntax/schema and semantics.    

In this case I think the guiding principle is intact, in that the use case can be delivered by multiple vendors.  It is ok for OC to adopt a syntax that looks more like one implementation than the others, when they are all different.   The alternative is for OC to contrive some schema that that no existing implementation has, just to be different.   I'd rather go with what operators have some rough consensus is "easy enough" to use.

dplore:
LGTM, moving to last call for Jan 16, 2025

xuqma:
Thanks Darren! I have fixed the revisions based on the previous check result.

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -1404,6 +1411,18 @@ module openconfig-bgp-policy {
         "When set-med is specified, this leaf is mandatory to set the
         appropriate action on the MED metric value.";
     }
+
+    leaf-list set-as-path {
```
akalluru1, line 1415: Sorry for the very late comment, Its not possible to store duplicate values in a leaf-list, i.e., for ex: If the AS path to be set is something like `10 10 10 20 30`, as per YANG 1.0 or 1.1. RFE in the config container. Any thoughts or comments?

dplore:
Hi @xuqma ,  still waiting on your comment or changes.  Probably this should just be changed to use the defined-sets for as-path.

ie: [/routing-policy/defined-sets/bgp-defined-sets/as-path-sets/as-path-set/config/as-path-set-member](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-defined-sets-bgp-defined-sets-as-path-sets-as-path-set-config-as-path-set-member)

release/models/bgp/openconfig-bgp-policy.yang
xuqma: Apologies for the delay. I will just repost my comment in the chat room- 

Thanks for the suggestion of switching to as-path-set reference. However, it seems that [as-path-set-mumber](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-defined-sets-bgp-defined-sets-as-path-sets-as-path-set-config-as-path-set-member) is also a leaf-list of string. Would we run into the same issue of not allowing same ASNs in the as-path?

Besides, after further investigation, looks like this new path is not required after all, since `as-path-expand` in Juniper is actually equivalent to `prepend-as-path` in Arista for eBGP export policies. Sorry it took us so long to came to the realization. Since Arista is the only vendor supporting this feature now, maybe we could hold off this proposal until we have another concrete use case? Please let me know what you think. Thank you! @dplore

release/models/bgp/openconfig-bgp-policy.yang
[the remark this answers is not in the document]
jhaas-pfrc: Juniper has strongly avoided features that outright do as-path surgery as it weakens global Internet security.

I'd suggest OC figure out whether it would be comfortable having a conversation with white house CISA about such features given that there's a push to more properly validate the attribute.

Issue #380, opened 2020-09-22
NTP server config should have reference to a NTP-auth-key


jmenok:
Today, OpenConfig defined a list of NTP servers and a list of NTP authentication keys assuming that each server might any of the keys. 
Typically a NTP server can be accessed only with one authentication key.
So from modelling point of view it would be better to have an optional reference in NTP server config specifying the NTP auth key, which should be used for this dedicated NTP server. 

Reference:
https://github.com/openconfig/public/blob/7ff359490559c11cbe63bcce27915dfdb5958ffa/release/models/system/openconfig-system.yang#L480

Verma-Anukul:
Any PR for this request

PR #1149, opened 2024-07-15
Adding NTP auth-key reference for NTP server config

Verma-Anukul:
Related to https://github.com/openconfig/public/issues/380

Issue :
NTP model has key-list, but there is no way to reference the same from
ntp servers.

Fix :
Added a new leafref in server config container to reference configured key.

## Tree view
```diff
      +--rw system
      |  +--rw ntp
      |  +--rw servers
      |     +--rw server* [address]
      |        +--rw address    -> ../config/address
      |        +--rw config
      |        |  +--rw address?            oc-inet:host
      |        |  +--rw port?               oc-inet:port-number
      |        |  +--rw version?            uint8
      |        |  +--rw association-type?   enumeration
      |        |  +--rw iburst?             boolean
      |        |  +--rw prefer?             boolean
      |        |  +--rw network-instance?   oc-ni:network-instance-ref
      |        |  +--rw source-address?     oc-inet:ip-address
+     |        |  +--rw key-id?             -> ../../../../ntp-keys/ntp-key/key-id
      |        +--ro state
      |           +--ro address?            oc-inet:host
      |           +--ro port?               oc-inet:port-number
      |           +--ro version?            uint8
      |           +--ro association-type?   enumeration
      |           +--ro iburst?             boolean
      |           +--ro prefer?             boolean
      |           +--ro network-instance?   oc-ni:network-instance-ref
      |           +--ro source-address?     oc-inet:ip-address
+     |           +--ro key-id?             -> ../../../../ntp-keys/ntp-key/key-id
      |           +--ro stratum?            uint8
      |           +--ro root-delay?         int64
      |           +--ro root-dispersion?    int64
      |           +--ro offset?             int64
      |           +--ro poll-interval?      uint32
```

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -51,7 +51,7 @@ module openconfig-system {
 
   revision "2024-07-15" {
     description
-      "Add auth key reference in ntp server configuration.";
+      "Added auth key reference in ntp server configuration.";
     reference "2.0.1";
   }
 
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -748,6 +748,15 @@ module openconfig-system {
       description
         "Source address to use on outgoing NTP packets";
     }
+
+    leaf key-id {
+      type leafref {
+        //path "/oc-sys:system/oc-sys:ntp/oc-sys:ntp-keys/oc-sys:ntp-key/oc-sys:key-id";
+        path "../../../../ntp-keys/ntp-key/key-id";
+      }
+      description
+        "Reference to NTP authentication key for this server.";
+    }
   }
 
   grouping system-ntp-server-state {
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -47,12 +47,12 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "2.0.1";
+  oc-ext:openconfig-version "2.1.0";
 
   revision "2024-07-15" {
     description
       "Added auth key reference in ntp server configuration.";
-    reference "2.0.1";
+    reference "2.1.0";
   }
 
   revision "2023-12-20" {
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -47,7 +47,13 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "2.0.0";
+  oc-ext:openconfig-version "2.0.1";
+
+  revision "2024-07-15" {
+    description
+      "Add auth key reference in ntp server configuration.";
+    reference "2.0.1";
+  }
 
   revision "2023-12-20" {
     description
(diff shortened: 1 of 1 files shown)


release/models/system/openconfig-system.yang
```diff
@@ -47,7 +47,13 @@ module openconfig-system {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "2.0.0";
+  oc-ext:openconfig-version "2.0.1";
```
dplore, line 50:
```suggestion
  oc-ext:openconfig-version "2.1.0";
```

release/models/system/openconfig-system.yang
```diff
@@ -748,6 +754,14 @@ module openconfig-system {
       description
         "Source address to use on outgoing NTP packets";
     }
+
+    leaf key-id {
+      type leafref {
```
dplore, line 759: I suggest we use the keychain model instead?
For example
```
    leaf keychain {
      type oc-keychain:keychain-ref;
      description
        "Reference to a keychain that should be used for NTP authentication.";
    }
```

dplore (commented):
Thanks for this contribution!  

Please also provide at least 2 references for implementations that support configuring ntp key per server.

Verma-Anukul:
> Thanks for this contribution!
> 
> Please also provide at least 2 references for implementations that support configuring ntp key per server.

References ->

https://github.com/sonic-net/sonic-buildimage/blob/master/src/sonic-yang-models/yang-models/sonic-ntp.yang
                leaf key {
                    description "NTP server key ID";
                    type leafref {
                        path /ntp:sonic-ntp/ntp:NTP_KEY/ntp:NTP_KEYS_LIST/ntp:id;
                    }
                }

https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/ntp-edit-system.html
peer address <key key-number> <prefer> <version value>


https://github.com/YangModels/yang/blob/main/vendor/cisco/xe/1661/Cisco-IOS-XE-ntp.yang
  grouping ntp-peer-server-attr-grouping {
    leaf burst {
      description
        "Send a burst when peer is reachable (Default)";
      type empty;
    }
    leaf iburst {
      description
        "Send a burst when peer is unreachable (Default)";
      type empty;
    }
    leaf key {
      description
        "Configure peer authentication key";
      type uint32;
    }

release/models/system/openconfig-system.yang
dplore: @Verma-Anukul thanks for the updates.  Can you update this to use the keychain model as well?

Verma-Anukul: @dplore  the suggestion here is to use keychain from openconfig-keychain.yang ? If so, ntp-keys defined in openconfig-system.yang won't be required any more right. 
And does any vendor reference for this, as I checked most of them uses ntp key configured specific for ntp model

awebsters: openconfig-keychain allows the use of many encryption algorithms whereas current NTP Authentication only allows MD5. As far as I can see the NTPv4 RFC only specifies MD5 as supported?

awebsters: @dplore a few concerns about using keychain

1. Not every customer will want to use keychain. We should allow them to use the keychain or the key-list under NTP.
2. The leaf we need is a reference to a key and thus cannot be a reference to a keychain. If the keychain is to be supported as an option, it needs to reference a specific key under a keychain.

Verma-Anukul: @dplore 
Please share your view

dplore: Ok, we can treat this PR like a bug-fix, since this addition is needed to make the existing model paths for NTP usable.  We can address adding keychain support in a future PR.

dplore:
Last call for August 7th, 2024.

awebsters:
@dplore can we get this change merged?

dplore:
Last call deadline in Aug 7, 2024.  It will merge then.

PR #1150, opened 2024-07-16
Add pf match rule for network-instance

dplore:
### Change Scope

* The operational use case is to match packets received on a network-instance and apply policy forwarding actions.
* This change only adds containers and leaves and is backwards compatible.

The following two paths are added
* `/network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/network-instance/config/source-network-instance`
* `/network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/network-instance/state/source-network-instance`

### Abbreviated Tree view
```
        +--rw policy-forwarding
        |  +--rw policies
        |  |     +--rw rules
        |  |           +--rw network-instance
        |  |           |  |  +--rw source-network-instance?   -> /network-instances/network-instance/config/name
        |  |           |  +--ro state
        |  |           |     +--ro source-network-instance?   -> /network-instances/network-instance/config/name
```

### Platform Implementations

* IOS XR Example configuration for [vrf-policy](https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5500/qos/b-ncs5500-qos-cli-reference/b-ncs5500-qos-cli-reference_chapter_010.html#wp2717587720)
* JunOS supports [firewall filter match conditions for routing-instance](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/topic-map/firewall-filter-match-conditions-and-actions-ptx.html)

### DIFF
--- a/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
+++ b/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
@@ -21,7 +21,13 @@ submodule openconfig-pf-forwarding-policies {
     "This submodule contains configuration and operational state
     relating to the definition of policy-forwarding policies.";
 
-  oc-ext:openconfig-version "0.6.1";
+  oc-ext:openconfig-version "0.7.0";
+
+  revision "2023-07-16" {
+    description
+      "Add rule to match all packets received on a network-instance.";
+    reference "0.7.0";
+  }
 
   revision "2023-04-25" {
     description
--- a/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
+++ b/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
@@ -274,13 +274,13 @@ submodule openconfig-pf-forwarding-policies {
   grouping pf-forwarding-policy-rule-ni-config {
     description
       "Forwarding policy rules for network-instance matching.";
-    leaf network-instance {
-      type leafref {
-        path "/network-instances/network-instance/config/name";
-      }
+
+    leaf input-match-all {
+      type boolean;
+      default false;
       description
-        "Reference to network-instance name used to match packets
-        for this rule.";
+        "If set to true, all packets received on a network instance
+        are matched by this rule.";
     }
   }
 
--- a/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
+++ b/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
@@ -163,6 +163,25 @@ submodule openconfig-pf-forwarding-policies {
             uses oc-pmatch:ipv6-protocol-fields-top;
             uses oc-pmatch:transport-fields-top;
 
+            container network-instance {
+              description
+                "Container for parameters relating to match
+                rules for network-instance.";
+              container config {
+                description
+                  "Container for configuration parameters relating to
+                  match rules for network-instance.";
+                uses pf-forwarding-policy-rule-ni-config;
+              }
+              container state {
+                config false;
+                description
+                  "Container for state parameters relating to
+                  match rules for network-instance.";
+                uses pf-forwarding-policy-rule-ni-config;
+              }
+            }
+
             container action {
               description
                 "The forwarding policy action to be applied for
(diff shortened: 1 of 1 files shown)


robshakir:
Since a `policy-forwarding` policy is already defined within a network-instance, is this required?

It seems like the use case that you're describing is more "apply this policy to all interfaces within a `network-instance`" -- which seems like it should maybe be something like `/network-instances/network-instance/policy-forwarding/config/apply-network-instance-policy`? Or alternatively something within the `interfaces` list that indicates all interfaces within the NI?

dplore:
> Since a `policy-forwarding` policy is already defined within a network-instance, is this required?
> 
> It seems like the use case that you're describing is more "apply this policy to all interfaces within a `network-instance`" -- which seems like it should maybe be something like `/network-instances/network-instance/policy-forwarding/config/apply-network-instance-policy`? Or alternatively something within the `interfaces` list that indicates all interfaces within the NI?

Agreed the rule is "in" a network-instance, but something is needed to decide which packets to match in the instance?  

Some more concepts that might convey the operational use case:
* match all packets in a network instance
* default packet match for "this" network instance
* match all interfaces (as suggested above)

I kinda like the semantics of "default packet match" which might be modeled as:

`/network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/network-instance/config/default`

Thoughts?

robshakir:
> Agreed the rule is "in" a network-instance, but something is needed to decide which packets to match in the instance?

Perhaps a bit confused as to the use case -- a policy `PBRPOLICY` is assigned within VRF "blue" for an interface using `/network-instances/network-instance[name=blue]/policy-forwarding/interfaces/interface[interface-id=eth0]/config/apply-forwarding-policy = PBRPOLICY`. In this case, I assume that the packets that are candidates to be matched by the policy are all input packets on interface `eth0` (which should be assigned to this network instance).

Today there's no way to specify a policy is applied to anything other than an interface. I think from below you're assuming that configuring a policy applies it, but I don't think this is specified anywhere. My point was that we need a new place to attach the policy if it's going to be that "any packet that enters this network-instance is eligible to be matched by this policy" since the policies today are only attached to interfaces.

Happy to sync in real-time here.

dplore:
@robshakir  and I discussed this offline.  It seems I should:

1. update the description for policy-forwarding rules to specify what happens if there are no rules defined (no packets are matched).
2. Regarding the use case to match all packets in a network-instance, one can write a rule to match all destination IPv4 addresses with a prefix "0.0.0.0/0" to achieve matching all incoming IP packets for a policy forwarding rule.  This will satisfy the use case.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -252,6 +277,19 @@ submodule openconfig-pf-forwarding-policies {
...
 
+  grouping pf-forwarding-policy-rule-ni-config {
+    description
+      "Forwarding policy rules for network-instance matching.";
+
+    leaf input-match-all {
+      type boolean;
+      default false;
+      description
+        "If set to true, all packets received on a network instance
```
LimeHat, line 288: "all received on a network instance" is ambiguous.
Does this include only the packets arriving on interfaces attached to the network-instance?

Or is there an assumption that this statement also covers other use cases (if so, which)? Different scenarios are possible, such as:
- packets that are redirected via another PBF rule for the next lookup
- packets arriving with encapsulation headers (e.g. mpls or vxlan) that indicate the NI for the next lookup/action,
- vrf leaking

LimeHat (changes requested):

LimeHat:
Also, I wasn't able to find this functionality in the provided JunOS reference; can you please double-check (maybe quote a paragraph)?

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: For the use case we have in mind, matching all IPv4/6 packets is sufficient, so using a prefix match rules for 0.0.0.0/0 and ::/0  works.   This change is not needed.

PR #1151, opened 2024-07-16
Remove unused top-level 'messages' container

earies:
* (M) system/openconfig-messages.yang
    - Remove unused `uses` of the top-level `messages` container

### Change Scope

It was noticed that a top-level `uses` has existed in `openconfig-messages`
since initial commit.  This has resulted in the following paths being exposed.

* `/system/messages`
* `/messages`

```
$ pyang openconfig-messages.yang -f tree
module: openconfig-messages
  +--rw messages
     +--rw config
     |  +--rw severity?   oc-log:syslog-severity
     +--ro state
     |  +--ro severity?   oc-log:syslog-severity
     |  +--ro message
     |     +--ro msg?        string
     |     +--ro priority?   uint8
     |     +--ro app-name?   string
     |     +--ro procid?     string
     |     +--ro msgid?      string
     +--rw debug-entries
        +--rw debug-service* [service]
           +--rw service    -> ../config/service
           +--rw config
           |  +--rw service?   identityref
           |  +--rw enabled?   boolean
           +--ro state
              +--ro service?   identityref
              +--ro enabled?   boolean
```

While the latter is unintended.  This change while backwards incompatible
(removing an entire subtree) was never intended and should mostly be unused by
implementations (if they were using it is incorrect).

The initial publish of this module was also incorrectly tagged at `0.0.1` so
this is the first bump to `0.1.0` to indicate we are still within the minor
version scope.

### Platform Implementations

N/A: This is for cleanup of an unintended result of initial modeling

### DIFF
--- a/release/models/system/openconfig-messages.yang
+++ b/release/models/system/openconfig-messages.yang
@@ -32,12 +32,18 @@ module openconfig-messages {
     /yang/system/openconfig-system.yang model, rather it provies the
     Operator with an alternative method of consuming messages.";
 
-  oc-ext:openconfig-version "0.0.1";
+  oc-ext:openconfig-version "0.1.0";
+
+  revision 2024-07-15 {
+    description
+      "Remove unused top-level messages container.";
+    reference "0.1.0";
+  }
 
   revision "2018-08-13" {
-      description
-        "Initial draft.";
-      reference "0.0.1";
+    description
+      "Initial draft.";
+    reference "0.0.1";
   }
 
   // identity statements
@@ -217,5 +223,4 @@ module openconfig-messages {
     uses debug-messages-top;
     }
   }
-  uses messages-top;
 }


robshakir:
LGTM, seems like an error -- thanks for catching it.

earies:
Also for reference - nearly every major implementation is producing a deviation for `/messages` today

https://github.com/aristanetworks/yang/blob/master/EOS-4.32.1F/release/openconfig/models/not-supported/arista-messages-notsupported-deviations.yang#L23-L26

https://github.com/nokia/7x50_YangModels/blob/master/latest_sros_24.7/openconfig/nokia-sr-openconfig-messages-deviations.yang#L36-L38

https://github.com/Juniper/yang/blob/master/23.4/23.4R1.10/openconfig/models/deviations/jnx-openconfig-dev.yang#L4705-L4709

dplore:
Thanks for catching this.  It is unintended.  To add further evidence, the [reference documentation for the messages module](https://github.com/openconfig/public/blob/d23641eeb2c576f3d07b67361d526b3a39ffa8fa/doc/messages_overview.md?plain=1#L40) only refers to the /system/messages tree.

dplore (approved):

PR #1153, opened 2024-07-18
Add aft encapsulation-headers

dplore:
### Change Scope

* Add AFT container representing next-hop encapsulation headers.   A list of encap-headers is defined which is used to represent a stack of one or more packets added to any packet matching a next-hop.
* Add containers for GRE, IP, UDP and MPLS encapsulation

* Later we will deprecate the existing AFT encapsulation containers for [ip-in-ip](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-afts-next-hops-next-hop-ip-in-ip) and [gre](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-afts-next-hops-next-hop-gre) and MPLS.

* Changes to decapsulation are not in scope of this PR as the operational use cases inspiring this PR are met with with existing decapsulation leaves.  
 
### Use cases
* One operational use case for these nexthops is for an SDN controller to use gRIBI to program match and action rules that cause packets to be encapsulated in MPLS in UDP packets.  The SDN controller chooses the encapsulation stack. 
* In another, future operational use case, policy-forwarding might be used to define encapsulations which are representing in the afts model upon instantiation into the forwarding tables of the system.  (Note, at this moment,  changes to the policy-forwarding model to support multiple, arbitrary encapsulations is not in scope for this PR.)

### Platform Implementations

Regarding MPLS in UDP encapsulation (which is the operational use case inspiring this change) is defined by RFC and supported by the following implementations.

* [RFC 7510](https://datatracker.ietf.org/doc/html/rfc7510) defines the MPLS in UDP feature specification.
* [Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/asr9k-r7-6/mpls/configuration/guide/b-mpls-cg-asr9000-76x/implementing-mpls-ldp.html) supports MPLS in UDP.
* [JunOS](https://www.juniper.net/documentation/us/en/software/junos/mpls/topics/topic-map/nexthop-based-dynamic-tunnels.html) nexthop-based-dynamic-tunnels supports MPLS in UDP.

### Tree view

we are at: /network-instances/network-instance/afts/

```diff
diff -U 10 ~/master-tree.txt ~/aft-encap-tree.txt
[... snip ...]
         |  |     +--ro ip-in-ip
         |  |     |  +--ro state
         |  |     |     +--ro src-ip?   oc-inet:ip-address
         |  |     |     +--ro dst-ip?   oc-inet:ip-address
         |  |     +--ro gre
         |  |     |  +--ro state
         |  |     |     +--ro src-ip?   oc-inet:ip-address
         |  |     |     +--ro dst-ip?   oc-inet:ip-address
         |  |     |     +--ro ttl?      uint8
+        |  |     +--ro encap-headers
+        |  |     |  +--ro encap-header* [index]
+        |  |     |     +--ro index     -> ../state/index
+        |  |     |     +--ro state
+        |  |     |     |  +--ro index?   uint8
+        |  |     |     |  +--ro type?    oc-aftt:encapsulation-header-type
+        |  |     |     +--ro gre
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?   oc-inet:ip-address
+        |  |     |     |     +--ro dst-ip?   oc-inet:ip-address
+        |  |     |     |     +--ro ttl?      uint8
+        |  |     |     +--ro ipv4
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?   oc-inet:ip-address
+        |  |     |     |     +--ro dst-ip?   oc-inet:ip-address
+        |  |     |     +--ro ipv6
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?   oc-inet:ip-address
+        |  |     |     |     +--ro dst-ip?   oc-inet:ip-address
+        |  |     |     +--ro mpls
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro traffic-class?      oc-mplst:mpls-tc
+        |  |     |     |     +--ro mpls-label-stack*   oc-mplst:mpls-label
+        |  |     |     +--ro udp-v4
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?         oc-inet:ip-address
+        |  |     |     |     +--ro dst-ip?         oc-inet:ip-address
+        |  |     |     |     +--ro dscp?           oc-inet:dscp
+        |  |     |     |     +--ro src-udp-port?   oc-inet:port-number
+        |  |     |     |     +--ro dst-udp-port?   oc-inet:port-number
+        |  |     |     |     +--ro ip-ttl?         uint8
+        |  |     |     +--ro udp-v6
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?         oc-inet:ip-address
+        |  |     |     |     +--ro dst-ip?         oc-inet:ip-address
+        |  |     |     |     +--ro dscp?           oc-inet:dscp
+        |  |     |     |     +--ro src-udp-port?   oc-inet:port-number
+        |  |     |     |     +--ro dst-udp-port?   oc-inet:port-number
+        |  |     |     |     +--ro ip-ttl?         uint8
+        |  |     |     +--ro vxlan
+        |  |     |        +--ro state
+        |  |     |           +--ro vni-label?               oc-evpn-types:evi-id
+        |  |     |           +--ro tunnel-src-ip-address?   oc-inet:ip-address
         |  |     +--ro interface-ref
         |  |        +--ro state
         |  |           +--ro interface?      -> /oc-if:interfaces/interface/name
         |  |           +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../
```

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -281,46 +281,46 @@ submodule openconfig-aft-common {
             }
 
             container state {
-              description 
+              description
                 "State parameters relating to encapsulation headers.";
               uses aft-common-nexthop-encap-headers-state;
             }
 
             container gre {
-              description 
+              description
                 "Container of nodes for GRE encapsulation.";
               container state {
-                description 
+                description
                   "State parameters relating to GRE encapsulation headers.";
                 uses aft-common-entry-nexthop-gre-state;
               }
             }
 
             container ip {
-              description 
+              description
                 "Container of nodes for IP encapsulation.";
               container state {
-                description 
+                description
                   "State parameters relating to IP encapsulation headers.";
                 uses aft-common-entry-nexthop-ip-state;
               }
             }
 
             container mpls {
-              description 
+              description
                 "Container of nodes for IP encapsulation.";
               container state {
-                description 
+                description
                   "State parameters relating to MPLS encapsulation headers.";
                 uses aft-common-entry-nexthop-mpls-state;
               }
             }
 
             container udp {
-              description 
+              description
                 "Container of nodes for UDP encapsulation.";
               container state {
-                description 
+                description
                   "State parameters relating to UDP encapsulation headers.";
                 uses aft-common-entry-nexthop-udp-state;
               }
@@ -338,7 +338,7 @@ submodule openconfig-aft-common {
       "Operational state parameters relating to encapsulation headers.";
     leaf index {
       type uint8;
-      description 
+      description
         "A pointer to an entry in an ordered list of encapsulation headers.";
     }
   }
@@ -459,7 +459,7 @@ submodule openconfig-aft-common {
 
          Note: a swap operation is reflected by entries in the
          popped-mpls-label-stack and pushed-mpls-label-stack nodes.
-         
+
          This node is deprecated and the encap-headers/encap-header tree
          should be used instead.";
     }
@@ -593,8 +593,8 @@ submodule openconfig-aft-common {
       type oc-inet:port-number;
       description
         "Source UDP port number to use for the UDP header of the encapsulated
-        packet. 
-        
+        packet.
+
         When the payload packet is MPLS, then RFC 7510 - Encapsulating MPLS
         in UDP should be followed.";
       reference
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -501,10 +501,14 @@ submodule openconfig-aft-common {
 
     leaf encapsulate-header {
       type oc-aftt:encapsulation-header-type;
+      status deprecated;
       description
         "When forwarding a packet to the specified next-hop the local
         system performs an encapsulation of the packet - adding the
-        specified header type.";
+        specified header type.
+        
+        This node is deprecated and encap-headers/encap-header/state/type
+        should be used instead.";
     }
 
     leaf decapsulate-header {
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -368,25 +368,15 @@ submodule openconfig-aft-common {
     }
 
     leaf-list pushed-mpls-label-stack {
-      type oc-mplst:mpls-label;
+      type oc-mplst:pushed-mpls-label-stack;
       ordered-by user;
       description
         "The MPLS label stack imposed when forwarding packets to the
-        next-hop
-        - the stack is encoded as a leaf list whereby the order of the
-          entries is such that the first entry in the list is the
-          label at the bottom of the stack to be pushed.
+         next-hop. See the description of the type 
+         oc-mplst:pushed-mpls-label-stack for encoding rules.
 
-        To this end, a packet which is to forwarded to a device using
-        a service label of 42, and a transport label of 8072 will be
-        represented with a label stack list of [42, 8072].
-
-        The MPLS label stack list is ordered by the user, such that no
-        system re-ordering of leaves is permitted by the system.
-
-        A swap operation is reflected by entries in the
-        popped-mpls-label-stack and pushed-mpls-label-stack nodes.";
-
+         Note: a swap operation is reflected by entries in the
+         popped-mpls-label-stack and pushed-mpls-label-stack nodes.";
     }
 
     leaf encapsulate-header {
@@ -464,9 +454,7 @@ submodule openconfig-aft-common {
     leaf src-ip {
       type oc-inet:ip-address;
       description
-        "The source IP address for the MPLS in UDP encapsulation may be
-        expressed using this leaf (src-ip) or if may be derived from
-        '../../interface-ref/state/subinterface'";
+        "The source IP address for the MPLS in UDP encapsulation.";
     }
 
     leaf dst-ip {
@@ -481,13 +469,13 @@ submodule openconfig-aft-common {
         "Destination IP address to use for the encapsulated packet.";
     }
 
-    leaf ttl {
+    leaf ip-ttl {
       type uint8;
       description
-        "This leaf reflects the configured/default TTL value that is used in the
-         outer header during packet encapsulation. When this leaf is not set,
-         the TTL value of the inner packet is copied over as the outer packet's
-         TTL value during encapsulation.";
+        "This leaf reflects the configured/default IP TTL value that is used
+         in the outer header during packet encapsulation. When this leaf is
+         not set, the TTL value of the inner packet is copied over as the
+         outer packet's IP TTL value during encapsulation.";
     }
 
     leaf-list pushed-mpls-label-stack {
@@ -495,16 +483,16 @@ submodule openconfig-aft-common {
       ordered-by user;
       description
         "The MPLS label stack imposed when forwarding packets to the
-        next-hop
-        - the stack is encoded as a leaf list where the first entry in the
-          list is the label at the bottom of the stack to be pushed.
+         next-hop.  The stack is encoded as a leaf list where the first
+         entry in the list is the label at the bottom of the stack to be
+         pushed.
 
-        For example, a packet with a label stack of two labels, the bottom
-        label being 42 and the top label being 8072 will be represented with
-        a leaf-list of [42, 8072].
+         For example, a packet with a label stack of two labels, the bottom
+         label being 42 and the top label being 8072 will be represented with
+         a leaf-list of [42, 8072].
 
-        The MPLS label stack list is ordered by the user, such that no
-        system re-ordering of leaves is permitted by the system.";
+         The MPLS label stack list is ordered by the user, such that no
+         system re-ordering of leaves is permitted by the system.";
     }
 
   }
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -349,6 +349,17 @@ submodule openconfig-aft-common {
               }
             }
 
+            container vxlan {
+              when "../state/type = 'oc-aftt:VXLAN'";
+              description
+                "Container of nodes for VXLAN encapsulation.";
+              container state {
+                description
+                  "State parameters relating to VXLAN encapsulation headers.";
+                uses aft-common-entry-nexthop-vxlan-state;
+              }
+            }
+
           }
         }
 
@@ -415,18 +426,26 @@ submodule openconfig-aft-common {
 
     leaf vni-label {
       type oc-evpn-types:evi-id;
+      status deprecated;
       description
         "Where applicable, the next hop label representing the virtual
         network identifier (VNI) for the forwarding entry. This leaf is
         applicable only to next-hops which include VXLAN encapsulation
-        header information";
+        header information.
+        
+        This leaf is deprecated.  The encap-headers/encap-header/xvlan tree
+        should be used instead.";
     }
 
     leaf tunnel-src-ip-address {
       type oc-inet:ip-address;
+      status deprecated;
       description
         "Where applicable this represents the vxlan tunnel source ip address.
-        For VXLAN this represents the source VTEP ip address";
+        For VXLAN this represents the source VTEP ip address.
+
+        This leaf is deprecated.  The encap-headers/encap-header/xvlan tree
+        should be used instead.";
     }
   }
 
@@ -661,6 +680,27 @@ submodule openconfig-aft-common {
          in the outer header during packet encapsulation. When this leaf is
          not set, the TTL value of the inner packet is copied over as the
          outer packet's IP TTL value during encapsulation.";
+    }
+  }
+
+  grouping aft-common-entry-nexthop-vxlan-state {
+    description
+      "VXLAN encapsulation applied on top of a packet.";
+
+    leaf vni-label {
+      type oc-evpn-types:evi-id;
+      description
+        "Where applicable, the next hop label representing the virtual
+        network identifier (VNI) for the forwarding entry. This leaf is
+        applicable only to next-hops which include VXLAN encapsulation
+        header information";
+    }
+
+    leaf tunnel-src-ip-address {
+      type oc-inet:ip-address;
+      description
+        "Where applicable this represents the vxlan tunnel source ip address.
+        For VXLAN this represents the source VTEP ip address";
     }
   }
 
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -262,8 +262,20 @@ submodule openconfig-aft-common {
             "Container for packet encapsulation headers.  When present, this
             container indicates that encapsulation of the packet matching the
             next-hop is performed using a stack of one or more packets defined
-            in the list encap-header.";
+            in the list encap-header.
+            
+            Each entry in the list must indicate an encapsulation type and 
+            populate a container with the parameters for that encapsulation
+            header.";
           list encap-header {
+            description
+              "A list of headers added on top of a packet.  The first entry
+              in the list is the inner-most packet at the bottom of the
+              stack.
+
+              For example, in an encapsulation stack with MPLS in UDP, the
+              first index in the list is the MPLS packet and the second
+              index is UDP.";
             key "index";
 
             leaf index {
@@ -272,12 +284,13 @@ submodule openconfig-aft-common {
               }
               description
                 "A unique index identifying an encapsulation header in a stack
-                of encapsulation headers. The first entry in the list is the
-                inner-most packet at the bottom of the stack.
+                of encapsulation headers.";
+            }
 
-                For example, in an encapsulation stack with MPLS in UDP, the
-                first, index in the list is the MPLS packet and the second
-                index is UDP.";
+            leaf type {
+              type oc-aftt:encapsulation-header-type;
+              description
+                "Defines which type of packet header should be used.";
             }
 
             container state {
@@ -287,6 +300,7 @@ submodule openconfig-aft-common {
             }
 
             container gre {
+              when "../type = 'oc-aftt:GRE'";
               description
                 "Container of nodes for GRE encapsulation.";
               container state {
@@ -296,9 +310,21 @@ submodule openconfig-aft-common {
               }
             }
 
-            container ip {
+            container ipv4 {
+              when "../type = 'oc-aftt:IPV4'";
               description
-                "Container of nodes for IP encapsulation.";
+                "Container of nodes for IPv4 encapsulation.";
+              container state {
+                description
+                  "State parameters relating to IP encapsulation headers.";
+                uses aft-common-entry-nexthop-ip-state;
+              }
+            }
+
+            container ipv6 {
+              when "../type = 'oc-aftt:IPV6'";
+              description
+                "Container of nodes for IPv6 encapsulation.";
               container state {
                 description
                   "State parameters relating to IP encapsulation headers.";
@@ -307,8 +333,9 @@ submodule openconfig-aft-common {
             }
 
             container mpls {
+              when "../type = 'oc-aftt:MPLS'";
               description
-                "Container of nodes for IP encapsulation.";
+                "Container of nodes for MPLS encapsulation.";
               container state {
                 description
                   "State parameters relating to MPLS encapsulation headers.";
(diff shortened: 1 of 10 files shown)


release/models/aft/openconfig-aft-common.yang
```diff
@@ -438,6 +457,58 @@ submodule openconfig-aft-common {
...
+    leaf dst-ip {
+      type oc-inet:ip-address;
+      description
+        "Destination IP address to use for the encapsulated packet.";
+    }
+
+    leaf dscp {
+      type oc-inet:ip-address;
+      description
+        "Destination IP address to use for the encapsulated packet.";
```
nandanarista, line 481: copy pasta

release/models/aft/openconfig-aft-common.yang
```diff
@@ -238,6 +244,19 @@ submodule openconfig-aft-common {
           }
         }
 
+        container mpls-in-udp {
+          description
+            "When specified, the packet has an MPLS in UDP header applied to it before
+            forwarding to the specified next-hop.";
```
robshakir, line 250: I think here we are implying that `encapsulate-header` should be set to `MPLS_IN_UDPV[46]` before this container is used, right?

robshakir: (This would be consistent with the `encapsulate-header` being set to `IPV4` when we do IPinIPv4.)

release/models/aft/openconfig-aft-common.yang
```diff
@@ -438,6 +457,58 @@ submodule openconfig-aft-common {
...
+         TTL value during encapsulation.";
+    }
+
+    leaf-list pushed-mpls-label-stack {
+      type oc-mplst:mpls-label;
+      ordered-by user;
+      description
+        "The MPLS label stack imposed when forwarding packets to the
+        next-hop
+        - the stack is encoded as a leaf list where the first entry in the
```
robshakir, line 499: Erroneous newline?

Should we make a typedef for this type and its description to ensure consistency across the usages?

release/models/aft/openconfig-aft-common.yang
dplore: > I think here we are implying that `encapsulate-header` should be set to `MPLS_IN_UDPV[46]` before this container is used, right?

Yes, like this snippet:
```yaml
openconfig-network-instance:
  network-instances: 
    - network-instance: "customerX_vpc1"
      afts: 
        next-hops:
          - next-hop: 123
            state:
              encapsulate-header: openconfig-aft-types:MPLS_IN_UDPV6
```

dplore: I'll update the description to reflect that this is a container and it is used when encapsulate-header is set to openconfig-aft-types:MPLS_IN_UDPV(4|6)

akalluru1:
should the idea proposed in https://github.com/openconfig/public/pull/1038#issuecomment-1912343030 be followed in this PR? `pushed-mpls-label-stack` in `mpls-in-udp` container is same as `pushed-mpls-label-stack` in next-hop's `state` container. If we follow the idea in the above link this can be avoided & In near future when `ip-in-udp` needs to be supported then if we rename current `mpls-in-udp` container to `udp` container then it can be easily extended without adding yet another container.

dplore:
> should the idea proposed in [#1038 (comment)](https://github.com/openconfig/public/pull/1038#issuecomment-1912343030) be followed in this PR? `pushed-mpls-label-stack` in `mpls-in-udp` container is same as `pushed-mpls-label-stack` in next-hop's `state` container. If we follow the idea in the above link this can be avoided & In near future when `ip-in-udp` needs to be supported then if we rename current `mpls-in-udp` container to `udp` container then it can be easily extended without adding yet another container.


Ok, so something like `leaf-list encapsulate-stack`  = `[MPLS, UDP]`

we are at: /network-instances/network-instance/afts/next-hops/next-hop/
```
         |  +--ro next-hops
         |  |  +--ro next-hop* [index]
         |  |    +--ro encapsulate-stack
         |  |      +--ro encap-type    (leaflist of type encapsulation-header-type ordered by user )

And then the fields for each encap in the stack would need to be set
     |  |    +--ro encap-headers
     |  |      +--ro encap-header  [index]
     |  |        +--leaf1 for encap type1
     |  |        +--leaf2 for encap type1
     |  |        +--leaf1 for encap type2
     |  |        +--leaf2 for encap type2
... etc ...
```
One then sets the appropriate leaves for each item in the encapsulate-stack. (an MPLS encap would set the MPLS things,  GRE would set the GRE things, etc...)

How does that sound?

dplore:
Here's a better defined, but still conceptual, tree which may answer your questions.  

For backwards compatibility we will deprecate the existing containers for `gre` and `ip-in-ip`.  The new schema would become the preferred schema.  

```
we are at	/network-instances/network-instance/afts/

     |  +--ro next-hops
     |  |  +--ro next-hop* [index]
     |  |    +--ro encapsulate-stack
     |  |    |  +--ro encap-headers
     |  |    |    +--ro encap-header* [index]  (list of type encapsulation-header-type)
     |  |    |      +--ro state
     |  |    |        +--ro index       (an integer identifying the order of encap)
     |  |    |        +--ro type        (encapsulation-header-type, such as IP, GRE, MPLS, UDP)
     |  |    |      +--ro ip-encap      (container when encapsulation-header-type = IP)
     |  |    |        +--ro state
     |  |    |          +--ro src-ip
     |  |    |          +--ro dst-ip
     |  |    |          +--ro ttl
     |  |    |          +--ro dscp
     |  |    |      +--ro gre-encap     (container when encapsulation-header-type = GRE)
     |  |    |        +--ro state
     |  |    |          +--ro src-ip
     |  |    |          +--ro dst-ip
     |  |    |          +--ro ttl
     |  |    |      +--ro mpls-encap    (container when encapsulation-header-type = MPLS)
     |  |    |        +--ro state
     |  |    |          +--ro pushed_label_stack*
     |  |    |      +--ro udp-encap     (container when encapsulation-header-type = UDP)
     |  |    |        +--ro state
     |  |    |          +--ro src-ip
     |  |    |          +--ro dst-ip
     |  |    |          +--ro src-port
     |  |    |          +--ro dst-port
     |  |    |          +--ro ttl
     |  |    |          +--ro dscp
```

akalluru1 (commented):
`pop-top-label` should also be deprecated and should be moved under `mpls-encap` container?

As the MPLS related leaves are moving to its own encap it would be better to move `vni-label` & `tunnel-src-ip-address` fields into a new VXLAN encap container?

dplore:
> 1. `pop-top-label` should also be deprecated and should be moved under `mpls-encap` container? `lsp-name` should also move to mpls encap?
> 2. Doesn't `encapsulate-header` leaf under next-hop state container need to be deprecated as the encap-headers container has been added?
> 3. As the MPLS related leaves are moving to its own encap it would be better to move `vni-label` & `tunnel-src-ip-address` fields into a new VXLAN encap container?

Items 1 and 2, done.

Regarding item 3,  does it make sense to just deprecate these leaves and have the aft indicate an encap-headers/encap-header list with MPLS in IP and [next-hop/state/origin-protocol](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-afts-next-hops-next-hop-state-origin-protocol) of `BGP`?

akalluru1:
> ```diff
> +        |  |     |  x--ro encapsulate-header?          oc-aftt:encapsulation-header-type
> ```

I don't have good knowledge on VxLAN about your question on item 3. But the changes you've made in the latest commit, i.e., creation of a new container for VxLAN encap makes sense & kind of preserves the old `VXLAN` encapsulation enum as well.

dplore:
Last-call for comments!  This is planned to merge on Sep 26,2024.

@rgwilton @nandanarista @earies @limehat for comments.

nandanarista (approved):
The overall hierarchy LGTM.

vishnureddybadveli (approved):
Overall changes looks good to me.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -244,11 +258,134 @@ submodule openconfig-aft-common {
...
+                description
+                  "State parameters relating to MPLS encapsulation headers.";
+                uses aft-common-entry-nexthop-mpls-state;
+              }
+            }
+
+            container udp {
+              when "../state/type = 'oc-aftt:UDPV4' or " +
+              "../state/type = 'oc-aftt:UDPV6'";
+              description
```
robshakir, line 344: Can we add something to the description here to say that this implies both an IP *and* a UDP header are applied? Maybe we should also add something to the IPv4 and IPv6 cases to say that this is just a bare IP header.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -348,40 +495,45 @@ submodule openconfig-aft-common {
...
+
+        This node must be supported in addition to the
+        encap-headers/encap-header tree.  A future release of OpenConfig
+        will deprecate this node in favor of the
+        encap-headers/encap-header subtree.";
     }
 
     leaf-list pushed-mpls-label-stack {
-      type oc-mplst:mpls-label;
+      type oc-mplst:mpls-label-stack;
```
robshakir, line 510: Hmm... this seems confusing to me. We have a `typedef` that says it should only be used in `leaf-list`. 

Are we re-using this typedef anywhere else? It seems like we probably just want it in one place (i.e., `pushed-mpls-label-stack`)?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -244,11 +258,134 @@ submodule openconfig-aft-common {
...
+              description
+                "Container of nodes for GRE encapsulation.";
+              container state {
+                description
+                  "State parameters relating to GRE encapsulation headers.";
+                uses aft-common-entry-nexthop-gre-state;
+              }
+            }
+
+            container ipv4 {
```
robshakir, line 308: Should we have IPv4 and IPv6 separately here? It means there is a change from the existing model where one could have an `ip-in-ip` encap that would be either `ip-in-ipv4` or `ip-in-ipv6` based on the src/dst addresses that were sent. Are we OK with having to have explicitly different leaves here?

robshakir: Note that having this as separate per-AFI `container` conflicts with the approach used for UDP below, where UDP is implying UDPv4 or UDPv6 and the `src-addr` leaf can be either AFI. Is there a reason not to have IPv[46] combined when UDPoIPv[46] is?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -244,11 +258,134 @@ submodule openconfig-aft-common {
           }
         }
 
+        container encap-headers {
```
robshakir, line 261: How does decapsulation work here?

For example, we have `pop-top-label` which is a decapsulation action. We are leaving `decapsulate-header` in place un-changed. So does this mean that we should leave anything that is doing decapsulation in the same place and not change where it lives (this means keeping `pop-top-label` where it is today within `next-hop/state`).

release/models/mpls/openconfig-mpls-types.yang
```diff
@@ -479,6 +485,19 @@ module openconfig-mpls-types {
     reference "RFC 3032 - MPLS Label Stack Encoding";
   }
 
+  typedef mpls-label-stack {
+    type mpls-label;
+  description
```
robshakir, line 490: Indentation.

nachikethas (approved):
LGTM.

Specifically, approving that the existing ip-in-ip encap/decap structures will be deprecated in a _future_ release. Re-iterating that we fully expect all vendors to support this until we've completed the migration.

release/models/aft/openconfig-aft-common.yang
dplore: I would like to leave the changes to decapsulation  in a follow up PR.  At the moment, the existing OC models meet our operational use case for decapsulation without changes.

`pop-top-label` is still in the same place in the tree.

release/models/mpls/openconfig-mpls-types.yang
dplore: done

release/models/aft/openconfig-aft-common.yang
robshakir: I would create a `grouping` that just includes this -- rather than the typedef. If you have a typedef that must be used in a leaf-list, then why not just make sure that it is consistently named, and has a consistent description.

release/models/aft/openconfig-aft-common.yang
robshakir: Disagree re: `pop-top-label` is in the same place in the tree. This change makes `pop-top-label` be available multiple times in the list. If it's solely in the same place in the tree then it would just be at the state location before. 

So, either we clarify how this works here, or we leave `pop-top-label` at exactly the same place is and was before, and *not* here.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -244,11 +258,156 @@ submodule openconfig-aft-common {
...
+          description
+            "Container for packet encapsulation headers.  When present, this
```
robshakir, line 263: Can we reword here -- the presence of the container (in a YANG sense) does not reflect anything. Something like "when options within this container are specified, this indicates that ...".

release/models/aft/openconfig-aft-common.yang
```diff
@@ -244,11 +258,156 @@ submodule openconfig-aft-common {
...
+            in the list encap-header.
+
+            Each entry in the list must indicate an encapsulation type and
+            populate a container with the parameters for that encapsulation
+            header.";
+
+          list encap-header {
+            description
+              "A list of headers added on top of a packet.  The first entry
+              in the list is the inner-most packet at the bottom of the
```
robshakir, line 275: "at the bottom of the stack" -- can we say "the inner-most is closer to the original packet"?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -244,11 +258,156 @@ submodule openconfig-aft-common {
...
+            next-hop is performed using a stack of one or more packets defined
+            in the list encap-header.
+
+            Each entry in the list must indicate an encapsulation type and
+            populate a container with the parameters for that encapsulation
+            header.";
+
+          list encap-header {
+            description
+              "A list of headers added on top of a packet.  The first entry
```
robshakir, line 274: Can we say what the "first entry" means here -- it seems like it is is the lowest index.

Must the indices start at 0?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -244,11 +258,156 @@ submodule openconfig-aft-common {
...
+                  "State parameters relating to MPLS encapsulation headers.";
+                uses aft-common-entry-nexthop-mpls-state;
+              }
+            }
+
+            container udp-v4 {
+              when "../state/type = 'oc-aftt:UDPV4'";
+              description
+                "Container of nodes for UDP in IPv4 encapsulation.  When this
+                container is used, an IPv4 packet with a UDP header is added
```
robshakir, line 351:
```suggestion
                container is used, an IPv4 header and a UDP header is added
```

Suggest this edit here and below. We're not adding a new packet right?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -244,11 +258,156 @@ submodule openconfig-aft-common {
...
+        }
+
         uses oc-if:interface-ref-state;
       }
     }
   }
 
+  grouping aft-common-nexthop-encap-headers-state {
+    description
+      "Operational state parameters relating to encapsulation headers.";
```
robshakir, line 396: (nit): blank line after this one.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -294,14 +453,24 @@ submodule openconfig-aft-common {
...
+        will deprecate this node in favor of the
+        encap-headers/encap-header subtree.";
     }
 
     leaf tunnel-src-ip-address {
       type oc-inet:ip-address;
       description
         "Where applicable this represents the vxlan tunnel source ip address.
-        For VXLAN this represents the source VTEP ip address";
+        For VXLAN this represents the source VTEP ip address.
```
robshakir, line 468:
```suggestion
        For VXLAN this represents the source VTEP IP address.
```

release/models/aft/openconfig-aft-common.yang
```diff
@@ -348,40 +517,45 @@ submodule openconfig-aft-common {
...
       description
         "Flag that controls pop action, i.e., the top-most MPLS label
-         should be popped from the packet when switched by the system.
+        should be popped from the packet when switched by the system.
 
         The top-most MPLS label associated with pop action is equal to
-        the label key used in 'mpls' AFT 'label-entry' list.";
+        the label key used in 'mpls' AFT 'label-entry' list.
+
+        This node must be supported in addition to the
```
robshakir, line 525: I don't think this is true based on your comments above. I think this node will not be deprecated.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -444,6 +618,120 @@ submodule openconfig-aft-common {
...
+        "The source IP address for IP/UDP encapsulation.";
+    }
+
+    leaf dst-ip {
+      type oc-inet:ip-address;
+      description
+        "Destination IP address for IP/UDP encapsulation.";
+    }
+
+    leaf ip-dscp {
```
robshakir, line 672: Can this just be named `dscp`? I don't think it is ambiguous is it?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -444,6 +618,120 @@ submodule openconfig-aft-common {
...
+        "IP DSCP value to use for the UDP header of the encapsulated
+         packet.";
+    }
+
+    leaf src-udp-port {
+      type oc-inet:port-number;
+      description
+        "Source UDP port number to use for the UDP header of the encapsulated
+         packet.  The source UDP port should be derived from the payload
+         packet entropy.  The extact methodology is implementation dependent,
```
robshakir, line 684:
```suggestion
         packet entropy.  The exact methodology is implementation dependent,
```

release/models/aft/openconfig-aft-common.yang
dplore: Ah, you're right, it was added to the encap-header/mpls container, where it does not belong.  I have removed it.

release/models/aft/openconfig-aft-common.yang
dplore: Done

release/models/aft/openconfig-aft-common.yang
dplore: done

release/models/aft/openconfig-aft-common.yang
dplore: done

release/models/aft/openconfig-aft-common.yang
dplore: done

release/models/aft/openconfig-aft-common.yang
dplore: done

release/models/aft/openconfig-aft-common.yang
dplore: done

release/models/aft/openconfig-aft-common.yang
dplore: Yes, we'll leave decap to a future PR and if that results in this leaf being deprecated, note it then.  So I removed this text.

release/models/aft/openconfig-aft-common.yang
dplore: Done

release/models/aft/openconfig-aft-common.yang
dplore: done

release/models/aft/openconfig-aft-common.yang
dplore: done

dplore (commented):
Addressed comments
- do not add pop-top-level to encap-header/mpls container
- refactor mpls-label-stack to be a grouping instead of typdef
- update descriptions

robshakir (approved):
Thanks for addressing my comments!

LGTM.

Issue #1135, opened 2024-06-18
Correction to loopback-mode-type FACILITY and TERMINAL


liulk:
https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-terminal-device.html

Currently:

  * FACILITY: A port **internal** loopback at ASIC level.
  * TERMINAL: A port **external** loopback at ASIC level.

Cisco reported that FACILITY and TERMINAL loopbacks should be reversed. @ahsaanyousaf confirmed it.

Proposed:

  * FACILITY: A port **external** loopback at ASIC level.
  * TERMINAL: A port **internal** loopback at ASIC level.

Background:

@aashaikh first added the definition here: https://github.com/openconfig/public/commit/4e0d236b04326c37275ea9d5c621ea51e7d8eeb8

@aashaikh first use of loopback-mode-type for terminal device loopback-mode leaf here: https://github.com/openconfig/public/commit/b05e0076a8716f4a21dd5029e22fc47d54b7d059

@sulrich changed the interface loopback-mode leaf from boolean to loopback-mode-type here: https://github.com/openconfig/public/pull/731

@sallylsy and @dplore added PHY and MAC specific enums: https://github.com/openconfig/public/pull/1034

liulk:
cc: @aaronmillisor

ahsaanyousaf:
cc: @ejbrever

LimeHat:
An ASIC can implement loopback functionality in the MAC layer or the (internal) PHY layer.
There's also an external PHY layer which corresponds to a transceiver.

One of the points in #1034 was that in some use cases, the actual loopback functionality is not necessarily required and the `FACILITY` is used to emulate the oper-up state of an interface (when the operator doesn't care about the data plane behavior of the interface)

ejbrever:
I agree the definitions of FACILITY and TERMINAL are reversed.

Maybe FACILITY|TERMINAL are not applicable to pluggable optics on a router the same way. In that case, perhaps dedicated enums for those would be better. But I would strongly suggest we do not deprecate FACILITY|TERMINAL as for optical networking they are industry standard names and we have a lot of OpenConfig implementations using those that we shouldn't break (at least ten across four vendors).

liulk:
Updated proposal, after discussion with @dplore:

* Deprecate: `FACILITY`, `TERMINAL` without changing their incorrect meanings.
* Add: `LOCAL` for internal and `REMOTE` for external.
[...]

PR #1155, opened 2024-07-24
Correction to loopback-mode-type FACILITY and TERMINAL #1135

liulk:
### Change Scope

* `release/models/optical-transport/openconfig-transport-types.yang` corrected description for FACILITY and TERMINAL
* NOT backwards compatible

### Platform Implementations

 * Cisco already implemented the corrected definition on both packet and optical devices: https://github.com/openconfig/public/issues/1135#issuecomment-2229397632
 * To my best knowledge, Arista, Juniper and Nokia packet devices implement the incorrect OC definition.
 * According to @ejbrever, Cisco, Ciena, Nokia, Infinera optical devices implement the correct definition.

### DIFF
--- a/release/models/optical-transport/openconfig-transport-types.yang
+++ b/release/models/optical-transport/openconfig-transport-types.yang
@@ -22,7 +22,14 @@ module openconfig-transport-types {
     "This module contains general type definitions and identities
     for optical transport models.";
 
-  oc-ext:openconfig-version "0.25.0";
+  oc-ext:openconfig-version "1.0.0";
+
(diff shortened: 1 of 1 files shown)


dplore:
This was reviewed at the July 23 OC Operators meeting without objection.  We'd like to get feedback from more vendors.  So far we have confirmation from a number of optical platform vendors and [from Cisco on their packet platforms](https://github.com/openconfig/public/issues/1135#issuecomment-2229397632).

Setting this to last call for merging on Aug 15, 2024

dplore (approved):

PR #1156, opened 2024-07-25
Clarify management and cpu interface type descriptions

dplore:
### Change Scope

* Add more detail to the descriptions for cpu and management interface types.
* Since the component tree for management and cpu interfaces both should trace up to a CPU component, more description and examples are helpful to ensure the intended types are used.

### DIFF
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -51,9 +51,16 @@ module openconfig-interfaces {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "3.7.1";
+  oc-ext:openconfig-version "3.7.2";
 
-  revision "2024-04-04" {
+  revision "2024-07-25" {
+      description
+        "Clarify cpu and management interface types.";
+      reference
+        "3.7.2";
+  }
+
+revision "2024-04-04" {
       description
         "Use single quotes in descriptions.";
       reference
@@ -720,9 +727,14 @@ module openconfig-interfaces {
       type boolean;
       description
         "When set to true, the interface is a dedicated
-        management interface that is not connected to dataplane
-        interfaces.  It may be used to connect the system to an
-        out-of-band management network, for example.";
+        management interface connected to the CPU and available
+        for external connectivity to the system.  This interface
+        is not connected to an INTEGRATED_CIRCUIT and therefore
+        not typically used for dataplane traffic.  It may be used
+        to connect the system to an out-of-band management network, 
+        for example.   On systems with a CONTROLLER_CARD, the PORT
+        components on the CONTROLLER_CARD are typically management
+        interfaces";
       oc-ext:telemetry-on-change;
     }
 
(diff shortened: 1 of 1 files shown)


release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -720,21 +727,26 @@ module openconfig-interfaces {
       type boolean;
       description
         "When set to true, the interface is a dedicated
-        management interface that is not connected to dataplane
-        interfaces.  It may be used to connect the system to an
-        out-of-band management network, for example.";
+        management interface connected to the CPU and available
```
LimeHat, line 730: I think `connected to the CPU` phrasing is imprecise. They're co-located on the same board, but the CPU (a chip) is not a direct owner of the interface and does not have a direct connection. If we start counting indirect connections, then any regular interface is connected to the CPU.

LimeHat (changes requested):

release/models/interfaces/openconfig-interfaces.yang
dplore: Perhaps all we can say that  is unique about a management connected interface is:

1. it is not connected to an INTEGRATED_CIRCUIT.
2. It is externally accessible to the device for user management access.

Is this more clear and precise enough to disambiguate from a 'cpu' interface?

LimeHat: That seems reasonable. 
FWIW I'm happy with the current definition of `management` leaf. In my view, ambiguity was related mostly to the definition `cpu` leaf (and the way it is used in gnmi-1.22 https://github.com/openconfig/featureprofiles/blob/4cff2c8fa1815887a481b0b051fbf077464c47ad/feature/platform/controllercard/tests/port/README.md?plain=1#L32-L33)

dplore: I fixed the description for gnmi-1.22 here: https://github.com/openconfig/featureprofiles/pull/3364

I also removed the "connected to the CPU" part of the description.

dplore:
Setting last call to Nov 5, 2024

release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -720,21 +727,25 @@ module openconfig-interfaces {
       type boolean;
       description
         "When set to true, the interface is a dedicated
-        management interface that is not connected to dataplane
-        interfaces.  It may be used to connect the system to an
-        out-of-band management network, for example.";
+        management interface for external connectivity to the system.
+        This interface is not connected to an INTEGRATED_CIRCUIT and
+        therefore not typically used for dataplane traffic.  It may be
```
s19nal, line 732: this is the only part of the description that seems slightly off to me.  I'm not sure I can come up with a better description, but why the use of capitalised 'INTEGRATED_CIRCUIT'?  I think I'm  likely missing the significance of this term - it makes sense that a dedicated management only interface is not designed to make use of the devices dataplane, but what is the relevance of not connecting to an IC? Surely there could be a separate IC specifically designed to handle management plane traffic?

s19nal (commented):
one very minor NIT and one 'simon doesn't really understand something' question :)

release/models/interfaces/openconfig-interfaces.yang
robshakir: Yes, I think this is a good question -- if this port is not connected to some forwarding ASIC, what is it connected to? It seems like it /could/ be connected to an ASIC, just there is no guarantee that it is, since it might just be a NIC-like local Ethernet port on the supervisor/routing-engine/CPM/....

(Simon, I think the relevance of the `INTEGRATED_CIRCUIT` being uppercase is that this is the identity in the `openconfig-platform` model for ASICs/NPUs.)

robshakir (changes requested):
Please make the changes Simon suggested, and then one question on the management interfaces.

PR #1158, opened 2024-08-03
Add preserve-next-hop for ebgp-multihop container

dplore:
### Change Scope
By default BGP changes the next hop of a BGP route to itself when advertising to neighbors.  `ebgp-nexthop/config/preserve-next-hop` is added to disable this behavior.    

This feature is commonly used in combination with BGP based VPN's (RFC4364).  

The full paths added are:

```
/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/ebgp-multihop/config/preserve-next-hop
/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/ebgp-multihop/state/preserve-next-hop

/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/ebgp-multihop/config/preserve-next-hop
/network-instances/network-instance/protocols/protocol/bgp/peer-groups/peer-group/ebgp-multihop/state/preserve-next-hop
```

### Platform Implementations

 * [Cisco IOS XR](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/asr9k-r7-6/routing/configuration/guide/b-routing-cg-asr9000-76x/implementing-bgp.html)
 * [JunOS no-nexthop-change](https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/topic-map/multihop-sessions.html)

### DIFF
--- a/release/models/bgp/openconfig-bgp-common.yang
+++ b/release/models/bgp/openconfig-bgp-common.yang
@@ -24,19 +24,25 @@ submodule openconfig-bgp-common {
     may be application to a subset of global, peer-group or neighbor
     contexts.";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-08-02" {
+    description
+      "Add preserve-next-hop to ebgp-multihop container.";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
+    reference "9.7.1";
   }
 
   revision "2023-12-28" {
     description
       "Add support for controling use of link-bandwidth extended
       community for BGP multipath.";
-          reference "9.7.0";
+    reference "9.7.0";
   }
 
   revision "2023-11-02" {
@@ -432,6 +438,16 @@ submodule openconfig-bgp-common {
         "Time-to-live value to use when packets are sent to the
         referenced group or neighbors and ebgp-multihop is enabled";
     }
+
+    leaf preserve-next-hop {
+     type boolean;
+     default "false";
+     description
+      "Override the default eBGP behavior of rewriting the next-hop, and
+      instead keep the next-hop learned from a neighbor.  This leaf can
+      only be configured if multihop enabled is true.";
+    }
+
   }
 
   grouping bgp-common-neighbor-group-route-reflector-config {
--- a/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
+++ b/release/models/bgp/openconfig-bgp-common-multiprotocol.yang
@@ -24,12 +24,18 @@ submodule openconfig-bgp-common-multiprotocol {
     for multiple protocols in BGP. The groupings are common across
     multiple contexts.";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-08-02" {
+    description
+      "Add preserve-next-hop to ebgp-multihop container.";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
+    reference "9.7.1";
   }
 
   revision "2023-12-28" {
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -21,12 +21,18 @@ submodule openconfig-bgp-common-structure {
     "This sub-module contains groupings that are common across multiple BGP
     contexts and provide structure around other primitive groupings.";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-08-02" {
+    description
+      "Add preserve-next-hop to ebgp-multihop container.";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
+    reference "9.7.1";
   }
 
   revision "2023-12-28" {
--- a/release/models/bgp/openconfig-bgp-global.yang
+++ b/release/models/bgp/openconfig-bgp-global.yang
@@ -27,12 +27,18 @@ submodule openconfig-bgp-global {
     "This sub-module contains groupings that are specific to the
     global context of the OpenConfig BGP module";
 
-  oc-ext:openconfig-version "9.7.1";
+  oc-ext:openconfig-version "9.8.0";
+
+  revision "2024-08-02" {
+    description
+      "Add preserve-next-hop to ebgp-multihop container.";
+    reference "9.8.0";
+  }
 
   revision "2023-12-28" {
     description
       "Clarification to last-prefix-limit-exceeded in description";
-          reference "9.7.1";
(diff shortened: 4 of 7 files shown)


release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -432,6 +438,16 @@ submodule openconfig-bgp-common {
...
         referenced group or neighbors and ebgp-multihop is enabled";
     }
+
+    leaf preserve-next-hop {
+     type boolean;
+     default "false";
+     description
+      "Override the default eBGP behavior of rewriting the next-hop, and
+      instead keep the next-hop learned from a neighbor.  This leaf can
+      only be configured if multihop enabled is true.";
```
LimeHat, line 448: The placement of this leaf and the multihop requirement are not correct. This only standardizes the limitation of a specific implementation. Fundamentally there's no tie between the two.

LimeHat (changes requested):
This should be configured with per afi/safi granularity and not tied to multihop.

Nokia ref: https://documentation.nokia.com/sr/24-7/7750-sr/cli-books/classic-cli-command-reference/classic-n-commands.html?hl=next-hop-unchanged#yc23860

Nvidia ref: https://docs.nvidia.com/networking/display/onyxv3104302/bgp+commands#src-132451846_BGPCommands-neighbornext-hop-unchanged

Another Cisco ref that shows per-afi/safi config
https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst3650/software/release/16-12/configuration_guide/rtng/b_1612_rtng_3650_cg/configuring_bgp_next_hop_unchanged.pdf

I know that Arista supports the config like in the above cisco reference as well, just wasn't able to find a public reference immediately.

release/models/bgp/openconfig-bgp-common.yang
dplore: Thanks for the feedback, moved to afi-safi hierarchy

dplore:
> This should be configurable with per afi/safi granularity and not tied to multihop.
> 
> Nokia ref: https://documentation.nokia.com/sr/24-7/7750-sr/cli-books/classic-cli-command-reference/classic-n-commands.html?hl=next-hop-unchanged#yc23860
> 
> Nvidia ref: https://docs.nvidia.com/networking/display/onyxv3104302/bgp+commands#src-132451846_BGPCommands-neighbornext-hop-unchanged
> 
> Another Cisco ref that shows per-afi/safi config https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst3650/software/release/16-12/configuration_guide/rtng/b_1612_rtng_3650_cg/configuring_bgp_next_hop_unchanged.pdf
> 
> I know that Arista supports the config like in the above cisco reference as well, just wasn't able to find a public reference immediately.

Moved to afi-safi container.  I also renamed the leaf to 'next-hop-unchanged' to align with something much closer to what implementations are already using.

robshakir:
This is still inconsistent with how next-hop-self is represented in OpenConfig -- where it is something that is within a policy not a leaf. I don't agree with this proposal unless we have a coherent approach within the models.

Note that implementations would already a need to be mapping the current approach to underlying config -- so either this is implemented (and thus can be extended), or unimplementable (and thus the current n-h-s approach should be deprecated).

robshakir (changes requested):
Requesting changes to make this consistent with next-hop self in the models.

dplore:
> Requesting changes to make this consistent with next-hop self in the models.

Sorry I do not understand the issue.  Could you provide an example of how BGP next-hop-self is represented in openconfig?   Maybe that will be enough for me to understand how to propose something that enable/disables that.  But if you have a suggestion in mind, I'd also be happy to see that too.

robshakir:
Next-hop self is specified through:

1. (`/routing-policy/policy-definitions/policy-definition/statements/statement/actions/bgp-actions/config/set-next-hop`)[https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-actions-bgp-actions-config-set-next-hop]
2. (`bgp-next-hop-type`)[https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#type-bgp-next-hop-type] enum values -- i.e., `SELF` here.

We have discussed this numerous times over the years -- since some implementations have this in config not policy, I'm not understanding why we would have one thing (n-h-s) be via policy and another (unchanged) be in neighbour config.

dplore:
> Next-hop self is specified through:
> 
> 1. (`/routing-policy/policy-definitions/policy-definition/statements/statement/actions/bgp-actions/config/set-next-hop`)[https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-actions-bgp-actions-config-set-next-hop]
> 2. (`bgp-next-hop-type`)[https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#type-bgp-next-hop-type] enum values -- i.e., `SELF` here.
> 
> We have discussed this numerous times over the years -- since some implementations have this in config not policy, I'm not understanding why we would have one thing (n-h-s) be via policy and another (unchanged) be in neighbour config.

Ah ok, I understand the reason and goal here.  But the semantics are different from the 100% of all the precedent that I can find (I spent a lot more time search and found a total of 6 references below, all of which have distinct configuration for this and apply on a per session basis).  

Is there an operational reason to need to craft policies to selectively choose which routes would cause next-hop self to be unchanged?  The operational use cases I know of are both the classic scenario of override RFC4271 behavior: EVPN and an internal one at Google.  The Google use case is for **R1** to  learn multi-path routes from an eBGP neighbor **R2** and we want **R1** to retain the next-hops/multipath information when sending the routes to iBGP peer **R3**.

* Cisco [IOS XR](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/24xx/routing/configuration/guide/b-routing-cg-asr9000-24xx/implementing-bgp.html?bookSearch=true)
* JunOS [no-nexthop-change](https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/topic-map/multihop-sessions.html). Note, this is implemented only at a hierarchy of eBGP
* Nokia SROS [next-hop-unchanged](https://documentation.nokia.com/sr/24-7/7750-sr/cli-books/classic-cli-command-reference/classic-n-commands.html?hl=next-hop-unchanged#yc23860)
* Nvidia [switch (config router bgp 65001 address-family l2vpn-evpn) # next-hop-unchanged](https://docs.nvidia.com/networking/display/onyxv3104302/bgp+commands#src-132451846_BGPCommands-neighbornext-hop-unchanged):
* Arista EOS [neighbor xyz123 next-hop-unchanged](https://www.arista.com/en/um-eos/eos-sample-configurations). (although I found other references in support articles that are not publicly viewable that if address-family evpn is configured, then this behavior is the default).
* Extreme Networks [neighbor next-hop-unchanged](https://documentation.extremenetworks.com/slxos/sw/20xx/20.3.2/commands/GUID-00D8537D-27AE-4F70-BF7E-349B860B5D8A.shtml)

I appreciate that it is elegant to have 'things related to next-hop-self' in the same subtree, but from a standpoint of precedent and ability to easily translate existing bgp policies into openconfig, applying on a per session is very straightforward.  If we apply via policy, then I guess it's just adding another statement with no condition and an action `next-hop-unchanged`.  Implementations however will have to parse the logic in context of routing-policy.  This will be be an issue if there are `statement/conditions` defined, which is something we aren't asking for and implementations don't offer (to my knowledge).

PR #1160, opened 2024-08-08
Update description of the model-name leaf.

SydneyCaulfeild:
### Change Scope

* Updating the description of the model-name leaf to help distinguish it from the part-no leaf.

### DIFF
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -546,8 +546,9 @@ module openconfig-platform {
     }
 
     leaf model-name {
-      type string;
-      description
+      when "../removable = 'true'" {
+        type string;
+        description
         "The hardware model name of a component. This is the hardware
         name that would be found in a device catalog and is often the
         orderable name of the component. This is required for all FRU
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -546,9 +546,8 @@ module openconfig-platform {
     }
 
     leaf model-name {
-      when "../removable = 'true'" {
-        mandatory true;
-      }
+      when "../removable = 'true'";
+      mandatory true;
       type string;
       description
         "Model name that would be found in a catalog of stock keeping
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -542,9 +542,13 @@ module openconfig-platform {
     leaf model-name {
       type string;
       description
-        "A human readable string describing the model of a component.
-        This string is optional and should only be populated if part-no
-        is also populated.";
+        "The model of a component. This should be the orderable part
+        number when the component is a FRU (field replaceable unit).
+        To distinguish this from part-no, the model-name is the model
+        that could be searched for online. The part-no is the unique
+        identifier of a part, but would not necessarily have meaning
+        to someone reading it. The model-name is optional and should
+        only be populated if part-no is also populated.";
     }
 
     leaf clei-code {
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -551,9 +551,9 @@ module openconfig-platform {
       }
       type string;
       description
-        "This is the component name that would be found in a catalog of 
-        stock keeping units (SKU) and should be the orderable name of
-        the component. This is required for all FRU (field replaceable unit)
+        "Model name that would be found in a catalog of stock keeping
+        units (SKU) and should be the orderable name of the component.
+        This is required for all FRU (field replaceable unit)
         components (i.e., /components/component/state/removable=TRUE).";
     }
 
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -549,11 +549,10 @@ module openconfig-platform {
       when "../removable = 'true'" {
         type string;
         description
-        "The hardware model name of a component. This is the hardware
-        name that would be found in a device catalog and is often the
-        orderable name of the component. This is required for all FRU
-        (field replaceable unit) components (i.e.,
-        /components/component/state/removable=TRUE).";
+        "This is the component name that would be found in a catalog of 
+        stock keeping units (SKU) and should be the orderable name of
+        the component. This is required for all FRU (field replaceable unit)
+        components (i.e., /components/component/state/removable=TRUE).";
     }
 
     leaf clei-code {
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -546,7 +546,7 @@ module openconfig-platform {
     }
 
     leaf model-name {
-      when "../removable = 'true' or ../type = 'CHASSIS'";
+      when "../removable = 'true' or ../type = 'oc-platform-types:CHASSIS'";
       mandatory true;
       type string;
       description
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -548,13 +548,8 @@ module openconfig-platform {
     leaf model-name {
       type string;
       description
-        "The model of a component. This should be the orderable part
-        number when the component is a FRU (field replaceable unit).
-        To distinguish this from part-no, the model-name is the model
-        that could be searched for online. The part-no is the unique
-        identifier of a part, but would not necessarily have meaning
-        to someone reading it. The model-name is optional and should
-        only be populated if part-no is also populated.";
+        "The model of a component. The model-name is optional and
+        should only be populated if part-no is also populated.";
     }
 
     leaf clei-code {
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,7 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.27.0";
+  oc-ext:openconfig-version "0.28.0";
+
+  revision "2024-08-08" {
+    description
+      "Update description of model-name leaf.";
+    reference "0.28.0";
+  }
 
   revision "2024-05-29" {
     description
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -546,7 +546,7 @@ module openconfig-platform {
     }
 
     leaf model-name {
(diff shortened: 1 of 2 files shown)


release/models/platform/openconfig-platform.yang
```diff
@@ -542,9 +548,13 @@ module openconfig-platform {
     leaf model-name {
       type string;
       description
-        "A human readable string describing the model of a component.
-        This string is optional and should only be populated if part-no
-        is also populated.";
+        "The model of a component. This should be the orderable part
+        number when the component is a FRU (field replaceable unit).
```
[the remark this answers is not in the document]
dplore: Here's an example (this is a hypothetical example, but I assure you this issue exists in production):

```
/components/component[name="fabric1"]/state/part-no = "123-4567890"
/components/component[name="fabric1"]/state/model-name =  "FAB-100-M2"
```

In this example, the part-no is referencing some part+revision number which is internal to the vendor hardware revision.  This number might change even when the SKU/orderable part does not change (for example due to minor hardware revisions that do not affect the feature set).   The model-name is a "marketing sku" which is familiar to humans, appears in parts catalogs and can be ordered.

Not all vendors do this.  Some vendors only expose a "marketing sku" which is orderable.  In this case the leaves are redundant:
```
/components/component[name="fabric1"]/state/part-no = "FAB-100-M2"
/components/component[name="fabric1"]/state/model-name =  "FAB-100-M2"
```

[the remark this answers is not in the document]
SydneyCaulfeild: For context, I also plan on updating the type of this leaf from a string to an identity (https://github.com/openconfig/public/pull/1161) so that everyone is on the same page on which model names are expected.

As another example, this Nokia 7250 IXR-e  chassis has a `state/description` of "250 IXR-10e" and a `state/part-no` of "3HE16997AARB01" (and no `state/model-name`). I would like for us to add some form of this "250-ixr-10e"  to the model-name identity so that I could use `state/model-name` path to determine which chassis this is (rather than parsing `state/description`). 

In summary, I would like `state/model-name` to solve 2 problems:

1. Provide a consistent location for the model name (some vendors use description, some use part-no, etc.) so that I only have to use one path.
2. Define what values we expect to see (using the identity) so that I don't have to parse any values.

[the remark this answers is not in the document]
LimeHat: I understand the idea, I just find it confusing to call something a part number and then proceed to say that it is not, in fact, a part number.

```In this example, the part-no is referencing some part+revision number which is internal to the vendor hardware revision. This number might change even when the SKU/orderable part does not change (for example due to minor hardware revisions that do not affect the feature set).```

That is true, however, only the revision changes. In our case, for example, the orderable part is the part number minus the revision (`3HE16997AA` in Sydney's example).

I'm not debating the leaf's overall meaning; I just think the new description is less clear than before.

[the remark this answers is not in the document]
Comment (SydneyCaulfeild) [reply release/models/platform/openconfig-platform.yang]: Ok yes I see your point. I have updated the description so that it doesn't say that model-name is the only one that could be searched online and I no longer call it a part number.

Do you have any other suggestions on what I could add to help explain the difference between part-no and model-name? I understand that both could be used to identify a component and potentially to search one up, but am not sure how else to explain that part-no is more often a "random" string whereas model-name is more of a "name" that has more meaning to someone who is reading it without a lot of background.

**Review by dplore — COMMENTED** [review]
some wordsmithing suggestions

File: release/models/platform/openconfig-platform.yang [hunk 2]
```diff
@@ -540,11 +546,13 @@ module openconfig-platform {
     }
 
     leaf model-name {
-      type string;
-      description
-        "A human readable string describing the model of a component.
-        This string is optional and should only be populated if part-no
-        is also populated.";
+      when "../removable = 'true'" {
```
earies, line 549: The type stmt must be defined outside of the when stmt - this conditional stmt has to be defined as a sibling of the leafs substatements (and can carry it's own separate description substatement)

SydneyCaulfeild: Oh I see, thanks!

release/models/platform/openconfig-platform.yang
```diff
@@ -540,11 +546,13 @@ module openconfig-platform {
...
     leaf model-name {
+      when "../removable = 'true'";
       type string;
       description
-        "A human readable string describing the model of a component.
-        This string is optional and should only be populated if part-no
-        is also populated.";
+        "This is the component name that would be found in a catalog of 
+        stock keeping units (SKU) and should be the orderable name of
+        the component. This is required for all FRU (field replaceable unit)
```
earies, line 554: Note: due to the conditional when statement, this leaf will _only_ be available for components with `removable=true` and never set for any other component.  I think you are just stating here that this leaf is mandatory in the description for such cases vs. setting `mandatory true;` to enforce.  Any reason not to set the `mandatory` stmt here? (This will then _only_ be mandatory when removable=true and not mandatory otherwise)

Nits: 
* `"This is the component name..."` sounds a bit confusing since this terminology is used for the actual key name itself.  Any reason you didn't just use `model name` here?
* For the path `s/TRUE/true/`
* Another confusing point is the there is a catchall identity today for the `type` named `FRU`.  I recall we spoke about this in a PR long back as to it's potential removal but it is likely there are various implementations where the `type` identities is not fully comprehensive thus folks may set to `FRU` with the `removable` leaf set to `false` (not really removable but no other identity suits without creating vast augments).  Something we should probably clean up in a separate PR to shape this and reduce further confusion

SydneyCaulfeild: Thanks for the feedback! I have set mandatory true so that it is mandatory whenever removable is true and is optional otherwise (at least I believe that is what my implementation is now doing).

@earies what do you mean by your comment "For the path s/True/true/"?

I decided to remove the comment about it being required for FRUs from the description because I believe making it mandatory when removable is true achieves the same thing, and as you mentioned might be more accurate since the type FRU is sometimes "misused". @dplore does this sound ok to you to use removable=true instead of type=FRU?

dplore:
model-name as a mandatory field when removable = true is the intent.  I think this is good.  It's a new precedent in OC as the few places we use mandatory are not conditional.  

I agree that `removable` is the better criteria than component type FRU which is overloaded and in practice couldn't be used as a conditional in practice because un-orderable parts are often represented using FRU as Ebben has noted.

dplore:
> model-name as a mandatory field when removable = true is the intent. I think this is good. It's a new precedent in OC as the few places we use mandatory are not conditional.

Well, looks like our tool-chain in general does not support the mandatory field inside a where clause.  (See the CI checks). I think a description stating this is mandatory when the component is 'removable' will need to suffice.

release/models/platform/openconfig-platform.yang
earies: > @earies what do you mean by your comment "For the path s/True/true/"?

The lexical representation for a boolean is represented as lower case true/false is all I mean here

https://datatracker.ietf.org/doc/html/rfc6020#section-9.5.1

earies:
> > model-name as a mandatory field when removable = true is the intent. I think this is good. It's a new precedent in OC as the few places we use mandatory are not conditional.
> 
> Well, looks like our tool-chain in general does not support the mandatory field inside a where clause. (See the CI checks). I think a description stating this is mandatory when the component is 'removable' will need to suffice.

This is failing because as-is currently is invalid YANG.  A `mandatory` stmt cannot be a substatment of a `when` clause and must exist as a sibling.  I believe when this is fixed, the toolchain should accept this as it is no different than other usage to date.

e.g.

```
 leaf model-name {
   type string;
   when "../removable = 'true'";
   mandatory true;
 }
```

SydneyCaulfeild:
Ok I have updated the mandatory stmt to not be nested in the when - @earies just want to confirm though that this leaf is still be optional even when ../removable is false?

Also, does anyone know why in [openconfig-vlan.yang](https://github.com/openconfig/public/blob/master/release/models/vlan/openconfig-vlan.yang#L185) I see an example of something nested in the when? Is it because descriptions are allowed in the when statement but not a mandatory?

earies:
> Ok I have updated the mandatory stmt to not be nested in the when - @earies just want to confirm though that this leaf is still be optional even when ../removable is false?

If `removable=false` then the `model-name` leaf is not permitted at all given the current logic

The following is permitted

```
<components xmlns="http://openconfig.net/yang/platform">
  <component>
    <name>C1</name>
    <removable>false</removable>
  </component>
  <component>
    <name>C2</name>
    <removable>true</removable>
    <model-name>M1</model-name>
  </component>
</components>
```

But either of the following is not

```
<components xmlns="http://openconfig.net/yang/platform">
  <component>
    <name>C1</name>
    <removable>true</removable>
  </component>
  <component>
    <name>C2</name>
    <removable>false</removable>
    <model-name>M1</model-name>
  </component>
</components>
```

Example yanglint validation on the latter:

```
libyang err : When condition "../removable = 'true'" not satisfied. (/openconfig-platform:components/component[name='C2']/model-name)
libyang err : Mandatory node "model-name" instance does not exist. (/openconfig-platform:components/component[name='C1'])
```

* If not removable then model-name is not permitted whatsoever, it is not optional
* If removable, then model-name is mandatory

> Also, does anyone know why in [openconfig-vlan.yang](https://github.com/openconfig/public/blob/master/release/models/vlan/openconfig-vlan.yang#L185) I see an example of something nested in the when? Is it because descriptions are allowed in the when statement but not a mandatory?

See the grammar defined in RFC 6020.  Only the following substatements are permitted and optional.

```
  when-stmt           = when-keyword sep string optsep
                           (";" /
                            "{" stmtsep
                                ;; these stmts can appear in any order
                                [description-stmt stmtsep]
                                [reference-stmt stmtsep]
                             "}")
```

release/models/platform/openconfig-platform.yang
```diff
@@ -540,11 +546,13 @@ module openconfig-platform {
     }
 
     leaf model-name {
+      when "../removable = 'true' or ../type = 'CHASSIS'";
```
dplore, line 549: Need to prefix the CHASSIS component type

SydneyCaulfeild: Done, thanks!

release/models/platform/openconfig-platform.yang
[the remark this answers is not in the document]
dplore: I think the latest description resolves the ambiguity for model-name.

dplore:
This was reviewed in the OC Community meeting on Sep 12, 2024 without objection.  Setting last call to Sep 13, 2024.

dplore (approved):

PR #1161, opened 2024-08-08
Add identity for model-name.

SydneyCaulfeild:
### Change Scope

* Add a MODEL_NAME identity that will be used as the type for the model-name leaf.

### DIFF
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -20,7 +20,13 @@ submodule openconfig-platform-common {
     "This modules contains common groupings that are used in multiple
     components within the platform module.";
 
-  oc-ext:openconfig-version "0.27.0";
+  oc-ext:openconfig-version "0.28.0";
+
+  revision "2024-08-08" {
+    description
+      "Update type of model-name leaf from string to MODEL_NAME identity.";
+    reference "0.28.0";
+  }
 
   revision "2024-05-29" {
     description
--- a/release/models/platform/openconfig-platform-types.yang
+++ b/release/models/platform/openconfig-platform-types.yang
@@ -22,8 +22,13 @@ module openconfig-platform-types {
     "This module defines data types (e.g., YANG identities)
     to support the OpenConfig component inventory model.";
 
-  oc-ext:openconfig-version "1.8.0";
+  oc-ext:openconfig-version "1.9.0";
 
+  revision "2024-08-08" {
+    description
+      "Add MODEL_NAME identity";
+    reference "1.9.0";
+  }
 
   revision "2024-04-30" {
     description
@@ -285,6 +290,11 @@ module openconfig-platform-types {
   }
 
   // identity statements
+  identity MODEL_NAME {
+    description
+      "Base identity for hardware model name.";
+  }
+
   identity OPENCONFIG_HARDWARE_COMPONENT {
     description
       "Base identity for hardware related components in a managed
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,7 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.27.0";
+  oc-ext:openconfig-version "0.28.0";
+
+  revision "2024-08-08" {
+    description
+      "Update type of model-name leaf from string to MODEL_NAME identity.";
+    reference "0.28.0";
+  }
 
   revision "2024-05-29" {
     description
@@ -540,7 +546,9 @@ module openconfig-platform {
     }
 
     leaf model-name {
-      type string;
+      type identityref {
+        base oc-platform-types:MODEL_NAME;
+      }
       description
         "A human readable string describing the model of a component.
         This string is optional and should only be populated if part-no


LimeHat:
How the identities are defined? Is the expectation that each vendor creates its own augmentation?

earies:
> How the identities are defined? Is the expectation that each vendor creates its own augmentation?

+1 to this question.  I'd like to understand the problem statement with this being a free-form string in reality at the implementations discretion.

If this is now an identityref, the only method to cover some complete list now involves augments across all hardware models on all implementations that every consumer needs to absorb and handle should there be some strict validation.  Any such augments should in reality also follow OC style but could do whatever they wish and breach the boundaries of the YANG language restrictions within OC (general augment problem)

This introduces more complexity in reality and it's a design pattern that you could question all over the model set.  I get the reason to want to enforce a better contract but I think leaving this as a (mostly) unbounded string is probably best

rgwilton:
I'm pretty much on the same page as @earies here.  I think that using identities over a string would probably be quite painful for implementations to manage across different platforms/etc.

I guess that the advantage of using identities is that it effectively forces an implementation to publish these ids up front (presumably they make the YANG available) rather than only being able to discover these at runtime.  But I'm still not sure whether it would really help or just creates a lot of extra work for perhaps relatively little benefit.

nleiva:
This is very timely. I've been exploring avenues to work with hardware module names. While I would love to have them well documented in a model, the complexity of naming conventions across vendors (*) poses challenges, so I see how this can get quickly out of control.

Somehow related, we are currently working on a project to map the **string** linecard model-name to relevant configuration information, such as its interface naming convention. For more details, see: https://github.com/openconfig/entity-naming/issues/33


(*)
**Juniper**: Flexible PIC Concentrators (**FPC**s) and Physical Interface Cards (**PIC**s), or  Modular Port Concentrators (**MPC**s) and Modular Interface Cards (**MIC**s), etc.
- MPC: MPC5EQ-40G10G, MPC7E-MRATE, MPC2E-3D-NG, MPC5E-100G10G, etc.
- MIC: MIC-3D-4OC3OC12-1OC48, MIC-3D-20GE-SFP, MIC-3D-20GE-SFP-E, MIC-3D-2XGE-XFP, MIC-3D-4XGE-XFP, MIC-3D-20GE-SFP, MIC3-3D-1X100GE-CFP, etc. (FRU model number)

**Nokia**: The Input/Output module (**IOM**) and Media Dependent Adapters (**MDA**s).
- IOM: IOM-E, IOM3-XP, IOM3-XP-B, IOM3-XP-C, IOM4-e,IOM4-e-B, etc. 
- MDA: me10-10gb-sfp+, me40-1gb-csfp, me10-10gb-sfp+, me40-1gb-csfp, m60-10/100eth-tx, c8-10/100eth-tx, etc. (Provisioned Type)

...

dplore:
Thanks for the comments @earies, @rgwilton and to @nleiva for the examples on the different patterns.  

There was some discussion on this in the Aug 13, 2024 operators meeting where concerns about the giant size of model-name identities given the breadth and long history of product catalogs make enumerating the list a burden of maintenance and bloat of the model.  It's not clear there is an operational use case to have the model-names as an identity vs. runtime discovery.     

@SydneyCaulfeild I realize this draft PR is to explore these questions.  Did you have a particular operational use case where having an identity is a significant advantage over runtime discovery?

SydneyCaulfeild:
I can see how trying to maintain a list of model-names for all devices would be challenging. 

An example use case to demonstrate the benefit of an identity instead of a string was that when I generate devpaths (i.e. unique identifiers) for a chassis component using the device's telemetry, I use the `state/model-name` value to determine which device it is. If vendors had to publish the model-name they are going to use in an identity, it would allow me to 1) perhaps be aware of what value they plan on using earlier rather than waiting until we have a device to check what is in the path and 2) use the identity as the "source of truth" so that I know what value they will be using and more importantly that it is not going to change.

@dplore If we're not going to use an identity, could I document somewhere that we would need to be notified if the model-name string is going to change (or maybe update the definition to say that it should not change)? If the chassis model-name string changes it would break my devpath generation for all components.

Based on the comments above though it sounds like using an identity would not be worth it, which I understand. @robshakir using an identity for model-name was brought up in a meeting we had in early August. Did you have any other use cases to support using an identity over a string? If not, are you ok if we leave this as a string?

earies:
> I can see how trying to maintain a list of model-names for all devices would be challenging.
> 
> An example use case to demonstrate the benefit of an identity instead of a string was that when I generate devpaths (i.e. unique identifiers) for a chassis component using the device's telemetry, I use the `state/model-name` value to determine which device it is. If vendors had to publish the model-name they are going to use in an identity, it would allow me to 1) perhaps be aware of what value they plan on using earlier rather than waiting until we have a device to check what is in the path and 2) use the identity as the "source of truth" so that I know what value they will be using and more importantly that it is not going to change.
> 
> @dplore If we're not going to use an identity, could I document somewhere that we would need to be notified if the model-name string is going to change (or maybe update the definition to say that it should not change)? If the chassis model-name string changes it would break my devpath generation for all components.
> 
> Based on the comments above though it sounds like using an identity would not be worth it, which I understand. @robshakir using an identity for model-name was brought up in a meeting we had in early August. Did you have any other use cases to support using an identity over a string? If not, are you ok if we leave this as a string?

I see where you are going with this as this creates a nice firm contract however I'd question how practical this is in reality and the gain vs. pain involved.

This problem statement can apply to many areas of the model-set across many model-sets where it is potentially unmanageable (models in need of constantly being updated), extensibility must be accounted for thus augments must be accepted (does everyone want to accept augments and deviations?) and not all values can be easily describable in similar fashion (model names I'd agree may be slightly more manageable to come up with but still what I consider a small change to a model like this w/ a potential for large impact)

The OC model-set is generally very loose and forgiving in definition which in reality just shifts the problems around.  There are many spots where things can be firmed up but it's generally left up to implementation behavior to go enforce what is not described (leading to divergence often)

The other option is to draw correlation in that values at some path (say a string) should come from a dictionary defined at some other path not bound by the YANG type (like a catalog, features, capabilities, etc.. - e.g. a loose out-of-band or annotated relationship).  As an identityref with strict validation, a new version of a model needs to be published for any updates.  The looser the type, the more flexibility to iterate at runtime and accept things outside the boundaries however your example refers to codegen which may be stricter.

> and more importantly that it is not going to change.

We often talk about model node/structure backwards incompatibilities but rarely talk about the impact of values changing especially if those values are something that could be used as a cross reference, key or just ultimately a dependence being built - this is mainly your point on consistency and predictability.

The reality is that change does occur for a variety of reasons - some good, some bad w/ varying effects so I think this is more about the pros and cons associated with being too strict vs. too loose and the overall impact to the community and implementors.

Am not discrediting the overall problem statement though as having an enumerated list of all possible options for a value is desirable - esp. when it comes to hardware model names.  YMMV as speaking on behalf of 1 implementation right now, there is often the dilemma to try to describe something as an enum or identity but the total amount of options may be too large or too dynamic in nature that falling back to a string is the best option overall.

SydneyCaulfeild:
Thank you all for your feedback! Based on the consistent feedback that changing this to an identity would not be worth the pain of maintaining a list of hardware models, we can leave this as a string. We will instead send requests to communicate what we are expecting the model-name value to be so that we are on the same page with vendors.

Here is an example of how the model-name would be used for the chassis component on a Nokia 7250 IXR-10e device:

* state/part-no is currently "3HE16997AARB01"
* state/description is currently "7250 IXR-10e"
* we would expect to see "7250 IXR-10e" (or something similar) in state/model-name since this leaf is meant to be the consistent location to put the hardware model.

PR #1165, opened 2024-08-15
Refactor non-compliant/overlapping YANG definition for BGP MED

earies:
    - Remove string type from bgp-set-med-type union
    - Add new bgp-set-med-action typedef to separate out action on MED
      values
    - Add new set-med-action leaf to be optionally utilized with set-med

### Change Scope

YANG 1.0 (RFC 6020) as well as YANG 1.1 (RFC 7950) specify that the legal
lexical representation for built-in integer types can be represented with an
optional sign.

https://www.rfc-editor.org/rfc/rfc6020.html#section-9.2.1

```
   An integer value is lexically represented as an optional sign ("+" or
   "-"), followed by a sequence of decimal digits.  If no sign is
   specified, "+" is assumed.
```

Example:

```
$ cat main.yang
module main {
  yang-version "1";
  namespace urn:m;
  prefix m;

  leaf value {
    type uint64;
  }
}

$ cat main.xml
<value xmlns="urn:m">+100</value>

$ yanglint -f xml main.xml main.yang
<value xmlns="urn:m">100</value>
```

The current union as specified interleaves a built-in integer type with a
string pattern regex that just so happens to overlap with allowable characters
for the uint32 type.

```
  type union {
    type uint32;
    type string {
      pattern '[+-][0-9]+';
      oc-ext:posix-pattern '^[+-][0-9]+$';
    }
    type enumeration {
      enum IGP {
        description "set the MED value to the IGP cost toward the
        next hop for the route";
      }
    }
  }
```

Therefore a "+100" would be matched by the `uint32` and the "+" character
dropped.  Since the integer is unsigned in this case, a "-100" would
fallthrough to the 2nd match.

Rather than encode the "action" and value into a string, this proposal suggests
separating out the action to a distinct optional node.  It appears that this is
the only such usage where there is a string with a pattern statement that
overlaps that of another permissible type encoding when also used in a union
with that type.

### Platform Implementations

This is a NBC however necessary to fix to remove ambiguity and improper handling

### DIFF
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -30,10 +30,10 @@ module openconfig-bgp-policy {
 
   oc-ext:openconfig-version "8.0.0";
 
-  revision "2024-08-15" {
+  revision "2024-08-23" {
     description
-      "Separate the ability to set the BGP MED into a distinct value
-      with optional action.";
+      "Separate the ability to set the BGP MED along with an appropriate
+      action.";
     reference "8.0.0";
   }
 
@@ -228,14 +228,17 @@ module openconfig-bgp-policy {
     description
       "Type definition for specifying how the BGP MED can be set in BGP
       policy actions.  The MED can be specified as a direct integer
-      value or setting it to the IGP cost.  When set to the integer
-      value without a MED 'action', the value will be explicitly set.
-      If a MED 'action' is defined, the previous value will be modified
-      according to the action (e.g. ADD or SUBTRACT).";
+      value or setting it to the IGP cost.  To be used in conjunction
+      with `bgp-set-med-action` in order to specify the appropriate
+      action to take on this value.";
   }
 
   typedef bgp-set-med-action {
     type enumeration {
+      enum SET {
+        description
+          "Action to set the MED to a specific value.";
+      }
       enum ADD {
         description
           "Action to increment the previous MED value.";
@@ -246,8 +249,8 @@ module openconfig-bgp-policy {
       }
     }
     description
-      "If specified, this action will alter a previous MED value by the
-      value defined by the bgp-set-med-type.";
+      "Specifies which action to take on the value specified by the
+      bgp-set-med-type.";
   }
 
   // grouping statements
@@ -1366,17 +1369,24 @@ module openconfig-bgp-policy {
 
     leaf set-med {
       type bgp-set-med-type;
+      must "../set-med-action" {
+        error-message
+          "set-med cannot be specified without a valid set-med-action";
+      }
       description
-        "Set the MED metric attribute in the route update.  Can be used
-        in conjunction with 'set-med-action' in order to alter a
-        previous value by the value specified here.";
+        "Set the MED metric attribute in the route update.  When set, a
+        valid `set-med-action` must be specified.";
     }
 
     leaf set-med-action {
       type bgp-set-med-action;
+      must "../set-med" {
+        error-message
+          "set-med-action cannot be specified without a set-med value";
+      }
       description
-        "Optionally used in conjunction with 'set-med' to alter a
-        previous value.";
+        "When set-med is specified, this leaf is mandatory to set the
+        appropriate action on the MED metric value.";
     }
   }
 
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -32,7 +32,8 @@ module openconfig-bgp-policy {
 
   revision "2024-08-15" {
     description
-      "";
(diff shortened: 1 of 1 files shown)


dplore:
This will be reviewed at the Aug 20, 2024 OC operators meeting

release/models/bgp/openconfig-bgp-policy.yang
```diff
@@ -210,23 +217,37 @@ module openconfig-bgp-policy {
...
+  typedef bgp-set-med-action {
+    type enumeration {
+      enum ADD {
+        description
+          "Action to increment the previous MED value.";
+      }
+      enum SUBTRACT {
+        description
+          "Action to decrement the previous MED value.";
+      }
```
dplore, line 246: I think what you have here works, but it would be more explicit if we add an enum `SET` or similar to indicate to change the value of med to be the integer specified (instead of add or subtract).  This way the bgp-set-med-action is always relevant when `set-med` is a uint32 value.

earies: We can very well enforce this new action leaf by doing such.  The current proposal was assuming if only `set-med` is set then it is an implicit "set" but if this leaf is specified in addition then the behavior changes to add/subtract.

If we add another enum variant here then it likely makes sense to gate a `when` statement as well otherwise we leave some undefined implementation behavior on the table

dplore: After thinking about this a few days  I do prefer the extending the enum and as you suggest, adding a `when` constraint because the relationship between the enum and the leaf are explicit (and still rather simple) in yang .  

Would you like to make this change?

earies: Updated via commit: https://github.com/openconfig/public/pull/1165/commits/3d1aba4f520ddd6f8481f0751653a78837d5edc4 to add an explicit `SET` action along with constraints between the 2 leafs.

Example stub validation (removal of either leaf):
```
<bgp-actions xmlns="http://openconfig.net/yang/bgp-policy">
  <config>
    <set-med>100</set-med>
    <set-med-action>ADD</set-med-action>
  </config>
</bgp-actions>

libyang err : set-med-action cannot be specified without a set-med value (/openconfig-bgp-policy:bgp-actions/config/set-med-action)

libyang err : set-med cannot be specified without a valid set-med-action (/openconfig-bgp-policy:bgp-actions/config/set-med)
```

NOTE: When `IGP` is specified, this now enables the ability to apply any action in its current state which is likely fine

dplore: Thanks for adding this!  

We reviewed again in Aug 27, 2024 operators meeting.  To make this backwards compatible as possible, the typedef bgp-set-med-action should have a default of 'SET'.     This way if someone is currently specifying an integer as the 'bgp-set-med-type', then we'll get the same behavior as today.

earies: I've been a little hesitant on the approach of using defaults in this fashion and would like to hear folks thoughts as this could be applied in a number of areas and there should likely be a bit of focus and guidance on default handling in general outside the scope of this PR.

On one hand, the approach suggested could provide a fraction of a transition path for a previous configuration.  The case where a precise positive value was specified only without prefix.  e.g. `200`

What would not transition whatsoever is anything that previously contained a `+/-` prefix even if the implementation skirted around this previously.

A few other thoughts

* If a default were set on `set-med-action` then there would no longer be any need for the `must` statements as `set-med-action` would always be present.  Since this value is always set, it can then exist w/o any value specified as well.
* When defaults are set in this fashion, we have an implicit behavior vs. enforcing explicit behavior.  The backwards incompatibilities introduced really needs to be understood and assessed imo - anything else may be luck with undesirable results.  A previous `+200` may have been treated one way previously but now still accepted w/ alternate behavior that needs explicit setting of the "action" node.
* Defaults in general tend to lack behavioral expectations with protocols such as gNMI (retrieval).  This is something already covered in NETCONF/RESTCONF.  A case of this is the expectations surrounding a "get" to the config branch where defaults are conveyed.  Today this should likely never return defaults but rather the mirrored state node would be emitted.

In addition to above regarding defaults is behavioral expectations surrounding what led to this PR.  If a value on a int type is set as `+/-` prefixed, the implementation really should report back any config/state as non `+` prefixed in the case of positive (dropping the prefix) or `-` prefixed (retained) in the case of negative.

dplore (approved):

release/models/bgp/openconfig-bgp-policy.yang
dplore: I am ok with leaving out a default.  When discussing with others offline I was also reminded of our bias to avoid defaults and prefer leaves being set explicitly by a caller.

PR #1174, opened 2024-08-29
Update reference to current IETF DRAFT for link-bandwidth community.

rszarecki:
The new revision (-08) of draft-ietf-idr-link-bandwidth is published by IETF. This new revision do not make any changes but "refresh" draft to be active. The draft-ietf-idr-link-bandwidth-07 expierd on September 6, 2018. This is cosmetic change

### DIFF
--- a/release/models/bgp/openconfig-bgp-types.yang
+++ b/release/models/bgp/openconfig-bgp-types.yang
@@ -642,7 +642,7 @@ module openconfig-bgp-types {
         - color:<CO bits>:<4b value> per draft-ietf-idr-segment-routing-te-policy
           section 3
         - link-bandwidth:<2 byte asn>:<bandwidth_value> per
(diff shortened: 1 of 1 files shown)


dplore:
This change is ok to make.  You'll need to bump up the version number to 6.0.1 and add a revision statement here: https://github.com/openconfig/public/pull/1174/files#diff-32e3ff8896cce4fbd6a6920426d642a6dfbac07a89adddeaf3990847534e50aeR27

dplore (commented):
@rszarecki ,  LGTM, but the version needs to be updated:

```
openconfig-bgp-types.yang: file updated but openconfig-version string not updated: "6.0.0"
```

dplore:
Reviewed in Dec 10, 2024 OC Operators meeting.  There is rough consensus that bumping the version number is not necessary.  We should only update if there is a change we want to adopt/integrate into the OC models.

PR #1176, opened 2024-09-05
Add fallback leaf for LACP in OC model

sallylsy:
### Change Scope

* Add a new leaf `fallback` under both config and state subtree if lacp interface
* The is a new leaf, and has no reference on any current leaves.
```
module: openconfig-lacp
  +--rw lacp
     +--rw interfaces
        +--rw interface* [name]
           +--rw config
           |  +--rw fallback?                    boolean
           +--ro state
           |  +--ro fallback?                    boolean
```
### Platform Implementations

 * SONIC HLD: https://github.com/sonic-net/SONiC/blob/18de1c1f71d1be495852fe57263bc5aaf56e1d42/doc/lag/LACP%20Fallback%20Feature%20for%20SONiC_v0.5.md?plain=1#L4
 * SONIC Implementation: https://github.com/sonic-net/sonic-swss/blob/42ea859ce982b21ff40c78c92a5ab21aafb52fad/cfgmgr/teammgr.cpp#L270
 * Cisco: https://www.cisco.com/c/en/us/td/docs/iosxr/ncs560/interfaces/72x/b-interfaces-hardware-component-cg-72x-ncs560/configuring_link_bundling.html#task_844BB227554D4C42A96F1AF8D5AC477C
 * Arista: https://arista.my.site.com/AristaCommunity/s/article/configuring-port-channel-lacp-fallback-on-arista-switches-2

### DIFF
--- a/release/models/lacp/openconfig-lacp.yang
+++ b/release/models/lacp/openconfig-lacp.yang
@@ -418,7 +418,8 @@ grouping aggregation-lacp-members-statistics {
       description
         "If the fallback is set to true, current LACP interface is
         able to establish a Link Aggregation (LAG) before it receives
-        LACP PDUs from its peer";
+        LACP PDUs from its peer, and fallback to a single port active
+        after the expiry of the timeout period.";
     }
 
     uses aggregation-lacp-global-config;
--- a/release/models/lacp/openconfig-lacp.yang
+++ b/release/models/lacp/openconfig-lacp.yang
@@ -26,7 +26,13 @@ module openconfig-lacp {
     managing aggregate interfaces.   It works in conjunction with
     the OpenConfig interfaces and aggregate interfaces models.";
 
-  oc-ext:openconfig-version "2.0.0";
+  oc-ext:openconfig-version "2.1.0";
+
+  revision "2024-09-24" {
+    description
+      "Add LACP fallback leaf under both config and state.";
+    reference "2.1.0";
+  }
 
   revision "2023-12-11" {
     description
@@ -405,6 +411,14 @@ grouping aggregation-lacp-members-statistics {
         "The MAC address portion of the node's System ID. This is
         combined with the system priority to construct the 8-octet
         system-id";
+    }
+
+    leaf fallback {
+      type boolean;
+      description
+        "If the fallback is set to true, current LACP interface is
+        able to establish a Link Aggregation (LAG) before it receives
+        LACP PDUs from its peer";
     }
 
     uses aggregation-lacp-global-config;


dplore:
Looks good to me. I set last-call to September 18, 2024.

robshakir:
Looks reasonable to me. Thanks for the references.

Do we also want to add a configurable timeout here (given that multiple implementations support it)?

dplore:
> Looks reasonable to me. Thanks for the references.
> 
> Do we also want to add a configurable timeout here (given that multiple implementations support it)?

There is a timeout in OC for lacp, but it is constrained to IEEE defined values, whereas it seems at least Cisco and Arista (refs above) allow a custom value.

https://github.com/openconfig/public/blob/6b19efbb170259b6fda5e1faa43188cb8a98337c/release/models/lacp/openconfig-lacp.yang#L170

@sallylsy do you need to set the timeout to something other than LONG/SHORT as supported today?

sallylsy:
Discussed with Darren offline. Cisco reference didn't mention timeout, only Arista did. We should be good.

sallylsy:
> > Looks reasonable to me. Thanks for the references.
> > Do we also want to add a configurable timeout here (given that multiple implementations support it)?
> 
> There is a timeout in OC for lacp, but it is constrained to IEEE defined values, whereas it seems at least Cisco and Arista (refs above) allow a custom value.
> 
> https://github.com/openconfig/public/blob/6b19efbb170259b6fda5e1faa43188cb8a98337c/release/models/lacp/openconfig-lacp.yang#L170
> 
> @sallylsy do you need to set the timeout to something other than LONG/SHORT as supported today?

I don't need additional timeout.

dplore:
Bumping last call date to Sep 13, 2024 as this is urgent.

release/models/lacp/openconfig-lacp.yang
```diff
@@ -407,6 +413,15 @@ grouping aggregation-lacp-members-statistics {
         system-id";
     }
 
+    leaf fallback {
+      type boolean;
+      description
+        "If the fallback is set to true, current LACP interface is
+        able to establish a Link Aggregation (LAG) before it receives
+        LACP PDUs from its peer, and fallback to a single port active
+        after the expiry of the timeout period.";
```
sallylsy, line 422: IEEE or IETF references would be better

dplore:
Reviewed in Sep 10, 2024 OC operators meeting with no objection.  If there is a standards or other non-vendor reference for this feature, that would be nice to include in a yang reference statement.  This is a non-blocking request.

release/models/lacp/openconfig-lacp.yang
dplore: I did some more searching and the fallback option does not appear to be documented in IEEE or other standards organization.  It is widely published as an option in device implementations.  I'll mark this resolved.

dplore (approved):

Issue #945, opened 2023-08-21
How to disable sending BGP communities at group and neighbor level


dplore:
Sorry to comment after this was merged, but I am realizing that the NONE option was deprecated by this change. I don't believe that was a good decision.

BGP configuration is based on a model of inheritance where neighbor settings override group settings, which in turn override global settings. Suppose the group or global configuration specifies that standard, extended and large communities should all be transmittable. The configuration of a specific neighbor should be able to override this and make no communities transmittable to that peer. But without a 'NONE' option, how do you do this? A send-community-type leaf-list with no elements is equivalent to no configuration of send-community-type at all, which should mean that the configuration is inherited from a higher context -- but clearly that's not what we want.

TLDR - it should not be assumed that an empty leaf-list is equivalent to 'NONE'.

_Originally posted by @nokia1adam in https://github.com/openconfig/public/issues/852#issuecomment-1684261880_

dplore:
The use case is to send communities at the global level and disable/not send communities for a specific peer-group or neighbor.  #852 unintentionally deprecated this ability.

If we like the leaf-list of enum design, then we could add  a 'send-community-enabled' boolean at  the [bgp global level](https://github.com/openconfig/public/blob/face342aaf2f65988b1bbd21a08faee72ec0d736/release/models/bgp/openconfig-bgp-common-multiprotocol.yang#L223) and at the [group/neighbor  level](https://github.com/openconfig/public/blob/face342aaf2f65988b1bbd21a08faee72ec0d736/release/models/bgp/openconfig-bgp-common.yang#L297).  The global could have a default of false and the group/neighbor levels would not have a default (so inheritance is preserved).

earies:
Taking a step back here, it seems this problem statement isn't unique to BGP communities (or even BGP) but rather configuration hierarchies, groupings and expected inheritance across years of implementation to template, reuse, override partial/all parent behavior, etc..

This seems like a wider problem statement that needs to be dug into vs. just a leaf-by-leaf behavior no?

jhaas-pfrc:
> Taking a step back here, it seems this problem statement isn't unique to BGP communities (or even BGP) but rather configuration hierarchies, groupings and expected inheritance across years of implementation to template, reuse, override partial/all parent behavior, etc..
> 
> This seems like a wider problem statement that needs to be dug into vs. just a leaf-by-leaf behavior no?

[...]

PR #1177, opened 2024-09-06
Restore NONE as an enum value for typedef community-type

dplore:
Proposed fix for https://github.com/openconfig/public/issues/945

### Change Scope
In the current OC models, there is ambiguity regarding how to use send-community-type to disable sending any communities at the bgp global, peer-group and neighbor levels.  The OC model description says the lack of the `send-community-type` node means communities should not be sent.  This is in conflict with the expected inheritance from global, peer-group and neighbor levels in BGP configuration.   With inheritance, the lack of a node means to inherit the configuration from the next higher level.

It was proposed in #945 that an explicitly defined send-community-type as an empty list could be used to indicate no communities should be sent.  But this is problematic for some encodings and likely toolchains because [RFC7590](https://datatracker.ietf.org/doc/html/rfc7950) states that XML encoded yang simply omits empty lists ([ref](https://github.com/openconfig/public/issues/945#issuecomment-2311652367)).

This proposal removes the deprecated status from the `NONE` value of the `community-type` enum.

### DIFF
--- a/release/models/bgp/openconfig-bgp-types.yang
+++ b/release/models/bgp/openconfig-bgp-types.yang
@@ -24,7 +24,13 @@ module openconfig-bgp-types {
     policy. It can be imported by modules that make use of BGP
     attributes";
 
-  oc-ext:openconfig-version "6.0.0";
+  oc-ext:openconfig-version "6.1.0";
+
+  revision "2024-09-06" {
+    description
+      "Restore NONE enum value for community-type.";
+    reference "6.1.0";
+  }
 
   revision "2024-02-01" {
     description
@@ -795,10 +801,8 @@ module openconfig-bgp-types {
       }
       enum NONE {
         description
-            "Do not send any community attribute.
-            This value has been deprecated because the node is now
-            a leaf-list.";
-        status deprecated;
+            "Do not send any community attribute. If this value is present
+            then no communities will be sent.";
       }
     }
     description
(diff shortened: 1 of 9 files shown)


release/models/bgp/openconfig-bgp-types.yang
```diff
@@ -795,10 +801,8 @@ module openconfig-bgp-types {
       }
       enum NONE {
         description
-            "Do not send any community attribute.
-            This value has been deprecated because the node is now
-            a leaf-list.";
-        status deprecated;
+            "Do not send any community attribute. If this value is present
+            then no communities will be sent.";
```
rszarecki, line 805: This value must not coexist with any other value in same community type list.
[NONE STANDARD] is ambiguous.

rszarecki (changes requested):

release/models/bgp/openconfig-bgp-types.yang
dplore: Done

s19nal (approved):
as discussed at OC operators meeting - removing a deprecated status from one enum in the leaf looks to be ok although it might be slightly contentious to call it non-breaking.
However re-instating the deprecated NONE enum solves the problem that was introduced by removing it in the first place, so I'm happy :)

nokia1adam:
@dplore, thanks for resolving this issue.

PR #1179, opened 2024-09-09
Clarify last-reboot-time is updated during a power-on event

### LINKED ISSUE #1171: Clarification on last-reboot-time, boot-time and up-time
- last-reboot-time:
> This reports the time of the last reboot of the component

Does this indicate the time at which the reboot was initiated by the user/ the component going down? or does it indicate the time at which the component came back up to oper-status ACTIVE?

- boot-time:
> This timestamp indicates the time that the system was last restarted.  

- up-time: 
> The amount of time since the network operating system was initialized

What is the difference between restarted and rebooted in this context? Linux boot-time usually refers to /proc/stat btime. Is that a valid for the boot-time leaf? For platforms with multiple components like Line Cards and Fabric Cards, (with and without CPU), there can be different boot-time/up-time per component. Would it be possible to have enhance ocplatform to contain a component-boot-time/component-up-time for individual components?

dplore:
Fixes #1171 

### Tree view
We are at `/components/component/state`
```diff
         |  +--ro last-poweroff-reason
         |  |  +--ro trigger?   component-last-poweroff-reason-trigger
         |  |  +--ro details?   string
         |  +--ro last-poweroff-time?             oc-types:timeticks64
         |  +--ro last-switchover-reason
         |  |  +--ro trigger?   component-redundant-role-switchover-reason-trigger
         |  |  +--ro details?   string
         |  +--ro last-switchover-time?           oc-types:timeticks64
         |  +--ro last-reboot-reason?             identityref
-        |  +--ro last-reboot-time?               oc-types:timeticks64
+        |  x--ro last-reboot-time?               oc-types:timeticks64
+        |  +--ro boot-time?                      oc-types:timeticks64
         |  +--ro switchover-ready?               boolean
```

### DIFF
--- a/release/models/platform/openconfig-platform-common.yang
+++ b/release/models/platform/openconfig-platform-common.yang
@@ -20,7 +20,14 @@ submodule openconfig-platform-common {
     "This modules contains common groupings that are used in multiple
     components within the platform module.";
 
-  oc-ext:openconfig-version "0.27.0";
+  oc-ext:openconfig-version "0.28.0";
+
+  revision "2024-09-09" {
+    description
+      "Clarify last-reboot-time is updated during a power-on
+      event.";
+    reference "0.28.0";
+  }
 
   revision "2024-05-29" {
     description
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -644,8 +644,8 @@ revision "2024-05-29" {
          - SYSTEM_INITIATED
          - POWER_FAILURE
          This field is only updated when power is shut off.  It is not
-         updated during reboots; those are tracked in the
-         'last-reboot-time' leaf.";
+         updated during reboots; those are tracked in the 'boot-time'
+         leaf.";
     }
 
     container last-switchover-reason {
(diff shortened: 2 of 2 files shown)


LimeHat:
looks good to me.

rolandphung:
With `last-poweroff-reason` and this semantic change to `last-reboot-time` to be defined to be similar to `boot-time`, should we consider deprecating `last-reboot-reason`?

AnandJyrm:
@dplore Wouldn't it be better to add a new leaf called boot-time/component-boot-time indicating the boot time of the component? openconfig-system boot-time could indicate the boot-time of active controller card in a modular system or chassis in case of a fixed system.

earies:
Few generic comments/questions

* Should verbiage/expectations be added around reboots surrounding image upgrades, ZTP re-instantiation, etc.. where such values may not be retained?
* For POSIX compatible systems, see influence/structure from utmpx - e.g. `last` command, `/var/log/wtmp` - are distinct leafs needed vs. a flexible structure, historical events, etc..?

dplore:
> last-reboot-reason

I think we want to keep `last-reboot-reason`.  This useful text for an operator should be populated specifically for a  "reboot" operation.   

Agreed that this description change gives us the same semantics as `/system/state/boot-time`.  But this leaf is at the component level.

A more consistently named model would use `/components/component/state/boot-time` for this.  I'll update this proposal to this new path, and deprecate `/components/component/state/last-reboot-time`.

@earies those are some good ideas!  However I would like to keep this PR scoped to clarify the intent of the existing model.  A future PR could be raised to keep a history and implement additional fields.

dplore:
This was reviewed at the Feb 4, 2025 OC operators meeting without comments.  Please send any comments you may have by Feb 13, 2025.

Issue #1184, opened 2024-09-18
Misleading typo in the description of routing-policy/.../conditions/match-protocol-instance/config/protocol-name


qm2k:
Description of leaf `/routing-policy/policy-definitions/policy-definition/statements/statement/conditions/match-protocol-instance/config/protocol-name` currently states (emphasis mine):

> The string must match one of /network-instances/network-instance/protocols/protocol/**identifier** in the local network instance.

However, leaf `protocol-identifier` in the same container already matches `/network-instances/network-instance/protocols/protocol/identifier`, which is one part of the key of the [enclosing collection](https://github.com/openconfig/public/blob/9b0a82ee10b6a598d124250f4df51a927c5916a9/release/models/network-instance/openconfig-network-instance.yang#L867), the other being `name`:
```yang
container protocols {
  description
    "The routing protocols that are enabled for this
    network-instance.";

  list protocol {
    key "identifier name";
    // ...
  }
}
````
Everything points to idea that authors wanted `protocol-name` under `match-protocol-instance` to point to the `name`, not `identifier`. We also have type mismatch otherwise, names being strings and identifiers being equivalent identityrefs. Therefore, the description needs to be corrected by changing the word `identifier` to the word `name`.

PR #1185, opened 2024-09-18
Correct typo in protocol-name description

qm2k:
Description stated that this leaf should match identifier of a protocol, but it should actually match name of a protocol.

Fixes: #1184

### Change Scope

Comment-only change preventing misinterpretation of the leaf meaning.

### DIFF
--- a/release/models/network-instance/openconfig-network-instance-policy.yang
+++ b/release/models/network-instance/openconfig-network-instance-policy.yang
@@ -25,7 +25,14 @@ module openconfig-network-instance-policy {
     actions) for the network instance model.  These statements are
     generally added to the routing policy model.";
 
-  oc-ext:openconfig-version "0.1.2";
+  oc-ext:openconfig-version "0.1.3";
+
+  revision "2024-09-19" {
+    description
+      "Correct typo in description of
+       match-protocol-instance/protocol-name.";
+    reference "0.1.3";
+  }
 
   revision "2023-07-25" {
     description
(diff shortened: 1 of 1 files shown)


release/models/network-instance/openconfig-network-instance-policy.yang
```diff
@@ -74,7 +74,7 @@ revision "2018-11-21" {
           on in the local network instance. The string
           must match one of /network-instances/
           network-instance/protocols/
-          protocol/identifier in the local network
+          protocol/name in the local network
```
earies, line 77: The typo correction is correct however it's worth pointing out that these constructs are heavily relying on description statements to describe the intent vs. formulating proper model restrictions to enforce existence of these list keys within a local NI context.

PR as-is LGTM but this modeling should be reworked as a subsequent issue as I'm noticing now.

It requires implementations to have to build special logic outside of the schema that can provide inconsistencies.  This identityref allows for _ANY_ protocol type to be referenced legally and any unbound string for the name.  Maybe the grouping should not be so loose especially when there is only 1 current usage

dplore (commented):
You will need to increment the 'patch' version:

ie:


`ext:openconfig-version "0.1.3";`

qm2k:
> You will need to increment the 'patch' version:
> 
> ie:
> 
> `ext:openconfig-version "0.1.3";`

Thank you, done.

release/models/network-instance/openconfig-network-instance-policy.yang
qm2k: Anything I can do to help it? I am implementer, and I needed this ambiguous point resolved. I would of course always welcome more clarity on the schema level, but for now correct description is better than nothing. I would prefer to keep an issue of matching against technically unbound space of potential protocol names separate.

rgwilton: LGTM

dplore (approved):

Issue #1064, opened 2024-03-08
AFT atomic and counters


dplore:
This issue is regarding counters in the afts tree (for example: [/network-instances/network-instance/afts/next-hops/next-hop/state/counters](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-afts-next-hops-next-hop-state-counters) )

Today the AFT tree contains state for routes and state for counters of packets sent via those routes.   The desired behavior for a subscribe to `/afts` on a device is to  stream the routes as they change (`ON_CHANGE`) and the counters on a periodic interval (`SAMPLE`).  In practice we can observe in many implementations that  subscription mode `TARGET_DEFINED` delivers this desired behavior.   

But this is formally at odds with the aft model which defines the `telemetry-atomic` annotation, for example, at the [next-hop](https://github.com/openconfig/public/blob/815341544e58548526e7f9bec4e0ed2d615b69ac/release/models/aft/openconfig-aft-common.yang#L169) list, below which are also counters.  A strict interpretation says atomic must apply to the nexthop data and the counters which is not clear how that would work, not really what we want and isn't what implementations deliver.    

So how should we address this?

Options might include:
- Update the [telemetry-atomic description](https://github.com/openconfig/public/blob/815341544e58548526e7f9bec4e0ed2d615b69ac/release/models/openconfig-extensions.yang#L154-L165) to say it does not apply to `counters` containers.  This is clear enough, but is not an explicit model.  It also doesn't provide a scalable way to subscribe to a child node in SAMPLE mode while some ancestor is in ON_CHANGE mode.
- Refactor the model to separate counters from the data which is intended to be atomic.  However, it seems the counters container seems to be in a good place from an organization of the data.
- Other ideas?

> **_NOTE:_** On a related note, there has been discussion regarding behavior of  subscription using 'SAMPLE' vs. `ON_CHANGE` for counters.    The summary I think so far is: ON_CHANGE could be interpreted as an event of the counter being refreshed/updated by the network operating system.  One could theoretically change that refresh rate if  config leaf(s) are exposed to allow this.  Others I think have pointed out that the concept of `ON_CHANGE` is incompatible or undesirable counter leafs or any leaf which is expected to be monotonically changing and especially if changing rapidly.    I don't think we need to consider that in this thread because we still don't want or need the atomic behavior for counters in either scenario.

PR #1186, opened 2024-09-19
Define `telemetry-atomic-exempt` annotation. Add `telemetry-atomic-exempt` annotation to AFT counters containers

romeyod:
* (M) openconfig-extensions.yang

Change Scope
* Define `telemetry-atomic-exempt` annotation.
* Add `telemetry-atomic-exempt` annotation to AFT counters containers/grouping

[AFT atomic and counters issue](https://github.com/openconfig/public/issues/1064)

### DIFF
--- a/release/models/openconfig-extensions.yang
+++ b/release/models/openconfig-extensions.yang
@@ -18,7 +18,13 @@ module openconfig-extensions {
     "This module provides extensions to the YANG language to allow
     OpenConfig specific functionality and meta-data to be defined.";
 
-  oc-ext:openconfig-version "0.5.1";
+  oc-ext:openconfig-version "0.5.2";
+
+  revision "2024-09-19" {
+    description
+      "Add telemetry-atomic-exempt annotation.";
+    reference "0.5.2";
+  }
 
   revision "2022-10-05" {
     description
@@ -154,7 +160,7 @@ module openconfig-extensions {
   extension telemetry-atomic {
     description
       "The telemetry-atomic annotation is specified in the context of
-      a subtree (containre, or list), and indicates that all nodes
+      a subtree (container, or list), and indicates that all nodes
       within the subtree are always updated together within the data
       model. For example, all elements under the subtree may be updated
       as a result of a new alarm being raised, or the arrival of a new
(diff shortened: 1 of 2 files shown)


dplore:
This makes sense to me.   @limehat @nandan-arista @earies @rgwilton for comment

release/models/openconfig-extensions.yang
```diff
@@ -164,6 +170,21 @@ module openconfig-extensions {
...
+      within the subtree are not always updated together within the data
+      model of the parent tree. All elements under the
+      subtree may not be updated as a result of a new alarm being raised,
+      or the arrival of a new protocol message that updates the parent
+      tree.
+
+      This annotation allows parent tree containers with telemetry-atomic
+      annotation to not be updated when a more frequently updated subtree.
+      For example, a counters container is present.";
+  }
```
earies, line 186: It should probably be noted that this extension is not to be used unless there is a parent that already contains `telemetry-atomic`

And on `telemetry-atomic`, while I don't believe we have a case today I believe the interpretation is a blanket "every child of where the annotation exists" is to be sent atomically.  If defined at some parent and as child subtrees grow, this could cause revisiting intentions (in other forums there have been similar discussions on just delivering the content of 1 level depth, etc...)

Other than that, I think this is a fine approach

Minor nits: alignment seems off at least in the UI diff rendering, sentence on L# 183-184 I'm not sure I'm grokking as-is...

romeyod: Thanks @earies 
I have updated the documentation to clarify that the parent should have `telemetry-atomic` extension

Not sure about the alignment issue. Let me know if you still see that.

earies: :+1:  - alignment looks fine now

romeyod: Great! Let me know if you have any other feedback as well. Thanks!

release/models/openconfig-extensions.yang
```diff
@@ -164,6 +170,23 @@ module openconfig-extensions {
       optimisations for sending or storing the corresponding data.";
   }
 
+  extension telemetry-atomic-exempt {
+    description
+      "The telemetry-atomic-exempt annotation is specified in the context
+      of a subtree (container, or list), and indicates that all nodes
```
dplore, line 176: I think telemetry-atomic-exempt could also be defined for an individual node.  While I don't have a use case for this at the moment, it seems allowing it is useful.

dplore (commented):
Just one comment to consider, otherwise LGTM

dplore:
Please also update the version statements for the many submodules.  You can see the list in the CI check results which are failing:   https://gist.github.com/OpenConfigBot/7b24c6fe579a3665eb73ad941f55d7e7

dplore:
Setting last call to Oct 24, 2024

release/models/openconfig-extensions.yang
romeyod: I agree. Updated the description to have "node".

dplore (approved):

PR #1187, opened 2024-09-20
Introducing a optical-switch model

mrevang:
### Change Scope

* This change proposes a new model for optical switches.
* As a new model backwards compatibility isn't a concern.

```
module: openconfig-optical-switch
  +--rw optical-switch
     +--rw config
     +--ro state
     +--rw optical-switch-topology
     |  +--ro state
     |     +--ro port-connectivity                  enumeration
     |     +--ro total-port-count                   int32
     |     +--ro total-optical-module-count         int32
     |     +--ro ingress-port-count                 int32
     |     +--ro egress-port-count                  int32
     |     +--ro ingress-start-index                int32
     |     +--ro egress-start-index                 int32
     |     +--ro inter-module-connection-support    boolean
     +--rw port-connections
     |  +--rw port-connection* [slot port]
     |     +--rw slot      -> ../config/slot
     |     +--rw port      -> ../config/port
     |     +--rw config
     |     |  +--rw slot?        string
     |     |  +--rw port?        string
     |     |  +--rw peer-port?   string
     |     |  +--rw peer-slot?   string
     |     +--ro state
     |        +--ro slot?        string
     |        +--ro port?        string
     |        +--ro peer-port?   string
     |        +--ro peer-slot?   string
     +--ro port-statuses
        +--ro port-status* [slot port]
           +--ro slot     -> ../state/slot
           +--ro port     -> ../state/port
           +--ro state
              +--ro slot?                  string
              +--ro port?                  string
              +--ro port-status?           enumeration
              +--ro port-status-message?   string
```

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,7 +3,7 @@ tagged with "Note" can be removed from the template.]
 
 [Note: Before this PR can be reviewed please agree to the CLA covering this
 repo. Please also review the contribution guide -
-https://github.com/openconfig/public/blob/master/doc/external-contributions-guide.md]
+https://github.com/openconfig/public/blob/master/doc/contributions-guide.md]
 
 ### Change Scope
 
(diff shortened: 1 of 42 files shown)


dplore:
[mrevang](https://github.com/mrevang) our only open comment is that there should be some reference to at least 2 implementations which support this kind of functionality.    This can be any documentation link that optical switch vendors say they support point to point programability on their devices.

mrevang:
Don't have those available, closing for now.

Issue #1188, opened 2024-09-22
Clarify transceiver input/output power leaves for standard and ZR use cases.


ejbrever:
Currently, the leaves for transceiver input power and output power are not clear and we should update the descriptions. Additionally there is a "when" statement that prevents the input power from correctly being utilized.

Specifically, this is the "when" statement where it is constrained to standard optics, but the input-power is still relevant for coherent channels to represent the total power (signal+noise):
https://github.com/openconfig/public/blob/9b0a82ee10b6a598d124250f4df51a927c5916a9/release/models/platform/openconfig-platform-transceiver.yang#L299

PR #1189, opened 2024-09-22
Define how physical channel power leaves are used.

ejbrever:
This change:
* Removes the "when" statement and allows the descriptions to clarify when/how the leaves should be used.
* This allows the input-power to be utilized for coherent or standard transceivers. For coherent transceivers, this allows the optical-channel to represent the signal power, while the physical-channel input-power to represent the signal+noise.

More info at:
https://github.com/openconfig/public/issues/1188

### DIFF
--- a/release/models/platform/openconfig-platform-transceiver.yang
+++ b/release/models/platform/openconfig-platform-transceiver.yang
@@ -66,7 +66,13 @@ module openconfig-platform-transceiver {
       specify a physical-channel within a TRANSCEIVER component
       (i.e. gray optic) that it is associated with.";
 
-  oc-ext:openconfig-version "0.14.0";
+  oc-ext:openconfig-version "0.15.0";
+
+revision "2024-09-21" {
+    description
+      "Clearly define how physical channel power leaves are used.";
+    reference "0.15.0";
+  }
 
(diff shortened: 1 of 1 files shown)


ejbrever:
@dplore @ahsaanyousaf can you please review? Thanks.

ahsaanyousaf (approved):

ahsaanyousaf:
Thanks for the changes, LGTM

release/models/platform/openconfig-platform-transceiver.yang
```diff
@@ -384,24 +407,6 @@ revision "2023-08-30" {
...
-    description
-      "Extended operational state data for physical client channels
-      for applications where the full physical channel config and
-      state are used. In some cases, such as when the physical
-      channel has a leafref to an optical channel component and the
-      module-functional-type is TYPE_DIGITAL_COHERENT_OPTIC this
-      grouping will NOT be used.";
-
     uses output-optical-frequency;
     uses optical-power-state;
```
dplore, line 411: indent should be updated

dplore:
Setting last-call for Oct 10, 2024

dplore (approved):

PR #1190, opened 2024-09-24
Add ingress policer to qos overview

dplore:
Adding documentation on how to use the `/qos` model to create an ingress policer

### DIFF
--- a/doc/qos_overview.md
+++ b/doc/qos_overview.md
@@ -1,10 +1,17 @@
 # Overview of OpenConfig QoS
 
-Contributors: aashaikh†, robjs†  
+Contributors: aashaikh†, robjs†, dloher†
 † @google.com  
-December ‘17
+September 2024
 
 # Overview of the QoS Model
+
+## Schema of the QoS Model
+
+The schema of the QoS model can be visualized with this diagram which
+highlights the relationships between the objects in the model.
+
+![QoS schema relationships](img/qos_schema.svg)
 
 The OpenConfig quality of service model is made up of two sets of definitions:
 
@@ -19,9 +26,11 @@ The OpenConfig quality of service model is made up of two sets of definitions:
    correspond to each of the instantiated elements (e.g., schedulers, or
    queues). This is located under `/qos/interfaces`.
 
-The flow of the QoS model is shown in the diagram below:
+## Flow of data through the QoS Model
 
-![Overview of QoS model](img/qos_layout.svg)
+The flow of packets through of the QoS model is shown in the diagram below.
+
+![QoS model data flow](img/qos_layout.svg)
 
 When a packet arrives at an interface it is initially classified according to
 the classifier that is described under
@@ -52,7 +61,7 @@ to virtual output queues that are instantiated for remote interfaces. In this
 case, statistics for each egress interface are reported within the ingress
 interface's entry in the `/qos/interfaces/interface` list.
 
-# Annotated QoS Example
+# Annotated QoS Examples
 
 ## Ingress Classification with Egress Scheduling
 
@@ -60,7 +69,7 @@ The example QoS configuration below shows the configuration of an interface,
 assumed to be facing a customer which has ingress classification based on DSCP
 markings. The same interface has an egress scheduler policy applied to it.
 
-```
+```json
 {
   #
   # The standard definition of an interface, assumed to be facing the customer.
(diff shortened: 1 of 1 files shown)


doc/qos_overview.md
```diff
@@ -422,3 +429,176 @@ markings. The same interface has an egress scheduler policy applied to it.
...
+        config:
+          interface-id: "PortChannel1.200"
+        input:
+          classifers:
+            - classifier:
+              config:
+                name: "dest_B"
+                type: "IPV4"
+          scheduler-policy:
+            state:
```
rgwilton, line 598: Is this a typo?  Should this be a config container?  Ditto below.

doc/qos_overview.md
```diff
@@ -422,3 +429,176 @@ markings. The same interface has an egress scheduler policy applied to it.
...
+assumed to be facing a customer which has ingress classification based on
+the next-hop group the packet is being sent to.  The same interface has an
+ingress scheduler policy applied to it which implementes a ONE_RATE_TWO_COLOR
+policer.
+
+In this scenario, the device is does not have hardware or software to implement
+in ingress queue.  To satisfy the OC schema requirements, a dummy or "fake"
+queue is created.
+
+```yaml
```
robshakir, line 445: Given that we don't have a spec for YANG modelled data being serialised to YAML, can we write this as JSON please?

doc/qos_overview.md
```diff
@@ -422,3 +429,176 @@ markings. The same interface has an egress scheduler policy applied to it.
   }
 }
 ```
+
+### Ingress Classification with Ingress Scheduling (Policer)
+
+The example QoS configuration below shows the configuration of an interface,
+assumed to be facing a customer which has ingress classification based on
+the next-hop group the packet is being sent to.  The same interface has an
+ingress scheduler policy applied to it which implementes a ONE_RATE_TWO_COLOR
```
robshakir, line 438:
```suggestion
ingress scheduler policy applied to it which implementes an `ONE_RATE_TWO_COLOR`
```

doc/qos_overview.md
```diff
@@ -422,3 +429,176 @@ markings. The same interface has an egress scheduler policy applied to it.
...
+policer.
+
+In this scenario, the device is does not have hardware or software to implement
```
robshakir, line 441: We should clarify that on a VOQ architecture this is needed. It's not a "dummy" queue if we have the ability to do this queueing on an egress linecard.

doc/qos_overview.md
```diff
@@ -52,15 +59,15 @@ to virtual output queues that are instantiated for remote interfaces. In this
...
 
-# Annotated QoS Example
+## Annotated QoS Examples
 
-## Ingress Classification with Egress Scheduling
+### Ingress Classification with Egress Scheduling
 
 The example QoS configuration below shows the configuration of an interface,
 assumed to be facing a customer which has ingress classification based on DSCP
 markings. The same interface has an egress scheduler policy applied to it.
```
robshakir, line 68: Can we expand a bit on what this achieves? i.e., there's an egress scheduler policy but it seems like it's material to the example that this is being used to police traffic to some expected output data rate.

doc/qos_overview.md
dplore: Yes, typo, fixed these instances

doc/qos_overview.md
dplore: Converted to legal JSON, so the comments are gone.    Do you want me to add them back in even though it cannot be validated?  😜

doc/qos_overview.md
dplore: done

doc/qos_overview.md
dplore: Added more description for this

doc/qos_overview.md
dplore: Oh, I did not intend this example to be related to  the ingress policer example below.

dplore:
@robshakir ready for your re-review

LimeHat:
I don't think this is the appropriate way to describe a canonical ingress policer.

While the scheduling typically supports rate-limiting mechanisms, in many implementations, there are separate policer entities (often implemented in HW) that are independent of schedulers; i.e., those are two different mechanisms.

The fake queuing/ingress scheduling approach feels more like a workaround for implementations where true ingress policers are not available (or they never implement "real" ingress scheduling and can perform this config translation into policing in native yang/cli model).
IMO it is fine if you want to describe that use case here, but stating that this is the (only) canonical way severely limits the range of covered scenarios and use cases.

dplore:
@robshakir Ready for your next review

doc/qos_overview.md
```diff
@@ -422,3 +429,278 @@ markings. The same interface has an egress scheduler policy applied to it.
...
+### Example 2: Ingress Classification with Ingress Scheduling (Policer) on a VOQ device
+
+The example QoS configuration below shows the configuration of an interface,
+assumed to be facing a customer which has ingress classification based on
+the next-hop group the packet is being sent to.  The same interface has an
+ingress scheduler policy applied to it which implements an
+`ONE_RATE_TWO_COLOR` policer.
+
+In this scenario, the device has a VOQ architecture is does not have hardware
+or software to implement in ingress queue.  To satisfy the OC schema
```
robshakir, line 442:
```suggestion
or software to implement in ingress queue.  To allow a consistent representation to be used across different architectures
```

doc/qos_overview.md
```diff
@@ -422,3 +429,278 @@ markings. The same interface has an egress scheduler policy applied to it.
...
+  "openconfig-qos": {
+    "classifers": [
+      {
+        "classifer": "“dest_A”",
+        "config": {
+          "name": "“dest_A”"
+        },
+        "terms": [
+          {
+            "term": null,
```
robshakir, line 462: Not sure what this `term` field that is set to `null` is -- the JSON here seems incorrect:

```json
{
  "qos": {
    "classifiers": {
      "classifier": [
        {
          "name": "Dest_A",
          "config": {
            "name": "Dest_A"
          },
          "terms": {
            "term": [
              {
                "id": "foo"
              }
            ]
          }
        }
      ]
    }
  }
}
```

I suggest that this is generated using pyangbind, ygot, or your-favourite-YANG-thingy -- you can also write it by hand and validate it using something like [`govalidate`](https://github.com/openconfig/oc-tools/blob/master/validate/govalidate/main.go) (although this needs some maintenance to add some new modules and update the docs).

robshakir: FYI, the above JSON validates: 

```
[2:47] robjs@aardvark:oc-tools/validate/govalidate   master ✔                                                                       242d ⚑ ◒  ⍉
▶ ./govalidate -fn foo.json && echo $? 
0
```

doc/qos_overview.md
```diff
@@ -422,3 +429,278 @@ markings. The same interface has an egress scheduler policy applied to it.
...
+        "forwarding-group": "input_dest_B",
+        "config": {
+          "name": "input_dest_B",
+          "output-queue": "dummy_input_queue_B"
+        }
+      }
+    ],
+    "queues": [
+      {
+        "queue": null,
```
robshakir, line 565: Same problem here as in `classifiers`.

doc/qos_overview.md
```diff
@@ -422,3 +429,278 @@ markings. The same interface has an egress scheduler policy applied to it.
...
+          "name": "dummy_input_queue_B"
+        }
+      }
+    ],
+    #
+    # Two scheduler policies are created, limit_1Gb and limit_2Gb
+    # and are associated with the dummy queue they are servicing.
+    "scheduler-policies": [
+      {
+        "scheduler-policy": null,
```
robshakir, line 582: Ditto, same issue w.r.t a list that is being set to `null`.

robshakir (changes requested):

doc/qos_overview.md
dplore: generated with ygot.
(ref: https://gist.github.com/dplore/8d5f91d9e0b1a76cdde47f7a4d53acc3)

PR #964, opened 2023-09-22
Relay agent helper source

marius-ore:
### Change Scope

* Replace leaf-list of helper addresses with a keyed list and add attribute for source interface or IP address
* This is a breaking change, as it deprecates the existing leaf-list
### Platform Implementations

 * Arista: [ip helper-address](https://www.arista.com/en/um-eos/eos-ipv4#xx1144379)
 * Junos (source interface): [relay-source](https://www.juniper.net/documentation/us/en/software/junos/subscriber-mgmt-sessions/topics/ref/statement/relay-source-edit-forwarding-options.html)
 * HPE FlexNetwork (source IP address), "Specifying the source IP address for relayed DHCP requests" in Configuring the DHCP relay agent: [HPE FlexNetwork 5130 EI Switch Series Layer 3—IP ServicesConfiguration Guide](https://techhub.hpe.com/eginfolib/networking/docs/switches/5130ei/5200-3942_l3-ip-svcs_cg/content/index.htm)
 * Cisco (source-interface): [Configuration Examples for DHCP Relay Server ID Override and Link Selection Option 82 Suboptions](https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/ipaddr_dhcp/configuration/15-sy/dhcp-15-sy-book/dhcp-relay-svr-option-82.html#GUID-4AF995F3-68E4-4EC3-B03D-B9896306ADC1)

Issue #951, opened 2023-09-04
Adding the option to configure source address/interface for a DHCP relay agent


marius-ore:
In some scenarios when configuring DHCP relay, it can be useful to specify a source IP address or a source interface whose address is to be used as the source address in the messages relayed to a DHCP server. I don’t see a way to specify that in the current model for `relay-agent`, so I propose deprecating the leaf list `helper-address` from `/relay-agent/dhcp(v6)/interfaces/interface/(config|state)/` and replacing it with a list `helper-addresses` keyed by IP helper address and containing attribute `source`, which is a union of `inet:ip-address` and `oc-if:interface-id`

Change summary:
```
 module: openconfig-relay-agent
   +--rw relay-agent 
      +--rw dhcp
      |  +--rw interfaces
      |  |  +--rw interface* [id]
      |  |  |  +--rw id               -> ../config/id
      |  |  |  +--rw config
      |  |  |  |  +--rw id?                        oc-if:interface-id
-     |  |  |  |  +--rw helper-address*   inet:ip-address
+     |  |  |  +--rw helper-addresses
+     |  |  |  |  +--rw helper-address* [address]
+     |  |  |  |  |  +--rw address?     -> ../config/address
+     |  |  |  |  |  +--rw config
+     |  |  |  |  |  |  +--rw address?           string  //ip address or hostname
+     |  |  |  |  |  |  +--rw source?            union   // source ip address or interface identifier
+     |  |  |  |  |  +--ro state
+     |  |  |  |  |  |  +--ro address?           string  //ip address or hostname
+     |  |  |  |  |  |  +--ro source?            union   // source ip address or interface identifier
      +--rw dhcpv6
      |  +--rw interfaces
      |  |  +--rw interface* [id]
      |  |  |  +--rw id               -> ../config/id
      |  |  |  +--rw config
      |  |  |  |  +--rw id?                        oc-if:interface-id
-     |  |  |  |  +--rw helper-address*   inet:ipv6-address
+     |  |  |  +--rw helper-addresses
+     |  |  |  |  +--rw helper-address* [address]
+     |  |  |  |  |  +--rw address?     -> ../config/address
+     |  |  |  |  |  +--rw config
+     |  |  |  |  |  |  +--rw address?           inet:ipv6-address
+     |  |  |  |  |  |  +--rw source?            union   // ipv6 address or interface identifier
+     |  |  |  |  |  +--ro state
+     |  |  |  |  |  |  +--ro address?           inet:ipv6-address
[...]

PR #1191, opened 2024-09-25
Relay Agent helper: add support for source, hostname, and network instance

rachidtt:
This is a follow-up and update to #964  #951  . More context can be found there

### Change Scope

* Replace “helper-address” leaf-list with keyed list
* Update address type to be `oc:inet:host` to support ip and hostname configuration
* Add support for source-interface or source-address configuration 
* Add option to specify network-instance 

### Platform Implementations

 * Arista: [ip helper-address](https://www.arista.com/en/um-eos/eos-ipv4#xx1144379) 

New tree (relevant changes):
```module: openconfig-relay-agent
  +--rw relay-agent
     +--rw dhcp
     |  +--rw interfaces
     |     +--rw interface* [id]
     |        +--rw id                          -> ../config/id
     |        +--rw config
     |        |  +--rw id?                 oc-if:interface-id
     |        |  +--rw enable?             boolean
     |        |  x--rw helper-address*     inet:ip-address
     |        |  +--rw helper-addresses
     |        |     +--rw helper-address* [address]
     |        |        +--rw address    -> ../config/address
     |        |        +--rw config
     |        |        |  +--rw address?            oc-inet:host
     |        |        |  +--rw local-address?      union
     |        |        |  +--rw network-instance?   oc-ni:network-instance-ref
     |        |        +--ro state
     |        |           +--ro address?            oc-inet:host
     |        |           +--ro local-address?      union
     |        |           +--ro network-instance?   oc-ni:network-instance-ref
     |        +--ro state
     |        |  +--ro id?                 oc-if:interface-id
     |        |  +--ro enable?             boolean
     |        |  x--ro helper-address*     inet:ip-address
     |        |  +--ro helper-addresses
     |        |  |  +--ro helper-address* [address]
     |        |  |     +--ro address    -> ../config/address
     |        |  |     +--ro config
     |        |  |     |  +--ro address?            oc-inet:host
     |        |  |     |  +--ro local-address?      union
     |        |  |     |  +--ro network-instance?   oc-ni:network-instance-ref
     |        |  |     +--ro state
     |        |  |        +--ro address?            oc-inet:host
     |        |  |        +--ro local-address?      union
     |        |  |        +--ro network-instance?   oc-ni:network-instance-ref
     +--rw dhcpv6
        +--rw interfaces
           +--rw interface* [id]
              +--rw id               -> ../config/id
              +--rw config
              |  +--rw id?                 oc-if:interface-id
              |  +--rw enable?             boolean
              |  x--rw helper-address*     inet:ipv6-address
              |  +--rw helper-addresses
              |     +--rw helper-address* [address]
              |        +--rw config
              |        |  +--rw local-address?      union
              |        +--ro state
              |           +--ro address?            oc-inet:host
              |           +--ro local-address?      union
              |           +--ro network-instance?   oc-ni:network-instance-ref
              +--ro state
              |  +--ro id?                 oc-if:interface-id
              |  +--ro enable?             boolean
              |  x--ro helper-address*     inet:ipv6-address
              |  +--ro helper-addresses
              |  |  +--ro helper-address* [address]
              |  |     +--ro address    -> ../config/address
              |  |     +--ro config
              |  |     |  +--ro address?            oc-inet:host
              |  |     |  +--ro local-address?      union
              |  |     |  +--ro network-instance?   oc-ni:network-instance-ref
              |  |     +--ro state
              |  |        +--ro address?            oc-inet:host
              |  |        +--ro local-address?      union
              |  |        +--ro network-instance?   oc-ni:network-instance-ref```

### DIFF
--- a/release/models/relay-agent/openconfig-relay-agent.yang
+++ b/release/models/relay-agent/openconfig-relay-agent.yang
@@ -10,8 +10,10 @@ module openconfig-relay-agent {
   // import some basic types
   import ietf-inet-types { prefix inet; }
   import ietf-yang-types { prefix yang; }
+  import openconfig-network-instance { prefix oc-ni; }
   import openconfig-interfaces { prefix oc-if; }
   import openconfig-extensions { prefix oc-ext; }
+  import openconfig-inet-types { prefix oc-inet; }
 
 
   // meta
@@ -27,7 +29,14 @@ module openconfig-relay-agent {
     packets.  The supports both DHCP and DHCPv6 and device-wide and
     per-interface settings.";
 
-  oc-ext:openconfig-version "0.1.2";
+  oc-ext:openconfig-version "0.2.0";
+
+  revision "2024-09-18" {
+    description
+      "Turn helper-address into a keyed list and add attributes for
+      source and hostname configuration";
+    reference "0.2.0";
+  }
 
   revision "2023-02-06" {
     description
@@ -624,6 +633,62 @@ module openconfig-relay-agent {
 
   }
 
+  grouping relay-agent-ipv4-helper-address-config {
+    description
+      "Configuration options for an IP helper-address";
+
+    leaf address {
+      type oc-inet:host;
+      description
+        "IP address to relay DHCP requests to";
+    }
+
+    leaf local-address {
+      type union {
+        type inet:ipv4-address;
+        type oc-if:interface-id;
+      }
+      description
+        "IP address or interface whose IP address is to be used as source in
+	relayed DHCP requests";
+    }
+
+    leaf network-instance {
+      type oc-ni:network-instance-ref;
+      description
+        "The network instance that this DHCP Server
+        is associated with";
+    }
+  }
+
+  grouping relay-agent-ipv6-helper-address-config {
+    description
+      "Configuration options for an IPv6 helper-address";
+
+    leaf address {
+      type oc-inet:host;
+      description
+        "IPv6 address to relay DHCPv6 requests to";
+    }
+
+    leaf local-address {
+      type union {
+        type inet:ipv6-address;
+        type oc-if:interface-id;
+      }
+      description
+        "IPv6 address or interface whose IPv6 address is to be used as source in
+	relayed DHCPv6 requests";
+    }
+
+    leaf network-instance {
+      type oc-ni:network-instance-ref;
+      description
+        "The network instance that this DHCP Server
+        is associated with";
+    }
+  }
+
   grouping relay-agent-ipv4-interfaces-config {
     description
       "Configuration data for interfaces enabled for relaying";
@@ -644,12 +709,52 @@ module openconfig-relay-agent {
 
     leaf-list helper-address {
       type inet:ip-address;
+      status deprecated;
       description
         "List of IPv4 or IPv6 addresses of DHCP servers to which the
         relay agent should forward DHCPv4 requests.  The relay agent is
         expected to forward DHCPv4/BOOTP requests to all listed
         server addresses when DHCPv4 relaying is enabled globally, or
         on the interface.";
+    }
+
+    container helper-addresses {
+      description
+        "Enclosing container for helper addresses";
+      list helper-address {
+	key "address";
+	description
+          "List of IPv4 addresses of DHCP servers to which the
+          relay agent should forward DHCPv4 requests.  The relay agent is
+          expected to forward DHCPv4/BOOTP requests to all listed
+          server addresses when DHCPv4 relaying is enabled globally, or
+          on the interface.";
+
+	leaf address {
+          type leafref {
+            path "../config/address";
+          }
+          description
+            "Reference to list key representing IP helper address";
+        }
+
+
+	container config {
+	  description
+	    "Configuration data for IP helper address";
+
+	  uses relay-agent-ipv4-helper-address-config;
+	}
+
+	container state {
+	  config false;
+
+	  description
+	    "Operational state for IP helper address";
+
+	  uses relay-agent-ipv4-helper-address-config;
+	}
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: release/models/relay-agent/openconfig-relay-agent.yang [hunk 1]
```diff
@@ -624,6 +633,62 @@ module openconfig-relay-agent {
...
+    description
+      "Configuration options for an IP helper-address";
+
+    leaf address {
+      type oc-inet:host;
+      description
+        "IP address to relay DHCP requests to";
+    }
+
+    leaf local-address {
```
dplore, line 646: In most of the OC models, we use source-address as the leaf describing the origin a device should use to create a packet.  

```suggestion
    leaf source-address {
```

release/models/relay-agent/openconfig-relay-agent.yang
```diff
@@ -624,6 +633,62 @@ module openconfig-relay-agent {
...
+    description
+      "Configuration options for an IPv6 helper-address";
+
+    leaf address {
+      type oc-inet:host;
+      description
+        "IPv6 address to relay DHCPv6 requests to";
+    }
+
+    leaf local-address {
```
dplore, line 674:
```suggestion
    leaf source-address {
```

dplore (changes requested):
One minor change suggested.  Also, check the linter output.  There needs to be a config and state container, and the config container must have config set to true.

Ref: https://gist.github.com/OpenConfigBot/0d4b55defd09566d4bd0bebb7efeb460

release/models/relay-agent/openconfig-relay-agent.yang
```diff
@@ -624,6 +633,148 @@ module openconfig-relay-agent {
...
+
+  grouping helper-addresses-ipv4-top {
+    description
+      "Top-level grouping for DHCPv4 helper-addresses on interfaces";
+
+    container helper-addresses {
+      description
+        "Enclosing container for helper addresses";
+      list helper-address {
+        key "address";
```
earies, line 700: It seems to me that keying off an IP address with a child leaf that has a network-instance ref prevents any case of IP address overlap and the following example EOS configuration (albeit likely rare in practice)

```
interface Ethernet1
   no switchport
   ip address 10.100.1.1/30
   ip helper-address 1.1.1.1 vrf VRF-A
   ip helper-address 1.1.1.1 vrf VRF-B
   ipv6 address 2001:db8:1::/64
```

release/models/relay-agent/openconfig-relay-agent.yang
```diff
@@ -624,6 +633,148 @@ module openconfig-relay-agent {
...
+
+    leaf address {
+      type oc-inet:host;
+      description
+        "IPv6 address to relay DHCPv6 requests to";
+    }
+
+    leaf source-address {
+      type union {
+        type inet:ipv6-address;
```
earies, line 676: Use OpenConfig types - is there any reason to keep the old IETF types as part of this module import any longer?

```suggestion
        type oc-inet:ipv6-address;
```

release/models/relay-agent/openconfig-relay-agent.yang
```diff
@@ -624,6 +633,148 @@ module openconfig-relay-agent {
...
+  }
+
+  grouping relay-agent-ipv6-helper-address-config {
+    description
+      "Configuration options for an IPv6 helper-address";
+
+    leaf address {
+      type oc-inet:host;
+      description
+        "IPv6 address to relay DHCPv6 requests to";
```
earies, line 671: The use of `oc-inet:host` implies a union of `domain-name` as well as `ipv4-address` _AND_ `ipv6-address`

earies (commented):
Please provide current implementation references for more than just 1 implementation

release/models/relay-agent/openconfig-relay-agent.yang
dplore: This should be oc-inet:ip-address and not host

release/models/relay-agent/openconfig-relay-agent.yang
dplore: agreed, this should be a composite key with ip-address and network-instance name

earies: To be fair, this can apply throughout the model set wherever a `network-instance-ref` was added post list structure/key definition however in various other cases, the list key is then a named (user-defined) object abstracting out the unique aspects which are carried as non-key children.

So that is 1 approach to the problem statement however one that I'll venture to bet misaligns w/ all implementations here (as well as elsewhere)

The other being the multi-key approach you suggest but it presents different design patterns across the model set

rachidtt: This makes sense to support ip overlap. I changed the key to use ip-address and network-instance.
The updated pyang looks like:
```    |        +--rw helper-addresses
     |           +--rw helper-address* [address network-instance]
     |              +--rw address             -> ../config/address
     |              +--rw network-instance    -> ../config/network-instance
     |              +--rw config
     |              |  +--rw address?            oc-inet:host
     |              |  +--rw network-instance?   oc-ni:network-instance-ref
     |              |  +--rw source-address?     union
     |              +--ro state
     |                 +--ro address?            oc-inet:host
     |                 +--ro network-instance?   oc-ni:network-instance-ref
     |                 +--ro source-address?     union
```

release/models/relay-agent/openconfig-relay-agent.yang
earies: How many implementations support domain-name for such use (that also store the object as such to reflect in config/state and not resolve at time of input)?  And is it a good idea to rely off DNS (and/or a local cache) in such operational scenarios?

IMO precise IP address types should suffice here.  Speaking on behalf of one implementation, we do not support DNS names for this purpose and baking that into a union is not deviate-able without replacing the entire type (and thus changing the underlying namespace for alternate NBI where this matters) - it is then left to runtime error handling which is also not conveyed at the data modeling layers

release/models/relay-agent/openconfig-relay-agent.yang
earies: I think this is probably the best approach where a list key contains an object such as an IP address _but_ I would suggest tagging this problem statement for a future OC community call so that there can be consistency across the model-set

For reference, another spot where I see the same issue:

```
module: openconfig-system
  +--rw system
     +--rw logging
        +--rw remote-servers
           +--rw remote-server* [host]
              +--rw host         -> ../config/host
              +--rw config
              |  +--rw host?                 oc-inet:host
              |  +--rw source-address?       oc-inet:ip-address
              |  +--rw network-instance?     oc-ni:network-instance-ref
              |  +--rw remote-port?          oc-inet:port-number
              |  +--rw transport-security?   boolean
              |  +--rw tls-profile-id?       string
              +--ro state
              |  +--ro host?                 oc-inet:host
              |  +--ro source-address?       oc-inet:ip-address
              |  +--ro network-instance?     oc-ni:network-instance-ref
              |  +--ro remote-port?          oc-inet:port-number
              |  +--ro transport-security?   boolean
              |  +--ro tls-profile-id?       string
```

Verma-Anukul:
@rachidtt 
Can you please take this to a closure, we are also working on similar extension, wants to use it

PR #1203, opened 2024-10-21
Add policy forwarding for next hop match and GUE encapsulation action

danameme:
### Change Scope

* Add support for policy forwarding match on next hop and GUE encapsulation action.
* Move the encapsulation-header-type to a common module so it can be used more generally.


release/models/aft/openconfig-aft-common.yang
```diff
@@ -303,7 +303,7 @@ submodule openconfig-aft-common {
             }
 
             container gre {
-              when "../state/type = 'oc-aftt:GRE'";
+              when "../state/type = 'oc-types:GRE'";
```
earies, line 306: Since there is a relocation of types to a separate model - the `openconfig-version` and `revision` must be updated to reflect this backwards incompatible change

robshakir: +1, but I don't see why we need to make this change. Why not leave these in AFT types?

dplore: The next PR from @danameme  will use them in policy-forwarding

danameme: It's used in policy-forwarding in this same PR.

danameme:
Create a new https://github.com/openconfig/public/pull/1208 due to merge conflicts from committed changes.

PR #1204, opened 2024-10-21
Add encapsulation header types referenced from AFT module.

danameme:
### Change Scope

* These encapsulation header types were referenced in openconfig-aft-common.yang but were not defined.


danameme:
I'm going to add these changes to https://github.com/openconfig/public/pull/1203 so closing this PR.

PR #1218, opened 2024-11-08
Adding enum to specifiy policy forwarding behaviour in case of decaps…

vishnureddybadveli:
…ulate a packet

### Change Scope

### Platform Implementations

   implementation output.
   implementation output.


release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -214,6 +221,11 @@ submodule openconfig-pf-forwarding-policies {
             "The policy is used only to classify incoming packets into corresponding
             network instances.";
         }
+        enum HW_LOOKUP_POLICY {
```
akalluru1, line 224: As this is an enum it cannot represent both the HW look ups, i.e., before or after. So, description seems to be causing some confusion.

PR #1219, opened 2024-11-09
Added next-hop-network-instance under mpls egress config.

vishnureddybadveli:

### Change Scope

### Platform Implementations

   implementation output.
   implementation output.

### DIFF
--- a/release/models/mpls/openconfig-mpls-static.yang
+++ b/release/models/mpls/openconfig-mpls-static.yang
@@ -217,6 +217,14 @@ submodule openconfig-mpls-static {
         "Next hop IP address for the LSP";
     }
 
+    leaf next-hop-network-instance {
+      type oc-ni:network-instance-ref;
+      description
+        "The network-instance in which to resolve the next-hop.
+         If not specified, then the next-hop is resolved in the 
+         enclosing network instance.";
+    }
+
     leaf push-label {
       type oc-mplst:mpls-label;
       status deprecated;


release/models/mpls/openconfig-mpls-static.yang
```diff
@@ -24,7 +24,15 @@ submodule openconfig-mpls-static {
     "Defines static LSP configuration";
 
 
-  oc-ext:openconfig-version "3.6.0";
+  oc-ext:openconfig-version "3.7.0";
+
+  revision "2024-11-12" {
+    description
+      "Added support for mpls route post decap to a specific
```
LimeHat, line 31: Can you please clarify this? What is `post decap` in this case?

In terms of mpls static LSPs, you have three different scenarios:
1) ingress 
2) transit
3) egress

Does `decap` implies the egress (where the top label is removed by a PE node, and the LSP is terminated)?

LimeHat (changes requested):

release/models/mpls/openconfig-mpls-static.yang
nandanarista: We should remove `post decap`. 
This is for egress LSPs. 
The intent is to do a lookup for the nexthop (to determine egress interface and rewrite) in a vrf that is different from the ingress interface's VRF.
For example: in some vpc use case ingress interface is in the default VRF as it's facing the providers network and the egress interface is in a customer specific VRF.

release/models/mpls/openconfig-mpls-static.yang
```diff
@@ -217,6 +225,14 @@ submodule openconfig-mpls-static {
         "Next hop IP address for the LSP";
     }
 
+    leaf next-hop-network-instance {
```
akalluru1, line 228: As some of the attributes, for ex: next-hop, push-label & etc..., in this grouping are `deprecated` wondering if this should be part of `lsp-next-hops` container.

release/models/mpls/openconfig-mpls-static.yang
vishnureddybadveli: May I kindly request 

> Ok. In that case, in addition to updating the description, I think this leaf should be moved from `static-lsp-common-config` to `static-lsp-egress-config`

I had updated description and can move leaf to egress config, may I kindly request if Nokia platform can support this? I could not find any reference from Nokia public documentation.

As a reference ARISTA support's this using following CLI command:
```
mpls static top-label vrf pop payload-type ipv4/6
```

AnnamalaiRajeev:
Hello Vishnu as discussed on the operators board. Please include vendor config and usecase. Limit augment to only egress LSP. thx

vishnureddybadveli:
> Hello Vishnu as discussed on the operators board. Please include vendor config and usecase. Limit augment to only egress LSP. thx

Rajeev, description is updated with only egress use case and one vendor config. We are looking for a second vendor that can implement similar functionality. Once we find second vendor I will work on the augment part.

release/models/mpls/openconfig-mpls-static.yang
LimeHat, line 228: This leaf is still placed in the wrong grouping.

Please move it from `static-lsp-common-config` to `static-lsp-egress-config`

LimeHat (changes requested):

release/models/mpls/openconfig-mpls-static.yang
Shashank-arista: Yes, please move this to `lsp-next-hops` container as we are supporting adding next-hop related information in this new container only..

vishnureddybadveli: Thank you, please bear with me:                                                                                                                                                     

I am trying to move it to /network-instances/network-instance/mpls/lsps/static-lsps/static-lsp/egress/lsp-next-hops/lsp-next-hop/config/

Ideally it should go to static-lsp-nexthop-common-config under config but has these common attributes defined across ingress,transit and egress.                                                                                                                                                        
I am trying to create additional static-lsp-nexthop-egress-config to restrict the vrf name.

release/models/mpls/openconfig-mpls-static.yang
dplore: The updates here look good to me

dplore:
@vishnureddybadveli , can you provide a config example from JunOS?  The referenced documentation sort of sounds like it matches this use case, but it's not clear.  An example config snippet would help.    Or alternatively, you could reference a different implementation if that is easier to provide an example.

vishnureddybadveli:
simon.gates@bt.com may I kindly request to update Cisco's CLI if they do support this feature. Thank you.

PR #1221, opened 2024-11-14
Add set-as-path-prepend/use-last-as

xuqma:
### Change Scope

* Add set-as-path-prepend/use-last-as to allow prepending last AS to the path.

### Platform Implementations

 * Implementation A: CiscoXR [prepend as-path most-recent](https://www.cisco.com/c/en/us/td/docs/routers/crs/software/crs_r4-0/routing/command/reference/rr40crs1book_chapter8.html#wp1928320454).
 * Implementation B: Arista routemap [set as-path prepend last-as](https://www.arista.com/en/um-eos/eos-acls-and-route-maps#xx1325439).

New paths added:

```
+ /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/actions/openconfig-bgp-policy:bgp-actions/set-as-path-prepend/config/use-last-as
+ /openconfig-routing-policy:routing-policy/policy-definitions/policy-definition/statements/statement/actions/openconfig-bgp-policy:bgp-actions/set-as-path-prepend/state/use-last-as
```

### DIFF
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -28,7 +28,13 @@ module openconfig-bgp-policy {
     It augments the base routing-policy module with BGP-specific
     options for conditions and actions.";
 
-  oc-ext:openconfig-version "8.0.0";
+  oc-ext:openconfig-version "8.1.0";
+
+  revision "2024-11-13" {
+    description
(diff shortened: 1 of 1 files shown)


dplore:
Reviewed in Dec 10, 2024 OC Operators meeting without objection.  Setting to last-call for Dec 20, 2024

dplore (approved):

PR #1223, opened 2024-11-14
Clarify that if no rules are present, no packets will be matched.

dplore:
### Change Scope
* Clarify policy-forwarding such that if no rules are present, no packets will be matched

### DIFF
--- a/release/models/policy-forwarding/openconfig-policy-forwarding.yang
+++ b/release/models/policy-forwarding/openconfig-policy-forwarding.yang
@@ -62,11 +62,14 @@ module openconfig-policy-forwarding {
 
     A forwarding-policy specifies the match criteria that it intends
     to use to determine the packets that it reroutes - this may
-    consist of a number of criteria, such as DSCP. The action of the
-    policy results in a forwarding action being applied to matching
-    packets. For example, decapsulating the packet from a GRE header.
-    In order to enact the policy based on particular interfaces - the
-    forwarding-policy is applied to an interface via referencing it
+    consist of a number of criteria, such as DSCP. The match criteria
+    is specified as rules.  If no rules are specified, then the policy
+    will not match any packets.
+
+    The action of the policy results in a forwarding action being applied
+    to matching packets. For example, decapsulating the packet from a GRE
+    header. In order to enact the policy based on particular interfaces -
+    the forwarding-policy is applied to an interface via referencing it
     within an 'apply-forwarding-policy' statement associated with an
     interface.
 
@@ -81,7 +84,13 @@ module openconfig-policy-forwarding {
     The forwarding action of the corresponding policy is set to
     PATH_GROUP and references the configured group of LSPs.";
 
-  oc-ext:openconfig-version "0.6.1";
+  oc-ext:openconfig-version "0.7.0";
+
+  revision "2024-11-14" {
+    description
+      "Clarify that if no rules are present, no packets will be matched.";
+    reference "0.7.0";
+  }
 
   revision "2023-04-25" {
     description


dplore:
Related to discussion on #1150

s19nal (approved):
LGTM - updated clarification reads well.

dplore:
We discussed this a second time in the OC Operators meeting on Nov 26, 2024.
It is observed that policy-forwarding and routing-policy should have the same semantics.  At the moment, both are ambiguous about the behavior when there are no rules (policy-forwarding) or no conditions ([routing-policy](https://github.com/openconfig/public/blob/c58e42f811de0e381f38781179cbcf61e34d3e0d/release/models/policy/openconfig-routing-policy.yang#L24)) defined.  

For routing-policy, vendor native implementations widely support policies which have no 'conditions' matching all packets.  

Example native routing policy configuration: https://gist.github.com/dplore/cff9a4c782f68734adaafff8952fc3cb   (Kudos to @oscargdd for sharing this investigation)


So I am updating this description to assert that no rules means match all packets.

xuqma (approved):
LGTM for BGP routing policies. Thanks Darren.

dplore:
After a bit more discussion, there was a request to also clarify what happens if there is a routing-policy statement with no condition AND no action, as well as a policy-definition  without any statements.  These should be treated as  errors  (grpc error: `INVALID_ARGUMENT`,   netconfig error `missing-element`).  

Note that [policy-result ](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-actions-config-policy-result) is an action.  So one may write a valid OC routing-policy which has for example, one statement which as no condition (matching all routes) and a `policy-result` of `ACCEPT_ROUTE`.   

I will update this pull request accordingly.

release/models/policy-forwarding/openconfig-policy-forwarding.yang
```diff
@@ -62,11 +62,14 @@ module openconfig-policy-forwarding {
...
     A forwarding-policy specifies the match criteria that it intends
     to use to determine the packets that it reroutes - this may
-    consist of a number of criteria, such as DSCP. The action of the
-    policy results in a forwarding action being applied to matching
-    packets. For example, decapsulating the packet from a GRE header.
-    In order to enact the policy based on particular interfaces - the
-    forwarding-policy is applied to an interface via referencing it
+    consist of a number of criteria, such as DSCP. The match criteria
+    is specified as rules.  If no rules are specified, then the policy
+    will match all packets.
```
robshakir, line 67: Why is this the right thing to do vs. having an explicit match-all type criteria?

The downside of this API design is that the absence of something means something implicitly -- it seems a better API design to have the presence of something indicate something, which would mean that we should rather have an explicit "match everything" (can be done with matching all src/dst IP addrs)?

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -51,11 +51,13 @@ module openconfig-routing-policy {
...
-    accept-route or reject-route, evaluation of the current policy definition
-    stops, and no further policy statements are evaluated. In case there is a
-    policy chain, no further policy definitions in the chain are evaluated.
+    corresponding action statement is executed.  A policy statement with no
+    conditions defined is evaluated as statisfied, meaning all routes are
+    matched.  If the action statement has a final disposition configured as
+    policy result, either accept-route or reject-route, evaluation of the
+    current policy definition stops, and no further policy statements are
+    evaluated. In case there is a policy chain, no further policy definitions
+    in the chain are evaluated.
```
robshakir, line 60: Interested in @xuqma's view here -- is it painful to add explicit "match everything" terms per my suggestion above?

robshakir (changes requested):

akalluru1:
> After a bit more discussion, there was a request to also clarify what happens if there is a routing-policy statement with no condition AND no action, as well as a policy-definition without any statements. These should be treated as errors (grpc error: `INVALID_ARGUMENT`, netconfig error `missing-element`).
> 
> Note that [policy-result ](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-actions-config-policy-result) is an action. So one may write a valid OC routing-policy which has for example, one statement which as no condition (matching all routes) and a `policy-result` of `ACCEPT_ROUTE`.
> 
> I will update this pull request accordingly.

Small clarification, About following:
```
what happens if there is a routing-policy statement with no condition AND no action, as well as a policy-definition without any statements. These should be treated as errors (grpc error: `INVALID_ARGUMENT`, netconfig error `missing-element`)
```
- Is this capturing 2 different scenarios, i.e., scenario 1: a routing-policy statement with no condition AND no action & scenario 2:  a policy-definition without any statements?
- In both the scenarios error (grpc error: `INVALID_ARGUMENT`, netconfig error `missing-element`) should be returned.

release/models/policy-forwarding/openconfig-policy-forwarding.yang
dplore: There is no right or wrong here.   This proposal is made based on precedent as referenced in  https://github.com/openconfig/public/pull/1223#issuecomment-2501782886 earlier in this PR.  The industry precedent is no-condition = matches any and all.  This was also discussed in the OpenConfig operator meeting on Nov 26, 2024 with some thoughtful conversation.    

The precedent is pretty strong where 5 out of 5 implement that no "condition" means match all.    (Not included in the gist of 4 examples is arista EOS, which also defines a [route map with no match statements as match all](https://www.arista.com/en/um-eos/eos-acls-and-route-maps))

I do like the idea of being explicit in general.  A number of changes in OC in the last few years have been due to ambiguity.  But what is better?  Do something different in OC or go with the precedent?   In this case, I think the precedent is so strong that it's better to go with no conditions means match all.  This change makes this decision explicit via description, but not in the data model/API.

release/models/policy/openconfig-routing-policy.yang
dplore: The proposed description change reads for policy-forwarding: `If no rules are specified, then the policy will match all packets.`

And for routing-policy:
`A policy statement with no conditions defined is evaluated as statisfied, meaning all routes are matched.`

dplore:
> > After a bit more discussion, there was a request to also clarify what happens if there is a routing-policy statement with no condition AND no action, as well as a policy-definition without any statements. These should be treated as errors (grpc error: `INVALID_ARGUMENT`, netconfig error `missing-element`).
> > Note that [policy-result ](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-routing-policy.html#routing-policy-policy-definitions-policy-definition-statements-statement-actions-config-policy-result) is an action. So one may write a valid OC routing-policy which has for example, one statement which as no condition (matching all routes) and a `policy-result` of `ACCEPT_ROUTE`.
> > I will update this pull request accordingly.
> 
> Small clarification, About following:
> 
> ```
> what happens if there is a routing-policy statement with no condition AND no action, as well as a policy-definition without any statements. These should be treated as errors (grpc error: `INVALID_ARGUMENT`, netconfig error `missing-element`)
> ```
> 
> * Is this capturing 2 different scenarios, i.e., scenario 1: a routing-policy statement with no condition AND no action & scenario 2:  a policy-definition without any statements?
> * In both the scenarios error (grpc error: `INVALID_ARGUMENT`, netconfig error `missing-element`) should be returned.

Two scenarios as you described.

release/models/policy-forwarding/openconfig-policy-forwarding.yang
dplore: @robshakir any response for us?

dplore:
@robshakir ready for your comment

release/models/policy-forwarding/openconfig-policy-forwarding.yang
```diff
@@ -62,14 +62,20 @@ module openconfig-policy-forwarding {
...
+
+    The action of the policy results in a forwarding action being applied
+    to matching packets. For example, decapsulating the packet from a GRE
+    header. In order to enact the policy based on particular interfaces -
+    the forwarding-policy is applied to an interface via referencing it
     within an 'apply-forwarding-policy' statement associated with an
     interface.
 
+    A policy-forwarding rule with no action should be treated as an error
+    (grpc error: INVALID_ARGUMENT, netconfig error missing-element).
```
robshakir, line 77:
```suggestion
    (grpc error: INVALID_ARGUMENT, NETCONF error missing-element).
```

We generally don't put in the text what the type of protocol is though.

release/models/policy-forwarding/openconfig-policy-forwarding.yang
robshakir: This seems reasonable, given the precedent. It's nicer to be explicit, but I also do not quite understand how we would be so in YANG.

e.g., we could have a flag with a specific match-everything condition, but it seems there's still some ambiguity remaining.

release/models/policy/openconfig-routing-policy.yang
```diff
@@ -51,11 +51,17 @@ module openconfig-routing-policy {
...
-    corresponding action statement is executed.  If the action
-    statement has a final disposition configured as policy result, either
-    accept-route or reject-route, evaluation of the current policy definition
-    stops, and no further policy statements are evaluated. In case there is a
-    policy chain, no further policy definitions in the chain are evaluated.
+    corresponding action statement is executed.  A policy statement with no
+    conditions defined is evaluated as statisfied, meaning all routes are
+    matched.  A routing-policy statement with no condition AND no action, or
+    a policy-definition without any statements should be treated as errors
+    (grpc error: INVALID_ARGUMENT, netconfig error missing-element).
```
robshakir, line 58: Same comment above.

robshakir (approved):

release/models/policy-forwarding/openconfig-policy-forwarding.yang
dplore: removed the references

release/models/policy/openconfig-routing-policy.yang
dplore: removed protocol references

PR #1224, opened 2024-11-14
Add policy-forwarding network-instance attachment point

dplore:
### Change Scope

Today policy-forwarding policies can only be attached to interfaces.  This change adds network-instance as an attachment point for policy-forwarding rules.  This causes packets in the specified network-instance to be evaluated against the specified policy-forwarding rules and actions.

## Tree diff view
We are at:
`/network-instances/network-instance/policy-forwarding/`

```diff
         |  +--rw interfaces
         |  |  +--rw interface* [interface-id]
         |  |     +--rw interface-id     -> ../config/interface-id
         |  |     +--rw config
         |  |     |  +--rw interface-id?                 oc-if:interface-id
         |  |     |  +--rw apply-forwarding-policy?      -> ../../../../policies/policy/config/policy-id
         |  |     |  +--rw apply-vrf-selection-policy?   -> ../../../../policies/policy/config/policy-id
         |  |     +--ro state
         |  |     |  +--ro interface-id?                 oc-if:interface-id
         |  |     |  +--ro apply-forwarding-policy?      -> ../../../../policies/policy/config/policy-id
         |  |     |  +--ro apply-vrf-selection-policy?   -> ../../../../policies/policy/config/policy-id
         |  |     +--rw interface-ref
         |  |        +--rw config
         |  |        |  +--rw interface?      -> /oc-if:interfaces/interface/name
         |  |        |  +--rw subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
         |  |        +--ro state
         |  |           +--ro interface?      -> /oc-if:interfaces/interface/name
         |  |           +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
+        |  +--rw network-instances
+        |  |  +--rw network-instance* [name]
+        |  |     +--rw name      -> ../config/name
+        |  |     +--rw config
+        |  |     |  +--rw name?                         -> /network-instances/network-instance/config/name
+        |  |     |  +--rw apply-forwarding-policy?      -> ../../../../policies/policy/config/policy-id
+        |  |     |  +--rw apply-vrf-selection-policy?   -> ../../../../policies/policy/config/policy-id
+        |  |     +--ro state
+        |  |        +--ro name?                         -> /network-instances/network-instance/config/name
+        |  |        +--ro apply-forwarding-policy?      -> ../../../../policies/policy/config/policy-id
+        |  |        +--ro apply-vrf-selection-policy?   -> ../../../../policies/policy/config/policy-id
         |  +--rw path-selection-groups
         |     +--rw path-selection-group* [group-id]
         |        +--rw group-id    -> ../config/group-id
         |        +--rw config
         |        |  +--rw group-id?   string
         |        |  +--rw mpls-lsp*   -> ../../../../../mpls/lsps/constrained-path/tunnels/tunnel/config/name
         |        +--ro state
         |           +--ro group-id?   string
         |           +--ro mpls-lsp*   -> ../../../../../mpls/lsps/constrained-path/tunnels/tunnel/config/name


### Platform Implementations

Cisco IOS XR
```
 vrf-policy
 vrf default address-family ipv4 policy type pbr input GRE-DECAP-PBR
```
Arista EOS
```
ip decap-group SRTE-GRE-DECAP
 tunnel type gre
 tunnel decap-ip 10.5.32.0
```

### DIFF
--- a/release/models/policy-forwarding/openconfig-pf-network-instances.yang
+++ b/release/models/policy-forwarding/openconfig-pf-network-instances.yang
@@ -18,7 +18,7 @@ submodule openconfig-pf-network-instances {
   description
     "This submodule contains groupings to attach
      a policy-forwarding policy to one or more network-instances.
-     
+
      When attached to a network-instance, the policy-forwarding rules
      should be evaluated against packets in the specified
      network-instances.";
@@ -121,10 +121,10 @@ submodule openconfig-pf-network-instances {
       }
       description
         "Apply the specific VRF selection policy on the network-instance.
-         
+
          The referenced policy MUST be of the type VRF_SELECTION_POLICY.
          The VRF selection policy may coexist with a policy-forwarding policy.
-         
+
          The policy specified in this leaf is used to specifically choose the
          L3VRF network instance that is used for specific input criteria of
          packets.";
--- a/release/models/policy-forwarding/openconfig-pf-network-instances.yang
+++ b/release/models/policy-forwarding/openconfig-pf-network-instances.yang
@@ -3,10 +3,9 @@ submodule openconfig-pf-network-instances {
     prefix "oc-pf";
   }
 
-  import openconfig-extensions { prefix "oc-ext"; }
-  import openconfig-interfaces { prefix "oc-if"; }
-  import openconfig-network-instance { prefix "oc-ni"; }
-
+  import openconfig-extensions {
+    prefix "oc-ext";
+  }
   include openconfig-pf-forwarding-policies;
 
   organization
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: release/models/policy-forwarding/openconfig-pf-network-instances.yang [hunk 1]
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  revision "2016-11-08" {
+    description
+      "Initial revision";
+    reference "0.0.1";
+  }
+
+
+  grouping pf-network-instance-structural {
```
rgwilton, line 98: @dplore, I couldn't see where this grouping is used, is part of the change set missing?

release/models/policy-forwarding/openconfig-pf-network-instances.yang
```diff
@@ -0,0 +1,194 @@
...
+    leaf apply-forwarding-policy {
+      type leafref {
+        // We are at /network-instances/network-instance/
+        // policy-forwarding/interfaces/interface/config/
+        // apply-forwarding-policy
+        path "../../../../policies/policy/config/policy-id";
+      }
+      description
+        "The policy to be applied on the network-instance. Packets ingress
+        on the referenced interface should be compared to the match
```
LimeHat, line 168: referenced interface?

release/models/policy-forwarding/openconfig-pf-network-instances.yang
dplore: yes, this a a draft PR / work in progress.  I'll switch it to "ready to review" shortly

release/models/policy-forwarding/openconfig-pf-network-instances.yang
```diff
@@ -0,0 +1,194 @@
...
+    leaf name {
+      type oc-ni:network-instance-ref;
+      description
+        "A unique identifier for the network-instance.";
+    }
+
+    leaf apply-forwarding-policy {
+      type leafref {
+        // We are at /network-instances/network-instance/
+        // policy-forwarding/interfaces/interface/config/
```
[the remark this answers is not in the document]
dplore: Thanks, fixed

release/models/policy-forwarding/openconfig-pf-network-instances.yang
dplore: fixed

dplore:
Last call for comments.  This will merge on Jan 6th, 2025

release/models/policy-forwarding/openconfig-pf-network-instances.yang
```diff
@@ -0,0 +1,133 @@
...
+      }
+      description
+        "The policy to be applied on the network-instance. Packets ingress
+         on the referenced interface should be compared to the match
+         criteria within the specified policy, and in the case that
+         these criteria are met, the forwarding actions specified
+         applied. These policies should be applied following quality of
+         service classification, and ACL actions if such entities are
+         referenced by the corresponding interface.";
+    }
```
LimeHat, line 113: This description still talks about the "referenced interface".

release/models/policy-forwarding/openconfig-pf-network-instances.yang
```diff
@@ -0,0 +1,133 @@
...
+    "OpenConfig working group
+     www.openconfig.net";
+
+  description
+    "This submodule contains groupings to attach
+     a policy-forwarding policy to one or more network-instances.
+
+     When attached to a network-instance, the policy-forwarding rules
+     should be evaluated against packets in the specified
+     network-instances.";
```
LimeHat, line 24: Similar comment to what I left in the previous PR #1150: what are the "packets in the specified network-instances"?

* does that mean that the policy applied to all interfaces associated with the network-instance on ingress (e.g. this change introduces only a new way of configuring the existing functionality), or
* are there any additional expectations w.r.t what packets should be matched (e.g packets redirected from another network-instance via a policy)?

LimeHat (commented):
I'd like to see a precise definition of how matching is supposed to work.

In addition, I think there's a formatting issue with the references in the first comment, can it be fixed?

release/models/policy-forwarding/openconfig-pf-network-instances.yang
dplore: The operational use case motivating this PR is for packets ingress to the device which arrive on interfaces (including virtual interfaces) which are mapped to a network-instance.  

I am not aware of an operational use case that we have for packets re-directed from another network-instance.

nandanarista:
Hi Darren, please note that EOS decap groups cannot be attached to a network instances, so not sure why we think that EOS decap group implements this proposal. 
https://www.arista.com/en/um-eos/eos-decap-groups

release/models/policy-forwarding/openconfig-pf-network-instances.yang
nandanarista: Policy forwarding entries are already defined within a network instance, so what does the enclosing network instance mean?

LimeHat: > The operational use case motivating this PR is for packets ingress to the device which arrive on interfaces (including virtual interfaces) which are mapped to a network-instance.

Thus, in terms of functionality, this configuration will be fully equivalent to applying a policy to all interfaces associated with the network-instance? We just adding a "simpler" way to configure this?

Can this be explicitly stated in the description? Thanks

dplore: Yes, that is correct.  I will make this explicit in the description

dplore: > Policy forwarding entries are already defined within a network instance, so what does the enclosing network instance mean?

It is true, but without any "apply-policy" configuration, it is ambiguous where the policy is applied?  Further, since OC has a way to apply-policy to interfaces, if none are specified, what should the behavior be?

I'd prefer an explicit leaf for this.   

I will change the name/description from a leaf-ref to network-instance instead to a boolean like 'all-interfaces'.   In our use case, we only want to apply the policy to the same network-instance the policy is defined in, and not some different network-instance.

robshakir (changes requested):
I tend to agree with the other reviewers here -- given that we are already in the context of a `network-instance` when we are applying a policy this looks to be:

 * syntatic sugar to apply to "all interfaces",
 * probably something that can be applied programmatically at the cost of adding some more content to the configuration request.

It looks like @nandanarista also highlighted this isn't supported in EOS. Can we look at whether we need to make this enhancement?

PR #1226, opened 2024-11-15
Expand mount-point/state/type to union of string and enum

mrevang:
### Change Scope

* This change is being made to add more structure to the content of mount-point/state/type. This will enable more predictable content for programming against.
* The underlying type is being updated from a string to a union (indentyref/string), so it's not backwards compatible. In practice, a string will ultimately still be decoded, so it is backward compatible in that regard.
### Platform Implementations

 * Before mount-point/state/type was added to the standard, it was a Google specific augment.

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -21,6 +21,7 @@ module openconfig-system {
   import openconfig-messages { prefix oc-messages; }
   import openconfig-license { prefix oc-license; }
   import openconfig-network-instance { prefix oc-ni; }
+  import openconfig-fs-types { prefix oc-fs-types; }
 
   // meta
   organization "OpenConfig working group";
@@ -54,7 +55,7 @@ module openconfig-system {
       "Expand mount-point type leaf to a union of string and enum.";
     reference "2.4.0";
   }
-  
+
   revision "2024-09-24" {
     description
       "Added mount-point type leaf to describe the type of file system.";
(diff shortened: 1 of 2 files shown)


release/models/system/openconfig-system.yang
```diff
@@ -428,11 +434,59 @@ module openconfig-system {
     }
 
     leaf type {
-      type string;
+      type union {
+        type string;
+        type enumeration {
```
earies, line 439: Few overall comments:

* Depending how this is looked at, it is a backwards incompatible change as you are ultimately changing the underlying YANG type which can have other effects outside of the encoded value representation
* I would propose if doing this to take the enum declaration out into a reusable typedef or better yet a set of expansible identities (and use an identityref rather)
* Reference to `/proc/filesystems` is useful here and there is also an implicit assumption of underlying OS in doing this (this already exists in other spots too)
* IMO drop the union - just keep an identityref then and implementations can augment in expanded identities to the base if necessary.  If you order the union as such then the rule of thumb is that the value matches the string as first match everytime as both are encoded identically

But if this is looked at as a mapping of the first column of something like `df` output (Filesystem) then you have a mish-mash of what can be represented here (This is likely what each implemtation is doing today) meaning alot of different string values (or expanded identities which then don't make as much sense imo)

If this is to be the `type` from something like the output of `mounts` then corresponds directly w/ the list you might see from `/proc/filesystems`

rgwilton: Wouldn't we also need the enumeration to be listed above the string in the union definition (I appreciate that you are suggesting that the union isn't needed at all), otherwise everything would always match against the string, regardless of the enum values?

dplore:
Thanks for the comments.  We reviewed in the Dec 3 operators meeting and there is some rough consensus that  making this  change to this very new leaf as an indentify-ref like @earies suggests is cleaner and still supports the operational use case.  We have good precedence of adding more identity refs as needed in areas like transceiver types.  

@mrevang could you revise this remove the string entirely and become an identity-ref?

dplore:
> Thanks for the comments. We reviewed in the Dec 3 operators meeting and there is some rough consensus that making this change to this very new leaf as an indentify-ref like @earies suggests is cleaner and still supports the operational use case. We have good precedence of adding more identity refs as needed in areas like transceiver types.
> 
> @mrevang could you revise this remove the string entirely and become an identity-ref?

The new yang would look like:
```
    leaf type {
      type identityref {
        base oc-fs-types:OPENCONFIG_FS_MOUNT;
      }

```

Then in a new file, such as release/models/system/openconfig-fs-types.yang, add identities for all the desired filesystem types.   These identities will be encoded the same as strings (should anyone have implemented it)

```
  identity OPENCONFIG_FS_MOUNT {
    description
      "Base identity for system filesystem mounting point types. 
      Derived identities are based on the linux mount api fs_type.";
    reference
      "https://docs.kernel.org/filesystems/mount_api.html";
  }

  identity UNKNOWN {
    base OPENCONFIG_FS_MOUNT;
    description
      "Indicates that the mount type could not be determined.
      Use of this identity SHOULD be avoided.";
  }

  identity CGROUP_FS {
    base OPENCONFIG_FS_MOUNT;
    description
      "cgroup filesystem type.";
  }

  identity DEBUGFS_FS {
    base OPENCONFIG_FS_MOUNT;
    description
      "debugfs filesystem type.";
  }
```
and so on

release/models/system/openconfig-system.yang
dplore: Evan made this change for us (moved the indentityref before the string)

dplore:
This will be reviewed at the next OC Operators meeting on July 15, 2025

ElodinLaarz:
Discussed at the OC Operators Meeting on July 15th. Seems generally fine. One suggestion is to change the description `Please prefer the identifyref over the string` to something of the form `The string type exists to support backwards compatibility, ...` to better reflect the desire for deprecating the string field.

PR #1231, opened 2024-12-05
fix description typo under openconfig-interfaces module interface-ref…

awebsters:
…-common group

### Change Scope

* Fixing misspelling in a leaf's description
* Backwards compatible: Yes


robshakir (approved):
Thanks very much for the fix, appreciated!

PR #1232, opened 2024-12-06
Update PIM documentation to clarify that only ipv4 AF is supported.

AnilKR123:
### Change Scope

Update PIM documentation to clarify that only ipv4 AF is supported.


AnilKR123:
I updated the patch revision. 

Thanks.

dplore (approved):

PR #1233, opened 2024-12-16
Add Bfd diagnostic code for when neighbor signaled session down.

AnilKR123:
### Change Scope

I have added a Bfd diagnostic code for when neighbor signals session down. This is specified in RFC5880 but is absent from the public model.

From RFC5880:

     A diagnostic code specifying the local system's reason for the
      last change in session state.  Values are:

         0 -- No Diagnostic
         1 -- Control Detection Time Expired
         2 -- Echo Function Failed
         3 -- Neighbor Signaled Session Down
         4 -- Forwarding Plane Reset
         5 -- Path Down
         6 -- Concatenated Path Down
         7 -- Administratively Down
         8 -- Reverse Concatenated Path Down

### DIFF
--- a/release/models/bfd/openconfig-bfd.yang
+++ b/release/models/bfd/openconfig-bfd.yang
@@ -164,27 +164,21 @@ module openconfig-bfd {
           "The BFD echo function failed - echo packets have not been
           received for the required period of time.";
       }
-      enum NEIGHBOR_DOWN {
-        value 3;
-        description
-          "The Bfd neighbor signaled session down - Bfd packet was received with
-           neighbor state down.";
-      }
       enum FORWARDING_RESET {
-        value 4;
+        value 3;
         description
           "The forwarding plane in the local system was reset - such
           that the remote system cannot rely on the forwarding state of
           the device specifying this error code.";
       }
       enum PATH_DOWN {
-        value 5;
+        value 4;
         description
           "Signalling outside of BFD specified that the path underlying
           this session has failed.";
       }
       enum CONCATENATED_PATH_DOWN {
-        value 6;
+        value 5;
         description
           "When a BFD session runs over a series of path segments, this
           error code indicates that a subsequent path segment (i.e.,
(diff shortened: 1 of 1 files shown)


release/models/bfd/openconfig-bfd.yang
```diff
@@ -157,35 +164,41 @@ module openconfig-bfd {
           "The BFD echo function failed - echo packets have not been
           received for the required period of time.";
       }
-      enum FORWARDING_RESET {
+      enum NEIGHBOR_DOWN {
```
dplore, line 167: To keep this as a non-breaking change, the original ordering of the enums must be preserved.  You can only add enums to the list.

dplore: I realize the enum value won't match the value in a packet on the wire, but that is not required.

AnilKR123:
I moved NEIGHBOR_DOWN to the end of the enums list. 

Thanks.

AnilKR123:
I would like to propose deprecating the bfd-diagnostic-code enum and creating a new enum, perhaps called rfc5880-bfd-diagnostic-code. The bfd-diagnostic-code enum is missing “Neighbor Signaled Session Down” diag, as specified in RFC5880:


A diagnostic code specifying the local system's reason for the
 last change in session state.  Values are:

     0 -- No Diagnostic
     1 -- Control Detection Time Expired
     2 -- Echo Function Failed
     3 -- Neighbor Signaled Session Down
     4 -- Forwarding Plane Reset
     5 -- Path Down
     6 -- Concatenated Path Down
     7 -- Administratively Down
     8 -- Reverse Concatenated Path Down



The objection to adding this diagnostic code to enum value 3 is that it would be a breaking change. However adding it to the end of the enum would not only make Neighbor Down not compliant, but also all of the enum values after 3 differ in relation to the RFC. This adds operational complexity since clients now need a parser to map between the RFC codes and OC public codes.

By adding a new enum for the diag codes, we could support the RFC compliant values. And each vendor can decide which leaf they want to support for now.

dplore:
OC Operator opinion is that it's more important to have a vendor neutral model where all vendors consistently implement the same OpenConfig model and there is only one way to model the BFD state.  It's also viewed as more churn to refactor the model than to add the enum to the end, even though it doesn't match the packets on the wire.    So this PR change is approved as is.  

Last call for Feb 25, 2025.

dplore:
@AnilKR123 one misc check is failing, can you fix it please?

```
openconfig-bfd.yang: file updated but openconfig-version string not updated: "0.3.0"
```

AnilKR123:
Hi @dplore, I have fixed the issue by bumping the version number. Thanks.

dplore (approved):

PR #1235, opened 2024-12-19
Fix enclosing container p2p-primary-paths name

Pull-eckermann:
Fix enclosing container p2p-primary-paths name.

Enclosing container p2p-primary-path name is not in the format described in https://www.openconfig.net/docs/guides/style_guide/#yang-style-conventions.

### DIFF
--- a/release/models/mpls/openconfig-mpls-te.yang
+++ b/release/models/mpls/openconfig-mpls-te.yang
@@ -32,6 +32,12 @@ submodule openconfig-mpls-te {
 
   oc-ext:openconfig-version "3.6.0";
 
+  revision "2024-12-19" {
+    description
+      "Fix enclosing container p2p-primary-paths name";
+    reference "3.6.1";
+  }
+
   revision "2024-06-19" {
     description
       "Added support with backward compatibility to configure ECMP
@@ -1334,7 +1340,7 @@ submodule openconfig-mpls-te {
     description
       "Top level grouping for p2p primary paths";
 
-    container p2p-primary-path {
+    container p2p-primary-paths {
       description
         "Primary paths associated with the LSP";
 


Pull-eckermann:
Can someone review it?

romeyod:
This seems like a breaking change to me. I understand the motivation of being yang style complaint but seems to me we have missed that opportunity with this one.
@dplore can correct me here.

jsterne:
> This seems like a breaking change to me. I understand the motivation of being yang style complaint but seems to me we have missed that opportunity with this one. @dplore can correct me here.

I'd tend to agree. Non backwards compatible changes shouldn't be made lightly. They impact everyone who has automated/integrated around this area. I'm doubtful this one is ever worth it. But I'd imagine a variety of opinions exist around something like this :-)

earies:
> > This seems like a breaking change to me. I understand the motivation of being yang style complaint but seems to me we have missed that opportunity with this one. @dplore can correct me here.
> 
> I'd tend to agree. Non backwards compatible changes shouldn't be made lightly. They impact everyone who has automated/integrated around this area. I'm doubtful this one is ever worth it. But I'd imagine a variety of opinions exist around something like this :-)

+1 - this is definitely a breaking change and quick glance looks like this subtree/list has already been shipping across various implementations for a while now.  Yes it slipped through the cracks but if we weigh the pros and cons here to going back and fixing this now, you would likely find a strong preference towards preserving compatibility vs. style conformance.

This is especially true for a good chunk of the industry that ingests OpenConfig modeled data unknowingly.  All that is seen is a breaking change to a request or an alternate dataset now being created as a result.

It's not to say we should be held to never iterating in such fashion but there is an elephant in the room here that needs proper attention on proper agreed upon operational transition methods and signals of such changes.

dplore:
Thank you for identifying this.  Unfortunately, as noted above, correcting style does not warrant a breaking change, especially in a long standing v1+ model.

PR #1236, opened 2024-12-24
Add acctz counters for login/logout records

nmahabaleshwar:
Add new counters for acctz records generated for login and logout records

### Change Scope

* Added source counters which will be incremented when login and logout records are generated.
* A new type called `session-service` is added which defines new enums `LOGIN` and `LOGOUT`.
* `service-request` has a new enum `SESSION_SERVICE`

### Platform Implementations

None

### DIFF
--- a/release/models/gnsi/openconfig-gnsi-acctz.yang
+++ b/release/models/gnsi/openconfig-gnsi-acctz.yang
@@ -27,7 +27,13 @@ module openconfig-gnsi-acctz {
     "This module provides counters of gNSI accountZ requests and responses and
     the quantity of data transferred.";
 
-  oc-ext:openconfig-version "0.3.0";
+  oc-ext:openconfig-version "0.4.0";
+
+  revision 2024-12-24 {
+    description
+      "Add counters for login/logout accounting records";
+    reference "0.4.0";
+  }
 
   revision 2024-10-07 {
     description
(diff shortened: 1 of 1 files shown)


release/models/gnsi/openconfig-gnsi-acctz.yang
```diff
@@ -96,6 +102,20 @@ module openconfig-gnsi-acctz {
       }
     }
   }
+  typedef session-service {
+    description "enum SessionService.SessionServiceType";
```
dplore, line 106: I don't see this in the acctz proto?

https://github.com/openconfig/gnsi/blob/a43096008ba06570eddbdd64531d775a9b25a4b3/acctz/acctz.proto#L114

nmahabaleshwar: Fixed this

haussli:
Did we discuss how/if to account for SESSION_STATUS_ONCE, et al?

I think that the commit message should be more verbose; it should describe the reason for this change, please.

nmahabaleshwar:
> Did we discuss how/if to account for SESSION_STATUS_ONCE, et al?
> 
> I think that the commit message should be more verbose; it should describe the reason for this change, please.

Did you mean the counter update for SESSION_STATUS_ONCE ? SESSION_STATUS_ONCE already has information about which operation is being accounted, gRPC vs command services and we already have counters for them.

I will update the commit description with more details on why this change is required.

haussli:
> > Did we discuss how/if to account for SESSION_STATUS_ONCE, et al?
> > I think that the commit message should be more verbose; it should describe the reason for this change, please.
> 
> Did you mean the counter update for SESSION_STATUS_ONCE ? SESSION_STATUS_ONCE already has information about which operation is being accounted, gRPC vs command services and we already have counters for them.
> 
> I will update the commit description with more details on why this change is required.

OK, understood; if login succeeds, the service type is known for the following record.

dplore:
This LGTM. @morrowc and @marcushines for any comments.

dplore:
Last call for comments is March 4th, 2025

morrowc (approved):

dplore (approved):

dplore:
FYI @mihirpitale-googler

PR #1238, opened 2025-01-02
Update telemetry-atomic description

romeyod:

Change Scope
* Update `telemetry-atomic` description to clarify the list container behavior.

### DIFF
--- a/release/models/openconfig-extensions.yang
+++ b/release/models/openconfig-extensions.yang
@@ -18,7 +18,13 @@ module openconfig-extensions {
     "This module provides extensions to the YANG language to allow
     OpenConfig specific functionality and meta-data to be defined.";
 
-  oc-ext:openconfig-version "0.6.0";
+  oc-ext:openconfig-version "0.7.0";
+
+  revision "2025-01-02" {
+    description
+      "Update telemetry-atomic description.";
+    reference "0.7.0";
+  }
 
   revision "2024-09-19" {
     description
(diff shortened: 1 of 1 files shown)


release/models/openconfig-extensions.yang
```diff
@@ -160,11 +166,17 @@ module openconfig-extensions {
...
       model. For example, all elements under the subtree may be updated
       as a result of a new alarm being raised, or the arrival of a new
-       protocol message.
+      protocol message. When the annotation is specified in the context
+      of a list of containers, it applies to individual list elements
+      (subtree or container and all elements under that subtree) of the
+      list and not to the entire list. For example, say ipv4-entry list
+      has two elements, prefix1 and prefix2, all elements under the
+      prefix1 subtree may be updated as a result of a next-hop group
+      change while prefix2 subtree is not updated.
```
dplore, line 179: Would it be more clear to say something like:
"the atomic property applies to the contents of each individual keyed entry in the list."

dplore:
Added to OC operator review for March 11, 2025

jsterne:
I'm finding the text in this proposed clarification a bit confusing. The first change actually removes a reference to lists and instead says leaf-list. Was that really intended? I'm pretty sure the intention is for this to apply to containers and lists.
The term "list of containers" is also somewhat confusing. I believe this clarification is trying to say that when a single list entry changes, only that specific list entry (and all its descendants) is updated (not all entries in the list).

dplore:
Perhaps the description could be clarified more like @jsterne recommends:

"When a single list entry changes, only that specific list entry and all its descendants are updated, not all entries in the list."

Is there some reason to exclude list?  That part seemed clear to me already.

romeyod:
> Perhaps the description could be clarified more like @jsterne recommends:
> 
> "When a single list entry changes, only that specific list entry and all its descendants are updated, not all entries in the list."
> 
> Is there some reason to exclude list? That part seemed clear to me already.

I have incorporated the feedback. Thanks @jsterne and @dplore 
Let me know what you think of the latest diffs.

jsterne:
That looks more clear to me now - thanks.

PR #1239, opened 2025-01-02
 Add MACSEC SCI

dplore:
### Change Scope

* Add a leaf to enable/disable MACSEC Secure Channel Identifier (SCI) 
* `/macsec/mka/policies/policy/config/include-sci`
* `/macsec/mka/policies/policy/state/include-sci`

### Platform Implementations

 * Cisco IOS XR [Supports SCI by default](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/security/79x/b-system-security-cg-cisco8000-79x/configuring-macsec.html#concept_pw3_wft_s3b)

 * [Arista EOS](https://www.arista.com/en/um-eos/eos-data-plane-security#unique_1576191211)
 ```
switch(config)# mac security
switch(config-mac-security)# profile TEST
switch(config-mac-security-profile-TEST)# sci
```

 * [Juniper JunOS](https://www.juniper.net/documentation/us/en/software/junos/security-services/topics/task/macsec.html)
```
[edit security macsec connectivity-association connectivity-association-name]
user@host# set include-sci
```

### DIFF
--- a/release/models/macsec/openconfig-macsec.yang
+++ b/release/models/macsec/openconfig-macsec.yang
@@ -18,10 +18,17 @@ module openconfig-macsec {
     "This module defines configuration and state data for
      MACsec IEEE Std 802.1AE-2018.";
 
-  oc-ext:openconfig-version "1.1.1";
+  oc-ext:openconfig-version "1.2.0";
   oc-ext:regexp-posix;
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
+
+  revision "2025-01-02" {
+    description
+      "Add include-sci to allow enable/disable of secure channel
+      identifiers.";
+    reference "1.2.0";
+  }
 
   revision "2023-08-03" {
     description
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: release/models/macsec/openconfig-macsec.yang [hunk 1]
```diff
@@ -627,6 +634,12 @@ revision "2023-06-08" {
          require an ICV";
     }
 
+    leaf include-sci {
```
robshakir, line 637: For the Cisco implementation, I assume that this is just not configurable to `false` -- but for the others this switch is needed? If so, LGTM.

robshakir (approved):

release/models/macsec/openconfig-macsec.yang
dplore: Yes, I would expect Cisco to accept a "true" value for this and reject a "false" whereas other implementations would accept either value.

romeyod:
should we add a corresponding leaf in the state path as well?

dplore:
> should we add a corresponding leaf in the state path as well?

Hi, it's already added to both, but I missed noting that in the PR description. I'll update it now.

dplore:
Setting last-call to Feb 18, 2025

PR #1240, opened 2025-01-03
clarification of use and semantics of metric and preference of next-h…

rszarecki:


dplore:
Thanks for this suggestion.  I do think it's worth clarifying that
1. Preference is ONLY used for protocol based selection for RIB population.    It is not used for RIB selection between next-hops within the static protocol.
2. If a static route contains multiple next-hops, only the lowest preference value is used.

We'll review on the Jan 7th , 2025 OC Operators call.

dplore:
In review on the Jan 7th , 2025 OC Operators call it was observed that 3 out of 4 implementations do support using preference (admin distance, etc) for next-hop selection within the same protocol.  ie:  One can configure 2 static routes with different preferences or with different metrics.

In this case I don't see a reason OC needs to create a more constrained view when the implementations broadly support this behavior already.   So I recommend closing this PR,  @rszarecki .

Issue #1241, opened 2025-01-03
Component upgrade state information


ejbrever:
I'd like to propose adding upgrade state information for components. This will allow details regarding upgrades to be exposed to users as an upgrade process can take many minutes to hours. This would be read-only information.

This could support components such as transceivers (i.e. ZR transceivers) or the device operating system.

These could be placed under a new grouping such as:
openconfig/platform/component/state/upgrade/

Proposed leaves would be:
new-firmware-version: string, the new firmware package version.
new-firmware-version-service-impacting: bool, true if the firmware would require a service impacting upgrade to take effect.
status: identityref (sw loading, in-progress, fail, complete)
step: string, describing the step the device is attempting to do.
step-percent-complete: oc-types:percentage, the percent of the step that is completed.
total-percent-complete: oc-types:percentage, the percent of the total upgrade that is completed.
start-time: timeticks64, the time when the upgrade was initiated.
duration: yang:counter64, the time elapsed since the upgrade was initiated.
stop-time: timeticks64, the time when the upgrade was initiated.
last-known-failure: string, describing the last known failure encountered.

@robshakir @ahsaanyousaf

proberts2022:
My current focus is around the use of this for transceivers...I may come back with more comments on the generality of this to other component types.
A few comments and questions for better understanding:
1. The new-version description indicates its available on the component.   To me this means the firmware image has been already successfully transferred into the component but is not yet the active image.   Is that the intent?
2. The duration field proposal above is how long the upgrade has been in progress but in PR#1242 the leaf description is "the duration of the upgrade" which I think can be interpreted in different ways.  Also is it your intention that once an upgrade operation is complete, that this leaf would report the total time the last upgrade required or it would not be reported since there is no upgrade in progress.   Perhaps the description of the leaf could be "The time elapsed since the upgrade was initiated or the duration of the last upgrade"
3. Why is the duration leaf tracking microseconds of elapsed time.   Seems overkill for an operation that might take seconds.  Do you see value on this granularity for debugging purposes?
4. Do you want a separate OPENCONFIG_UPGRADE_STATUS of UPGRADE_ABORT or is UPGRADE_FAIL sufficient?
5. For my education, does the openconfig model include 'action' definitions?   Basically, how does one of these transceiver upgrades get trigger via openconfig?

ejbrever:
1. Correct, that could only be specified once an image was transferred.
2. Thanks for the suggestion on the description, updated to that.
[...]

PR #1242, opened 2025-01-03
Component upgrade state.

ejbrever:
Related to https://github.com/openconfig/public/issues/1241, this is a proposal to add upgrade state information for components.

Example output:

module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--ro state
        |  +--ro name?                     string
        <......>
        |  +--ro upgrade
        |     +--ro new-firmware-version?                     string
        |     +--ro new-firmware-version-service-impacting?   boolean
        |     +--ro status?                                   identityref
        |     +--ro step?                                     string
        |     +--ro step-percent-complete?                    oc-types:percentage
        |     +--ro total-percent-complete?                   oc-types:percentage
        |     +--ro start-timestamp?                          oc-types:timeticks64
        |     +--ro duration?                                 yang:counter64
        |     +--ro stop-timestamp?                           oc-types:timeticks64
        |     +--ro last-known-failure?                       string

@robshakir @ahsaanyousaf could you please take a look?

### DIFF
--- a/release/models/platform/openconfig-platform-types.yang
+++ b/release/models/platform/openconfig-platform-types.yang
@@ -22,8 +22,13 @@ module openconfig-platform-types {
     "This module defines data types (e.g., YANG identities)
     to support the OpenConfig component inventory model.";
 
-  oc-ext:openconfig-version "1.8.0";
+  oc-ext:openconfig-version "1.9.0";
 
+  revision "2025-01-03" {
+    description
+      "Add OPENCONFIG_UPGRADE_STATUS";
+    reference "1.9.0";
+  }
 
   revision "2024-04-30" {
     description
@@ -550,6 +555,35 @@ module openconfig-platform-types {
     base COMPONENT_REBOOT_REASON;
     description
       "The component reboots due to critical errors.";
+  }
+
+  identity OPENCONFIG_UPGRADE_STATUS {
+    description
+      "Base entity for component upgrade status.";
+  }
+
+  identity UPGRADE_FILE_LOADING {
+    base OPENCONFIG_UPGRADE_STATUS;
+    description
+      "The file is being loaded on the component.";
+  }
+
+  identity UPGRADE_IN_PROGRESS {
+    base OPENCONFIG_UPGRADE_STATUS;
+    description
+      "The upgrade is in-progress.";
+  }
+
+  identity UPGRADE_COMPLETE {
+    base OPENCONFIG_UPGRADE_STATUS;
+    description
+      "The upgrade was completed.";
+  }
+
+  identity UPGRADE_FAIL {
+    base OPENCONFIG_UPGRADE_STATUS;
+    description
+      "The upgrade failed.";
   }
 
   typedef component-redundant-role {
--- a/release/models/platform/openconfig-platform.yang
+++ b/release/models/platform/openconfig-platform.yang
@@ -65,7 +65,13 @@ module openconfig-platform {
     (presence or absence of a component) and state (physical
     attributes or status).";
 
-  oc-ext:openconfig-version "0.30.0";
+  oc-ext:openconfig-version "0.31.0";
+
+  revision "2025-01-03" {
+    description
+      "Add upgrade state information.";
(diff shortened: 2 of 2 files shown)


LimeHat:
Are there other use cases (in addition to transceiver FW upgrades)?
If not, shall we consider placement of this grouping in `/components/component/transceiver` (or introduce a new `when` condition based on the component type)?

ejbrever:
@LimeHat I was thinking this could apply well to generic OS installs in which we don't have this information elsewhere either (and our workflows are still retrieving it via CLI). If folks would prefer to scope this down, my main need is around firmware upgrades on transceivers though, so that would be okay for me.

gprasat:
step-percent-complete vs total-percent-complete --> The number of steps involved in a component upgrade is not standard between components/vendors. Is there any advantage in having this represented as two different parameters ?

release/models/platform/openconfig-platform-types.yang
```diff
@@ -564,6 +569,41 @@ module openconfig-platform-types {
       "The component reboots due to critical errors.";
   }
 
+  identity OPENCONFIG_INSTALL_STATUS {
```
jsterne, line 572: What is the expected status when a device (e.g. router) first boots up and there hasn't been any firmware install on a transceiver? Suppress the install status leaf (i.e. don't return it)? Or do we need some sort of INSTALL_IDLE status?

release/models/platform/openconfig-platform-types.yang
```diff
@@ -564,6 +569,41 @@ module openconfig-platform-types {
...
+      "Base entity for component install status.";
+  }
+
+  identity INSTALL_FILE_LOADING {
+    base OPENCONFIG_INSTALL_STATUS;
+    description
+      "The file is being loaded on the component.";
+  }
+
+  identity INSTALL_IN_PROGRESS {
```
jsterne, line 583: There can be 3 separate steps in an overall install:
1) transfer
2) activate (with the option to not commit)
3) commit
Do we consider the state to be INSTALL_IN_PROGRESS from when step 1 starts until the new firmware is committed? In other words, if an operator does an activate with the "no commit" option, and then operates the optic for several hours/days without committing the newly active bank, then the install status will stay as IN_PROGRESS during those hours/days until the bank containing the new firmware is finally committed? Note that the user could switch banks multiple times before finally deciding to commit the bank containing the new firmware. And they might decide to never commit the new firmware (may instead start a new install on top of it without ever comitting it).

release/models/platform/openconfig-platform.yang
```diff
@@ -726,6 +731,83 @@ revision "2024-05-29" {
...
+          "The percent complete for the entire install process.";
+      }
+
+      leaf start-timestamp {
+        type oc-types:timeticks64;
+        description
+          "The timestamp when the install started.";
+      }
+
+      leaf duration {
```
jsterne, line 788: It feels like there is some overlap/duplication in having a start-timestamp, stop-timestamp, and duration. And would duration constantly update every second? (i.e. be very noisy if subscriber to ON_CHANGE)? I'd propose we drop the duration leaf here.

release/models/platform/openconfig-platform.yang
```diff
@@ -726,6 +731,83 @@ revision "2024-05-29" {
...
+        description
+          "Description of the current in-progress step of the install.";
+      }
+
+      leaf step-percent-complete {
+        type oc-types:percentage;
+        description
+          "The percent complete for the in-progress step.";
+      }
+
```
jsterne, line 775: The most value for % complete, start-timestamp and end-timestamp is for the transfer step. And it isn't so clear that it is realistic (or important) to know the % complete for activate, commit or abort. I'd suggest we rework this to have transfer-start-time, transfer-end-time, and transfer-percent-complete.

ElodinLaarz:
Discussed in the OC Operators Meeting Jan 6th 2026:

I see some merge conflicts, as well as some unresolved comments, can we address those and then re-raise for discussion?

PR #1248, opened 2025-01-23
Updated UDP enums related to openconfig-aft-types.

tengyiG:
### Change Scope

* Added UDPV4 and UDPV6 while keeping enums of UDP for backward compatibility.
* Yes, I kept the UDP enum.

### DIFF
--- a/release/models/aft/openconfig-aft-types.yang
+++ b/release/models/aft/openconfig-aft-types.yang
@@ -102,9 +102,8 @@ module openconfig-aft-types {
           "The encapsulation header is a VXLAN packet header";
       }
       enum UDP {
-        deprecated;
         description
-          "The encapsulation header is UDP packet header.  This node
+          "DEPRECATED. The encapsulation header is UDP packet header. This node
           is deprecated in favor of UDPV4 and UDPV6 nodes.";
       }
       enum UDPV4 {
--- a/release/models/aft/openconfig-aft-types.yang
+++ b/release/models/aft/openconfig-aft-types.yang
@@ -108,11 +108,11 @@ module openconfig-aft-types {
       }
       enum UDPV4 {
         description
-          "The encapsulation header is UDPV4 packet header.";
+          "The encapsulation header is a UDP + IPv4 header.";
       }
       enum UDPV6 {
         description
-          "The encapsulation header is UDPV6 packet header.";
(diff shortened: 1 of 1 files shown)


release/models/aft/openconfig-aft-types.yang
```diff
@@ -99,6 +99,14 @@ module openconfig-aft-types {
         description
           "The encapsulation header is UDP packet header.";
       }
```
dplore, line 101:
```suggestion
        deprecated;
        description
          "The encapsulation header is UDP packet header.  This node
          is deprecated in favor of UDPV4 and UDPV6 nodes.";
      }
```

dplore (changes requested):
Also need to bump the version to 1.3.0 here:

https://github.com/openconfig/public/pull/1248/files#diff-fefb9524b37f8f5c6637274d1eaa085ac01e294c5b3bf02ef8e4d37a12cba43bR19

dplore:
Thanks for this.  This is essentially a bug fix as it was always intended to have UDPV4 and UDPV6 enums here.  I will merge this on Feb 4th, pending OC Operator review.

release/models/aft/openconfig-aft-types.yang
```diff
@@ -97,7 +103,16 @@ module openconfig-aft-types {
...
       enum UDP {
         description
-          "The encapsulation header is UDP packet header.";
+          "DEPRECATED. The encapsulation header is UDP packet header. This node
+          is deprecated in favor of UDPV4 and UDPV6 nodes.";
+      }
+      enum UDPV4 {
+        description
+          "The encapsulation header is UDPV4 packet header.";
```
robshakir, line 111: Suggest that we say here "The encapsulation header is a UDP + IPv4 header." given that there's not really a single "UDPV4" header. We decided that combining UDP and IPv[46] together was the right thing to do since there wasn't a use case for just a UDP header with no corresponding IP header.

release/models/aft/openconfig-aft-types.yang
```diff
@@ -97,7 +103,16 @@ module openconfig-aft-types {
...
+      }
+      enum UDPV6 {
+        description
+          "The encapsulation header is UDPV6 packet header.";
```
robshakir, line 115: Same comment as above.

dplore (approved):

Issue #1251, opened 2025-02-03
Clarification on Bfd up-transitions and failure-transitions.


AnilKR123:
In the  GNMI Path: /openconfig-bfd:bfd/interfaces/interface/peers, the state container contains an attribute “failure-transitions” that tracks the number of times that the BFD session has transitioned out of the UP state. In the echo and async sub-containers, there is an attribute for “up-transitions”, which tracks the  number of times that the adjacency with the neighbor has transitioned into the UP state." I would like some clarification as to why the failure-transitions are tracked as one attribute in the state container, while there are 2 attributes tracking up-transitions, one for echo and the other for async.

dplore:
Hi @AnilKR123, I don't see a history where this is explained.  However, it looks unintentional to me?  

I would expect to see 
```
/bfd/interfaces/interface/peers/peer/state/failure-transitions
/bfd/interfaces/interface/peers/peer/state/up-transitions
```

I don't think these leaves are relevant as I think the session state would be reported at the `/bfd/interfaces/interface/peers/peer/state` container and not within the mode container?

```
/bfd/interfaces/interface/peers/peer/state/echo/up-transitions	
/bfd/interfaces/interface/peers/peer/state/async/up-transitions	
```

If you would like to see this change, please feel free to submit a pull request for review

AnilKR123:
Hi Darren,

I submitted a pr to address this: https://github.com/openconfig/public/pull/1252

Thanks,

Anil

PR #1252, opened 2025-02-05
Move the up-transitions leaf from the async and echo sub-containers into the state container.

AnilKR123:
### Change Scope

Currently there is an up-transitions leaf in both the async and echo subcontainers:

```/bfd/interfaces/interface/peers/peer/state/echo/up-transitions```
```/bfd/interfaces/interface/peers/peer/state/async/up-transitions```

I have moved this leaf into the state container, so that it is grouped with the failure-transitions leaf:

```/bfd/interfaces/interface/peers/peer/state/failure-transitions```
```/bfd/interfaces/interface/peers/peer/state/up-transitions```

 * Discussed here: [Clarification on Bfd up-transitions and failure-transitions.](https://github.com/openconfig/public/issues/1251)

### DIFF
--- a/release/models/bfd/openconfig-bfd.yang
+++ b/release/models/bfd/openconfig-bfd.yang
@@ -26,7 +26,14 @@ module openconfig-bfd {
     "An OpenConfig model of Bi-Directional Forwarding Detection (BFD)
     configuration and operational state.";
 
-  oc-ext:openconfig-version "0.3.0";
+  oc-ext:openconfig-version "0.3.1";
+
+  revision "2025-02-05" {
+    description
+      "Move the up-transitions leaf from the asynch and echo
+       sub-containers into the state container.";
+    reference "0.3.1";
+ }
 
   revision "2024-03-05" {
     description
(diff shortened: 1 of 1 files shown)


AnilKR123:
Hi @dplore, 

I have launched this pr to address the issue discussed in: https://github.com/openconfig/public/issues/1251.

Thanks,

Anil

AnilKR123:
Hi @dplore, what is pending on this PR? Merging seems to be blocked.

dplore:
This will be reviewed in the Feb 25, 2025 OC Operators meeting.  If there is no objection, we'll set to "last-call" status for any further comments and then merge after the last-call period ends (typically two weeks).

The change as proposed is a breaking change because a leaf is deleted (moved).  Unless there's a strong reason to delete this leaf, it should instead have a status field set to`deprecated`.  Please also update the description to indicate it's deprecated and what should be used in it's place so the information is easy to reference in documentation.

AnilKR123:
Thank you for the clarification. I deprecated the leaf instead of deleting it, and updated the description to reflect that as well as specify the new leaf to be used in its place.

dplore (approved):

PR #1253, opened 2025-02-10
Refactor IS-IS Graceful Restart nodes

earies:
    - Refactor IS-IS GR leaves according to RFC, remove defaults
    - Removal of non-standard `interface-time-expirations` leaf

### Change Scope

This change involves:

* Removal of any graceful-restart timer defaults.  There are no such defaults
  agreed upon across implementations and usage can vary per use-case.
* Addition of `units` statements to represent correct time unit
* Refactor grouping names to align w/ the usage (level/system)
* Removal of `interface-time-expirations` leaf as there is only 1 referenced
  implementation.  Suggestion would be an augmentation or usage of native
  modeling in such cases vs. everyone else deviating.
* Correct description statements/RFC references

This change is backwards incompatible though likely reduced impact due to
implementation status.

### Platform Implementations

* RFC compliance
* Removal of `interface-time-expirations` leaf as there was only 1 prior
  implementation reference.

### DIFF
--- a/release/models/isis/openconfig-isis.yang
+++ b/release/models/isis/openconfig-isis.yang
@@ -54,7 +54,15 @@ module openconfig-isis {
             +-> { levels config }
             +-> { level adjacencies }";
 
-  oc-ext:openconfig-version "1.7.0";
+  oc-ext:openconfig-version "2.0.0";
+
+  revision "2025-02-10" {
+    description
+      "Remove graceful-restart restart-time defaults and update
+      description statements, reference and grouping naming.  Removal of
+      non-standard interface-time-expirations leaf.";
+    reference "2.0.0";
+  }
 
   revision "2024-02-28" {
     description
@@ -938,29 +946,24 @@ module openconfig-isis {
         "When this leaf is set to TRUE, planned restart procedures as
         described in RFC 8706 are not used.";
       reference
-      "RFC 5706: Restart Signaling for IS-IS";
+        "IETF RFC 8706, Section 3.2: Restart Signaling for IS-IS";
     }
-
-    reference
-      "RFC 5306: Restart Signaling for IS-IS; RFC 5706: Restart Signaling
-      for IS-IS";
   }
 
-  grouping isis-graceful-restart-level-config {
+  grouping isis-graceful-restart-global-level-config {
     description
       "This grouping defines ISIS graceful restart configuration relevant
-      for ISIS level/LSDB";
+      for ISIS at global system and per-level/LSDB hierarchies.";
 
     leaf restart-time {
       type uint16;
-      default 30;
+      units "seconds";
       description
-        "Value of RFC5306/RFC8706 T2 timer";
+        "Value of T2 or T3 timers at either the LSDB/level or global
+        system context.";
+      reference
+        "IETF RFC 8706, Section 3.1: Restart Signaling for IS-IS";
     }
-
-    reference
-      "RFC 5306: Restart Signaling for IS-IS; RFC 5706: Restart Signaling
-      for IS-IS";
   }
 
   grouping isis-graceful-restart-interface-config {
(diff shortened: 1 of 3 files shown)


dplore:
@rgwilton @nandanarista @limehat any comments on this one?

release/models/isis/openconfig-isis.yang
```diff
@@ -938,29 +946,24 @@ module openconfig-isis {
...
-      for ISIS level/LSDB";
+      for ISIS at global system and per-level/LSDB hierarchies.";
 
     leaf restart-time {
       type uint16;
-      default 30;
+      units "seconds";
       description
-        "Value of RFC5306/RFC8706 T2 timer";
+        "Value of T2 or T3 timers at either the LSDB/level or global
```
rszarecki, line 962: it can't be T2 OR T3. This is ambigous. 

The RFC8706 state:
```
This is initialized to 65535 seconds but is set to the minimum of the remaining times of received IIHs containing a Restart TLV with the Restart Acknowledgement (RA) set and an indication that the neighbor has an adjacency in the UP state to the restarting router.
```
This is NOT configurable timer. it is always initialized to 2^16, and the adjusted as derived from adjacencies.

This **IS T2**.

release/models/isis/openconfig-isis.yang
```diff
@@ -970,21 +973,12 @@ module openconfig-isis {
 
     leaf interface-timer {
       type uint16;
+      units "seconds";
       description
-        "Value of RFC5306/RFC8706 T1 timer";
-    }
-
-    leaf interface-time-expirations {
```
rszarecki, line 977: Why we need to remove it and make this change breaking?

- Implementation may opmit it if do not support. it is NOT a mandatory leaf.
- this is RFC8706 standard attribute  "Therefore, timer T1 is canceled after some predetermined number of expirations (which MAY be 1)". 
As it is not commonly supported as configuration option I would not include it in OpenConfig in first place. But since it is alredy here, IMHO removal is not worth of breaking change.

earies: As mentioned in the description, there is only evidence of 1 implementation which in itself does not justify addition.  If there are more, I would like to hear from implementors.  As far as breaking, that implementation can choose to deliver via native modeling or an augmentation should folks want to absorb augmentations.

Sure - implementors can treat many nodes as optional but the reality is it exists, was published without the gating factors of >= 2 implementations and folks choose not to absorb deviations creating a messy perception and unwarranted demand towards compliance often.

Note the previous datatype is also an unbound `int64`.  If a state leaf is warranted, I'd suggest a separate PR to bring that back w/ proper review.

release/models/isis/openconfig-isis.yang
dplore: in other words, this should be a state only leaf?  Is that right?

ElodinLaarz: 1. We need to update the description to only reference `T2`, if that is the expectation?
2. Do we want to remove the config leaf as part of this PR? (If this is going to be a breaking change, already.)

release/models/isis/openconfig-isis.yang
ElodinLaarz: Cleaning up a model via a breaking change is probably unnecessary here?

An option could be to deprecate the leaf, if we just want to signal "we don't want more vendors to implement this," if there's no intent to use it going forwards. I don't know of any use cases, myself.

earies: Fair - addressed via: https://github.com/openconfig/public/pull/1253/commits/3f6e78ff40adca704746914ccd07116dc67851b3 and version changed to minor `1.8.0` bump vs. `2.0.0`

release/models/isis/openconfig-isis.yang
rszarecki: - This shall be configurable T2 timer. As it was before this PR.

- It cant be "T2 or T3" because then it is unclear which one it is. There is a reason why RFC has T2 distinct then T3.
- T3 is not configurable. IF we want to expose it, it shall be RO and it shall be new leaf.

PR #1260, opened 2025-02-25
Add decap mpls-in-gre

dplore:
### Change Scope

* Add a leaf to configure a policy-forwarding action to decapsulate MPLS in GRE packets.

`/network-instances/network-instance/policy-forwarding/policies/policy/rules/rule/action/config/decapsulate-mpls-in-gre`

### Platform Implementations

* Arista EOS - decap-group
* EOS native config creates a system wide configuration to decap packets with a specific destination IP address (decap-ip)
```
switch(config)#ip decap-group decap-gp-1
switch(config)#tunnel type gre
switch(config)#tunnel decap-ip 100.1.1.2
switch(config)#tunnel overlay mpls qos map mpls-traffic-class to traffic-class
```

 * [Cisco IOS XR - MPLS in GRE decap](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/mpls/24xx/configuration/guide/b-mpls-cg-cisco8000-24xx.pdf)
 Cisco IOS XR native config requires a tunnel interface in addition to a Cisco PBR rule.  
 ```
Router # configure
Router(config)# interface tunnel-ip1
Router(config-if)# ipv4 address 112.0.0.2 255.255.255.0
Router(config-if)# tunnel mode gre ipv4 decap
Router(config-if)# tunnel source Loopback 0
Router(config-if)# tunnel destination 10.0.0.1

! PBR Configuration for GRE Tunnel Decapsulation
Router(config)# class-map type traffic match-all test_gre1
Router(config-cmap)# match protocol gre
Router(config-cmap)# match destination-address ipv4 50.0.0.1 255.255.255.255
Router(config-cmap)# match source-address ipv4 10.0.0.1 255.255.255.255
Router(config-cmap)# end-class-map
Router(config)# policy-map type pbr P1-test
Router(config-pmap)# class type traffic test_gre1
Router(config-pmap-c)#decapsulate gre
Router(config-pmap-c)# end-policy-map
Router(config)# vrf-policy vrf default address-family ipv4 policy type pbr input P1-test
```

### DIFF
--- a/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
+++ b/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
@@ -21,7 +21,13 @@ submodule openconfig-pf-forwarding-policies {
     "This submodule contains configuration and operational state
     relating to the definition of policy-forwarding policies.";
 
-  oc-ext:openconfig-version "0.6.1";
+  oc-ext:openconfig-version "0.7.0";
+
+  revision "2025-02-21" {
+    description
+      "Add MPLS-in-GRE decapsulation action.";
+    reference "0.7.0";
+  }
 
   revision "2023-04-25" {
     description
@@ -385,6 +391,16 @@ submodule openconfig-pf-forwarding-policies {
       description
         "When this leaf is set to true, the local system should remove
         the UDP header from the packet matching the rule.
+        Following the decapsulation it should subsequently forward the
+        encapsulated packet according to the underlying MPLS label.";
+    }
+
+    leaf decapsulate-mpls-in-gre {
+      type boolean;
+      default false;
+      description
+        "When this leaf is set to true, the local system should remove
+        the GRE header from the packet matching the rule.
         Following the decapsulation it should subsequently forward the
         encapsulated packet according to the underlying MPLS label.";
     }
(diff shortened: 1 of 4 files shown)


rgwilton:
Please can you clarify why this needs to be a specific mpls-in-gre-decaps, and not just a gre-decaps.  E.g., looking at your Cisco config example there isn't any mention of MPLS config, although perhaps that is a slightly different config example.

dplore:
> Please can you clarify why this needs to be a specific mpls-in-gre-decaps, and not just a gre-decaps. E.g., looking at your Cisco config example there isn't any mention of MPLS config, although perhaps that is a slightly different config example.

I believe this is needed to provide a hint that the payload is expected to be MPLS and forwarded as such. However, it is a good question for implementation owners to answer.  Do you need this hint for MPLS or is the existing [decapsulate-gre ](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-policy-forwarding-policies-policy-rules-rule-action-config-decapsulate-gre)  sufficient?

arkaab (approved):

vishnureddybadveli (approved):
Changes are good to me.

dplore:
As GRE has a[ protocol type field](https://datatracker.ietf.org/doc/html/rfc2784#section-2.4), it is assumed that the node which does GRE encapsulation would set that to the Ethernet MPLS protocol value (`0x8847`) and this path is not needed. I will close this PR.  If there is some need for it, please feel free to add a comment.

vishnureddybadveli:
+1 to darren.

PR #1264, opened 2025-03-06
Add interface-transitions and link-transitions counters

rgwilton:
### Change Scope

* Deprecates carrier-transitions counter (because it has the wrong name), replacing with interface-transitions counter with the same definition.
* Add a new link-transitions counter that reports the number of underlying link transitions (i.e., matching the historical vendor definition of carrier-transitions).

### Platform Implementations

 * Cisco, IOS XR:
 * https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/Interfaces/b-interfaces-hardware-component-cr-8000/m-global-interface-commands.html, "interface transitions" & "carrier transitions" 
 * From the discussion last month it sounded like JunOS also has similar counters available.  At least one of them is published as "carrier transitions"

Tree change:
```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw config
        +--ro state
        |  +--ro counters
        |     x--ro carrier-transitions?     oc-yang:counter64
        |     +--ro interface-transitions?   oc-yang:counter64
        |     +--ro link-transitions?        oc-yang:counter64
```

### DIFF
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -766,8 +766,7 @@ module openconfig-interfaces {
 
         Discontinuities in the value of this counter can occur
         at re-initialization of the management system, and at
-        other times as indicated by the value of
-        'last-clear'.";
+        other times as indicated by the value of 'last-clear'.";
       reference
         "RFC 2863: The Interfaces Group MIB - ifHCInOctets.
         RFC 4293: Management Information Base for the
@@ -795,8 +794,7 @@ module openconfig-interfaces {
 
         Discontinuities in the value of this counter can occur
         at re-initialization of the management system, and at
-        other times as indicated by the value of
-        'last-clear'.";
+        other times as indicated by the value of 'last-clear'.";
       reference
         "RFC 2863: The Interfaces Group MIB - ifHCInUcastPkts.
         RFC 4293: Management Information Base for the
@@ -812,8 +810,7 @@ module openconfig-interfaces {
 
         Discontinuities in the value of this counter can occur
         at re-initialization of the management system, and at
-        other times as indicated by the value of
-        'last-clear'.";
+        other times as indicated by the value of 'last-clear'.";
       reference
         "RFC 2863: The Interfaces Group MIB - ifHCInBroadcastPkts.
         RFC 4293: Management Information Base for the
@@ -830,8 +827,7 @@ module openconfig-interfaces {
 
         Discontinuities in the value of this counter can occur
         at re-initialization of the management system, and at
-        other times as indicated by the value of
-        'last-clear'.";
+        other times as indicated by the value of 'last-clear'.";
       reference
         "RFC 2863: The Interfaces Group MIB - ifHCInMulticastPkts.
         RFC 4293: Management Information Base for the
@@ -851,8 +847,7 @@ module openconfig-interfaces {
 
         Discontinuities in the value of this counter can occur
         at re-initialization of the management system, and at
-        other times as indicated by the value of
-        'last-clear'.";
+        other times as indicated by the value of 'last-clear'.";
       reference
         "RFC 2863: The Interfaces Group MIB - ifInErrors.
         RFC 4293: Management Information Base for the
@@ -870,9 +865,7 @@ module openconfig-interfaces {
 
         Discontinuities in the value of this counter can occur
         at re-initialization of the management system, and at
-        other times as indicated by the value of
-        'last-clear'.";
-
+        other times as indicated by the value of 'last-clear'.";
 
       reference
         "RFC 2863: The Interfaces Group MIB - ifInDiscards.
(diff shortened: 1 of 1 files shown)


dplore:
Added to March 11, 2025 OC operator review

jsterne:
We should consider clarifying the description to indicate whether this counts down->up + up->down or just down->up.   Although I don't think there is consistency on this amongst vendors.

release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -1055,6 +1075,21 @@ module openconfig-interfaces {
       oc-ext:telemetry-on-change;
     }
 
+    leaf link-transitions {
+      type oc-yang:counter64;
+      description
+        "This is the number of times that the underlying link state
+        (e.g., at the optical receiver) has transitioned to/from
```
dplore, line 1082: Based on feedback from operators meeting, we should change this to be specifically "and".  This way it's clear and consistent about what is being counted.     The language could be:

"from up to down and from down to up" to be more clearly worded.

rgwilton: (+@jsterne) I've reworked the descriptions to try and be a lot more explicit on the expected behaviour, please can you check if you think that this text is better.

A couple more points:

1. I've not added either of these counters to sub-interfaces because the existing carrier-transitions counter is already marked as being deprecated.

2. link-transitions is marked as oc-ext:telemetry-on-change, since being notified that the underlying interface is not completely stable may be worthy of being notified of.

release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -1048,10 +1055,56 @@ module openconfig-interfaces {
...
+        "The total number of times the interface state (oper-status)
+        has either transitioned to 'UP' state from any other state, or
+        from state 'UP' to any other state.  I.e., an interface flap
+        from UP to DOWN back to UP increments the counter by 2.
+        Transitions between any other interface states other than to
+        or from 'UP' state are not included in the counter.
+
+        Discontinuities in the value of this counter can occur
+        when the device is re-initialization as indicated by the
+        value of 'last-clear'.";
```
jsterne, line 1082: re-initialized iso re-initialization?
Does this discontinuity note also apply to link-transitions?

release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -1048,10 +1055,56 @@ module openconfig-interfaces {
...
+        has been applied that could suppress an update to the interface
+        'oper-status' and corresponding interface-transitions counter.
+
+        The counter is incremented both when the link transitions
+        to 'UP' state from any other link state and also when the link
+        transitions from 'UP' state to any other link state, i.e., an
+        interface flap from UP to DOWN back to UP increments the
+        counter by 2.
+
+        Implementations are not required to count all transitions,
```
jsterne, line 1101: Did Jeff's point about this possibly also apply to interface-transitions? He brought up some point (I think) about the system coming up or maybe a linecard coming up?

release/models/interfaces/openconfig-interfaces.yang
```diff
@@ -1048,10 +1055,56 @@ module openconfig-interfaces {
...
+        e.g., if they are below the level of granularity monitored by
+        the system, and hence may not tally with the equivalent counter
+        on the remote end of the link.
+
+        Reported relative to the time the device restarted or
+        the last-clear time, whichever is most recent.";
```
jsterne, line 1107: Does this 'relative to time' also apply to interface-transitions?

jsterne:
@rgwilton I like your wording around UP and DOWN.

release/models/interfaces/openconfig-interfaces.yang
rgwilton: Good spot.  I just directly copied this text from in-fcs-errors.  I will fix in both places.

release/models/interfaces/openconfig-interfaces.yang
rgwilton: I don't think that it should.

Link transitions could happen at high frequency and there are good reasons that such events might get lost or be suppressed and not necessarily truly represent the number of transitions.

But I think that interface-transitions is different, this counter should only be changed when the interface-state field is changed, which I think systems should be able to properly sequence.  Otherwise it feels like it would be getting too much into implementation details.

release/models/interfaces/openconfig-interfaces.yang
rgwilton: I will check and align this text.

rgwilton:
@dplore, Darren, is it possible to mark this for last-call please?

release/models/interfaces/openconfig-interfaces.yang
dplore: Agreed, all interface transitions are expected to be set and observed by the overall device management and control software.

dplore (approved):

PR #1265, opened 2025-03-12
Use IP address-family specific types for udp-v4 and udp-V6 encapsulations in AFT

danameme:
### Change Scope

* This change is based on review comment on [PR#1234](https://github.com/openconfig/public/pull/1234) about the re-use of `udp-v4` encapsulation header container.
* Use `ipv4-address` type for src-ip and dst-ip fields in `udp-v4` encapsulation header.
* Use `ipv6-address` for src-ip and dst-ip fields in `udp-v6` encapsulation header.
* The new types and old `ip-address` are both strings.

### Tree View

```diff
@@ -3594,50 +3594,50 @@
         |  |     |     |  +--ro type?    oc-aftt:encapsulation-header-type
         |  |     |     +--ro gre
         |  |     |     |  +--ro state
         |  |     |     |     +--ro src-ip?   oc-inet:ip-address
         |  |     |     |     +--ro dst-ip?   oc-inet:ip-address
         |  |     |     |     +--ro ttl?      uint8
         |  |     |     +--ro ipv4
         |  |     |     |  +--ro state
         |  |     |     |     +--ro src-ip?   oc-inet:ip-address
         |  |     |     |     +--ro dst-ip?   oc-inet:ip-address
         |  |     |     +--ro ipv6
         |  |     |     |  +--ro state
         |  |     |     |     +--ro src-ip?   oc-inet:ip-address
         |  |     |     |     +--ro dst-ip?   oc-inet:ip-address
         |  |     |     +--ro mpls
         |  |     |     |  +--ro state
         |  |     |     |     +--ro traffic-class?      oc-mplst:mpls-tc
         |  |     |     |     +--ro mpls-label-stack*   oc-mplst:mpls-label
         |  |     |     +--ro udp-v4
         |  |     |     |  +--ro state
-        |  |     |     |     +--ro src-ip?         oc-inet:ip-address
-        |  |     |     |     +--ro dst-ip?         oc-inet:ip-address
+        |  |     |     |     +--ro src-ip?         oc-inet:ipv4-address
+        |  |     |     |     +--ro dst-ip?         oc-inet:ipv4-address
         |  |     |     |     +--ro dscp?           oc-inet:dscp
         |  |     |     |     +--ro src-udp-port?   oc-inet:port-number
         |  |     |     |     +--ro dst-udp-port?   oc-inet:port-number
         |  |     |     |     +--ro ip-ttl?         uint8
         |  |     |     +--ro udp-v6
         |  |     |     |  +--ro state
-        |  |     |     |     +--ro src-ip?         oc-inet:ip-address
-        |  |     |     |     +--ro dst-ip?         oc-inet:ip-address
+        |  |     |     |     +--ro src-ip?         oc-inet:ipv6-address
+        |  |     |     |     +--ro dst-ip?         oc-inet:ipv6-address
         |  |     |     |     +--ro dscp?           oc-inet:dscp
         |  |     |     |     +--ro src-udp-port?   oc-inet:port-number
         |  |     |     |     +--ro dst-udp-port?   oc-inet:port-number
         |  |     |     |     +--ro ip-ttl?         uint8
         |  |     |     +--ro vxlan
         |  |     |        +--ro state
         |  |     |           +--ro vni-label?               oc-evpn-types:evi-id
         |  |     |           +--ro tunnel-src-ip-address?   oc-inet:ip-address
         |  |     +--ro interface-ref
         |  |        +--ro state
         |  |           +--ro interface?      -> /oc-if:interfaces/interface/name
         |  |           +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
         |  +--ro oc-aftsummary:aft-summaries
         |     +--ro oc-aftsummary:ipv4-unicast
         |     |  +--ro oc-aftsummary:protocols
         |     |     +--ro oc-aftsummary:protocol* [origin-protocol]
         |     |        +--ro oc-aftsummary:origin-protocol    -> ../state/origin-protocol
         |     |        +--ro oc-aftsummary:state
         |     |           +--ro oc-aftsummary:origin-protocol?   identityref
         |     |           +--ro oc-aftsummary:counters

```

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,13 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "2.8.0";
+  oc-ext:openconfig-version "2.9.0";
+
+  revision "2025-03-12" {
+    description
+      "Use IP address-family specific types for UPD-V4 and UDP-V6 encapsulations in AFT.";
+    reference "2.9.0";
+  }
 
   revision "2024-09-19" {
     description
@@ -376,7 +382,7 @@ submodule openconfig-aft-common {
                   "State parameters relating to UDP in IPv4 encapsulation
                   headers.";
 
-                uses aft-common-entry-nexthop-encap-udp-state;
+                uses aft-common-entry-nexthop-encap-udp-v4-state;
               }
             }
 
@@ -392,7 +398,7 @@ submodule openconfig-aft-common {
                   "State parameters relating to UDP in IPv6 encapsulation
                   headers.";
 
-                uses aft-common-entry-nexthop-encap-udp-state;
+                uses aft-common-entry-nexthop-encap-udp-v6-state;
               }
             }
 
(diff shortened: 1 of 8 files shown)


dplore:
Thanks Dan for fixing this.  This was missed on our earlier reviews.  Fixing this should in practice be a non-breaking change.

Setting to last call for March 18, 2025.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -679,20 +685,78 @@ submodule openconfig-aft-common {
   }
 
 
-  grouping aft-common-entry-nexthop-encap-udp-state {
+  grouping aft-common-entry-nexthop-encap-udp-v4-state {
```
earies, line 688: Suggest just branching off where things differ into its own groupings

e.g. ipv4 src/dst in one, ipv6 src/dst in another

Then have a common grouping where things do not differ and use accordingly (reducing duplication where you dont have to)

earies:
> Thanks Dan for fixing this. This was missed on our earlier reviews. Fixing this should in practice be a non-breaking change.
> 
> Setting to last call for March 18, 2025.

LGTM other than a quick comment I provided on reducing duplication

YANG-wise, this is backwards incompatible but implementation wise, I highly doubt this is breaking anything today.... If we are not going to follow YANG backwards compatible rules 100% then we should work on defining what types of changes are/arent compat so expectations are set accordingly

dplore:
> > Thanks Dan for fixing this. This was missed on our earlier reviews. Fixing this should in practice be a non-breaking change.
> > Setting to last call for March 18, 2025.
> 
> LGTM other than a quick comment I provided on reducing duplication
> 
> YANG-wise, this is backwards incompatible but implementation wise, I highly doubt this is breaking anything today.... If we are not going to follow YANG backwards compatible rules 100% then we should work on defining what types of changes are/arent compat so expectations are set accordingly

Yes, it is formally a breaking change, because the regex is now more restrictive than before.  But in practice and when used properly, a mismatched address family IP address should not be used anyways.    In addition, this is so new as to not be implemented yet in any production code yet.

Regarding yang model version numbering and OC release versioning, what does the community think?  Shall we increment the major version so we are strictly consistent?   This would result in major increment for the model to v3.0.0 and OC release (when created) to 6.0.0. (because OC release 5.0.0 does contain the v2.8.0 openconfig-aft-common.yang model with these ip-address leaves).

release/models/aft/openconfig-aft-common.yang
danameme: Ok, I've updated as suggested. Thanks.

dplore (approved):

PR #1266, opened 2025-03-17
Clarify node deprecation intent

dplore:
Clarify the intent of the status deprecate node by referencing the yang 1.0 standard and stating the operational expectations for deprecated status nodes.

### DIFF
--- a/doc/releases.md
+++ b/doc/releases.md
@@ -71,14 +71,20 @@ As a side note, these releases are compatible with the notion of
 
 6.  Wherever possible, it is RECOMMENDED to make backward compatible API changes
     (e.g. deprecating leaves via the
-    [status statement](https://www.rfc-editor.org/rfc/rfc7950#section-7.21.2))
+    [status statement](https://datatracker.ietf.org/doc/html/rfc6020#section-7.19.2))
     for at least one minor release prior to a non-backward compatible API change
     in order to ease the transition to the new API. These leaves are then
     expected to be removed or modified in the next major version release. NOTE:
     This guideline may change once OpenConfig operators gain more experience
     managing breaking changes.
 
-7.  Release documentation should include the list of models and their version
+7. [Deprecated](https://datatracker.ietf.org/doc/html/rfc6020#section-7.19.2)
+   nodes must be supported until they are deleted from the models. The deprecated
+   status serves as a hint that the node will be removed in some future release of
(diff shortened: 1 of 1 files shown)


robshakir (approved):

doc/releases.md
```diff
@@ -71,14 +71,20 @@ As a side note, these releases are compatible with the notion of
...
+    [status statement](https://datatracker.ietf.org/doc/html/rfc6020#section-7.19.2))
     for at least one minor release prior to a non-backward compatible API change
     in order to ease the transition to the new API. These leaves are then
     expected to be removed or modified in the next major version release. NOTE:
     This guideline may change once OpenConfig operators gain more experience
     managing breaking changes.
 
-7.  Release documentation should include the list of models and their version
+7. [Deprecated](https://datatracker.ietf.org/doc/html/rfc6020#section-7.19.2)
+   nodes must be supported until they are deleted from the models. The deprecated
```
earies, line 82: Assume the following scenario:

The model was never implemented but now implementation begins and the model contains a deprecated node in conjunction with the forward transition path (replacement node) or it is deprecated without any replacement (no longer needed)

The implementor could very well choose to just choose the forward path and not implement the deprecated node unless there is some ultimate reason for backwards compatibility.

Should this then be a "must" ?

earies: Also "must" here seems to imply all implementations have to support that node (there could be reasons why the implementation cannot support the node irregardless of the status stmt)

dplore: For the general case, we'd like to have simultaneous support for both the old and new leaves to facilitate a transition operationally.  This transition period is likely long lived (as we haven't deleted deprecated nodes yet in OpenConfig).

If a required leaf is impossible for a device to support, then it should be dealt with a yang deviation removing the leaf.  This can be painful for operators to support.  In our case we generate the expected OC (containing a deprecated leaf) to use for all devices.  When one of the devices doesn't support the standard (deprecated)  leaffor configuration, the alternate (new) OC can be generated and replaces the expected (old, deprecated) OC (in a gnmi query in our case) before pushing to a device.   

Of course, one or two is not a big deal, but we have a few hundred so far managed in as a list of "deviations" in the [featureprofiles deviation module](https://github.com/openconfig/featureprofiles/blob/main/internal/deviations/deviations.go).

Issue #1254, opened 2025-02-10
Bfd failure-transitions leaf.


AnilKR123:
In the GNMI Path: /openconfig-bfd:bfd/interfaces/interface/peers/peer/state, the ‘failure-transitions’ leaf is defined as the number of times that a Bfd session has transitioned out of the UP state. This means that it includes transitions from both UP -> DOWN and UP -> ADMIN_DOWN. When a Bfd session is in the ADMIN_DOWN state, this means that the session was administratively disabled, so this is not a session failure. Should failure-transitions be changed to only track session failures (UP -> DOWN transitions)?

dplore:
IMO, yes, [failure-transitions](https://github.com/openconfig/public/blob/a1722856e7e576484eaebd43129269c5298f81b9/release/models/bfd/openconfig-bfd.yang#L323-L328) should only track transitions which are due to a failure and not administratively down session transitions. 
 I'd appreciate any additional community feedback.

dplore:
Set last call to Feb 25, 2025 so both OC operator reviews will have a chance to comment

PR #1267, opened 2025-03-18
Update BFD documentation to clarify session failure.

AnilKR123:
### Change Scope

The current definition of a BFD session failure in OpenConfig is when the session transitions out of the UP state. This includes transitions from both UP -> DOWN and UP -> ADMIN_DOWN. When a BFD session is administratively disabled, this is not a session failure. Session failures should only track UP -> DOWN transitions. I have updated the descriptions of the ‘failure-transitions’ and ‘last-failure-time’ leaves to clarify this.

Discussed in #1254

### DIFF
--- a/release/models/bfd/openconfig-bfd.yang
+++ b/release/models/bfd/openconfig-bfd.yang
@@ -31,7 +31,9 @@ module openconfig-bfd {
   revision "2025-03-18" {
     description
       "Clarify BFD session failure as a transition
-      from the UP state to the DOWN state.";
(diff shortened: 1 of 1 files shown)


dplore (commented):
Could you also expand on the description to highlight that admin down is not included?

dplore:
Reviewed in Mar 25, 2025 OC operators meeting without objection.  Set for last call April 9, 2025

dplore (approved):

PR #1269, opened 2025-03-19
Add accept action to policy-forwarding supported actions

ihebboubaker:
### Change Scope

* The 'accept' action is to be included in the set of actions supported by policy-forwarding.

### DIFF
--- a/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
+++ b/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
@@ -21,7 +21,13 @@ submodule openconfig-pf-forwarding-policies {
     "This submodule contains configuration and operational state
     relating to the definition of policy-forwarding policies.";
 
-  oc-ext:openconfig-version "0.7.0";
+  oc-ext:openconfig-version "0.8.0";
+
+  revision "2025-03-19" {
+    description
+      "Added accept action to policy-forwarding actions rules.";
+    reference "0.8.0";
+  }
 
   revision "2024-11-14" {
     description
@@ -267,6 +273,14 @@ submodule openconfig-pf-forwarding-policies {
       default false;
       description
         "When this leaf is set to true, the local system should drop
+        packets that match the rule.";
+    }
+
+    leaf accept {
+      type boolean;
+      default false;
+      description
+        "When this leaf is set to true, the local system should accept
         packets that match the rule.";
     }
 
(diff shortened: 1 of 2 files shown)


release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -270,6 +276,14 @@ submodule openconfig-pf-forwarding-policies {
         packets that match the rule.";
     }
 
+    leaf accept {
```
robshakir, line 279: What happens when this leaf is set to false? Packets should be dropped?

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -270,6 +276,14 @@ submodule openconfig-pf-forwarding-policies {
...
+      type boolean;
+      default false;
```
robshakir, line 281: I don't agree that the default should be false -- it will mean that any existing policy-forwarding rule will automatically start dropping packets, which is going to be harmful. We should consider:

 * having no default -- since we've found that having a default value creates an implicit specification.
 * making this a `drop` leaf since it seems more common that one would want to forward a packet vs. drop it.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -270,6 +276,14 @@ submodule openconfig-pf-forwarding-policies {
...
+      description
```
robshakir, line 282: I'd like some clarification either in this leaf, or elsewhere, as to how this coexists with ACLs. What is the order of operations -- policy-forwarding and then ACL or ACL then policy-forwarding? Can both be used on a particular link?

robshakir (changes requested):

dplore (commented):
In routing-policy there are [ACCEPT, REJECT and NEXT_STATEMENT](https://github.com/openconfig/public/blob/50a6b4075923a2cf8b46b86f1f3cd7ecb17ea0f8/release/models/policy/openconfig-routing-policy.yang#L203C11-L222) results per policy.  NEXT_STATEMENT is default.  Perhaps we should follow a similar pattern here.

LimeHat (commented):
Can you please explain the use case? It is not clear what is the purpose of the proposed action.

dplore:
@ihebboubaker please do state some operational use cases where the existing model doesn't meet the needs.

PR #1273, opened 2025-03-30
Auto generated link bandwidth

rszarecki:
### Change Scope

Introduces OpenConfig schema to enable automatic generation of link-bandwidth extended community based on underlaying interface speed and peer-group and neighbour level.
Not available on global level because it is applicable only for sessions soucerd to physical interface, and not applicable to interfaces of uncnown speed like "loopback", "irb/svi", etc.
```
  +--rw network-instances
        +--rw protocols
              +--rw bgp
                 +--rw neighbors
                    +--rw neighbor* [neighbor-address]
                       +--rw auto-link-bandwidth
                          +--rw import
                             +--rw config
                             |  +--rw enabled?          boolean
                             |  +--rw hold-down-time?   uint16
                             |  +--rw transitive?       boolean
                             +--rw state
                                +--rw enabled?          boolean
                                +--rw hold-down-time?   uint16
                                +--rw transitive?       boolean
```
```
  +--rw network-instances
        +--rw protocols
              +--rw bgp
                 +--rw peer-groups
                    +--rw peer-group* [peer-group-name]
                       +--rw auto-link-bandwidth
                          +--rw import
                             +--rw config
                             |  +--rw enabled?          boolean
                             |  +--rw hold-down-time?   uint16
                             |  +--rw transitive?       boolean
                             +--rw state
                                +--rw enabled?          boolean
                                +--rw hold-down-time?   uint16
                                +--rw transitive?       boolean
```
The schema is extedable to export direction and link-bandwidth regenaration.

This change is backward compatible.

#### JUNIPER JUNOS, JUNOS-EVO [link-bandwidth autosense](https://www.juniper.net/documentation/us/en/software/junos/ai-ml-evo/topics/topic-map/bgp-link-bandwidth.html#concept_kjn_vyx_2bc__section_a1p_ngh_gcc)
```
set protocols bgp group group-name link-bandwith auto-sense
set protocols bgp group group-name neighbor link-bandwith auto-sense
set protocols bgp group group-name link-bandwith auto-sense hold-down time-in-seconds
```
#### CISCO IOS-XR [DMZ link-bandwidth](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/bgp/cumulative/command/reference/b-bgp-cr-cisco8000/m-bgp-commands-8k.html#wp2318208407)
```
Router(config)# router bgp 1
Router(config-bgp)#neighbor 10.67.89.01 
Router(config-bgp-nbr)#dmz-link-bandwidth 
```
ARISTA EoS [Enabling BGP Link Bandwidth attribute setting inbound](https://www.arista.com/en/support/toi/eos-4-24-1f/14534-bgp-ucmp#enabling-bgp-link-bandwidth-attribute-setting-inbound)
```
[no|default] neighbor <peer|peer-group> link-bandwidth auto
[no|default] neighbor <peer|peer-group> link-bandwidth update-delay <0-600>
```

### DIFF
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -346,7 +346,7 @@ submodule openconfig-bgp-common-structure {
         uses auto-gen-link-bandwidth-config;
       }
       container state {
-        description: "";
+        description "";
         uses auto-gen-link-bandwidth-config;
       }
     }
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -351,18 +351,18 @@ submodule openconfig-bgp-common-structure {
         configuration.";
       container config {
         description
-          "Container used to provides configuration
+          "Container used to provide configuration
           parameters related to auto-generation of link-bandwidth
-          extended-community attribute for lerned routes.
-          Apply only for routes that do not have link-bandwidth
+          extended-community attribute for learned routes.
+          Applies only for routes that do not have link-bandwidth
           extended-community.";
         uses auto-gen-link-bandwidth-config;
       }
       container state {
         description
-          "Container used to provides state
+          "Container used to provide state
           parameters related to auto-generation of link-bandwidth
-          extended-community attribute for lerned routes.
+          extended-community attribute for learned routes.
           Apply only for routes that do not have link-bandwidth
           extended-community.";
         uses auto-gen-link-bandwidth-config;
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -338,7 +338,7 @@ submodule openconfig-bgp-common-structure {
       provides configuration and state parameters relating to
       auto-generation of link-bandwidth extended-community
       attribute.";
-      
+
     container import {
       description
         "Container used to provide configuration and state
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -325,4 +325,31 @@ submodule openconfig-bgp-common-structure {
     }
   }
 
+   grouping bgp-common-structure-neighbor-group-auto-gen-link-bandwidth {
+    description:
+      "Common grouping used for both groups and neighbors which
+      provides configuration and state parameters relating to
+      auto-generation of link-bandwidth extended-community
+      attribute.";
+    container import {
+      description: 
+        "Container used to provides configuration and state
+        parameters relating tovauto-generation of link-bandwidth 
+        extended-community attribute for lerned routes.
+        Apply only for routes that do not have link-bandwidth 
+        extended-community.
+        Any policy action related to link-bandwidth 
+        extended-community takes precedence over this
+        configuration";
+      container config {
+        description: "";
+        uses auto-gen-link-bandwidth-config;
+      }
+      container state {
+        description: "";
+        uses auto-gen-link-bandwidth-config;
+      }
+    }
+  }
+  
 }
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -359,6 +359,7 @@ submodule openconfig-bgp-common-structure {
           extended-community.";
         uses auto-gen-link-bandwidth-config;
       }
+
       container state {
         description
           "Container used to provide state
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -344,7 +344,7 @@ submodule openconfig-bgp-common-structure {
         "Container used to provide configuration and state
         parameters related to auto-generation of link-bandwidth
         extended-community attribute for learned routes.
-        Apply only for routes that do not have link-bandwidth
+        Applies only for routes that do not have link-bandwidth
         extended-community.
         Any policy action related to link-bandwidth
         extended-community takes precedence over this
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -349,6 +349,7 @@ submodule openconfig-bgp-common-structure {
         Any policy action related to link-bandwidth
         extended-community takes precedence over this
         configuration.";
+
       container config {
         description
           "Container used to provide configuration
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -326,13 +326,13 @@ submodule openconfig-bgp-common-structure {
   }
 
    grouping bgp-common-structure-neighbor-group-auto-gen-link-bandwidth {
-    description:
+    description
       "Common grouping used for both groups and neighbors which
       provides configuration and state parameters relating to
       auto-generation of link-bandwidth extended-community
       attribute.";
     container import {
-      description: 
+      description 
         "Container used to provides configuration and state
         parameters relating tovauto-generation of link-bandwidth 
         extended-community attribute for lerned routes.
@@ -342,7 +342,7 @@ submodule openconfig-bgp-common-structure {
         extended-community takes precedence over this
         configuration";
       container config {
-        description: "";
+        description "";
         uses auto-gen-link-bandwidth-config;
       }
       container state {
@@ -351,5 +351,5 @@ submodule openconfig-bgp-common-structure {
       }
     }
   }
-  
+
 }
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -325,7 +325,7 @@ submodule openconfig-bgp-common-structure {
     }
   }
 
-   grouping bgp-common-structure-neighbor-group-auto-gen-link-bandwidth {
+  grouping bgp-common-structure-neighbor-group-auto-gen-link-bandwidth {
     description
       "Common grouping used for both groups and neighbors which
       provides configuration and state parameters relating to
@@ -359,6 +359,15 @@ submodule openconfig-bgp-common-structure {
           extended-community.";
         uses auto-gen-link-bandwidth-config;
       }
+    }
+  }
+
+  grouping bgp-common-structure-neighbor-group-auto-link-bandwidth {
+      container auto-generated-link-bandwidth {
+      description
+        "Parameters related to automatic generation link-bandwidth
+        extended community base on underlaying interface speed.";
+      uses bgp-common-structure-neighbor-group-auto-gen-link-bandwidth;
     }
   }
 
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -21,7 +21,14 @@ submodule openconfig-bgp-common-structure {
     "This sub-module contains groupings that are common across multiple BGP
     contexts and provide structure around other primitive groupings.";
 
-  oc-ext:openconfig-version "9.8.0";
+  oc-ext:openconfig-version "9.9.0";
+
+  revision "2025-03-30" {
+    description
+      "Add schema for auto-generated link-bandwidth extended community
+      for learned routes.";
+    reference "9.9.0";
+  }
 
   revision "2024-09-06" {
     description
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -332,21 +332,31 @@ submodule openconfig-bgp-common-structure {
       auto-generation of link-bandwidth extended-community
       attribute.";
     container import {
-      description 
+      description
         "Container used to provides configuration and state
-        parameters relating tovauto-generation of link-bandwidth 
+        parameters related to auto-generation of link-bandwidth
         extended-community attribute for lerned routes.
-        Apply only for routes that do not have link-bandwidth 
+        Apply only for routes that do not have link-bandwidth
         extended-community.
-        Any policy action related to link-bandwidth 
+        Any policy action related to link-bandwidth
         extended-community takes precedence over this
         configuration";
       container config {
-        description "";
+        description
+          "Container used to provides configuration
+          parameters related to auto-generation of link-bandwidth
+          extended-community attribute for lerned routes.
+          Apply only for routes that do not have link-bandwidth
+          extended-community.";
         uses auto-gen-link-bandwidth-config;
       }
       container state {
-        description "";
+        description
+          "Container used to provides state
+          parameters related to auto-generation of link-bandwidth
+          extended-community attribute for lerned routes.
+          Apply only for routes that do not have link-bandwidth
+          extended-community.";
         uses auto-gen-link-bandwidth-config;
       }
     }
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -363,7 +363,7 @@ submodule openconfig-bgp-common-structure {
   }
 
   grouping bgp-common-structure-neighbor-group-auto-link-bandwidth {
-      container auto-generated-link-bandwidth {
+      container auto-link-bandwidth {
       description
         "Parameters related to automatic generation link-bandwidth
         extended community base on underlaying interface speed.";
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -368,6 +368,7 @@ submodule openconfig-bgp-common-structure {
       }
 
       container state {
+        config false;
         description
           "Container used to provide state
           parameters related to auto-generation of link-bandwidth
@@ -380,7 +381,10 @@ submodule openconfig-bgp-common-structure {
   }
 
   grouping bgp-common-structure-neighbor-group-auto-link-bandwidth {
-      container auto-link-bandwidth {
+    description
+      "Parameters related to automatic generation link-bandwidth
+      extended community base on underlaying interface speed.";
+    container auto-link-bandwidth {
       description
         "Parameters related to automatic generation link-bandwidth
         extended community base on underlaying interface speed.";
--- a/release/models/bgp/openconfig-bgp-common-structure.yang
+++ b/release/models/bgp/openconfig-bgp-common-structure.yang
@@ -348,7 +348,7 @@ submodule openconfig-bgp-common-structure {
         extended-community.
         Any policy action related to link-bandwidth
         extended-community takes precedence over this
-        configuration";
+        configuration.";
       container config {
         description
           "Container used to provides configuration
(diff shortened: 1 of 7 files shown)


release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
     }
   }
 
+  grouping bgp-common-structure-neighbor-group-auto-gen-link-bandwidth {
+    description
+      "Common grouping used for both groups and neighbors which
+      provides configuration and state parameters relating to
+      auto-generation of link-bandwidth extended-community
+      attribute.";
```
robshakir, line 340:
```suggestion
      attribute.";
      
```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
...
+    container import {
+      description
+        "Container used to provides configuration and state
```
robshakir, line 343:
```suggestion
        "Container used to provide configuration and state
```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
...
+        parameters related to auto-generation of link-bandwidth
+        extended-community attribute for lerned routes.
```
robshakir, line 345:
```suggestion
        extended-community attribute for learned routes.
```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
...
+        Apply only for routes that do not have link-bandwidth
```
robshakir, line 346:
```suggestion
        Applies only for routes that do not have link-bandwidth
```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
...
+        extended-community.
+        Any policy action related to link-bandwidth
+        extended-community takes precedence over this
+        configuration";
```
robshakir, line 350:
```suggestion
        configuration.";
```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
...
+      container config {
+        description
+          "Container used to provides configuration
```
robshakir, line 353:
```suggestion
          "Container used to provide configuration
```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
...
+          parameters related to auto-generation of link-bandwidth
+          extended-community attribute for lerned routes.
```
robshakir, line 355:
```suggestion
          extended-community attribute for learned routes.
```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
...
+          Apply only for routes that do not have link-bandwidth
```
robshakir, line 356:
```suggestion
          Applies only for routes that do not have link-bandwidth
```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
...
+          extended-community.";
+        uses auto-gen-link-bandwidth-config;
+      }
+      container state {
+        description
+          "Container used to provides state
+          parameters related to auto-generation of link-bandwidth
+          extended-community attribute for lerned routes.
```
robshakir, line 364:
```suggestion
          extended-community attribute for learned routes.
```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,50 @@ submodule openconfig-bgp-common-structure {
...
+          "Container used to provides configuration
+          parameters related to auto-generation of link-bandwidth
+          extended-community attribute for lerned routes.
+          Apply only for routes that do not have link-bandwidth
+          extended-community.";
+        uses auto-gen-link-bandwidth-config;
+      }
+      container state {
+        description
+          "Container used to provides state
```
robshakir, line 362:
```suggestion
          "Container used to provide state
```

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,30 @@ submodule openconfig-bgp-common {
   }
 
 
+  grouping auto-gen-link-bandwidth-config {
+    description
+      "Grouping of parameters related to automatic generation of
+      link-bandwidth extended community.";
```
robshakir, line 866:
```suggestion
      link-bandwidth extended community.";
      
```

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,30 @@ submodule openconfig-bgp-common {
...
+    leaf enabled {
+      description
+        "Generate link-bandwidth extnded community for BGP routes that do not
+        have link-bandwidth extnded community attached.";
+      type boolean;
+      default false;
+    }
```
robshakir, line 873:
```suggestion
    }
    
```

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,30 @@ submodule openconfig-bgp-common {
...
 
 
+  grouping auto-gen-link-bandwidth-config {
+    description
+      "Grouping of parameters related to automatic generation of
+      link-bandwidth extended community.";
+    leaf enabled {
+      description
+        "Generate link-bandwidth extnded community for BGP routes that do not
+        have link-bandwidth extnded community attached.";
```
robshakir, line 870:
```suggestion
        have link-bandwidth extended community attached.";
```

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,30 @@ submodule openconfig-bgp-common {
...
+      type boolean;
+      default false;
+    }
+    leaf hold-down-time {
+      description
+        "Suppress frequent changes in the link-bandwidth value when bandwidth
+        increases.";
+      type uint16;
+    }
```
robshakir, line 879:
```suggestion
    }
    
```

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,30 @@ submodule openconfig-bgp-common {
...
+    leaf enabled {
+      description
+        "Generate link-bandwidth extnded community for BGP routes that do not
+        have link-bandwidth extnded community attached.";
+      type boolean;
+      default false;
+    }
+    leaf hold-down-time {
+      description
+        "Suppress frequent changes in the link-bandwidth value when bandwidth
```
robshakir, line 876: Can we perhaps clarify this further -- i.e., rather than saying that we just suppress frequent changes, can we say what this value means?

I assume that it is the number of <some units of time> that we wait between subsequent updates of the link bandwidth extcommm?

You can specify the <some units of time> using `units seconds;` in the leaf description.

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,30 @@ submodule openconfig-bgp-common {
...
+        increases.";
+      type uint16;
+    }
+    leaf transitive {
+      description
+        "Generated link-bandwidth attribute of transitive type. The non-transitive
```
robshakir, line 882:
```suggestion
        "Generate link-bandwidth extended community as a transitive value. The non-transitive
```

Main suggestion here is to make sure that it's clear that we're generating the extcomm vs. a new attr.

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,30 @@ submodule openconfig-bgp-common {
...
+        is default and matches original draft.";
```
robshakir, line 883: I believe that [-11](https://www.ietf.org/archive/id/draft-ietf-idr-link-bandwidth-11.html) of the draft now supports creating transitive and non-transitive extcomms.

robshakir:
One comment -- is this approach (based on interface bandwidth) sufficient for the use cases that we see? If you check the internal "bgp-ucmp-ext.yang" directory (cs it -- or ping me :-)) then there were some other use cases that needed to supported in historical discussions.

jhaas-pfrc:
It should be noted that for junos that the auto sense only works for ebgp peers:
[regress@10.53.63.94#](mailto:regress@10.53.63.94) commit
[edit protocols bgp group int]
  'link-bandwidth'
    Link-bandwidth auto-sense is valid only for single-hop EBGP neighbor
error: commit failed: (statements constraint check failed)

Normally we'd model this as a constraint on the configuration.

release/models/bgp/openconfig-bgp-common.yang
rszarecki: please have a look.

rszarecki:
> One comment -- is this approach (based on interface bandwidth) sufficient for the use cases that we see? If you check the internal "bgp-ucmp-ext.yang" directory (cs it -- or ping me :-)) then there were some other use cases that needed to supported in historical discussions.

This is actually complementary to this internal "bgp-ucmp-ext.yang". This PR is handling **inbound**/**import** direction when incoming routes do not have link-bandwidth attached, and local system is expected to perform wECMP/UCMP loadbalancing. While "bgp-ucmp-ext.yang" addresses **outbound**/**export** direction via policy action. See line 225 "// Augmentation to add UCMP policy outbound actions". 
The link-bandwidth "regeneration" for advertisment to peer, base on Local-RIB content is not subject of this PR.

The two other things "bgp-ucmp-ext.yang" is doing are:
-  enables wECMP (aka UCMP) for bgp multipathing. We alredy have this in public OpenConfig -  [merged PR1028](https://github.com/openconfig/public/pull/1028). 
-  allows to send and recive non-transitive link-bandwidth extended community over eBGP session.  This is not subject of this PR. As of today I know of only one implementation that needs this functionality.

rszarecki:
> It should be noted that for junos that the auto sense only works for ebgp peers: [regress@10.53.63.94#](mailto:regress@10.53.63.94) commit [edit protocols bgp group int] 'link-bandwidth' Link-bandwidth auto-sense is valid only for single-hop EBGP neighbor error: commit failed: (statements constraint check failed)
> 
> Normally we'd model this as a constraint on the configuration.

- Why its is valid only for EBGP? I think single-hop IBGP with local-address assotiated with "real" interface (not irb, vsi, rbi, loopback) cold be resonable use case. Perhaps JUNOS is too restrictive?
- if session is "floating" form loopback/irb/vsi/etc indeed it is hard to determine bandwidth, and IMHO implemenattion has all right to reject such config.

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,51 @@ submodule openconfig-bgp-common-structure {
...
+    container import {
+      description
+        "Container used to provide configuration and state
+        parameters related to auto-generation of link-bandwidth
+        extended-community attribute for learned routes.
+        Applies only for routes that do not have link-bandwidth
+        extended-community.
+        Any policy action related to link-bandwidth
+        extended-community takes precedence over this
+        configuration.";
```
robshakir, line 351:
```suggestion
        configuration.";

```

release/models/bgp/openconfig-bgp-common-structure.yang
```diff
@@ -325,4 +332,51 @@ submodule openconfig-bgp-common-structure {
...
+      container config {
+        description
+          "Container used to provide configuration
+          parameters related to auto-generation of link-bandwidth
+          extended-community attribute for learned routes.
+          Applies only for routes that do not have link-bandwidth
+          extended-community.";
+        uses auto-gen-link-bandwidth-config;
+      }
```
robshakir, line 360:
```suggestion
      }

```

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,39 @@ submodule openconfig-bgp-common {
...
 
 
+  grouping auto-gen-link-bandwidth-config {
+    description
+      "Grouping of parameters related to automatic generation of
+      link-bandwidth extended community.";
+
+    leaf enabled {
+      description
+        "Generate link-bandwidth extnded community for BGP routes that do not
```
robshakir, line 870:
```suggestion
        "Append a link-bandwidth extended community for BGP routes that do not
```

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,39 @@ submodule openconfig-bgp-common {
...
+        have link-bandwidth extended community attached.";
+      type boolean;
+      default false;
+    }
+
+    leaf hold-down-time {
+      description
+        "Suppress for a nuber of seconds changes in the link-bandwidth value
```
robshakir, line 878:
```suggestion
        "Suppress for a number of seconds changes in the link-bandwidth value
```

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,39 @@ submodule openconfig-bgp-common {
...
+
+    leaf enabled {
+      description
+        "Generate link-bandwidth extnded community for BGP routes that do not
+        have link-bandwidth extended community attached.";
+      type boolean;
+      default false;
+    }
+
+    leaf hold-down-time {
```
robshakir, line 876: Is it better to name this `suppress-time` rather than `hold-down-time` since we're not really holding the bandwidth "down" are we?

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,39 @@ submodule openconfig-bgp-common {
...
+      description
+        "Suppress for a nuber of seconds changes in the link-bandwidth value
+        when bandwidth increases. Every time bandwidth of link increases,
```
robshakir, line 879: Does this timer apply only when the bandwidth increases, what about when it decreases?

release/models/bgp/openconfig-bgp-common.yang
```diff
@@ -853,4 +860,39 @@ submodule openconfig-bgp-common {
...
+        it must remain constant for hold-down-time before it is reflected
+        into link-bandwidth extended community of routes lerned on this
```
robshakir, line 881:
```suggestion
        into link-bandwidth extended community of routes learned on this
```

robshakir (commented):
Generally LGTM, one question and a few editorial nits.

robshakir:
> > One comment -- is this approach (based on interface bandwidth) sufficient for the use cases that we see? If you check the internal "bgp-ucmp-ext.yang" directory (cs it -- or ping me :-)) then there were some other use cases that needed to supported in historical discussions.
> 
> This is actually complementary to this internal "bgp-ucmp-ext.yang". This PR is handling **inbound**/**import** direction when incoming routes do not have link-bandwidth attached, and local system is expected to perform wECMP/UCMP loadbalancing. While "bgp-ucmp-ext.yang" addresses **outbound**/**export** direction via policy action. See line 225 "// Augmentation to add UCMP policy outbound actions". The link-bandwidth "regeneration" for advertisment to peer, base on Local-RIB content is not subject of this PR.

Thanks for the analysis -- I'd missed that this is just for routes that are received and not specifying what we do for our exported. I made a small wording suggestion that makes this more obvious, but this makes sense.

> The two other things "bgp-ucmp-ext.yang" is doing are:
> 
> * enables wECMP (aka UCMP) for bgp multipathing. We alredy have this in public OpenConfig -  [merged PR1028](https://github.com/openconfig/public/pull/1028).

Great, thanks.

> * allows to send and recive non-transitive link-bandwidth extended community over eBGP session.  This is not subject of this PR. As of today I know of only one implementation that needs this functionality.

Yes, I think in ~2016 when I wrote this there was much more variance in the implementation choices -- so this is probably something that can be deprecated at some point given the draft was updated to handle re-advertisement.

thanks for the change!

dplore:
Do we want to also model at the individual neighbor level?

jhaas-pfrc:
> > It should be noted that for junos that the auto sense only works for ebgp peers: [regress@10.53.63.94#](mailto:regress@10.53.63.94) commit [edit protocols bgp group int] 'link-bandwidth' Link-bandwidth auto-sense is valid only for single-hop EBGP neighbor error: commit failed: (statements constraint check failed)
> > Normally we'd model this as a constraint on the configuration.
> 
>     * Why its is valid only for EBGP? I think single-hop IBGP with local-address assotiated with "real" interface (not irb, vsi, rbi, loopback) cold be resonable use case. Perhaps JUNOS is too restrictive?

Link-local peering has a number of standards-based sharp edges.  There's work going on in IDR right now to try to address some of that, eventually draft-ietf-idr-linklocal-capability.  It's adopted but hasn't published ietf-idr-00.

As an example, how exactly is resolution intended to work in such circumstances?  If you have a ibgp served nexthop, for purposes of link-bw auto-assignment what you really care about is that the nexthop is directly attached.  eBGP gives you that property by default for first-party nexthops.  ibgp... less guaranteed.

Thus, if this is a real use case I think you'll both want to wait to see how the sharp edges of link local peering resolve themselves in a vendor neutral fashion.

> 
>     * if session is "floating" form loopback/irb/vsi/etc indeed it is hard to determine bandwidth, and IMHO implemenattion has all right to reject such config.

Also, ebgp multihop.

release/models/bgp/openconfig-bgp-common.yang
rszarecki: We do. We suppres changes when bandwidth is increased. But wne it is decreased we do not - react "immedietly". When goes down we want te evaculate load ASAP, and prevent/reduce congestion. When it goes UP, we can wait and see if this is stable or up just for few seconds. This will added flapping memeber going up-down-up-...
I was contemplating hold-up-timer also, but decided against - no implementation on industry, too strict. I thin implementation will do the best they can.
- EoS 'update-dely' says "delay in generation of route advertisement updates when the link bandwidth extended community value **keeps increasing**"
- Junos 'hold-down' is hold down...

release/models/bgp/openconfig-bgp-common.yang
rszarecki: See previous comment.

rszarecki:
> Do we want to also model at the individual neighbor level?

It is alredy here.

rszarecki:
> eBGP gives you that property by default for first-party nexthops. ibgp... less guaranteed.

yeah. But bottom line, I can configure iBGP session in on using link-addresses, ttl=1, and knowing remote peer is doing next-hop-self (bcoz it is my AS - my admin domain), use this feature. Here is a rope, you can hang yourself...
EBGP also can serves me 3rd pary NH...

I agree this is pramarilly EBGP and this is how we are coing to consime it.

rszarecki:
> Do we want to also model at the individual neighbor level?

We have it alredy defined for `neighbor` and `peer-group`. I intentionally skipped global, as it will cover also any IBGP session. So probably never used.

I think question would be if we need it on per AFI-SAFI level. I do see some value and it is feasible. But since this is not availabe in any implementation, just keep it simple.

dplore:
Reviewed in Apr 22, 2025 OC Operator meeting without objection, setting last call for comments.  This will merge on May 6, 2025

dplore (approved):

PR #1274, opened 2025-03-31
Static route next network instance

rszarecki:
### Change Scope

This change introduces **TWO** new option for static route:

#### 1. Redirecting packet to subsequent destination lookup in other network-instance.
This specify network-instance in which packet's destination address should be looked up, to to find next-hop. 
It must not be combined with `next-hop`, `nh-network-instance` and `recurce` leaf must be ignored.
The `metric` and `preference` semantics is the same as for `next-hop`

```
    +--rw network-instances
          +--rw protocols
                +--rw static-routes
                   +--rw static* [prefix]
                      +--rw next-hops
                         +--rw next-hop* [index]
                            +--rw config
                               +--rw index?                   string
+                              +--rw next-network-instance?   -> /network-instances/network-instance/config/name
                               +--rw next-hop?                union
                               +--rw nh-network-instance?     -> /network-instances/network-instance/config/name
                               +--rw recurse?                 boolean
                               +--rw metric?                  uint32
                               +--rw preference?              uint32
```
Pleas enote that multiple elements of `next-network-instance` and `next-hop`  types could be present on sinle next-hop list keyed by unique`index`.  Therefore ECMP netween next-hop and next-network-instance is expressible. Also, it is possible to express `next-network-instance` as fallback for `next-hop` (IP or interface-ref).

This change is non-disruptive.

##### Platform Implementations

 * Juniper: [Enabling Internet Access for Layer 3 VPNs
](https://www.juniper.net/documentation/us/en/software/junos/vpn-l3/topics/topic-map/l3-vpns-internet-access.html) 

```
[edit]
routing-instances {
    vpna {
        instance-type vrf;
        routing-options {
            static {
                route 0.0.0.0/0 next-table inet.0;
            }
        }
```

 * Cisco XR: 
implementation output.

```
router static
  vrf vrf_A
    address ipv4 unicast
      172.168.50.0/24 vrf vrf_B 
      172.168.50.0/24 vrf vrf_C

```

####  2. Lookup for static route's next-hop in other network-instance

```
    +--rw network-instances
          +--rw protocols
                +--rw static-routes
                   +--rw static* [prefix]
                      +--rw next-hops
                         +--rw next-hop* [index]
                            +--rw config
                               +--rw index?                   string
                               +--rw next-network-instance?   -> /network-instances/network-instance/config/name
                               +--rw next-hop?                union
+                              +--rw nh-network-instance?     -> /network-instances/network-instance/config/name
                               +--rw recurse?                 boolean
                               +--rw metric?                  uint32
                               +--rw preference?              uint32
```

This change is non-disruptive.

##### Platform Implementations

* Arista EoS [static-inter-vrf-route-support](https://www.arista.com/en/support/toi/eos-4-23-1f/14397-static-inter-vrf-route-support#configuration)

```
ip route 1.1.0.0/24 egress-vrf bar 2.2.2.2
```

 * Cisco XR [VFR for static routes](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/routing/cumulative/command/reference/b-routing-cr-cisco8000/m-f63682-adhoc.html#wp1021470314)
implementation output.

```
router static
vrf vrf_A
address ipv4 unicast
172.168.50.0/24 vrf vrf_B 192.168.1.2
```

### DIFF
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -313,25 +313,25 @@ module openconfig-local-routing {
 
         container next-network-instance {
           description
-          "The network instance in which context packet destination address
-          lookup shall be performed.";
-          
+            "The network instance in which context packet destination address
+            lookup shall be performed.";
+
           container config {
             leaf next-network-instance {
               description
                 "The network instance in which context packet destination address
-                lookup shall be performed."
+                lookup shall be performed.";
               type leafref {
                 path "../../../../../../../config/name";
               }
             }
-          } 
+          }
 
           container state {
             leaf next-network-instance {
               description
                 "The network instance in which context packet destination address
-                lookup is be performed."
+                lookup is be performed.";
               type leafref {
                 path "../../../../../../../state/name";
               }
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -235,13 +235,13 @@ module openconfig-local-routing {
     leaf next-network-instance {
       description
         "Instead of finding the next-hop for a destination address in the current
-	      network-instance, look up the destination address in the routing table
-	      of the specified network-instance.
-	      For example, let 'next-network-instance' = 'DEFAULT'.  If a packet arrives
+	network-instance, look up the destination address in the routing table
+	of the specified network-instance.
+	 For example, let 'next-network-instance' = 'DEFAULT'.  If a packet arrives
         in the 'BLUE' network-instance and the destination address matches this
         route, the destination address will be looked up again in the 'DEFAULT'
-	      network-instance to find the next-hop.  
-	      This leaf is mutually exclusive with 'next-hop', 'nh-network-instance' 
+	network-instance to find the next-hop.
+	This leaf is mutually exclusive with 'next-hop', 'nh-network-instance'
         leaves";
       type leafref {
         path "/network-instances/network-instance/config/name";
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -234,10 +234,15 @@ module openconfig-local-routing {
 
     leaf next-network-instance {
       description
-        "The network instance in which context packet destination address
-        lookup shall be performed.
-        This leaf is mutualy exclusive with 'next-hop', 'nh-network-instance',
-        'recurse' leaves";
+        "Instead of finding the next-hop for a destination address in the current
+	      network-instance, look up the destination address in the routing table
+	      of the specified network-instance.
+	      For example, let 'next-network-instance' = 'DEFAULT'.  If a packet arrives
+        in the 'BLUE' network-instance and the destination address matches this
+        route, the destination address will be looked up again in the 'DEFAULT'
+	      network-instance to find the next-hop.  
+	      This leaf is mutually exclusive with 'next-hop', 'nh-network-instance' 
+        leaves";
       type leafref {
         path "/network-instances/network-instance/config/name";
       }
@@ -272,8 +277,8 @@ module openconfig-local-routing {
         nh-network-instance.
         For example, let 'nh-network-instance' = 'DEFAULT'.  If a
         packet arrives on the 'BLUE' network-instance and the
-        destination address matches this route, look up the
-        configured next-hop in the 'DEFAULT' network-instance.
+        destination address matches this route, look up the IP
+        configured as next-hop in the 'DEFAULT' network-instance.
         This leaf is mutually exclusive with next-network-instance
         leaf";
       type leafref {
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -380,7 +380,7 @@ module openconfig-local-routing {
             container must not be used.";
 
           list next-hop {
-            
+
             key "index";
 
             description
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -232,13 +232,21 @@ module openconfig-local-routing {
         only support a numeric value for this string. ";
     }
 
+    leaf next-network-instance {
+      description
+        "The network instance in which context packet destination address
+        lookup shall be performed.
+        This leaf is mutualy exclusive with 'next-hop', 'nh-network-instance',
+        'recurse' leaves";
+      type leafref {
+        path "../../../../../../../../network-instance/config/name";
+      }
+    }
+
     leaf next-hop {
       type union {
         type inet:ip-address;
         type local-defined-next-hop;
-        type leafref {
-          path "../../../../../../../../../network-instance/config/name";
-        }
       }
       description
         "The next-hop that is to be used for the static route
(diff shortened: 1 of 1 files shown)


dplore (commented):
Some additional description in the yang I think is needed / helpful to understand the complete expectation.

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -304,6 +311,34 @@ module openconfig-local-routing {
           uses local-static-state;
         }
 
+        container next-network-instance {
+          description
+            "The network instance in which context packet destination address
```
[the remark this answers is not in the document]
dplore: Here's a suggestion from the prior PR: https://github.com/openconfig/public/pull/1259#issuecomment-2702128635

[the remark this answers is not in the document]
rszarecki: Good catch. When I thinked about it, It appears to be perfectly logical to specify mix of N x next-hop and M x Next-network-instance. and sort it out unig metric.

And thne possibly ECMP between next-hop and nexty-network-instance. ...
something like:

```
    leaf next-hop {
      type union {
        type inet:ip-address;
        type local-defined-next-hop;
        **type leafref {
          path "../../../../../../../../config/name";
        }**
      }
```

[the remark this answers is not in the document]
rszarecki: I did major change to schema. Plese review.

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -229,17 +236,39 @@ module openconfig-local-routing {
       type union {
         type inet:ip-address;
         type local-defined-next-hop;
+        type leafref {
```
robshakir, line 239: YANG nit: since this is in the context of some particular network instance -- i.e., `/network-instances/network-instance[name=fish]/protocols/protocol[name=STATIC][identifier=15169]/static-routes/static[prefix=1.1.1.1/32]/next-hops/next-hop[index=1]/config/next-hop` then `../../../../../../../../config/name` resolves to `/network-instances/network-instance[name=fish]/config/name` which can only ever equal `fish` (i.e., we're resolving to only one NI).

You need to go one level deeper and then add `network-instance/config/name` which will expand to all network instances on the device.

robshakir, line 239: YANG nit: this requires YANG 1.1 -- I think it should be supported, @dplore -- what is our current stance on this?

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -229,17 +236,39 @@ module openconfig-local-routing {
...
-        or a pre-defined next-hop type - for instance, DROP or
-        LOCAL_LINK. When this leaf is not set, and the interface-ref
+        - this may be specified as:
+          -  an IP address or
+          - a pre-defined next-hop type - for instance, DROP or
+        LOCAL_LINK or
+          - next network-instance in which packet's destination
+          address should be looked up, to to find next-hop. It
+          must not be combined with nh-network-instance and
+          recurce leaf must be ignored.
```
robshakir, line 252:
```suggestion
          recurse leaf must be ignored.
```

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -229,17 +236,39 @@ module openconfig-local-routing {
...
         interface.";
     }
 
+    leaf nh-network-instance {
+      description
+        "Network-instance in which IP address of next-hop should
+        be looked up inorder to find egress interface. This attribute
+        is only valid if next-hop is given as IP address.
+        This attribute may be combined with recurse attribute of any
+        value, in which case recurse setting applys to
```
robshakir, line 265:
```suggestion
        value, in which case recurse setting applies to
```

robshakir (commented):
Do we need both of these options @rszarecki? I think from your examples there is vendor support, but I'm just wondering whether we think that there's a need to have support for both in OC, or whether we want to be more opinionated about it.

Thanks for the change.

rszarecki:
> Do we need both of these options @rszarecki? I think from your examples there is vendor support, but I'm just wondering whether we think that there's a need to have support for both in OC, or whether we want to be more opinionated about it.
> 
> Thanks for the change.

We do need change 1 -  "Redirecting packet to subsequent destination lookup in other network-instance."

release/models/local-routing/openconfig-local-routing.yang
rszarecki: Done.

rszarecki: Changed to avoid union w/ leafref.
Please have a look.
There is text in description that make new leaf anf existing `next-hop` leaf mutually exclusive.

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -225,19 +232,49 @@ module openconfig-local-routing {
         only support a numeric value for this string. ";
     }
 
+    leaf next-network-instance {
+      description
+        "The network instance in which context packet destination address
+        lookup shall be performed.
+        This leaf is mutualy exclusive with 'next-hop', 'nh-network-instance',
+        'recurse' leaves";
+      type leafref {
```
dplore, line 241:
```suggestion
        "Instead of finding the next-hop for a destination address in the current
        network-instance, look up the destination address in the routing table
        of the specified network-instance.
        For example, let 'nh-network-instance' = 'DEFAULT'.  If a packet arrives in
        the 'BLUE' network-instance and the destination address matches this
        route, the destination address will be looked up again in the 'DEFAULT'
        network-instance to find the next-hop.  
        This leaf is mutually exclusive with 'next-hop', 'nh-network-instance',
        'recurse' leaves";
      type leafref {
```

dplore: Adding an example helps clarify the description I think.

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -225,19 +232,49 @@ module openconfig-local-routing {
...
+    leaf nh-network-instance {
+      description
+        "Network-instance in which IP address of next-hop should
+        be looked up, to find egress interface. This attribute
+        is only valid if next-hop is given as IP address.
+        This attribute may be combined with recurse attribute of any
+        value, in which case recurse setting applies to
+        nh-network-instance.
+        This leaf is mutualy exclusive with next-network-instance
+        leaf";
```
dplore, line 274:
```suggestion
        "Network-instance in which IP address of next-hop should
        be looked up, to find egress interface. This attribute
        is only valid if next-hop is given as IP address.
        This attribute may be combined with recurse attribute of any
        value, in which case recurse setting applies to
        nh-network-instance.
        For example, let 'nh-network-instance' = 'DEFAULT'.  If a
        packet arrives on the 'BLUE' network-instance and the
        destination address matches this route, look up the
        configured next-hop in the 'DEFAULT' network-instance.
        This leaf is mutually exclusive with next-network-instance
        leaf";
```

release/models/local-routing/openconfig-local-routing.yang
dplore: What if the next-hop in the next-network-instance is an IP address?  Shouldn't recurse be allowed?

dplore (commented):
I suggested some examples to help clarify how these leaves are supposed to work.  Let me know if they make sense.

Also a question regarding the `recurse` leaf for next-network-instance.

dplore (approved):

PR #1279, opened 2025-04-07
Fix enclosing container name for openconfig-mpls-te.yang p2p-primary-path list

jardel-lucca-venko:
`/openconfig-network-instance:network-instances/network-instance/mpls/lsps/constrained-path/tunnels/tunnel/p2p-tunnel-attributes/p2p-primary-path/p2p-primary-path`

### Change Scope

* This small change fixes `p2p-primary-path` list enclosing container name so it matches conventional OC naming (i.e., plural of enclosed list name). What makes this particularly distracting when interacting with this datamodel is that the `p2p-secondary` counterpart of this container (`.../p2p-tunnel-attributes/p2p-secondary-paths`) is following the convention correctly.

### DIFF
--- a/release/models/mpls/openconfig-mpls-te.yang
+++ b/release/models/mpls/openconfig-mpls-te.yang
@@ -30,12 +30,12 @@ submodule openconfig-mpls-te {
     signaling protocol or mechanism (see related submodules for
     signaling protocol-specific configuration).";
 
-  oc-ext:openconfig-version "3.6.1";
+  oc-ext:openconfig-version "4.0.0";
 
   revision "2025-04-07" {
     description
       "Fix enclosing container name for p2p-primary-path list";
-    reference "3.6.1";
+    reference "4.0.0";
   }
 
   revision "2024-06-19" {
--- a/release/models/mpls/openconfig-mpls-te.yang
+++ b/release/models/mpls/openconfig-mpls-te.yang
@@ -30,7 +30,13 @@ submodule openconfig-mpls-te {
     signaling protocol or mechanism (see related submodules for
     signaling protocol-specific configuration).";
 
-  oc-ext:openconfig-version "3.6.0";
+  oc-ext:openconfig-version "3.6.1";
+
+  revision "2025-04-07" {
+    description
+      "Fix enclosing container name for p2p-primary-path list";
+    reference "3.6.1";
+  }
 
   revision "2024-06-19" {
     description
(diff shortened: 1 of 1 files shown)


release/models/mpls/openconfig-mpls-te.yang
```diff
@@ -30,7 +30,13 @@ submodule openconfig-mpls-te {
     signaling protocol or mechanism (see related submodules for
     signaling protocol-specific configuration).";
 
-  oc-ext:openconfig-version "3.6.0";
+  oc-ext:openconfig-version "3.6.1";
```
earies, line 33: Would have to be 4.0.0 w/ what is proposed here

release/models/mpls/openconfig-mpls-te.yang
```diff
@@ -1334,7 +1340,7 @@ submodule openconfig-mpls-te {
     description
       "Top level grouping for p2p primary paths";
 
-    container p2p-primary-path {
+    container p2p-primary-paths {
```
earies, line 1343: Consistency is great and desired but this is a massive backwards incompatble break by doing this.

One alternative transition path would be to introduce the new container in parallel leveraging all of the same direct descendants but that is if this is justifiable in the first place (the negative impact is likely far larger than the benefit given implementation already)

jardel-lucca-venko: Thanks for your review.
I understand this is somewhat impactful. I'd like to emphasize that this PR is not merely for cosmetic reasons though. This subtle inconsistency causes confusion that oftentimes is tricky to get for those beginning to work with this part of the datamodel.

earies: Yea if you peruse the model set, I think we'd find handfuls of these and other nomenclature items that ideally should be corrected but since there is implementation and adoption, it becomes less of a priority unless there is some functionality missing or prevented by doing such.

There was already a previous attempt at this btw here: https://github.com/openconfig/public/pull/1235

dplore:
Agreed, unfortunately we should leave this non-style compliant container due to this being a breaking change for what is otherwise a cosmetic/style only issue.

Generally speaking, we reserve the use of breaking changes for fixing schema which cannot meet it's intended  operational function (ie: it's truly broken).  Even then, the preferred implementation of such a fix is to deprecate the broken node(s) and introduce new nodes, making the change non-breaking.  The theory is that in some period of time (measured in years) the deprecated node can be deleted, relieving the burden of supporting broken schema.  (sometimes even broken schema is used and deployed)

PR #1280, opened 2025-04-09
Update lag-speed definition in openconfig-if-aggregate.yang

AmrNJ:
### Change Scope

* The OC model definition of lag-speed leaf is changed to reflect the current behavior. 
* Lag-speed will now be calculated on the aggregate speed of forwarding-viable member interfaces as opposed to just the active interfaces as used before.

### DIFF
--- a/release/models/interfaces/openconfig-if-aggregate.yang
+++ b/release/models/interfaces/openconfig-if-aggregate.yang
@@ -148,7 +148,7 @@ module openconfig-if-aggregate {
       units Mbps;
       description
         "Reports effective speed of the aggregate interface,
-        based on speed of active member interfaces";
+        based on speed of forwarding-viable member interfaces";
     }
 
     leaf-list member {


release/models/interfaces/openconfig-if-aggregate.yang
```diff
@@ -148,7 +148,7 @@ module openconfig-if-aggregate {
       units Mbps;
       description
         "Reports effective speed of the aggregate interface,
-        based on speed of active member interfaces";
+        based on speed of forwarding-viable member interfaces";
     }
```
dplore, line 152: Please add more description to make it easy for the reader to discover the definition for forwarding-viable.  Also we need to continue to cover the case for `oper-status`.  For example,

for interfaces where /interfaces/interface/state/forwarding-viable = true and the /interfaces/interface/state/oper-status is UP.

dplore (changes requested):

release/models/interfaces/openconfig-if-aggregate.yang
LimeHat: >for interfaces where /interfaces/interface/state/forwarding-viable = true and the /interfaces/interface/state/oper-status is UP.

perhaps keep the word active in the definition? i.e. `active and forwarding-viable=true`.

if interface is down it will be inactive, also `active` covers other cases such as lacp down.

AmrNJ: Thanks, updated.

release/models/interfaces/openconfig-if-aggregate.yang
```diff
@@ -148,7 +148,7 @@ module openconfig-if-aggregate {
       units Mbps;
       description
         "Reports effective speed of the aggregate interface,
-        based on speed of active member interfaces";
+        based on speed of active (operationally up) and forwarding-viable member interfaces";
```
dplore, line 151: In discussion at OC Operators meeting, it was mentioned that LACP backup mode should also affect the lag-speed.  (ie: a member in 'backup' state should not be included in lag-speed).   I think it's helpful to enumerate the conditions we know of here in the description to help increase clarity for everyone.

dplore:
The content looks good.  Please format to 80 char max line length and remove any trailing whitespace (CI check is failing on this)

dplore:
Also, need to bump the version number and add a change description here: 

https://github.com/openconfig/public/pull/1280/files#diff-5b052c18962e8f00ce623edfa67b62e759b03f114986ff4bba4ce28702ed08a0R26-R32

release/models/interfaces/openconfig-if-aggregate.yang
```diff
@@ -23,12 +23,12 @@ module openconfig-if-aggregate {
...
-  oc-ext:openconfig-version "2.4.4";
+  oc-ext:openconfig-version "2.4.5";
 
-  revision "2022-06-28" {
+  revision "2025-04-22" {
     description
-      "Remove reference to invalid oc-ift type check";
-    reference "2.4.4";
+      "Updated the definition of lag-speed leaf";
+    reference "2.4.5";
```
dplore, line 31: We don't want to remove the existing revision references, just add the new one

dplore (changes requested):
just need to fix the revision statement (don't remove the existing one, just add a new one)

dplore (approved):

PR #1282, opened 2025-04-16
Change pathz telemetry to be per policy rather than per path

marcushines:

### Change Scope

* Refactoring the pathz yang telemetry to be based on per policy rather than by path.
This has been requested by multiple vendors as per path is not needed and causes a very large memory and state management issue

* this is not backwards compatible but currently these counters should be in production

### DIFF
--- a/release/models/gnsi/openconfig-gnsi-pathz.yang
+++ b/release/models/gnsi/openconfig-gnsi-pathz.yang
@@ -34,6 +34,12 @@ module openconfig-gnsi-pathz {
 
   oc-ext:openconfig-version "0.3.0";
 
+  revision 2025-04-15 {
+    description
+      "Refactor gnsi counters to be only per policy.";
+    reference "0.4.0";
+  }
+
   revision 2024-02-13 {
     description
       "Major style updates and move to openconfig/public from openconfig/gnsi.
@@ -109,62 +115,43 @@ module openconfig-gnsi-pathz {
     description
       "A collection of counters collected by the gNSI.pathz module.";
 
-    container gnmi-pathz-policy-counters {
-      config false;
+    container policy {
       description
-        "A collection of per-OpenConfig path counters.";
+        "Container for a collection of per-Pathz policy counters.";
 
-      uses gnmi-pathz-policy-xpath-success-failure-counters;
-    }
-  }
-
-  grouping gnmi-pathz-policy-xpath-success-failure-counters {
-    description
-      "A collection of per-OpenConfig path counters.";
-
-    container paths {
-      description
-        "Container for a collection of per-OpenConfig path counters.";
-
-      list path {
+      list policy {
         description
-          "List for a collection of per-OpenConfig path counters.";
+          "List for a collection of per-Pathz policy counters.";
         key "name";
         leaf name {
           type leafref {
             path "../state/name";
           }
           description
-            "A OpenConfig schema path the counter were
-            collected for.
-
-            For documentation on the naming of paths, see
-            https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md";
+            "A OpenConfig Pathz policy the counters were
+            collected for.";
         }
         container state {
           description
-            "Operational state for per-OpenConfig path counters.";
+            "Operational state for per-Pathz policy counters.";
           leaf name {
             type string;
             description
-              "A OpenConfig schema path the counter were
-              collected for.
-
-              For documentation on the naming of paths, see
-              https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md";
+              "A per-Pathz policy the counters were
+              collected for.";
           }
           container reads {
             description
               "The counter were collected while
-              performing a read operation on the
-              schema path.";
+              performing a read operations on the
+              policy.";
             uses counters;
           }
           container writes {
             description
               "The counter were collected while
-              performing a write operation on the
-              schema path.";
+              performing a write operations on the
+              policy.";
             uses counters;
           }
         }
--- a/release/models/gnsi/openconfig-gnsi-pathz.yang
+++ b/release/models/gnsi/openconfig-gnsi-pathz.yang
@@ -32,7 +32,7 @@ module openconfig-gnsi-pathz {
     OpenConfig-path-based authorization policies installed on a networking
     device.";
 
-  oc-ext:openconfig-version "0.3.0";
+  oc-ext:openconfig-version "0.4.0";
 
   revision 2025-04-15 {
     description
@@ -115,7 +115,8 @@ module openconfig-gnsi-pathz {
     description
       "A collection of counters collected by the gNSI.pathz module.";
 
-    container policy {
+    container policies {
+      config false;
       description
         "Container for a collection of per-Pathz policy counters.";
 
@@ -128,7 +129,7 @@ module openconfig-gnsi-pathz {
             path "../state/name";
           }
           description
-            "A OpenConfig Pathz policy the counters were
+            "A OpenConfig Pathz policy name the counters were
             collected for.";
         }
         container state {
--- a/release/models/gnsi/openconfig-gnsi-pathz.yang
+++ b/release/models/gnsi/openconfig-gnsi-pathz.yang
@@ -36,7 +36,7 @@ module openconfig-gnsi-pathz {
 
   revision 2025-04-15 {
     description
-      "Refactor gnsi counters to be only per policy.";
+      "Refactor gNSI counters to be only per policy.";
     reference "0.4.0";
   }
 
@@ -144,14 +144,14 @@ module openconfig-gnsi-pathz {
           container reads {
             description
               "The counter were collected while
-              performing a read operations on the
+              performing read operations on the
               policy.";
             uses counters;
           }
           container writes {
             description
               "The counter were collected while
-              performing a write operations on the
+              performing write operations on the
               policy.";
             uses counters;
           }


release/models/gnsi/openconfig-gnsi-pathz.yang
```diff
@@ -109,62 +115,44 @@ module openconfig-gnsi-pathz {
...
-  grouping gnmi-pathz-policy-xpath-success-failure-counters {
-    description
-      "A collection of per-OpenConfig path counters.";
-
-    container paths {
-      description
-        "Container for a collection of per-OpenConfig path counters.";
+        "Container for a collection of per-Pathz policy counters.";
 
-      list path {
```
robshakir, line 129: If it's implemented, rather than replace the paths container entirely, you can mark it as `deprecated`.  If not, then this LGTM.

release/models/gnsi/openconfig-gnsi-pathz.yang
```diff
@@ -109,62 +115,44 @@ module openconfig-gnsi-pathz {
...
-              https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md";
+              "A per-Pathz policy the counters were
+              collected for.";
           }
           container reads {
             description
               "The counter were collected while
-              performing a read operation on the
-              schema path.";
+              performing a read operations on the
```
robshakir, line 147:
```suggestion
              performing read operations on the
```

release/models/gnsi/openconfig-gnsi-pathz.yang
```diff
@@ -109,62 +115,44 @@ module openconfig-gnsi-pathz {
...
+              policy.";
             uses counters;
           }
           container writes {
             description
               "The counter were collected while
-              performing a write operation on the
-              schema path.";
+              performing a write operations on the
```
robshakir, line 154:
```suggestion
              performing write operations on the
```

release/models/gnsi/openconfig-gnsi-pathz.yang
```diff
@@ -32,7 +32,13 @@ module openconfig-gnsi-pathz {
     OpenConfig-path-based authorization policies installed on a networking
     device.";
 
-  oc-ext:openconfig-version "0.3.0";
+  oc-ext:openconfig-version "0.4.0";
+
+  revision 2025-04-15 {
+    description
+      "Refactor gnsi counters to be only per policy.";
```
robshakir, line 39:
```suggestion
      "Refactor gNSI counters to be only per policy.";
```

robshakir (commented):
```
module: openconfig-gnsi-pathz

  augment /oc-sys:system:
    +--ro gnmi-pathz-policies
       +--ro policies
          +--ro policy* [instance]
             +--ro instance    -> ../state/instance
             +--ro state
                +--ro instance?     enumeration
                +--ro version?      version
                +--ro created-on?   created-on
    +--ro gnmi-pathz-policy-version?      version
    +--ro gnmi-pathz-policy-created-on?   created-on
    +--ro gnmi-pathz-policy-counters
       +--ro paths
          +--ro path* [name]
             +--ro state
                +--ro name?     string
                +--ro reads
                |  +--ro access-rejects?       oc-yang:counter64
                |  +--ro last-access-reject?   oc-types:timeticks64
                |  +--ro access-accepts?       oc-yang:counter64
                |  +--ro last-access-accept?   oc-types:timeticks64
                +--ro writes
                   +--ro access-rejects?       oc-yang:counter64
                   +--ro last-access-reject?   oc-types:timeticks64
                   +--ro access-accepts?       oc-yang:counter64
                   +--ro last-access-accept?   oc-types:timeticks64
```

LGTM, couple of editorial suggestions.

earies:
For reference, the updated proposed tree:

```
module: openconfig-gnsi-pathz

  augment /oc-sys:system:
    +--ro gnmi-pathz-policies
       +--ro policies
          +--ro policy* [instance]
             +--ro instance    -> ../state/instance
             +--ro state
                +--ro instance?     enumeration
                +--ro version?      version
                +--ro created-on?   created-on
    +--ro gnmi-pathz-policy-version?      version
    +--ro gnmi-pathz-policy-created-on?   created-on
    +--ro policies
       +--ro policy* [name]
          +--ro state
             +--ro name?     string
             +--ro reads
             |  +--ro access-rejects?       oc-yang:counter64
             |  +--ro last-access-reject?   oc-types:timeticks64
             |  +--ro access-accepts?       oc-yang:counter64
             |  +--ro last-access-accept?   oc-types:timeticks64
             +--ro writes
                +--ro access-rejects?       oc-yang:counter64
                +--ro last-access-reject?   oc-types:timeticks64
                +--ro access-accepts?       oc-yang:counter64
                +--ro last-access-accept?   oc-types:timeticks64
```

Some comments to consider:

1. I would like to solicit thoughts around the last comment in the potential gRPC server subtree refactoring here: https://github.com/openconfig/public/pull/1096#issuecomment-2777211735
2. re: the comment above, I would forsee services as such possibly split into a few subtrees that should be classified by top-level containers that align directly with the service intent - e.g.
* Global `gnsi` related items located under a path such as: `/system/gnsi` or `/system/aaa/gnsi`
* Per NBI state/stats under its respective service endpoint: `/system/grpc-servers/grpc-server/services/gnsi`
3. In this model we have a list of policy instances, version, created-on which is good (I'd probably suggest an anchor point such as above rather) but we have augments to grpc-server state of versions w/ created-on as well.  Is this not global to the element and not "per server"?  In theory you could have multiple listening servers for gNSI services but the policy is global tmk
4. If we consider audit trails, we now have a list of policies w/ r/w accept/reject counters associated to each policy.  I believe this is incomplete as what you really want to know is given a policy, what users had hits against what rule.  e.g. something more like

```
  +--ro gnsi
     +--ro pathz
        +--ro policies
           +--ro policy* [instance]
              +--ro instance      enumeration
              +--ro version?      version
              +--ro created-on?   created-on
              +--ro users
                 +--ro user* [name]
                    +--ro name     string
                    +--ro rules
                       +--ro rule* [id]
                          +--ro id          string
                          +--ro counters
                             +--ro matches?   yang:counter64
```

earies:
> gnsi pathz only applies to gnmi... what other server do you think it would relate to? 

```
    +--ro gnmi-pathz-policy-version?      version
    +--ro gnmi-pathz-policy-created-on?   created-on
```

So the above should only be visible under a `grpc-server` that has the `GNMI` service correct?  Such a scenario imo would fall under something like (2.b) above where service state could be bound via when clauses to the services under a given server.

But I was moreso getting at the version and created-on are applied globally to the element and not per grpc-server instance.  If one wanted to say create 2 listening grpc-servers, 1 in global RIB, 1 in management they would share the same global policy thus shouldnt the applied policy be outside these servers and more well suited under something like `/system/...` ?

> as far as 4. we don't care about which users are doing what - that is what acctz is for

ok, if we scratch the users out of this, should we not pair up the nomenclature to the proto (e.g. `id` vs. `name`) and count "matches" and "last-match timestamp" against a given `AuthorizationRule` rather?  The "accept" and "reject" counters seem unnecessary here since the action in the rule already specifies this

LimeHat:
+1 to the `per policy per server` vs `per policy` question.

>gnsi pathz only applies to gnmi... what other server do you think it would relate to?

Each node can have multiple gnmi-servers (for instance, when an operator has multiple mgmt vrfs, inband, oob, etc). This is quite common.

release/models/gnsi/openconfig-gnsi-pathz.yang
```diff
@@ -109,62 +115,44 @@ module openconfig-gnsi-pathz {
...
-              https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-path-conventions.md";
+              "A per-Pathz policy the counters were
+              collected for.";
           }
           container reads {
             description
               "The counter were collected while
-              performing a read operation on the
-              schema path.";
+              performing read operations on the
```
LimeHat, line 147: what, exactly, is a `read operation` on the policy?

Is it a single RPC allowed by the policy (regardless of a number of paths authorized, e.g. a GetRequest with 10 allowed paths will be counted as +1?)

release/models/gnsi/openconfig-gnsi-pathz.yang
```diff
@@ -109,62 +115,44 @@ module openconfig-gnsi-pathz {
...
+              policy.";
             uses counters;
           }
           container writes {
             description
               "The counter were collected while
-              performing a write operation on the
-              schema path.";
+              performing write operations on the
```
LimeHat, line 154: same question as for the reads, please specify the exact definition of what should be counted here (number of set requests, number of individual operations in each request, etc)

LimeHat (changes requested):

rnhaddad:
+1 to the per policy per server vs per policy question. 

I think the current model assumes that an instance of the service (e.g. pathz) is per grpc-server. In some cases this may make sense, but in others it does not. For example: pathz is supposed to rotate the policy **for the node** and not for the server it is tied to, regardless of how many grpc-server instances are configured (note that one may still configure gNMI across multiple servers for different reasons - the model allows it).

release/models/gnsi/openconfig-gnsi-pathz.yang
```diff
@@ -109,62 +115,44 @@ module openconfig-gnsi-pathz {
     description
       "A collection of counters collected by the gNSI.pathz module.";
 
-    container gnmi-pathz-policy-counters {
+    container policies {
```
charanjith-anet, line 118: This should be called 'pathz' so that the yang tree looks as follows:

/system/grpc-servers/grpc-server
    +--ro pathz
       +--ro policy* [name]
          +--ro state

brianneville:
Nice to see the move away from xpaths :)

I have a few initial thoughts on this proposal:

1. On the notion of per-policy things. 
The pathz spec doesn't deal with multiple policies.
From the point of view of the spec, there is only one active policy on the system (unfinalized rotations can introduce the sandbox policy, but it doesnt make sense to track counters for the sandboxed policy, as this is not actually doing enforcement).
   The decision of "what happens to a policy after it has been rotated away and is no longer active" (i.e. is it stored on the device somewhere? is it deleted/overwritten?) is implementation-specific, not defined in the spec. 
In either case however, there's no reason why a non-active policy should ever sporadically become useful again, as the device can only have a new policy imposed via the Rotate RPC - it cannot be instructed to pick up a policy file from disk.
Is the idea of this PR that the YANG tree should be keeping a record of each distinct policy that was ever rotated onto the system and then track counters for each one?
  
   In that case we would only update the list entry corresponding to the the most recent policy, and then we would also have a bunch of stale entries in the list corresponding to the previous policies.
I dont know if i can see the value in keeping stale list entries around. For example, if you're interested in seeing information about historical accesses to the system, then accounting logs (acctz) would be a much more useful source of information, as those logs would include the gNMI path requested, timestamp, and information about the client.

   To me it would make more sense if we only tracked counters for the active policy instance (i.e. a container instead of a list), and then as soon as a rotation occurred we just reset the counters to 0 and start over.

2. how should policies be named/identified?
   Using the version/created-on timestamp to identify it is not a good option imo as its not guaranteed that the version/created-on are updated every time the policy is changed (as the gNSI client could set the force_overwrite field to forcibly update a policy without changing version https://github.com/openconfig/gnsi/blob/fef60085e9600881689196b52cff5dd62aa1a4c1/pathz/pathz.proto#L128 )
If we were to use the version/created-on to identify a policy, then we would either:
      
     i). need to reset the counters associated with a policy once a Rotate RPC completes, so that they reflect the policy changing (this aligns more with what i was saying in point 1 above)
or
      ii). need to accept that the counters for a particular policy may be an accumulation of any counters associated with past policies that have now been force-overwritten.

   The other option would be that the server delegates a name to the policy (e.g. monotonically incrementing version number) - and then either the gNMI client or the YANG schema would need to find some way to associate that with version/created-on information.

release/models/gnsi/openconfig-gnsi-pathz.yang
dplore: This should be called 'pathz' so that the yang tree looks as follows:

```
/system/grpc-servers/grpc-server
    +--ro pathz
       +--ro policies
           +--ro policy* [name]
              +--ro state
```

earies: See previous comments - this should not be "per gRPC server instance" so I believe the anchor point is incorrect first and foremost.

The gNSI suite should be looked at hollisically of which there are some gRPC APIs to interact (there could very well be others both native or in other OC methods)

But wherever the anchor point ends up - +1 to the last comment in order to follow the encapsulation of lists inside containers

dplore: I see,  your suggestion could perhaps be stated as reserving the  `/system/grpc-servers/grpc-server` subtree to paths which are generic across any/all grpc servers (which means it will only have things like a connection list and packet/byte counters).

The anchor for pathz can be at the system level since pathz is a "system-wide" entity.

```
  augment /oc-sys:system/:
  +--ro pathz
    +--ro policies
       +--ro policy* [name]
          +--ro state
             +--ro name?     string
             +--ro last-read-reject?   oc-types:timeticks64
             +--ro last-read-accept?   oc-types:timeticks64
             +--ro last-write-reject?   oc-types:timeticks64
             +--ro last-write-accept?   oc-types:timeticks64
             +--ro counters
             |  +--ro read-rejects?       oc-yang:counter64
             |  +--ro read-accepts?       oc-yang:counter64
                +--ro write-rejects?       oc-yang:counter64
                +--ro write-accepts?       oc-yang:counter64
```

release/models/gnsi/openconfig-gnsi-pathz.yang
dplore: Suggested description:

"The number of gnmi.Get and gnmi.Subscribe RPC's that match the policy and were accepted or rejected. "

LimeHat:
>Is the idea of this PR that the YANG tree should be keeping a record of each distinct policy that was ever rotated onto the system and then track counters for each one?

Considering that the state is ephemeral, that doesn't seem particularly useful either. 

I'd vote for simplification (count things only for one active policy; TBD with or without per-server granularity)

release/models/gnsi/openconfig-gnsi-pathz.yang
LimeHat: @earies 
> so I believe the anchor point is incorrect first and foremost.

It is not necessarily incorrect, but there's a mismatch between the PR description and the anchor point.

You can apply the same policy to multiple servers, but still implement per-server counters, especially if (as @dplore suggested) the purpose of this PR is to count number of accepted/denied RPCs (which hasn't been confirmed yet). 

After changes in openconfig/gnsi/pull/219, and combined with the idea of doing per-RPC counters, we are essentially in the authz-like territory which already implements per-server counters 
https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-authz.yang 

Which kind of brings the question, do we really need all these separate counters? Might as well just do a combined per-grpc statistics for both authz and pathz rejects.

release/models/gnsi/openconfig-gnsi-pathz.yang
brianneville: By "accept/rejected" do you mean "permitted/denied by the policy"?
its not really that binary - a gNMI read operation can be a mix of permitted and denied paths.
For example, if the policy is "`permit read /a` and `deny read /a/b`", and `/a/b` exists in YANG, then a request to read `/a` will be partially permitted, partially denied.

jsterne:
There is a mention of Subscribe RPC above. But is access control actually applied at *subscription* time or is it conceptually applied on the notification data stream? For example: if a user has permission to access interface "foo" but not interface "bar", but interface "bar" doesn't exist yet, and the user subscribes to "all interfaces" (which is only foo at the moment), what happens when interface "bar" is later added to the config? So I'm not sure that a counter for 'rejects' should even apply to Subscribe?

LimeHat:
> But is access control actually applied at subscription time or is it conceptually applied on the notification data stream?

both, depending on the policy and the subscription content.

see the recent discussion in https://github.com/openconfig/gnsi/pull/219

>if a user has permission to access interface "foo" but not interface "bar", but interface "bar" doesn't exist yet, and the user subscribes to "all interfaces" (which is only foo at the moment), what happens when interface "bar" is later added to the config? So I'm not sure that a counter for 'rejects' should even apply to Subscribe?

after the latest changes, this subscription will be rejected (regardless of the presence of interface "bar")

dplore:
Reviewed in May 13, 2025 OC Operators meeting.  Moving to "waiting for author status" so @marcushines can respond to the comments above.   One question is, is the goal to count accepted/rejected RPC's or paths?  (ie: how should an RPC with 10 paths, 9 are accepted and 1 is rejected be counted?)

PR #1284, opened 2025-04-25
Added PASSIVE as enum option for mode, it will stop all router advert…

shasingh27:
…isement and solicitation, it will only receive advertisement and solicitation

### Change Scope

### Platform Implementations

   implementation output.
   implementation output.


release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -1098,6 +1104,11 @@ revision "2023-06-30" {
             this interface.  Responses to router solicitation messages will
             be transmitted.";
         }
+        enum PASSIVE {
+          description
+            "The router advertisement neighbor discovery messages are not
+            transmitted on this interface.";
+        }
```
dplore, line 1111: This sounds like the same thing as `DISABLE_UNSOLICITED_RA`.  Can you describe how this is different from DISABLE_UNSOLICITED_RA?

dplore (changes requested):

LimeHat (changes requested):
Leaf `enable` was created exactly for this purpose
https://github.com/openconfig/public/blob/71eae392cc60e24466fe06208dd380686524859e/release/models/interfaces/openconfig-if-ip.yang#L1053-L1061

See also https://github.com/openconfig/featureprofiles/issues/2981#issuecomment-2354486208

There's no need to duplicate this functionality.

shasingh27:
> Leaf `enable` was created exactly for this purpose
> 
> https://github.com/openconfig/public/blob/71eae392cc60e24466fe06208dd380686524859e/release/models/interfaces/openconfig-if-ip.yang#L1053-L1061
> 
> See also [openconfig/featureprofiles#2981 (comment)](https://github.com/openconfig/featureprofiles/issues/2981#issuecomment-2354486208)
> 
> There's no need to duplicate this functionality.

suppress leaf was deprecated, suppress functionality was basically suppressing the sending of periodic and non-periodic advertisements.
it can still receive periodic and non-periodic advertisements

the leaf enable, when introduced:
leaf enable {
type boolean;
default true;
description
"If set to false, all IPv6 router advertisement functions are
disabled. The local system will not transmit router advertisement
messages and will not respond to router solicitation messages.";
}
it stops all the functionality, basically stopping the sending and receiving of both periodic and non-periodic advertisements.

release/models/interfaces/openconfig-if-ip.yang
shasingh27: when you set mode as DISABLE_UNSOLICITED_RA, it stops sending advertisements periodically, but solicited RAs(when requested) will still be advertised.

LimeHat:
>it stops all the functionality, basically stopping the sending and receiving of both periodic and non-periodic advertisements.

how did you come to that conclusion? the description talks only about originating the RAs: unsolicited (periodic) and solicited (non-periodic)

there's no mention of "receiving". in fact, a local system cannot control what it receives. 

`enable=false` is a direct replacement of the previous `suppress=true`.

PR #1287, opened 2025-04-29
Static wecmp

rszarecki:
### Change Scope
This change introduce WECMP weight attribute for static route next-hop. 
The new `wecmp-weight`leaf has significance only if multiple next-hops of given prefix form ECMP group, AND all of this next-hops have wecmp-weight leaf specified. Else, wecmp-weight should be ignored.

The wecmp-weight could be either uint64  or "AUTO".
The value 'AUTO' sets value to be equall to egress interface bandwidt expressed in bps (not Bps as BGP link-bandwidth extended-community does).

This leaf is valid only if recurse leaf is set to false, or if interface-ref is specified.
```
  +--rw network-instances
        +--rw protocols
              +--rw static-routes
                 +--rw static* [prefix]
                    +--rw next-hops
                       +--rw next-hop* [index]
                          +--rw index            -> ../config/index
                          +--rw config
                          |  +--rw index?          string
                          |  +--rw next-hop?       union
                          |  +--rw recurse?        boolean
                          |  +--rw metric?         uint32
                          |  +--rw preference?     uint32
+                         |  +--rw wecmp-weight?   union
                          +--ro state
                          |  +--ro index?          string
                          |  +--ro next-hop?       union
                          |  +--ro recurse?        boolean
                          |  +--ro metric?         uint32
                          |  +--ro preference?     uint32
+                         |  +--ro wecmp-weight?   union
```

### Platform Implementations

* Cisco XR: [Configure Native UCMP for Static Routing](https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5500/routing/75x/b-routing-cg-ncs5500-75x/implementing-static-routes.html#id_15043)
Note: In XR implementation presence of 'metric' with equal value set WECMP in "Local" mode what is functional equivalent to "AUTO". Lack of 'metric' make simple ECMP.
* Arista EoS: [link to documentation](https://www.arista.com/en/support/toi/eos-4-32-1f/19859-static-route-ucmp-local-forwarding)
```
ip route 10.0.0.1/8 192.168.2.2 weight 5
ip route 10.0.0.1/8 192.168.1.2 weight 3
```

### DIFF
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -124,6 +124,17 @@ module openconfig-local-routing {
       link-layer address discovery against";
   }
 
+  identity LOCAL_DEFINED_WEIGHT {
+    description
+      "A base identity type of local defined next-hops weight";
+  }
+
+  identity AUTO {
+    base LOCAL_DEFINED_WEIGHT;
+    description
+      "Inherit weight from egress interface bandwidth";
+  }
+
   // typedef statements
 
   typedef local-defined-next-hop {
@@ -133,6 +144,15 @@ module openconfig-local-routing {
     description
       "Pre-defined next-hop designation for locally generated
       routes";
+  }
+
+  typedef local-defined-weight {
+    type identityref {
+      base LOCAL_DEFINED_WEIGHT;
+    }
+    description
+      "Pre-defined wECMP weight for locally generated
+      static route next-hops";
   }
 
   // grouping statements
@@ -188,6 +208,26 @@ module openconfig-local-routing {
         inherited from the default preference of the implementation for
         static routes.";
     }
+
+    leaf wecmp-weight {
+      type union {
+        type uint64;
+        type local-defined-weight";
+      }
+      description
+        "The weight of next-hop used for WECMP (a.k.a. UCMP, WCMP). This
+        leaf has significance only if multiple next-hops of given prefix
+        form ECMP group, AND all of this next-hops have wecmp-weight leaf
+        specified. Else, wecmp-weight should be ignored.
+        The value "AUTO" sets value to be equall to egress interface
+        bandwidt expressed in bps (not Bps as BGP link-bandwidth
+        extended-community does).
+        It is recommended to express explicit value in bps, what alows
+        mixing next-hops with explicit value and next-hops with "AUTO".
+        This leaf is valid only if recurse leaf is set to false, or 
+        if interface-ref is specified.";
+    }
+
   }
 
   grouping local-static-config {
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -212,7 +212,7 @@ module openconfig-local-routing {
     leaf wecmp-weight {
       type union {
         type uint64;
-        type local-defined-weight";
+        type local-defined-weight;
       }
       description
         "The weight of next-hop used for WECMP (a.k.a. UCMP, WCMP). This
--- a/release/models/local-routing/openconfig-local-routing.yang
+++ b/release/models/local-routing/openconfig-local-routing.yang
@@ -219,7 +219,7 @@ module openconfig-local-routing {
         leaf has significance only if multiple next-hops of given prefix
         form ECMP group, AND all of this next-hops have wecmp-weight leaf
         specified. Else, wecmp-weight should be ignored.
-        The value "AUTO" sets value to be equall to egress interface
+        The value 'AUTO' sets value to be equall to egress interface
         bandwidt expressed in bps (not Bps as BGP link-bandwidth
         extended-community does).
         It is recommended to express explicit value in bps, what alows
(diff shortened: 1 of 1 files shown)


dplore:
scheduled to review at OC Operators May 13, 2025 meeting

dplore:
Reviewed in May 13, 2025 OC Operators meeting without objection.  I have a couple of grammar/syntax suggestions for the description, but otherwise LGTM.

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -43,7 +43,14 @@ module openconfig-local-routing {
     protocol-specific policy after importing the route into the
     protocol for distribution (again via routing policy).";
 
-  oc-ext:openconfig-version "2.2.0";
+  oc-ext:openconfig-version "2.3.0";
+
+  revision "2025-05-09" {
+    description
+      "Add wECMP support to static rote by wecmp-weight next-hop
+      attribute.";
```
dplore, line 51:
```suggestion
      "Add wecmp-weight leaf to support wECMP with static routes.";
```

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -131,6 +138,18 @@ module openconfig-local-routing {
       link-layer address discovery against";
   }
 
+  identity LOCAL_DEFINED_WEIGHT {
+    description
+      "A base identity type of localy defined next-hops weight.";
```
dplore, line 143:
```suggestion
      "A base identity type of locally defined next-hop weight.";
```

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -195,6 +223,26 @@ module openconfig-local-routing {
...
+        leaf has significance only if multiple next-hops of given prefix
+        form ECMP group, and all of this next-hops have ecmp-weight leaf
+        specified. Else, wecmp-weight should be ignored.
+        The value 'AUTO' sets value to be equal to egress interface
+        bandwidth expressed in bps (not Bps as BGP link-bandwidth
+        extended-community does).
+        It is recommended to express explicit value in bps, what allows
+        mixing next-hops with explicit value and next-hops with 'AUTO'.
+        This leaf is valid only if recurse leaf is set to false, or
+        if interface-ref is specified.";
```
dplore, line 243:
```suggestion
        It is recommended to express explicit value in bps to be compatible
        if mixing next-hops with wecmp-weight = 'AUTO'.
        This leaf is valid only if recurse leaf is set to false, or
        if interface-ref is specified.";
```

dplore (commented):
minor updates on wording suggested

dplore (approved):

dplore:
Last call for comments!  This will be merged on  May 27, 2025.

(See the [OpenConfig Operator Project Board](https://github.com/orgs/openconfig/projects/6/views/2) for PR status tracking)

robshakir:
One other technical question:

 * we now have `preference` and `wecmp-weight` -- but preference is specified per-nexthop.
 * should we rather have `preference` specified for the static route overall and then the `wecmp-weight` specified per `next-hop`? if not, how do these two things relate together -- are they incompatible with one another?

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -195,6 +222,26 @@ module openconfig-local-routing {
         inherited from the default preference of the implementation for
         static routes.";
     }
+
+    leaf wecmp-weight {
+      type union {
+        type uint64;
```
LimeHat, line 228: is 0 a legal value? if yes, what's the desired behavior?

robshakir:
@rszarecki PTAL at the comments, once you've updated this we'll re-enter last-call once addressed.

(Reviewed in OC operators meeting 2025-06-03.)

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -195,6 +221,26 @@ module openconfig-local-routing {
...
+        type local-defined-weight;
+      }
+      description
```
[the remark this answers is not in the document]
rszarecki: The description has suggestion to set explicit value in bps in order to allow mixing.

I think it is legitimate configuration. The client should know what it is doing.

I'm of opinion that we should not put unnecessary constrains. I would say that with software as client we can entrust sanity on client side. More of this would be human.

release/models/local-routing/openconfig-local-routing.yang
rszarecki: Good catch.

In my opinion 0 is illegal.

Will update description. Unless there is better way to express limitation.

rszarecki:
> One other technical question:
> 
> 
> 
>  * we now have `preference` and `wecmp-weight` -- but preference is specified per-nexthop.
> 
>  * should we rather have `preference` specified for the static route overall and then the `wecmp-weight` specified per `next-hop`? if not, how do these two things relate together -- are they incompatible with one another?

The 'wECMP-weight' and 'preference' are not in conflict at all.
1st 'metric' are used to select all **equall** next-hops.
2nd 'preference' are used to decide if which next-hops are system-wide best.( admin distance)
3rd if there are 2 or more static route next-hops selected as system-wide best, then 'wECMP-weight' is used to define proportion of load distribution between this next-hops.

dplore:
@rszarecki looks good now.  Can you resolve the file version conflicts?

Issue #1298, opened 2025-05-01
Clarify meaning of ecn-marked-pkts and ecn-marked-octets


rgwilton:
This PR https://github.com/openconfig/public/pull/960, introduced ECN counters to OpenConfig with two separate counters for different implementations.

The ecn-selected-pkt and ecn-selected-octets matches the Arista implementation.

The PR also introduced ecn-marked-pkts and ecn-marked-octets and cited Cisco IOS XR as an example of an implementation that implements these.  However, Cisco IOS XR does not implement the counters as defined in that PR.  More specifically, it will increment the marked packet if ECN is supported for the packet and it experiences congestion, regardless of whether it previously experienced congestion.

I.e., it changes one of the rows in the table for the PR from:

FROM:

ingressing packets ECN bits | queue utilization | ecn-selected-pkts, ecn-selected-octets | ecn-marked-pkts , ecn-marked-octets
-- | -- | -- | --
CE (11) | above RED max-treshold | increments | not increments

TO:

ingressing packets ECN bits | queue utilization | ecn-selected-pkts, ecn-selected-octets | ecn-marked-pkts , ecn-marked-octets
-- | -- | -- | --
CE (11) | above RED max-treshold | increments | **increments**

This issue (and the associated PR) is to help determine whether other vendors implement the ecn-marked-* counts as currently specified, or whether the definition can be updated to match IOS XR, if this is the only implementation.

rgwilton:
@rszarecki, tagged as the author of the original PR.

rszarecki:
> [@rszarecki](https://github.com/rszarecki), tagged as the author of the original PR.

I updated description in https://github.com/openconfig/public/pull/960 by:\
"Cisco implementation is similar to ecn-marked-pkts, ecn-marked-octets, but it increses a "RED ecn marked & transmitted" also for packte arricing with CE and enqueued when queue utilization is anove threshold."
This way at least PR description is true.

pandurangan-arista:
Starting from 4.34.0F, Arista platforms DCS-7280R, DCS-7280R2, DCS-7500R, DCS-7500R2, DCS-7280R3, DCS-7500R3 and DCS-7800R implement support for counting (only exposed via CLI as of now) as defined in PR #960 originally. We are planning to follow this up with corresponding openconfig changes, so that it matches  PR #960  specification.

PR #1299, opened 2025-05-01
Proposed change associated with #1298

rgwilton:
### Change Scope

* Clarify meaning of ecn-marked-pkts and ecn-marked-octets
* Not backwards compatible, but unclear if there are any existing implementations.
* See #1298 for more details
*
### Platform Implementations

 * Implementation A: Cisco IOS XR

This PR is intended to act as a discussion point to determine whether this change is acceptable.

### DIFF
--- a/release/models/qos/openconfig-qos-interfaces.yang
+++ b/release/models/qos/openconfig-qos-interfaces.yang
@@ -403,13 +403,23 @@ submodule openconfig-qos-interfaces {
     leaf ecn-marked-pkts {
       type oc-yang:counter64;
       description
-        "number of packets for which ECN codepoint has been changed from ECT to CE";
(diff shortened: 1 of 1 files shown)


LimeHat:
>Not backwards compatible, but unclear if there are any existing implementations.

we (Nokia) do have an implementation.

rgwilton:
> > Not backwards compatible, but unclear if there are any existing implementations.
> 
> we (Nokia) do have an implementation.

Do you know if your implementation exactly matches the specification.  I.e., you only increase the counter if the packet is newly marked and wasn't previously marked?

LimeHat:
yes, it is implemented as currently defined; ce(11) marked packets do not increment the counters

PR #1305, opened 2025-05-12
Added OpenConfig Support for Static Multicast RPF Routes

arun-arista:
* (M) release/models/multicast/openconfig-mrib.yang
* (M) release/models/policy/openconfig-policy-types.yang

### Change Scope

* In the multicast forwarding table (MFIB), each (*,G) and (S, G) route has an expected incoming interface for traffic. This interface is determined through a RPF lookup. To influence this lookup, static RPF routes can be manually configured. These static routes are stored in the MRIB and are used during the RPF check to ensure multicast traffic is forwarded correctly.
* Today it is not possible to set a static RPF route for multicast via OpenConfig. The proposal is to have a container named static-mrib for supporting rpf routes. There exists a static-routes container under protocol. Hence, we would like to have the static-mrib container at the same level, under protocol. At the same time, we had to consider adding another top level container that allows for config/state to globally configure all MRIB routes. Hence we propose another container named routes that goes under static-mrib. This way we will be able to create /network-instances/network-instance/protocols/protocol/static-mrib/{config|state} if needed, in future.

* The pyang output is given below.
```
  +--rw network-instances
        +--rw protocols
              +--rw static-mrib
                 +--rw routes
                    +--rw route* [prefix]
                       +--rw prefix       -> ../config/prefix
                       +--rw config
                       |  +--rw prefix?   openconfig-inet-types:ip-prefix
                       +--ro state
                       |  +--ro prefix?   openconfig-inet-types:ip-prefix
                       +--rw next-hops
                          +--rw next-hop* [index]
                             +--rw index            -> ../config/index
                             +--rw config
                             |  +--rw index?           string
                             |  +--rw next-hop-addr?   openconfig-inet-types:ip-address
                             |  +--rw preference?      uint8
                             +--ro state
                             |  +--ro index?           string
                             |  +--ro next-hop-addr?   openconfig-inet-types:ip-address
                             |  +--ro preference?      uint8
                             +--rw interface-ref
                                +--rw config
                                |  +--rw interface?      -> /oc-if:interfaces/interface/name
                                |  +--rw subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
                                +--ro state
                                   +--ro interface?      -> /oc-if:interfaces/interface/name
                                   +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
``` 
### Platform Implementations

 * Arista: https://www.arista.com/en/um-eos/eos-multicast-architecture
 
 * CLI Configuration:
```
switch(config)# router multicast
switch(config-router-multicast)# ipv4
switch(config-router-multicast-ipv4)# rpf route {source_prefix | source_address | mask}{rpf_interface | rpf_neighbor>} [admin_distance]
```

### DIFF
--- a/release/models/multicast/openconfig-mrib.yang
+++ b/release/models/multicast/openconfig-mrib.yang
@@ -1,0 +1,180 @@
+module openconfig-mrib {
+  yang-version "1";
+
+  // namespace
+  namespace "http://openconfig.net/yang/mrib";
+
+  prefix "oc-mrib";
+
+  // import some basic types/interfaces
+  import openconfig-extensions { prefix "oc-ext"; }
+  import openconfig-interfaces { prefix "oc-if"; }
+  import openconfig-inet-types { prefix "openconfig-inet-types"; }
+  import openconfig-interfaces { prefix "openconfig-interfaces"; }
+
+  // meta
+  organization
+    "OpenConfig working group";
+
+  contact
+    "OpenConfig working group
+    www.openconfig.net";
+
+  description
+    "Support for Static MRIB Routes.";
+
+  oc-ext:openconfig-version "0.1.0";
+
+  revision "2025-05-12" {
+    description
+      "Initial revision";
+    reference "0.1.0";
+  }
+
+  grouping mrib-next-hop-config {
+    description
+      "Configuration parameters relating to the next-hop entry.";
+
+    leaf index {
+      type string;
+      description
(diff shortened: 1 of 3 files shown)


romeyod:
@navaneethyv @rgwilton @earies can we get some inputs here? Thanks!

PR #1307, opened 2025-05-13
Add support for unicast Reverse Path Forwarding (URPF) check.

rszarecki:
### Change Scope

* This change introduces support for uRPF - the security techniqyes protecting against source addrress spoofing - described in:
  - [rfc2827](https://datatracker.ietf.org/doc/html/rfc2827).
  - [rfc3704](https://datatracker.ietf.org/doc/html/rfc3704).
  - [rfc5635](https://datatracker.ietf.org/doc/html/rfc5635).
  - [rfc8704](https://datatracker.ietf.org/doc/html/rfc8704)
The uRPF Protocol-independent, network-instance functionality with per ingress-interface configuration. This change adds following tree:
```
module: openconfig-interfaces
  +--rw interfaces
     +--rw interface* [name]
        +--rw subinterfaces
           +--rw subinterface* [index]
              +--rw oc-ip:ipv4
                 +--ro oc-ip:state
                    +--ro oc-ip:counters
                      +--ro oc-ip:urpf-drop-pkts?         oc-yang:counter64
                      +--ro oc-ip:urpf-drop-bytes?        oc-yang:counter64
               | +--rw oc-ip:urpf
               |    +--rw oc-ip:config
               |    |  +--rw oc-ip:enabled?               boolean
               |    |  +--rw oc-ip:mode?                  oc-ni-types:urpf-mode
               |    |  +--rw oc-ip:allow-default-route?   boolean
               |    |  +--rw oc-ip:allow-drop-next-hop?   boolean
               |    |  +--rw oc-ip:allow-feasible-path?   boolean
               |    +--rw oc-ip:state
               |       +--rw oc-ip:enabled?               boolean
               |       +--rw oc-ip:mode?                  oc-ni-types:urpf-mode
               |       +--rw oc-ip:allow-default-route?   boolean
               |       +--rw oc-ip:allow-drop-next-hop?   boolean
               |       +--rw oc-ip:allow-feasible-path?   boolean
              +--rw oc-ip:ipv6
                 +--ro oc-ip:state
                    +--ro oc-ip:counters
                      +--ro oc-ip:urpf-drop-pkts?         oc-yang:counter64
                      +--ro oc-ip:urpf-drop-bytes?        oc-yang:counter64
                 +--rw oc-ip:urpf
                    +--rw oc-ip:config
                    |  +--rw oc-ip:enabled?               boolean
                    |  +--rw oc-ip:mode?                  oc-ni-types:urpf-mode
                    |  +--rw oc-ip:allow-default-route?   boolean
                    |  +--rw oc-ip:allow-drop-next-hop?   boolean
                    |  +--rw oc-ip:allow-feasible-path?   boolean
                    +--rw oc-ip:state
                       +--rw oc-ip:enabled?               boolean
                       +--rw oc-ip:mode?                  oc-ni-types:urpf-mode
                       +--rw oc-ip:allow-default-route?   boolean
                       +--rw oc-ip:allow-drop-next-hop?   boolean
                       +--rw oc-ip:allow-feasible-path?   boolean
 ````
* Change is backward compatible

### Platform Implementations

 * Juniper JUNOS: [Understanding Unicast RPF (Routers)](https://www.juniper.net/documentation/us/en/software/junos/security-services/topics/topic-map/interfaces-configuring-unicast-rpf.html)
 * Cisco IOS-XR: [Implementing URPF](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/security/24xx/configuration/guide/b-system-security-cg-cisco8000-24xx/implementing-urpf.html)
 * Arista EoS: [Unicast Reverse Path Forwarding (uRPF)](https://www.arista.com/en/um-eos/eos-ipv4#xx1149080)

### CODE REVIEW

File: release/models/interfaces/openconfig-if-ip.yang [hunk 1]
```diff
@@ -1224,6 +1244,54 @@ revision "2023-06-30" {
...
+    leaf enabled {
+      type boolean;
+      default false;
+      description
+        "Turns ON/OFF packet's source address verification against content of
+        FIB. This is one of anti-spoofing protection techniques.";
+    }
+    leaf mode {
+      type urpf-mode;
+      default STRICT;
```
dplore, line 1259: Here and throughout, we have an OC bias for new models against specifying defaults, unless this is defined by RFC.  I recommend we remove the default and require the client to specify the node they want.

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -1224,6 +1246,79 @@ revision "2023-06-30" {
...
+        LPM any route.";
+    }
+    leaf allow-feasible-path {
+      type boolean;
+      default false;
+      description
+        "The routing system may select subset of equally-good path as result of
+        tie-break (e.g. BGP RID) or ECMP width limits. If set to true, packet
+        are accepted if source address LPM to any equally-good path, even if it
+        is not selected for forwading.";
```
dplore, line 1290: I'm not sure I understand this description completely.  

If the source address LPM matches an entry in a routing protocol RIB (BGP local RIB or ISIS RIB), but not in the forwarding table, it should pass the URPF check and forwarded?  Is that that intention?

rszarecki:
No. Let say we have 7 BGP paths, equally good, but since multipath is
DISABLED only one is the best. So local speaker install only one if it as
best and forwarding TOWARD given A address.

However any of peer speaker can forward traffic sourced at A toward
destination Z.

Feasible path allows to accept traffic coming from A from any of peers, not
only one.

--------------
Rafal Szarecki

release/models/interfaces/openconfig-if-ip.yang
dplore, line 1290:
```suggestion
        "The routing system may select subset of all learned paths.  For example,
        BGP without multi-path enabled will install only one of many possible ECMP
        paths into the FIB or ECMP multipath limits may select only a subset of all
        available paths. If set to true, the URPF check passes if the source address
        LPMs to any path in the RIB, even if it is not selected for forwarding in the
        FIB.";
```

Thanks for the more detailed explanation.  I suggest adding this wording to
make the intent even more clear.

dplore (approved):

release/models/interfaces/openconfig-if-ip.yang
rszarecki: It is bit more complex, but yes.

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -1224,6 +1246,79 @@ revision "2023-06-30" {
...
+    leaf allow-drop-next-hop {
+      type boolean;
+      description
+        "If set to false, and packets source address LPM the route with DROP
+        next-hop URPF fails and packet is discarded, as if it would not
+        LPM any route.";
+    }
+    leaf allow-feasible-path {
+      type boolean;
+      default false;
```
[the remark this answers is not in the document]
rszarecki: Ack

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -1224,6 +1246,79 @@ revision "2023-06-30" {
...
+    leaf enabled {
+      type boolean;
+      default false;
+      description
+        "Turns ON/OFF packet's source address verification against content of
+        FIB. This is one of anti-spoofing protection techniques.";
+    }
+    leaf mode {
+      type urpf-mode;
+      description
```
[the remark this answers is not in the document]
rszarecki: not quite undersand ...

Are you asking to add to description?

release/models/interfaces/openconfig-if-ip.yang
```diff
@@ -1224,6 +1246,79 @@ revision "2023-06-30" {
     }
   }
 
+  grouping urpf-config {
+    description
+        "Grouping configuration coniguration of URPF";
```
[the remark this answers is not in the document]
rszarecki: done

dplore:
Hi @rszarecki sorry, one more nit to pick for the path.  We should follow the existing style where all subinterface counters appear directly under a `state/counters` container:
ie:
`/interfaces/interface/subinterfaces/subinterface/ipv4/state/counters/`

The current proposal introduces another container  `.../urpf/state/counters/...`

dplore:
I'm choosing to expedite merging this as the change is relatively straightforward and a long established feature and we have a dependent, pending PR (#1320) and featureprofiles test (https://github.com/openconfig/featureprofiles/pull/3809) dependent on these changes.

akalluru1:
Some comments:
1. one minor comment/nit
Summary/description of the PR only refers to only one path, i.e., `/network-instances/network-instance/interfaces/interface[name]/subinterfaces/subinterface[index]/ipv4/urpf` & ipv6 urpf. But when I look at the full generated model urpf config was added even for `/network-instances/network-instance/interfaces/interface[name]/routed-vlan/ipv4/urpf` & ``/network-instances/network-instance/interfaces/interface[name]/tunnel/ipv4/urpf` (similarly ipv6 urpf as well). I don't think there is anything wrong with it. Just for completeness, can we list these paths as well in the PR description or add complete list of flat YANG paths added with this change.
2. EOS specific constraints: 
    - In loose urpf mode, no support for disabling `allow-drop-next-hop`
    - In strict urpf mode, no support for enabling `allow-drop-next-hop`
    - no support for disabling `allow-feasible-path` irrespective of urpf mode

rszarecki:
But when I look at the full generated model urpf config was added even for `/network-instances/network-instance/interfaces/interface[name]/routed-vlan/ipv4/urpf` & ``/network-instances/network-instance/interfaces/interface[name]/tunnel/ipv4/urpf` (similarly ipv6 urpf as well).

This are unintentional. Leaked from early modeling effort.

Will fix by ne PR, removing it.

rszarecki:
> 1. one minor comment/nit
>    Summary/description of the PR only refers to only one path, i.e., `/network-instances/network-instance/interfaces/interface[name]/subinterfaces/subinterface[index]/ipv4/urpf` & ipv6 urpf. But when I look at the full generated model urpf config was added even for `/network-instances/network-instance/interfaces/interface[name]/routed-vlan/ipv4/urpf` & ``/network-instances/network-instance/interfaces/interface[name]/tunnel/ipv4/urpf` (similarly ipv6 urpf as well). I don't think there is anything wrong with it. Just for completeness, can we list these paths as well in the PR description or add complete list of flat YANG paths added with this change.

- The added paths and leafs are under `/interfaces/interface/...` tree's branch. **Not** under `/network-instances/...`. I modified description by removing "network-instance" from text to avoid comfusion.
- Yes, this change expands to routed-vlan and tunnel as well as subinterfaces under access-point. Updated description.

> 2. EOS specific constraints:
>    
>    * In loose urpf mode, no support for disabling `allow-drop-next-hop`
>    * In strict urpf mode, no support for enabling `allow-drop-next-hop`

So handling of `DROP` next-hop is not configurable and is 1:1 tide to mode. 

>    * no support for disabling `allow-feasible-path` irrespective of urpf mode

Not an issue but strange. 

Generally as long as implementation is rejecting gNMI Set() with a) unsupported combination of values of supported leaves, b) unsoported leaves , this is not a problem.

PR #1308, opened 2025-05-15
Add GRE and MPLS encap-headers to network-instance/static

dplore:
### Change Scope

* Add GRE and MPLS encap-headers to network-instance/static
 
### Platform Implementations
While device implementations today widely support encapsulation of MPLS, GRE and MPLS into GRE packets this PR provides a method to configure a static next-hop-group for the purpose of encapsulation.  Using static next-hops for MPLS in GRE encapsulation is a new feature being pursued from multiple device implementations to unify a single vendor neutral way to configure this encapsulation.  

* Arista EOS supports [MPLS in GRE encapsulation](https://www.arista.com/en/um-eos/eos-nexthop-groups) 
* Cisco IOS XR supports [MPLS in GRE encapsulation](https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst9400/software/release/17-2/configuration_guide/mpls/b_172_mpls_9400_cg/configuring_mpls_over_gre.pdf)
* JunOS supports [MPLS in GRE encapsulation](https://www.juniper.net/documentation/us/en/software/junos/interfaces-encryption/topics/topic-map/configuring-flexible-tunnel-interfaces.html)

### Tree view
```diff
         +--rw static
         |  +--rw next-hop-groups
         |  |  +--rw next-hop-group* [name]
         |  |     +--rw name         -> ../config/name
         |  |     +--rw config
         |  |     |  +--rw name?   string
         |  |     +--ro state
         |  |     |  +--ro name?   string
         |  |     +--rw next-hops
         |  |        +--rw next-hop* [index]
         |  |           +--rw index     -> ../config/index
         |  |           +--rw config
         |  |           |  +--rw index?   -> ../../../../../../next-hops/next-hop/index
         |  |           +--ro state
         |  |              +--ro index?   -> ../../../../../../next-hops/next-hop/index
         |  +--rw next-hops
         |     +--rw next-hop* [index]
         |        +--rw index            -> ../config/index
         |        +--rw config
         |        |  +--rw index?                   string
         |        |  +--rw next-network-instance?   -> /network-instances/network-instance/config/name
         |        |  +--rw next-hop?                union
         |        |  +--rw nh-network-instance?     -> /network-instances/network-instance/config/name
         |        |  +--rw recurse?                 boolean
         |        |  +--rw metric?                  uint32
         |        |  +--rw preference?              uint32
         |        +--ro state
         |        |  +--ro index?                   string
         |        |  +--ro next-network-instance?   -> /network-instances/network-instance/config/name
         |        |  +--ro next-hop?                union
         |        |  +--ro nh-network-instance?     -> /network-instances/network-instance/config/name
         |        |  +--ro recurse?                 boolean
         |        |  +--ro metric?                  uint32
         |        |  +--ro preference?              uint32
         |        +--rw encap-headers
         |           +--rw encap-header* [index]
         |              +--rw index     -> ../config/index
         |              +--rw config
         |              |  +--rw index?   uint8
         |              |  +--rw type?    oc-aftt:encapsulation-header-type
         |              +--ro state
         |              |  +--ro index?   uint8
         |              |  +--ro type?    oc-aftt:encapsulation-header-type
         |              +--rw udp-v4
         |              |  +--rw config
         |              |  |  +--rw src-ip?         oc-inet:ipv4-address
         |              |  |  +--rw dst-ip?         oc-inet:ipv4-address
         |              |  |  +--rw dscp?           oc-inet:dscp
         |              |  |  +--rw src-udp-port?   oc-inet:port-number
         |              |  |  +--rw dst-udp-port?   oc-inet:port-number
         |              |  |  +--rw ip-ttl?         uint8
         |              |  +--ro state
         |              |     +--ro src-ip?         oc-inet:ipv4-address
         |              |     +--ro dst-ip?         oc-inet:ipv4-address
         |              |     +--ro dscp?           oc-inet:dscp
         |              |     +--ro src-udp-port?   oc-inet:port-number
         |              |     +--ro dst-udp-port?   oc-inet:port-number
         |              |     +--ro ip-ttl?         uint8
+        |              +--rw gre
+        |              |  +--rw config
+        |              |  |  +--rw src-ip?   oc-inet:ip-address
+        |              |  |  +--rw dst-ip?   oc-inet:ip-address
+        |              |  |  +--rw ttl?   oc-inet:ip-address
+        |              |  +--ro state
+        |              |     +--ro src-ip?   oc-inet:ip-address
+        |              |     +--ro dst-ip?   oc-inet:ip-address
+        |              |  |  +--rw ttl?   oc-inet:ip-address
+        |              +--rw mpls
+        |                 +--rw config
+        |                 |  +--rw traffic-class?   oc-mplst:mpls-tc
+        |                 |  +--rw label?      oc-mplst:mpls-label
+        |                 +--ro state
+        |                 |  +--rw traffic-class?   oc-mplst:mpls-tc
+        |                 |  +--rw label?      oc-mplst:mpls-label
```

### Example configuration for static next-hops with 1 MPLS label inside GRE encapsulation
```json
{
  "next-hops": {
    "next-hop": [
      {
        "config": {
          "index": "Dest A-NH1",
          "metric": 10,
          "next-hop": "1.2.3.4"
        },
        "encap-headers": {
          "encap-header": [
            {
              "config": {
                "index": 1
                "type": "GRE",
              },
              "index": 1,
              "gre": {
                "config": {
                  "dst-ip": "1.1.1.1",
                  "src-ip": "2.2.2.2",
                  "dscp": 8,
                  "ttl": 32
                }
              }
            },
            {
              "config": {
                "index": 2
                "type": "MPLS",
              },
              "index": 2,
              "mpls": {
                "config": {
                  "label": 100,
                  "traffic-class": 1
                }
              }
            }
          ]
        },
        "index": "Dest A-NH1"
      }
    ]
  }
}
```

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -329,7 +329,7 @@ submodule openconfig-aft-common {
                 description
                   "State parameters relating to GRE encapsulation headers.";
 
-                uses aft-common-entry-nexthop-encap-gre-config;
+                uses aft-common-entry-nexthop-gre-state;
               }
             }
 
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -270,7 +270,7 @@ submodule openconfig-aft-common {
             description
               "State parameters relating to GRE encapsulation.";
 
-            uses aft-common-entry-nexthop-gre-state;
+            uses aft-common-entry-nexthop-encap-gre-config;
           }
         }
 
@@ -323,7 +323,7 @@ submodule openconfig-aft-common {
                 description
                   "State parameters relating to GRE encapsulation headers.";
 
-                uses aft-common-entry-nexthop-gre-state;
+                uses aft-common-entry-nexthop-encap-gre-config;
               }
             }
 
@@ -627,7 +627,7 @@ submodule openconfig-aft-common {
     }
   }
 
-  grouping aft-common-entry-nexthop-gre-state {
+  grouping aft-common-entry-nexthop-encap-gre-config {
     description
       "GRE encapsulation applied on a IPv4 and IPv6 next-hop.";
 
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -28,7 +28,7 @@ submodule openconfig-aft-common {
   revision "2025-05-15" {
     description
       "Add GRE and MPLS to encap-headers.";
-    reference "3.0.0";
+    reference "3.1.0";
   }
 
   revision "2025-03-12" {
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,13 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "3.0.0";
+  oc-ext:openconfig-version "3.1.0";
+
+  revision "2025-07-17" {
+    description
+      "Add GRE and MPLS to encap-headers.";
+    reference "3.1.0";
+  }
 
   revision "2025-03-12" {
     description
@@ -691,7 +697,7 @@ submodule openconfig-aft-common {
         "The value of the MPLS traffic class (TC) bits, formerly known as the
          EXP bits.";
     }
-    leaf mpls-label {
+    leaf label {
       type oc-mplst:mpls-label;
       description
         "A MPLS label value.";
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -640,9 +640,11 @@ submodule openconfig-aft-common {
     leaf src-ip {
       type oc-inet:ip-address;
       description
-        "The source IP address for the GRE encapsulation may be expressed
-        using this leaf (src-ip) or if may be derived from
-        '../../interface-ref/state/subinterface'";
+        "The src-ip for the encapsulation is statically configured using this
+        leaf. If not populated the src-ip is expected to be dynamically
+        derived from the subinterface that the packet is forwarded from. If
+        an implementation requires a static src-ip, then it should return
+        an error if this is unspecified.";
     }
 
     leaf dst-ip {
@@ -668,9 +670,11 @@ submodule openconfig-aft-common {
     leaf src-ip {
       type oc-inet:ip-address;
       description
-        "The source IP address for the GRE encapsulation may be expressed
-        using this leaf (src-ip) or if may be derived from
-        '../../interface-ref/state/subinterface'";
+        "The src-ip for the encapsulation is statically configured using this
+        leaf. If not populated the src-ip is expected to be dynamically
+        derived from the subinterface that the packet is forwarded from. If
+        an implementation requires a static src-ip, then it should return
+        an error if this is unspecified.";
     }
 
     leaf dst-ip {
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,13 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "3.0.0";
+  oc-ext:openconfig-version "3.1.0";
+
+  revision "2025-05-15" {
+    description
+      "Add GRE and MPLS to encap-headers.";
+    reference "3.0.0";
+  }
 
   revision "2025-03-12" {
     description
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,13 +23,7 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "3.1.0";
-
-  revision "2025-05-15" {
-    description
-      "Add GRE and MPLS to encap-headers.";
-    reference "3.1.0";
-  }
+  oc-ext:openconfig-version "3.0.0";
 
   revision "2025-03-12" {
     description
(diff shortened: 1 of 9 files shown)


release/models/network-instance/openconfig-network-instance-static.yang
```diff
@@ -283,6 +289,52 @@ module openconfig-network-instance-static {
...
+
+            uses oc-aft:aft-common-entry-nexthop-encap-mpls-state;
+          }
+
+          container state {
+            config false;
+            description
+              "State parameters relating to GRE encapsulation headers.";
+
+            uses oc-aft:aft-common-entry-nexthop-encap-mpls-state;
```
danameme, line 334: Two comments:

1. Does not have to done/fixed in this PR

I see the MPLS header leaves are being re-used from AFT but the "mpls-label-stack" name is a bit confusing. This is defined as a union and is actually just a single label. If the user needs to configure a label stack, they will have multiple indexes in the "encap-header" with each header having an entry that has the fields label, traffic class. 

2. Unrelated to this PR

Do you think Bottom of Stack and TTL fields are needed in the AFT? Probably there's no use-case for adding these fields in the static next-hop config.

danameme: LGTM on PR.

robshakir: I think Dan's comment #1 is worth considering -- do we require N different `encap-header` list entries to create N labels in the stack, or is it acceptable to have one entry that adds N labels?

With the current approach, Dan's point that you can't set TC for each label seems reasonable.

On 2., it's not clear to me that we need to explicitly set BoS -- since this could create some "odd" things -- like, if we set BoS=1 on a header that isn't the last MPLS header in the stack. We could allow explicit TTL setting, but I'm not aware of a use-case for this currently.

dplore: I'll change the MPLS encap-header to be a single label, so a stack of labels would require multiple entries in the encap-headers list.

I think we'll need to specify TTL to be set for GRE headers when encapsulating MPLS in GRE.  We have a use case for this now.  (We can discuss the details offline).

danameme: Agreed, on not explicitly setting bottom of stack bit. Copying MPLS TTL to the outer header sounds good to me.

danameme: LGTM

danameme (approved):

dplore:
Updated the PR description to show the tree.

dplore:
@robshakir ready for your review

release/models/aft/openconfig-aft-common.yang
```diff
@@ -655,6 +661,25 @@ submodule openconfig-aft-common {
...
 
+  grouping aft-common-entry-nexthop-encap-gre-config {
+    description
+      "Configuration of GRE encapsulation applied on next-hop.";
+
+    leaf src-ip {
+      type oc-inet:ip-address;
+      description
+        "The source IP address for the GRE encapsulation may be expressed
+        using this leaf (src-ip) or if may be derived from
```
robshakir:
```suggestion
        using this leaf (src-ip) or it may be derived from
```

release/models/aft/openconfig-aft-common.yang
```diff
@@ -655,6 +661,25 @@ submodule openconfig-aft-common {
...
+        '../../interface-ref/state/subinterface'";
```
robshakir: It's not quite clear what this means. Do you intend to say that the source IP is set based on the interface the policy is applied to? I recall that there's some discussion of wanting to have policies that are not just applied to interfaces.

Also, is this widely supported?

release/models/network-instance/openconfig-network-instance-static.yang
robshakir: It feels like we just made something very implicit here -- where is it defined what the behaviour is w.r.t TTL copying?

release/models/aft/openconfig-aft-common.yang
dplore: Honestly, I cut and paste this from the existing aft definition for next-hop with GRE encapsulation, introduced at #1038 

I read this intent as the src-ip for the encapsulation may be statically configured using this leaf.  If not populated the src-ip is expected to be dynamically derived from the subinterface that the packet is forwarded from.  If an implementation requires a static src-ip, then it should return an error if this is unspecified.  

How does that sound for a description?

dplore: Updated description to be more explicit.  @romeyod for any addition comment.

romeyod: Is there a reason there is no "ip-ttl" leaf in the gre container?

romeyod: > I read this intent as the src-ip for the encapsulation may be statically configured using this leaf. If not populated the src-ip is expected to be dynamically derived from the subinterface that the packet is forwarded from. If an implementation requires a static src-ip, then it should return an error if this is unspecified.

src-ip can be statically configured or specified as local interface (whose ip addr should be used while doing gre encapsulation)

ElodinLaarz:
Discussed in the meet (June 17th, 2025).

There are some changes to fix the example JSON config.
We still need to get some feedback from the vendors to see if the behavior mentioned in [this comment thread](https://github.com/openconfig/public/pull/1308#discussion_r2137913628) is widely expected. (To be discussed at the community meet.)

Otherwise, seems good.

release/models/aft/openconfig-aft-common.yang
dplore: I see, so the "local interface" you are expecting is the ingress interface where the policy-forwarding forwarding is attached (eg, the fully qualified path being: `/network-instances/network-instance/policy-forwarding/interfaces/interface/interface-ref/config/subinterface`)

dplore: I decided to refactor the src-ip description to only specify the source IP and remove the "if unspecified" description.  This simplifies the definition and makes it explicit, avoiding any "default-like" behavior.

dplore: Also added the ip-ttl leaf

dplore:
> Just a nit: some nodes in the tree view in the description do not match with the changes:
> 
> * MPLS label should be named: "label" instead of "mpls-label"
> * TTL should be named: "ttl" instead of "ip-ttl"

Thanks for the review @trlongth I fixed the tree view in the PR description to align with the model.

dplore:
@robshakir this one is ready for your review

PR #1311, opened 2025-05-23
Add two new enums for scheduler-policy priority.

sallylsy:
### Change Scope

* Add two new enums for qos scheduler-policy priority.
### Platform Implementations

 * Implementation from Dell: [Egress QoS](https://infohub.delltechnologies.com/en-us/l/dell-enterprise-sonic-quality-of-service-qos/egress-qos-flow/#:~:text=With%20a%20scheduler%20policy%2C%20the,burst%20size%20(pbs)%20parameters.)
 * Implementation from Cicso: [Weighted deficit round robin](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/mds9000/sw/7_3/configuration/qos/cisco_mds9000_qos_config_guide/configuring_qos.pdf). [Weighted round robin](https://www.cisco.com/en/US/docs/switches/lan/csb_switching_general/olh/Sx200/1.3.0/en/Qos_Queues.html#:~:text=When%20the%20queuing%20mode%20is%20Weighted%20Round%20Robin%2C%20queues%20are,higher%20queues%20in%20strict%20priority.)

### DIFF
--- a/release/models/qos/openconfig-qos-elements.yang
+++ b/release/models/qos/openconfig-qos-elements.yang
@@ -35,7 +35,13 @@ submodule openconfig-qos-elements {
       packets for transmission, including policer and shaper
       functions";
 
-  oc-ext:openconfig-version "0.11.2";
+  oc-ext:openconfig-version "0.11.3";
+
+  revision "2025-05-22" {
+    description
+      "Add two new enums for scheduler-policy priority.";
+    reference "0.11.3";
+  }
 
   revision "2023-10-13" {
     description
@@ -1230,6 +1236,18 @@ submodule openconfig-qos-elements {
             "This scheduler term is considered as a strict priority
             term - such that packets that arrive in the queue are
             immediately serviced.";
+        }
+        enum DWRR {
+          description
+            "This scheduler term is considered as a deficit weighted
+            round robin term - such that packets that arrive in the queue are
+            serviced based on deficit counter with weighted round robin fashion.";
+        }
+        enum WRR {
+          description
+            "This scheduler term is considered as a deficit weighted
+            round robin term - such that packets that arrive in the queue are
+            serviced based on weighted round robin fashion.";
         }
       }
       description
--- a/release/models/qos/openconfig-qos-elements.yang
+++ b/release/models/qos/openconfig-qos-elements.yang
@@ -1245,7 +1245,7 @@ submodule openconfig-qos-elements {
         }
         enum WRR {
           description
-            "This scheduler term is considered as a deficit weighted
+            "This scheduler term is considered as a weighted
             round robin term - such that packets that arrive in the queue are
             serviced based on weighted round robin fashion.";
         }


release/models/qos/openconfig-qos-elements.yang
```diff
@@ -1231,6 +1237,18 @@ submodule openconfig-qos-elements {
...
         }
+        enum DWRR {
+          description
+            "This scheduler term is considered as a deficit weighted
+            round robin term - such that packets that arrive in the queue are
+            serviced based on deficit counter with weighted round robin fashion.";
+        }
+        enum WRR {
+          description
+            "This scheduler term is considered as a deficit weighted
```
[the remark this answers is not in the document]
sallylsy: Done

release/models/qos/openconfig-qos-elements.yang
```diff
@@ -1231,6 +1237,18 @@ submodule openconfig-qos-elements {
...
+            round robin term - such that packets that arrive in the queue are
+            serviced based on deficit counter with weighted round robin fashion.";
+        }
+        enum WRR {
+          description
+            "This scheduler term is considered as a weighted
+            round robin term - such that packets that arrive in the queue are
+            serviced based on weighted round robin fashion.";
+        }
       }
```
rgwilton, line 1252: Are there any strong references that provide a strong definition of what WRR or DWRR actually means?  My fear is that at least DWRR may end up being quite platform/hardware specific and if that is the case I'm thinking that we may need to pull that out in the description.

release/models/qos/openconfig-qos-elements.yang
```diff
@@ -1231,6 +1237,18 @@ submodule openconfig-qos-elements {
             term - such that packets that arrive in the queue are
             immediately serviced.";
         }
+        enum DWRR {
+          description
+            "This scheduler term is considered as a deficit weighted
+            round robin term - such that packets that arrive in the queue are
+            serviced based on deficit counter with weighted round robin fashion.";
+        }
+        enum WRR {
```
rgwilton, line 1246: With the existing model, it seems like setting priority was optional, and at least for our mappings, I think that we defaulted to WRR if STRICT priority wasn't explicitly set.  Is the intention with this change that priority should always be set?  Would it make sense to mark WRR as the default so that this change is backwards compatible from an existing implementation perspective (with the underlying assumption that other implementations map similarly)?

release/models/qos/openconfig-qos-elements.yang
ghananigans: +1. 

opening a featureprofiles PR with a functional test README would also help clarify the behaviour of these new enum values. 

For examples, see the following tests:
* https://github.com/openconfig/featureprofiles/tree/main/feature/qos/otg_tests/egress_strict_priority_scheduler_test
* https://github.com/openconfig/featureprofiles/tree/main/feature/qos/otg_tests/egress_strict_priority_scheduler_with_bursty_traffic_test

release/models/qos/openconfig-qos-elements.yang
ghananigans: It seems like OpenConfig prefers not to define defaults, at least according to https://github.com/openconfig/public/pull/1165#discussion_r1813524909.

release/models/qos/openconfig-qos-elements.yang
```diff
@@ -35,7 +35,13 @@ submodule openconfig-qos-elements {
       packets for transmission, including policer and shaper
       functions";
 
-  oc-ext:openconfig-version "0.11.2";
+  oc-ext:openconfig-version "0.11.3";
+
+  revision "2025-05-22" {
+    description
+      "Add two new enums for scheduler-policy priority.";
```
ghananigans, line 42: Can we specify the two new enums in the description? "Add WRR and DWRR enums ..."

release/models/qos/openconfig-qos-elements.yang
```diff
@@ -35,7 +35,13 @@ submodule openconfig-qos-elements {
...
+    reference "0.11.3";
```
ghananigans, line 43: Following https://github.com/openconfig/public/blob/master/doc/semver.md#general-guidelines-for-versioning-openconfig-yang-modules, should this be `0.12.0`?

release/models/qos/openconfig-qos-elements.yang
xavier-contreras: Aruba seems to also use the DWRR term: https://arubanetworking.hpe.com/techdocs/AOS-CX/AOSCX-CLI-Bank/cli_832x/Content/QoS_cmds/dwrr-queue.htm

ElodinLaarz:
Reviewed in the OC Operators Meeting on July 22nd.

https://github.com/openconfig/public/pull/1311/files#r2223216393 points out a third vendor-mention of DWRR and WRR.

In general, so long as we are able to show that something is used across multiple vendors, we do not require some other standardization for an operational use case. So, the current references are sufficient for demonstrating the terms as vendor-independent (and folks can continue to review the model change, itself).

PR #1313, opened 2025-05-28
Add ip-ttl policy forwarding action

dplore:
### Change Scope

* Add a new leaf to the policy-forwarding action container `ip-ttl` which re-writes the matching packet's ttl value.

### Platform Implementations

 * [Arista EOS traffic policy, set ttl](https://www.arista.com/en/um-eos/eos-traffic-management#topic_r14_l2q_ksb)
* IOS XR defines [ACL based forwarding](https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5500/ip-addresses/711x/configuration/guide/b-ip-addresses-cg-ncs5500-711x/implementing-access-lists-and-prefix-lists.html#id_60000) which is similar to this capability.
* JunOS defines [filter based forwarding](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/concept/firewall-filter-option-filter-based-forwarding-overview.html).

### DIFF
--- a/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
+++ b/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
@@ -430,13 +430,13 @@ submodule openconfig-pf-forwarding-policies {
     leaf ip-ttl {
       type uint8;
       description
-        "If set, this leaf causes the local system to re-write the IP TTL of
-         matching packets to the specified value and then passed to the next
-         stage of packet processing.
+        "If set, this leaf causes the local system to not decrement the TTL
+        and instead re-write the IP TTL of matching packets to the specified
+        value.
 
-         When this leaf is not set, the TTL value of the matching packet is
-         not affected by this policy and should be processed according to
-         normal forwarding rules.";
+        When this leaf is not set, the TTL value of the matching packet is
+        not affected by this policy and should be processed according to
+        normal forwarding rules.";
     }
 
   }
--- a/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
+++ b/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
@@ -21,7 +21,13 @@ submodule openconfig-pf-forwarding-policies {
     "This submodule contains configuration and operational state
     relating to the definition of policy-forwarding policies.";
 
-  oc-ext:openconfig-version "0.7.0";
+  oc-ext:openconfig-version "0.8.0";
+
+  revision "2025-05-27" {
+    description
+      "Add action to set IP TTL.";
+    reference "0.8.0";
+  }
 
   revision "2024-11-14" {
     description
@@ -404,6 +410,16 @@ submodule openconfig-pf-forwarding-policies {
         the rule. Following the decapsulation it should subsequently forward the
         encapsulated packet according to the underlying IPv4 or IPv6 header.";
     }
+
+    leaf ip-ttl {
+      type uint8;
+      description
+        "If set, this leaf causes the local system to re-write the IP TTL of
+         matching packets to the specified value. When this leaf is not set,
+         the TTL value of the matching packet is not affected by this policy
+         and should be processed according to normal forwarding rules.";
+    }
+
   }
 
   grouping pf-forwarding-policy-action-encapsulate-gre {
--- a/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
+++ b/release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
@@ -415,9 +415,12 @@ submodule openconfig-pf-forwarding-policies {
       type uint8;
       description
         "If set, this leaf causes the local system to re-write the IP TTL of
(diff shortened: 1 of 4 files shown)


robshakir (approved):
LGTM.

@danameme can you check that this meets the use case you've discussed with Darren please?

LimeHat:
Is this a new iteration of #1263 ?

In the previous PR we discussed several open questions, such as:
- when this leaf is applicable (pure IP traffic vs various encap/decap scenarios)
- at what stage of packet processing the rewrite is performed
etc

None of these questions are answered here, and I'd like to reiterate that all of this must be explicit unless you don't want to end up with 5 different implementations.

For example:

- Implementation A performs a TTL set only on outer header when doing encap
- Implementation B performs TTL rewrite only on a single (outer) IP header-
- Implementation C performs TTL rewrite only on inner header after decap
and so on and so forth.

rgwilton:
I'm not sure that Cisco XR is a good 2nd implementation reference here because XR doesn't allow rewriting to an arbitrary TTL in PBR.  Do you know if there are other implementations of this other than Arista that could be cited, or is the proposal to add this based on a single vendor configuration leaf?

saxenarahul13 (approved):
LGTM

dplore:
> Is this a new iteration of #1263 ?
> 
> In the previous PR we discussed several open questions, such as:
> 
> * when this leaf is applicable (pure IP traffic vs various encap/decap scenarios)
> * at what stage of packet processing the rewrite is performed
>   etc
> 
> None of these questions are answered here, and I'd like to reiterate that all of this must be explicit unless you want to end up with 5 different implementations.
> 
> For example:
> 
> * Implementation A performs a TTL set only on outer header when doing encap
> * Implementation B performs TTL rewrite only on a single (outer) IP header
> * Implementation C performs TTL rewrite only on inner header after decap
>   and so on and so forth.

Yes, this obsoletes #1263.  I will mark it as such.

The operational use case we are solving for is:

A customer sends a BGP packet to an edge router.  The goal is the edge router will transparently encapsulate and forward this packet to a remote BGP speaker.   

The proposed way to use this OC to solve for the use case is:
A client configures the edge router with a policy-forwarding rule which has a destination IP of the BGP speaker and an action to set TTL to 1 which applies to the ingress packet (the customer packet).   A static route is also configured on the edge router.  The destination IP is the the BGP speaker and the next-hop is configured to a static next-hop-group.   This static next-hop-group contains encap-headers (MPLS in GRE or MPLS in UDP in our use case).

The proposal here is the TTL is re-written based on policy-forwarding rules.  It is not limited to encapsulation actions.  But the use case does involve encapsulation.

LimeHat:
All I'm saying is that these rules should be added to the description of the leaf, or to a separate HLD document. Namely
1) on ingress you have a plain IP packet (you don't do any decap before the rewrite action)
2) what is the TTL of an incoming packet? do we need to perform TTL rewrite before any normal processing (which, if ingress packet TTL=1, would be sending a TTL expired response), or skip a normal TTL processing altogether? do you expect a normal decrement (-1) after the rewrite (doesn't appear so since you mentioned you want to set TTL=1..)
3) rewrite is an independent action which is not combined with any other actions in the policy rule*
*this is, by the way, an interesting deviation from the concept of policy forwarding: in this case you don't define any forwarding-related actions in the policy.

There are many nuances and all of them should be spelled out.

dplore:
> All I'm saying is that these rules should be added to the description of the leaf, or to a separate HLD document. Namely
> 
> 1. on ingress you have a plain IP packet (you don't do any decap before the rewrite action)
> 2. what is the TTL of an incoming packet? do we need to perform TTL rewrite before any normal processing (which, if ingress packet TTL=1, would be sending a TTL expired response), or skip a normal TTL processing altogether? do you expect a normal decrement (-1) after the rewrite (doesn't appear so since you mentioned you want to set TTL=1..)
> 3. rewrite is an independent action which is not combined with any other actions in the policy rule*
>    *this is, by the way, an interesting deviation from the concept of policy forwarding: in this case you don't define any forwarding-related actions in the policy.
> 
> There are many nuances and all of them should be spelled out.

Thanks for the feedback and agreed, I'll update this.

Note, there is a work in progress to expand OC to include more match + actions where the actions may not include forwarding actions.  This PR is in line with this larger expansion.  A PR and doc describing this will appear here in the coming weeks.

rgwilton:
> Note, there is a work in progress to expand OC to include more match + actions where the actions may not include forwarding actions. This PR is in line with this larger expansion. A PR and doc describing this will appear here in the coming weeks.

I think that it would be really helpful to see that wider document and PR.  It feels somewhat that new forwarding paradigms are being partially specified in OpenConfig via the addition of a few new nodes here and there in the Open Config config data model that don't always seem to currently have existing wide vendor implementations.  I think that this may mean that there is a relatively high risk of non conforming implementations because there isn't a sufficiently precise specification of the expected behaviour, i.e., not just for an individual leaf but how they are all expected to work together.

nandanarista:
lgtm

trlongth:
Hi @dplore, is anything else needed to merge this PR?

PR #1314, opened 2025-05-30
Add exclude-prefix/port leaf list

amitarista:
### Change Scope

- Add exlude-prefix in oc-defined-sets:ipv4-prefix-sets container
- Add exlude-prefix in oc-defined-sets:ipv6-prefix-sets container
- Add exclude-port in oc-defined-sets:port-sets container

Users may sometimes need to allow traffic from a broad subnet while excluding specific subnets. To achieve this, they must manually subdivide the larger subnet into smaller non-overlapping subnets that exclude the exceptions. As the number of exceptions increases, this process becomes increasingly complex and cumbersome. Hence, proposing a new attribute, exclude-prefix, which allows users to specify subnets to be excluded from the prefix list, simplifying subnet management and reducing manual effort. Similarly, exclude-port will provide the list of range of ports that has to be excluded from the port list.

### Tree view

```diff
module: openconfig-defined-sets
   +--rw defined-sets
      +--rw ipv4-prefix-sets
      |  +--rw ipv4-prefix-set* [name]
      |     +--rw name      -> ../config/name
      |     +--rw config
      |     |  +--rw name?             string
      |     |  +--rw description?      string
      |     |  +--rw prefix*           oc-inet:ipv4-prefix
+     |     |  +--rw exclude-prefix*   oc-inet:ipv4-prefix
      |     +--ro state
      |        +--ro name?             string
      |        +--ro description?      string
      |        +--ro prefix*           oc-inet:ipv4-prefix

+     |        +--ro exclude-prefix*   oc-inet:ipv4-prefix

      +--rw ipv6-prefix-sets
      |  +--rw ipv6-prefix-set* [name]
      |     +--rw name      -> ../config/name
      |     +--rw config
      |     |  +--rw name?             string
      |     |  +--rw description?      string
      |     |  +--rw prefix*           oc-inet:ipv6-prefix
+     |     |  +--rw exclude-prefix*   oc-inet:ipv6-prefix
      |     +--ro state
      |        +--ro name?             string
      |        +--ro description?      string
      |        +--ro prefix*           oc-inet:ipv6-prefix
+     |        +--ro exclude-prefix*   oc-inet:ipv6-prefix
      +--rw port-sets
         +--rw port-set* [name]
            +--rw name      -> ../config/name
            +--rw config
            |  +--rw name?           string
            |  +--rw description?    string
            |  +--rw port*           oc-pkt-match-types:port-num-range
+           |  +--rw exclude-port*   oc-pkt-match-types:port-num-range
            +--ro state
               +--ro name?           string
               +--ro description?    string
               +--ro port*           oc-pkt-match-types:port-num-range
+              +--ro exclude-port*   oc-pkt-match-types:port-num-range
```

### Platform Implementations

in Arista EOS, keyword - `except` is used to exclude prefixes or ports from the set. 

sets ( a.k.a. Field Set ) with IPv4 prefixes are configured using following syntax:
```
traffic-policies
   field-set ipv4 prefix <field-set-name>
      A.B.C.D/E      IPv4 prefix
      except         Add an exception prefix
```
Similarly for IPv6 prefixes:
```
traffic-policies
   field-set ipv6 prefix <field-set-name>
```
For Ports
```
traffic-policies
   field-set l4-port <field-set-name>
      <0-65535>       port values(s) or range(s) of port values
      except               except
```

Example:
```
traffic-policies
   field-set ipv4 prefix foo
      10.0.0.0/8
      except 10.10.0.0/16 10.20.0.0/16
```

### Are there any restrictions on the Inputs?
As long as the `exlude-prefix` entries are of the appropriate type ( IPv4, IPv6 or Port ), there are no additional restrictions. The `exclude-prefix` doesn't need to be a subset of the `prefix`. If for any reason, exclude-prefix and prefix are disjoint sets, the exclude-prefix list will be simply ignored as it will have no operational impact.

### DIFF
--- a/release/models/defined-sets/openconfig-defined-sets.yang
+++ b/release/models/defined-sets/openconfig-defined-sets.yang
@@ -27,6 +27,12 @@ module openconfig-defined-sets {
 
   oc-ext:openconfig-version "1.0.0";
 
+  revision "2025-05-30" {
+    description
+      "Add support for exclude-prefix";
+    reference "1.0.1";
+  }
+
   revision "2022-12-14" {
     description
       "Initial version of the defined set model";
@@ -57,6 +63,13 @@ module openconfig-defined-sets {
          "A user defined list of IPv4 prefixes to be used in match
          conditions. Each entry is a IPv4 + mask combination.";
     }
+
+    leaf-list exclude-prefix {
+       type oc-inet:ipv4-prefix;
+       description
+         "A user defined list of IPv4 prefixes to be excluded from
+         match conditions. Each entry is a IPv4 + mask combination.";
+    }
   }
 
   grouping ipv6-prefix-sets-config {
@@ -79,6 +92,13 @@ module openconfig-defined-sets {
        description
          "A user defined list of IPv6 prefixes to be used in match
          conditions. Each entry is a IPv6 + mask combination.";
+    }
+
+    leaf-list exclude-prefix {
+       type oc-inet:ipv6-prefix;
+       description
+         "A user defined list of IPv6 prefixes to be excluded from
+         match conditions. Each entry is a IPv6 + mask combination.";
     }
   }
 
@@ -103,6 +123,13 @@ module openconfig-defined-sets {
         description
           "A user defined set of ports to be
           used in the match conditions.";
+      }
+
+      leaf-list exclude-port {
+        type oc-pkt-match-types:port-num-range;
+        description
+          "A user defined set of ports to be excluded
+          from the match conditions.";
       }
   }
 
--- a/release/models/defined-sets/openconfig-defined-sets.yang
+++ b/release/models/defined-sets/openconfig-defined-sets.yang
@@ -25,7 +25,7 @@ module openconfig-defined-sets {
     for example, in network access control lists (i.e., filters,
     rules, etc.) in the matching fields.";
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "1.0.1";
 
   revision "2025-05-30" {
     description
--- a/release/models/defined-sets/openconfig-defined-sets.yang
+++ b/release/models/defined-sets/openconfig-defined-sets.yang
@@ -25,12 +25,12 @@ module openconfig-defined-sets {
     for example, in network access control lists (i.e., filters,
     rules, etc.) in the matching fields.";
 
-  oc-ext:openconfig-version "1.0.1";
+  oc-ext:openconfig-version "1.1.0";
 
   revision "2025-05-30" {
     description
       "Add support for exclude-prefix";
-    reference "1.0.1";
+    reference "1.1.0";
   }
 
   revision "2022-12-14" {
@@ -61,14 +61,20 @@ module openconfig-defined-sets {
        type oc-inet:ipv4-prefix;
        description
          "A user defined list of IPv4 prefixes to be used in match
-         conditions. Each entry is a IPv4 + mask combination.";
+         conditions. Each entry is a IPv4 + mask combination. Addresses
+         within these prefixes will be considered for matching,
+         unless they also fall within any of the 'exclude-prefix' entries.";
     }
 
     leaf-list exclude-prefix {
        type oc-inet:ipv4-prefix;
        description
-         "A user defined list of IPv4 prefixes to be excluded from
-         match conditions. Each entry is a IPv4 + mask combination.";
+         "A user-defined list of IPv4 prefixes to be excluded from the
+         'prefix' list. Each entry is an IPv4 address with a mask.
+         Let P be the union of all addresses in the 'prefix' list,
+         and E be the union of all addresses in the 'exclude-prefix' list.
+         The effective match set is ( P - E ), i.e., addresses in 'prefix'
+         that are not in 'exclude-prefix'.";
     }
   }
 
@@ -91,14 +97,20 @@ module openconfig-defined-sets {
        type oc-inet:ipv6-prefix;
        description
          "A user defined list of IPv6 prefixes to be used in match
-         conditions. Each entry is a IPv6 + mask combination.";
+         conditions. Each entry is a IPv6 + mask combination. Addresses
+         within these prefixes will be considered for matching,
+         unless they also fall within any of the 'exclude-prefix' entries.";
     }
 
     leaf-list exclude-prefix {
        type oc-inet:ipv6-prefix;
        description
-         "A user defined list of IPv6 prefixes to be excluded from
-         match conditions. Each entry is a IPv6 + mask combination.";
+         "A user-defined list of IPv6 prefixes to be excluded from the
+         'prefix' list. Each entry is an IPv6 address with a mask.
+         Let P be the union of all addresses in the 'prefix' list,
+         and E be the union of all addresses in the 'exclude-prefix' list.
+         The effective match set is ( P - E ), i.e., addresses in 'prefix'
+         that are not in 'exclude-prefix'.";
     }
   }
 
@@ -121,15 +133,19 @@ module openconfig-defined-sets {
       leaf-list port {
         type oc-pkt-match-types:port-num-range;
         description
-          "A user defined set of ports to be
-          used in the match conditions.";
+          "A user defined set of ports to be used in the match conditions.
+          Ports listed here will be considered for matching, unless they
+          are also included in 'exclude-ports'.";
       }
 
       leaf-list exclude-port {
         type oc-pkt-match-types:port-num-range;
         description
-          "A user defined set of ports to be excluded
-          from the match conditions.";
+          "A user-defined list of ports to be excluded from the 'ports'
+          list. Let P be the union of the ports in the 'ports' list,
+          and E be the union of ports in the 'exclude-ports' list.
+          The effective match set is ( P - E ) i.e., ports in 'ports'
+          that are not in 'exclude-ports'.";
       }
   }
 
--- a/release/models/defined-sets/openconfig-defined-sets.yang
+++ b/release/models/defined-sets/openconfig-defined-sets.yang
@@ -61,7 +61,7 @@ module openconfig-defined-sets {
        type oc-inet:ipv4-prefix;
        description
          "A user defined list of IPv4 prefixes to be used in match
-         conditions. Each entry is a IPv4 + mask combination. Addresses
+         conditions. Each entry is an IPv4 + mask combination. Addresses
          within these prefixes will be considered for matching,
          unless they also fall within any of the 'exclude-prefix' entries.";
     }
@@ -97,7 +97,7 @@ module openconfig-defined-sets {
        type oc-inet:ipv6-prefix;
        description
          "A user defined list of IPv6 prefixes to be used in match
-         conditions. Each entry is a IPv6 + mask combination. Addresses
+         conditions. Each entry is an IPv6 + mask combination. Addresses
          within these prefixes will be considered for matching,
          unless they also fall within any of the 'exclude-prefix' entries.";
     }
@@ -141,11 +141,11 @@ module openconfig-defined-sets {
       leaf-list exclude-port {
         type oc-pkt-match-types:port-num-range;
         description
-          "A user-defined list of ports to be excluded from the 'ports'
-          list. Let P be the union of the ports in the 'ports' list,
-          and E be the union of ports in the 'exclude-ports' list.
-          The effective match set is ( P - E ) i.e., ports in 'ports'
-          that are not in 'exclude-ports'.";
+          "A user-defined list of ports to be excluded from the 'port'
+          list. Let P be the union of the ports in the 'port' list,
+          and E be the union of ports in the 'exclude-port' list.
+          The effective match set is ( P - E ) i.e., ports in 'port'
+          that are not in 'exclude-port'.";
       }
   }
 
--- a/release/models/defined-sets/openconfig-defined-sets.yang
+++ b/release/models/defined-sets/openconfig-defined-sets.yang
@@ -29,7 +29,7 @@ module openconfig-defined-sets {
 
   revision "2025-05-30" {
     description
-      "Add support for exclude-prefix";
+      "Add support for exclude-prefix and exclude-port";
     reference "1.1.0";
   }
 
@@ -135,7 +135,7 @@ module openconfig-defined-sets {
         description
           "A user defined set of ports to be used in the match conditions.
           Ports listed here will be considered for matching, unless they
-          are also included in 'exclude-ports'.";
+          are also included in 'exclude-port'.";
       }
 
       leaf-list exclude-port {


robshakir:
@amitarista - Please can you check and sign the CLA?

amitarista:
> @amitarista - Please can you check and sign the CLA?

CLA has been signed. Thanks!

LimeHat:
>Arista: https://www.arista.com/en/um-eos/eos-ethernet-ports#topic_djn_qrh_jcc ( under traffic-policies field-set )

1. can you point me to a specific example there? The only potentially relevant thing I see there is the keyword `except`, but it is unclear if it is the same thing. The command description says
> The traffic-policies field-set creates field sets of prefixes, ports, or VLAN IDs to load from a location set by a URL. Reuse large field sets without copying and pasting each field set into a configuration. 

 which doesn't seem relevant.

2. Can you please also indicate the valid or invalid combinations of the proposed actions? Are there any restrictions for the values (e.g. does the `exclude-prefix` value have to be within the limits of another `prefix` value?)

3. Is there a practical use case for this? Is this "optimization" done solely for non-automated ops?

4. A few concerns come to mind
- We introduce complexity under the premise of simplification. I'd argue that generating prefix-lists on the client side is not that hard in most cases. 
- This will be hard to troubleshoot, esp. when you add multiple prefixes and multiple exclusions
- In many cases, exclusion can be handled as a separate entry. e.g. if you intend to create an acl rule that says drop traffic destined to ports `1024-2048` and `4096-8192`, but allow `1337`, you can create an entry to allow 1337 in front of your `drop` rule.

amitarista:
> > One of the key advantages of using the exclude-prefix option is the efficient utilization of TCAM space. Without this option, achieving the same functionality would require additional entries arranged in a specific order, resulting in increased TCAM consumption
> 
> Can you please elaborate on this? How does this help to optimize TCAM utilization? Are you claiming that a prefix, or a port range, with an `exclude` statement can fit in the single TCAM entry?



> > One of the key advantages of using the exclude-prefix option is the efficient utilization of TCAM space. Without this option, achieving the same functionality would require additional entries arranged in a specific order, resulting in increased TCAM consumption
> 
> Can you please elaborate on this? How does this help to optimize TCAM utilization? Are you claiming that a prefix, or a port range, with an `exclude` statement can fit in the single TCAM entry?

Without `exclude-prefix`, multiple ACL entries must be created, leading to a proportional increase in hardware resource usage, including TCAM space and counters. Its benefits become more significant as scale increases.

LimeHat:
ACL entries and TCAM entries are not the same. An ACL entry can consume multiple TCAM entries. A standard/naïve ACL implementation will have to do exactly that to implement this proposal.
And this is where hidden complexity comes into play: potentially we are talking about exploding the number of used TCAM entries in a significant way, while providing no visibility to user and creating a false impression that this does not increase resource usage.

ElodinLaarz:
For model augmentations, we generally prefer to see evidence of support from at least two vendor implementations. It's noted that Arista uses the `except` keyword for this kind of functionality. Are there analogous functions or similar support in other vendor systems?

---

[This might be a silly question; so, apologies in advance.]

The proposal mentions that there are no restrictions on the input prefixes. However, set subtraction is not commutative, which leads to some questions about the intended behavior when configured prefixes and exclude-prefixes result in nested or overlapping IP 'intervals'.

(I am secretly a mathematician; so, if this notation is confusing, let me know, and I can provide concrete CIDR addresses.)

Consider a set of IP addresses ordered as follows: `a1 < b1 < c1 < c2 < b2 < a2`.
Consider the subnets `[a1, a2]`, `[b1, b2]`, and `[c1, c2]`.

What would be the outcome when configuring the intervals `[a1, a2] U [c1, c2]`, whilst excluding the interval `[b1, b2]`?

The union `[a1, a2] U [c1, c2]` simplifies to `[a1, a2]` because `[c1, c2]` is contained within `[a1, a2]`.

If we then apply the exclusion:
> `([a1, a2] U [c1, c2]) - [b1, b2]`
> `= [a1, a2] - [b1, b2]`
> `= [a1, b1) U (b2, a2]`

However, if the operations are interpreted to do exclusion first:
> `([a1, a2] - [b1, b2]) U [c1, c2]`
> `= ([a1, b1) U (b2, a2]) U [c1, c2]`
> `= [a1, b1) U [c1, c2] U (b2, a2]`

Could you clarify:
*   Is there a predefined order for these operations (e.g., are all `prefix` entries unioned first, and then all `exclude-prefix` entries are unioned and subtracted from the first result)? Longest prefix matching to rules? Or is there another evaluation order?

---

Finally, regarding the `state` paths for `prefix` and `exclude-prefix`, is the expectation that the read paths will return the same values as those configured? Or can there be some reduction or canonicalization of these values by the router, where the state might reflect the *effective* set of IP addresses after all inclusion and exclusion operations have been resolved?

amitarista:
> For model augmentations, we generally prefer to see evidence of support from at least two vendor implementations. It's noted that Arista uses the `except` keyword for this kind of functionality. Are there analogous functions or similar support in other vendor systems?

I hope other vendors can add info on that. Could you please tag them to the review?

> [This might be a silly question; so, apologies in advance.]
> 
> The proposal mentions that there are no restrictions on the input prefixes. However, set subtraction is not commutative, which leads to some questions about the intended behavior when configured prefixes and exclude-prefixes result in nested or overlapping IP 'intervals'.
> 
> (I am secretly a mathematician; so, if this notation is confusing, let me know, and I can provide concrete CIDR addresses.)
> 
> Consider a set of IP addresses ordered as follows: `a1 < b1 < c1 < c2 < b2 < a2`. Consider the subnets `[a1, a2]`, `[b1, b2]`, and `[c1, c2]`.
> 
> What would be the outcome when configuring the intervals `[a1, a2] U [c1, c2]`, whilst excluding the interval `[b1, b2]`?
> 
> The union `[a1, a2] U [c1, c2]` simplifies to `[a1, a2]` because `[c1, c2]` is contained within `[a1, a2]`.
> 
> If we then apply the exclusion:
> 
> > `([a1, a2] U [c1, c2]) - [b1, b2]`
> > `= [a1, a2] - [b1, b2]`
> > `= [a1, b1) U (b2, a2]`
> 
> However, if the operations are interpreted to do exclusion first:
> 
> > `([a1, a2] - [b1, b2]) U [c1, c2]`
> > `= ([a1, b1) U (b2, a2]) U [c1, c2]`
> > `= [a1, b1) U [c1, c2] U (b2, a2]`
> 
It will be ([a1, a2] U [c1, c2]) - [b1, b2]
> Could you clarify:
> * Is there a predefined order for these operations (e.g., are all `prefix` entries unioned first, and then all `exclude-prefix` entries are unioned and subtracted from the first result)? Longest prefix matching to rules? Or is there another evaluation order?

You got it right.  The packet should match on the [ Union of Prefixes in the prefix-list - Union of Prefixes in the exclude-list ]

> 
> Finally, regarding the `state` paths for `prefix` and `exclude-prefix`, is the expectation that the read paths will return the same values as those configured? Or can there be some reduction or canonicalization of these values by the router, where the state might reflect the _effective_ set of IP addresses after all inclusion and exclusion operations have been resolved?

Read paths will return the same values as those configured.

ElodinLaarz:
If you look at previous PRs to change the model, you'll find that the proposer of the PR shows why the model should include the update by demonstrating (at least) one other vendor with the same property. This helps us avoid introducing vendor-specific implementations into the model.

Certainly if anyone knows of another instance, they're free to add it here, but the onus is on the original author to demonstrate that the request is a 'reasonably common' feature.

---

Seems as reasonable of a choice as any another. It does complicate the resolution, though. (Now, entire prefixes that are included in the `prefix` config can be overwritten with the `exclude` prefixes.)

And presumably this `U{additions} - U{exclusions}` applies the same for the ports?

Is there any reason we don't want to impose restrictions for consistency checks?
(We don't have any for the prefixes, as-is, but only adding can't 'overwrite' other bits of config in some way you may not intend.)

In any case, can you add some explanation to the comments in the file change with the expectation on how to hand the resolution of "all additions" and then "all exclusions" to make it clear to implementers?

release/models/defined-sets/openconfig-defined-sets.yang
```diff
@@ -55,7 +61,20 @@ module openconfig-defined-sets {
        type oc-inet:ipv4-prefix;
        description
          "A user defined list of IPv4 prefixes to be used in match
-         conditions. Each entry is a IPv4 + mask combination.";
+         conditions. Each entry is a IPv4 + mask combination. Addresses
```
ElodinLaarz: nit: `is "an" IPv4`.

release/models/defined-sets/openconfig-defined-sets.yang
```diff
@@ -78,7 +97,20 @@ module openconfig-defined-sets {
        type oc-inet:ipv6-prefix;
        description
          "A user defined list of IPv6 prefixes to be used in match
-         conditions. Each entry is a IPv6 + mask combination.";
+         conditions. Each entry is a IPv6 + mask combination. Addresses
```
ElodinLaarz: nit: `is "an" IPv6`.

release/models/defined-sets/openconfig-defined-sets.yang
```diff
@@ -101,8 +133,19 @@ module openconfig-defined-sets {
...
-          used in the match conditions.";
+          "A user defined set of ports to be used in the match conditions.
+          Ports listed here will be considered for matching, unless they
+          are also included in 'exclude-ports'.";
+      }
+
+      leaf-list exclude-port {
+        type oc-pkt-match-types:port-num-range;
+        description
+          "A user-defined list of ports to be excluded from the 'ports'
```
ElodinLaarz: Replace `ports` -> `port` and `exclude-ports` -> `exclude-port`.

amitarista:
> LGTM (after minor comments addressed).

Thanks for the comments, updated.

release/models/defined-sets/openconfig-defined-sets.yang
```diff
@@ -101,8 +133,19 @@ module openconfig-defined-sets {
       leaf-list port {
         type oc-pkt-match-types:port-num-range;
         description
-          "A user defined set of ports to be
-          used in the match conditions.";
+          "A user defined set of ports to be used in the match conditions.
+          Ports listed here will be considered for matching, unless they
+          are also included in 'exclude-ports'.";
```
ElodinLaarz: `exclude-ports` -> `exclude-port`.

release/models/defined-sets/openconfig-defined-sets.yang
```diff
@@ -25,7 +25,13 @@ module openconfig-defined-sets {
     for example, in network access control lists (i.e., filters,
     rules, etc.) in the matching fields.";
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "1.1.0";
+
+  revision "2025-05-30" {
+    description
+      "Add support for exclude-prefix";
```
ElodinLaarz: "... and exclude-port".

ElodinLaarz:
@dplore could you move this to ready-to-discuss? :)

dplore:
Set last call for July 24, 2025

dplore:
OC operator approval only needs the operational use cases and feature precedent (or evidence of active feature development) across more than one implementation to show feasibility and an indication of vendor neutrality.  A description of TCAM optimization is not required to approve this PR.  

The generalized operational use case is parity with existing, CLI based, prefix lists which are deployed today and used in  policy-forwarding, qos classifiers and ACLs.

LimeHat:
Are we saying that any proposal can have unsubstantiated claims and the community is totally fine with it? Authors are not required to provide details or any evidence of the truth behind it?  
I find this troubling, but if that's the direction the community wants to take, sure. 

> The generalized operational use case is parity with existing, CLI based, prefix lists which are deployed today and used in policy-forwarding, qos classifiers and ACLs.

Nothing related to policy-forwarding or qos is present in this PR.

I also believe that OC community should consider operational consequences and best practices and have a bit of a strategy. 
Rubber stamping everything that has 1.5 implementations is not a great strategy.

dplore:
@amitarista can you remove the TCAM savings assertion?  We only need operational use cases to justify the addition of the exclude leaf.

amitarista:
> @amitarista can you remove the TCAM savings assertion? We only need operational use cases to justify the addition of the exclude leaf.

Removed. Thanks!

nleiva:
Thanks for the proposal and for outlining the motivation behind `exclude-prefix` and `exclude-port.` 

I wonder if introducing exclusion logic at the model level adds unnecessary complexity. In many cases, couldn’t we achieve the same outcome by defining separate sets? One for what we want to allow, and another for what we want to deny or drop? This might offer clearer semantics and keep the model simpler for users and implementers alike.

Curious to hear if you've run into limitations with that approach. Thanks.

ElodinLaarz:
Discussed at the OC Operators Meeting on August 5th; there are still a few comments that folks want to leave / resolve; so, we'll push out the last-call to next week.

dplore:
Here is an example use case.  As a network operator, I want to set the next-hop-group all traffic inbound from the Internet except for a few prefixes that I trust.  The resulting prefix list is much easier to read than the inverse.  We prefer this on the device because it's easier to debug.

The implementation details of the rule are very device specific.  Expansion may or may not be required depending on the scenario and device.  As far as observability where a large amount of expansion occurs in a classic TCAM scenario, this should be observed indirectly by the resource utilization.  

[/components/component/integrated-circuit/utilization/resources/resource/state/used](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-integrated-circuit-utilization-resources-resource-state-used)

Here's an example config in OC JSON with just a few prefixes excluded from a larger range.

```json
{
  "defined-sets": {
    "ipv4-prefix-sets": {
      "ipv4-prefix-set": [
        {
          "config": {
            "name": "AllExceptAFew",
            "prefix": [
              "0.0.0.0/0"
            ]
            "exclude-prefix": [
              "192.0.2.0/24", "198.51.100.0/24", "203.0.113.0/24"
            ]
          },
          "name": "AllExceptAFew"
        }
      ]
    }
  },
  "network-instances": {
    "network-instance": [
      {
        "config": {
          "name": "DEFAULT"
        },
        "name": "DEFAULT",
        "policy-forwarding": {
          "interfaces": {
            "interface": [
              {
                "config": {
                  "apply-forwarding-policy": "Set Next Hop Group",
                  "interface-id": "eth0"
                },
                "interface-id": "eth0"
              }
            ]
          },
          "policies": {
            "policy": [
              {
                "config": {
                  "policy-id": "Set NHG"
                },
                "policy-id": "Set NHG",
                "rules": {
                  "rule": [
                    {
                      "action": {
                        "config": {
                          "next-hop-group": "NHG-1"
                        }
                      },
                      "config": {
                        "sequence-id": 1
                      },
                      "ipv4": {
                        "config": {
                          "destination-address-prefix-set": "AllExceptAFew"
                        }
                      },
                      "sequence-id": 1
                    }
                  ]
                }
              }
            ]
          }
        }
      }
    ]
  }
}
```


Note that [/defined-sets](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-defined-sets.html#defined-sets) do not have permit or deny statements. Instead an action (drop, forward, set-nhg, etc) is defined within the `/routing-policy`, `/network-instances/network-instance/policy-forwarding`,  `/qos/classifiers`, `/acl` that references the  prefix-set.

dplore (approved):
After a few discussions, the operational use case for this configuration still stands.

PR #1315, opened 2025-06-12
Define style for use of default node

dplore:
### Change Scope
This updates the OC yang style guide to avoid use the use of defaults.  Defaults are often a source of conflict between implementations.  It more clear and vendor neutral to simply require the caller to specify the value they want to use.  The exception to this is when the feature specification (such as an IETF RFC) defines a default as a requirement.  In this case, the OC model should define the defaults in line with the feature specification.

### DIFF
--- a/doc/openconfig_style_guide.md
+++ b/doc/openconfig_style_guide.md
@@ -264,9 +264,14 @@ including how modules should be structured and parsed.
 
 ### `default`
 
-The use of default should be avoided.  Instead a leaf should be explicitly
-set. Defaults may be defined when they are explicitly required by the feature
-specification.
+The use of default should be avoided. Defaults are often a source of conflict
+between implementations where there is no clear and correct requirement. In
+these cases, it is more vendor neutral to simply require the caller to
+specify the value they want to use.
+
+Defaults may be defined when they are explicitly required by the feature
+specification (such as an IETF RFC or IEEE standard).  In these cases the OC
+model should define the defaults in line with the feature specification.
 
 ### `list`
 
--- a/doc/openconfig_style_guide.md
+++ b/doc/openconfig_style_guide.md
@@ -1,10 +1,9 @@
 
 # YANG authoring guidelines for OpenConfig models
 
-**Contributors:** Anees Shaikh, Rob Shakir, Kristian Larsson<br>
+**Contributors:** Anees Shaikh, Rob Shakir, Kristian Larsson, Darren Lohr<br>
 **October 26, 2015**<br>
-*Updated: June 2, 2019*
-
+*Updated: June 12, 2025*
 
 ## Background
 This document describes conventions adopted in the OpenConfig operator group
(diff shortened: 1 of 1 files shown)


doc/openconfig_style_guide.md
```diff
@@ -256,11 +258,16 @@ identity LC_CONNECTOR {
...
 ```
 
-
 ## YANG language usage
 Language rules describe guidelines on use of specific YANG language statements,
 including how modules should be structured and parsed.
 
+### `default`
+
+The use of default should be avoided.  Instead a leaf should be explicitly
```
ghananigans: Can we also include the rationale for this recommendation? the details you have in this PR's description seems to be useful here as well

dplore: Good idea, I've updated the text

doc/openconfig_style_guide.md
```diff
@@ -256,11 +258,21 @@ identity LC_CONNECTOR {
...
 including how modules should be structured and parsed.
 
+### `default`
+
+The use of default should be avoided. Defaults are often a source of conflict
+between implementations where there is no clear and correct requirement. In
+these cases, it is more vendor neutral to simply require the caller to
+specify the value they want to use.
+
+Defaults may be defined when they are explicitly required by the feature
```
earies, line 272: Shall we consider another slightly obscure case that was pushed into a couple of models a few years back

https://github.com/openconfig/public/blob/master/release/models/network-instance/openconfig-network-instance.yang#L1374-L1386
https://github.com/openconfig/public/blob/master/release/models/system/openconfig-system-grpc.yang#L117-L129

Both are cases for list keys where:

* If you are looking to retrieve OpenConfig modeled state but did not configure via OpenConfig and a list instance is singular within an implementation then the implementation has guidance to emit said default value in this scenario
* For the caller, it is a contract of sorts to use the default value when configuring a singular list instance or the implementation does not support named entities (good example is the NI protocol list in most implementations does not support > 1 protocol instance)

The 'default' here is rather guidance/contractual

doc/openconfig_style_guide.md
```diff
@@ -256,11 +258,21 @@ identity LC_CONNECTOR {
...
+specification (such as an IETF RFC or IEEE standard).  In these cases the OC
+model should define the defaults in line with the feature specification.
```
ghananigans, line 274: When a default is defined, should we also require a featureprofiles test that validates the default? Is such a test feasible?

doc/openconfig_style_guide.md
```diff
@@ -1,10 +1,9 @@
 
 # YANG authoring guidelines for OpenConfig models
 
-**Contributors:** Anees Shaikh, Rob Shakir, Kristian Larsson<br>
+**Contributors:** Anees Shaikh, Rob Shakir, Kristian Larsson, Darren Lohr<br>
```
robshakir:
```suggestion
**Contributors:** Anees Shaikh, Rob Shakir, Kristian Larsson, Darren Loher  
```

doc/openconfig_style_guide.md
robshakir: Ideally, yes -- but also we have some decoupling between `featureprofiles` and the models themselves -- i.e., all `featureprofiles` tests use OpenConfig models, but all OpenConfig models do not (necessarily) have `featureprofiles` tests.

doc/openconfig_style_guide.md
robshakir: I think this is a bit niche -- the style guide will not cover every scenario.

dplore:
OC Operators suggested also elaborating on what happens if a required node is left out of a config.  (The implementation may throw an error (preferred) or may fill in it's own, "target defined" default.  Operators are advised to specify all the values they want configured.

doc/openconfig_style_guide.md
dplore: These are good examples to individual use cases which might recur and a default could be used.  However, I think the way to account for these is to have a good description embedded in the model justifying how and why a default is defined.  These are not common enough IMO to include in the style guide.

doc/openconfig_style_guide.md
dplore: This is a good thought!  However, [openconfig/featureprofiles](https://github.com/openconfig/featureprofiles/tree/main) is expected to remain a subset of all OpenConfig models for the foreseeable future simply due to the size of all the OC models compared to the size and effort to expand featureprofiles tests.   

That said, there are a few featureprofiles which include coverage for OC and implementation defined defaults.

doc/openconfig_style_guide.md
dplore: Done

robshakir:
I disagree that we should say that the target should ever fill in required information that is missing but does not have a default. This makes the API contract exceptionally unclear. If a field is required and it is not supplied in the input config then this should report an error.

ElodinLaarz:
Discussed at the July 1st OC Operators Meeting -- This does technically allow the use of defaults that are not specified by an RFC, etc. (since it doesn't say "must not" use) but there was agreement that the language is fine as-is.

PR #1316, opened 2025-06-18
Expand zone index regex for ip-address-zoned

mukherso:
**Files changed:**

- (M) release/models/types/openconfig-inet-types.yang

**Change Scope**

- Updated the regex for the zone index in the typedefs ipv4-address-zoned and ipv6-address-zoned from %[a-zA-Z0-9_]+ to %[a-zA-Z0-9_.-/]+.

- This change allows additional special characters ., -, and / in the zone index to better support real-world interface names such as Ethernet1/2/3, Port-Channel10, and Ethernet1.1.
 
- The change is backward compatible, as it only relaxes the regex constraints to include more valid zone identifiers.

Relevant pyang output snippet for context:

```
module: openconfig-inet-types
  typedef ipv4-address-zoned {
    type union {
      type inet:ipv4-address;
      type string {
        pattern '[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3}(%[a-zA-Z0-9_.-/]+)';
      }
    }
  }
  typedef ipv6-address-zoned {
    type union {
      type inet:ipv6-address;
      type string {
        pattern '([0-9a-fA-F]{1,4}:){7}[0-9a-fA-F]{1,4}(%[a-zA-Z0-9_.-/]+)?';
      }
    }
  }

```

**Platform Implementations**

[Arista EOS User Guide](https://www.arista.com/en/um-eos/eos-ethernet-ports#xx1142461) contains examples of interface names with /, ., and - characters.


release/models/types/openconfig-inet-types.yang
```diff
@@ -154,10 +160,10 @@ module openconfig-inet-types {
     type string {
       pattern
         '([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\.([0-9]|'
-        + '[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3}(%[a-zA-Z0-9_]+)';
+        + '[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3}(%[a-zA-Z0-9_.-/]+)';
```
earies: Another possible character thinking off the top of my head would be `:`

release/models/types/openconfig-inet-types.yang
```diff
@@ -154,10 +160,10 @@ module openconfig-inet-types {
     type string {
       pattern
         '([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\.([0-9]|'
-        + '[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3}(%[a-zA-Z0-9_]+)';
+        + '[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3}(%[a-zA-Z0-9_.-/:]+)';
       oc-ext:posix-pattern
         '^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\.([0-9]|'
-        + '[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3}(%[a-zA-Z0-9_]+))$';
+        + '[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3}(%[a-zA-Z0-9_.-/:]+))$';
```
ghananigans: This pattern looks like it'll allow `IP_ADDRESS%:.-` which doesn't seem like a valid interface name for a zone index. 

Would a regex like `%[a-zA-Z0-9_]([.-/:][a-zA-Z0-9_]+)?` be more appropriate for the zone index portion?

release/models/types/openconfig-inet-types.yang
```diff
@@ -31,7 +31,13 @@ module openconfig-inet-types {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.7.0";
+  oc-ext:openconfig-version "0.7.1";
```
earies: 0.8.0

regexp-tests/openconfig-inet-types-test.yang
```diff
@@ -27,6 +27,12 @@ module openconfig-inet-types-test {
     pt:pattern-test-fail "256.255.255.255";
     pt:pattern-test-fail "1.1.1.256";
     pt:pattern-test-fail "256.1.1.1%eth0";
+    pt:pattern-test-pass "192.168.1.1%Ethernet1/2";
+    pt:pattern-test-pass "10.0.0.1%Port-Channel10";
+    pt:pattern-test-pass "172.16.0.1%Ethernet1.1";
+    pt:pattern-test-pass "10.10.10.10%ns:eth1";
+    pt:pattern-test-fail "10.0.0.1%zone#1";
+    pt:pattern-test-fail "10.1.1.1%:.-";
```
ghananigans: Can we split this test to individually test each symbol? Test cases with the following pattern would be great for each delimiting symbol: `ADDRESS%(IF_NAME)?[:./-](SUB_IF_NAME)?`.

For example, `10.1.1.1%-`, `10.1.1.1%Port-`, `10.1.1.1%-Channel1` should fail, but `10.1.1.1%Port-Channel1` should pass. 

Same for IPv6.

ghananigans:
Thanks!

ghananigans (approved):

regexp-tests/openconfig-inet-types-test.yang
```diff
@@ -16,17 +16,45 @@ module openconfig-inet-types-test {
...
+    pt:pattern-test-fail "10.0.0.1%.eth0";
+    pt:pattern-test-fail "10.0.0.1%..";
+    // ADDRESS%(IF_NAME)?[/](SUB_IF_NAME) tests
+    pt:pattern-test-pass "192.168.1.1%Ethernet1/2";
+    pt:pattern-test-pass "10.10.10.10%uplink/4";
+    pt:pattern-test-fail "10.10.10.10%/";
+    pt:pattern-test-fail "10.10.10.10%/uplink";
+    pt:pattern-test-fail "10.10.10.10%uplink/";
+    pt:pattern-test-fail "10.10.10.10%uplink//2";
+    // ADDRESS%(IF_NAME)?[-](SUB_IF_NAME) tests
```
dplore, line 51: This should also add a set of test strings for `_` since that is allowed in the pattern

dplore (changes requested):
Please add a set of test strings for `_` since that is allowed in the pattern.  Otherwise, LGTM

dplore (approved):

dplore:
Last call for comments ends Aug 26, 2025, then this will merge.

PR #1324, opened 2025-06-30
Fix additions of NI-referencing next-hops in openconfig-local-routing.

robshakir:
```
 * (M) release/models/local-routing/openconfig-local-routing.yang
  - Remove leaves that require the network-instance model from the
    base openconfig-local-routing. Modules should be standalone so
    that they can be re-used outside of network-instance if required.
 * (A) release/models/local-routing/openconfig-local-routing-network-instance.yang
  - Add network-instance referencing leaves through augments.
 * (M) release/models/local-routing/.spec.yml
 * (M) release/models/network-instance/.spec.yml
   - Update spec to ensure that a build rule checks the new module.
```

PR #1325, opened 2025-07-01
Add next-hop-group to local-routing-network-instance.

dplore:
### Change Scope

This is a follow up to #1324, refactoring of yang model groupings for openconfig-local-routing.  This moves the `static/next-hop-group` grouping into the local-routing-network-instances module, allowing it to be reusable in modules that do not include network-instances.

No paths are changed, just the module in which they are contained.

```diff
         |     +--rw static-routes
         |     |  +--rw static* [prefix]
         |     |     +--rw prefix            -> ../config/prefix
         |     |     +--rw config
         |     |     |  +--rw prefix?        inet:ip-prefix
         |     |     |  +--rw set-tag?       oc-pt:tag-type
         |     |     |  +--rw description?   string
         |     |     +--ro state
         |     |     |  +--ro prefix?        inet:ip-prefix
         |     |     |  +--ro set-tag?       oc-pt:tag-type
         |     |     |  +--ro description?   string
-        |     |     +--rw next-hop-group
+        |     |     +--rw oc-loc-rt-netinst:next-hop-group
        |     |        +--rw oc-loc-rt-netinst:config
-        |     |     |  |  +--rw name?   -> /network-instances/network-instance/static/next-hop-groups/next-hop-group/name
+        |     |        |  +--rw oc-loc-rt-netinst:name?   -> ../../../../../../../oc-ni:static/next-hop-groups/next-hop-group/config/name
        |     |        +--ro oc-loc-rt-netinst:state
-        |     |     |  |  +--rw name?   -> /network-instances/network-instance/static/next-hop-groups/next-hop-group/name
+        |     |        |  +--rw oc-loc-rt-netinst:name?   -> ../../../../../../../oc-ni:static/next-hop-groups/next-hop-group/config/name
```

### DIFF
--- a/release/models/local-routing/openconfig-local-routing-network-instance.yang
+++ b/release/models/local-routing/openconfig-local-routing-network-instance.yang
@@ -21,14 +21,7 @@ module openconfig-local-routing-network-instance {
     "This module adds leaves that relate to network-instances
      to the local routing subtree.";
 
-  oc-ext:openconfig-version "1.1.0";
-
-  revision "2025-07-01" {
-    description
-      "Add next-hop-group, which contains a leafref to network-instances
-      for local routing module.";
-    reference "1.1.0";
-  }
+  oc-ext:openconfig-version "1.0.0";
 
   revision "2025-06-30" {
     description
@@ -89,6 +82,33 @@ module openconfig-local-routing-network-instance {
     }
   }
 
+  grouping static-next-hop-group-networkinstance {
+    container next-hop-group {
+      description
+        "Configuration and state parameters relating to the
+        next-hop-group. In the future, this container will replace the container
+        /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops.
+        If a statically configured next-hop-group is used for a static route
+        prefix then the 'static-routes/static/next-hops' container must not
+        be used.";
+
+      container config {
+        description
+          "Configuration parameters relating to the next-hop-group.";
+
+        uses next-hop-group-config;
+      }
+
+      container state {
+        config false;
+        description
+          "Operational parameters relating to the next-hop-group.";
+
+        uses next-hop-group-config;
+      }
+    }
+  }
+
   grouping next-hop-group-config {
     description
       "Grouping of configuration parameters for a static route's reference
@@ -96,7 +116,7 @@ module openconfig-local-routing-network-instance {
 
     leaf name {
       type leafref {
-        path "/network-instances/network-instance/static/next-hop-groups/next-hop-group/name";
+        path "/oc-ni:network-instances/oc-ni:network-instance/oc-ni:static/oc-ni:next-hop-groups/oc-ni:next-hop-group/oc-ni:config/oc-ni:name";
       }
 
       description
@@ -104,32 +124,12 @@ module openconfig-local-routing-network-instance {
     }
   }
 
-  container next-hop-group {
+  augment "/oc-ni:network-instances/oc-ni:network-instance/oc-ni:protocols/oc-ni:protocol/oc-ni:static-routes/oc-ni:static" {
     description
-      "Configuration and state parameters relating to the
-      next-hop-group. In the future, this container will replace the container
-      /network-instances/network-instance/protocols/protocol/static-routes/static/next-hops.
-      If a statically configured next-hop-group is used for a static route
-      prefix then the 'static-routes/static/next-hops' container must not
-      be used.";
+      "Add network-instance leaves to static routing next-hop container.";
 
-    container config {
-      description
-        "Configuration parameters relating to the next-hop-group.";
-
-      uses next-hop-group-config;
-    }
-
-    container state {
-      config false;
-      description
-        "Operational parameters relating to the next-hop-group.";
-
-      uses next-hop-group-config;
-    }
+    uses static-next-hop-group-networkinstance;
   }
-
-
 
   augment "/oc-ni:network-instances/oc-ni:network-instance/oc-ni:protocols/oc-ni:protocol/oc-ni:static-routes/oc-ni:static/oc-ni:next-hops/oc-ni:next-hop/oc-ni:config" {
     description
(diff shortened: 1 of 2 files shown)


release/models/local-routing/openconfig-local-routing-network-instance.yang
```diff
@@ -21,7 +21,14 @@ module openconfig-local-routing-network-instance {
     "This module adds leaves that relate to network-instances
      to the local routing subtree.";
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "1.1.0";
+
+  revision "2025-07-01" {
+    description
+      "Add refernce to static next-hop-group and augment the local routing
```
robshakir:
```suggestion
      "Add reference to static next-hop-group and augment the local routing
```

release/models/local-routing/openconfig-local-routing-network-instance.yang
```diff
@@ -82,6 +89,58 @@ module openconfig-local-routing-network-instance {
...
+  }
+
+  grouping next-hop-group-config {
+    description
+      "Grouping of configuration parameters for a reference to a
+      static next-hop-group.";
+
+    leaf name {
+      type leafref {
+        path "/oc-ni:network-instances/oc-ni:network-instance/oc-ni:static/oc-ni:next-hop-groups/oc-ni:next-hop-group/oc-ni:config/oc-ni:name";
```
robshakir: This reference allows for referencing any network instance, not just the local one. Is this your intent?

Even if it is, I think this is a bug since if you have network-instance `BLUE` with next-hop-group `FISH` and network-instance `RED` with next-hop-group `FISH` then `FISH` will be an ambiguous value here.

Either this should be a relative path, or we should have a leaf that tells us the name of the `network-instance`.

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -43,7 +43,14 @@ module openconfig-local-routing {
     protocol-specific policy after importing the route into the
     protocol for distribution (again via routing policy).";
 
-  oc-ext:openconfig-version "3.0.0";
+  oc-ext:openconfig-version "3.1.0";
+
+  revision "2025-07-01" {
+    description
+      "Add refernce to static next-hop-group and augment the local routing
```
robshakir:
```suggestion
      "Add reference to static next-hop-group and augment the local routing
```

release/models/local-routing/openconfig-local-routing-network-instance.yang
dplore: The current use cases assume the same local network-instance.  Now that you point it out, I see how with only the  next-hop-group name as the key, it is ambiguous which network instance should be used.   (References: [a few featureprofiles tests](https://github.com/search?q=repo%3Aopenconfig%2Ffeatureprofiles+%22%2Fnetwork-instances%2Fnetwork-instance%2Fstatic%2Fnext-hop-groups%2Fnext-hop-group%2Fconfig%2Fname%22&type=code))

I will make it a relative reference.

If in the future we wanted to allow this to be any network instance (because I can see this possibly occurring in the future), could we add a network-instance-name leaf and still use the relative path for the next-hop-group name?  Conceptually it seems like we could not?  In this scenario, a non-breaking change could be to introduce a new container with an absolute path reference and network-instance-name.  While that's a little messy, I am not aware of implementations that support a different network instance for the next-hop-group at this time and we don't have the operational use case, so I don't think it's a good idea to introduce a new leaf, just to future proof this. 

@danameme FYI

robshakir: No, you can't do that.

If you have a relative reference (i.e., `../../../config/name` to get you to `/network-instances/network-instance/config/name` -- didn't check the # of `..`s) then you're saying that it can only be within that relative path from the current node, so that `..` path would get you to  `/network-instances/network-instance[name=DEFAULT]` so the only valid `config/name` from here is `DEFAULT`.

If you want to allow a specified network-instance name then you need to have two leaves.

```yang
leaf network-instance {
   // allows you to specify any valid network instance on the device
   type leafref { path "/network-instances/network-instance/config/name"; }
} 

leaf next-hop-group {
   type leafref { 
     // use the network-instance leaf's name as the network-instance `name` list key.
     //
     // current() means take the current position in the tree of this leaf
     // .. takes one step "out" of the next-hop-group leaf
     // network-instance refers to our sibling node
     path "/network-instance/network-instance[name=current()/../network-instance]/static/next-hops/next-hop";
   }
}
```

But this will mandate that you need to create a `network-instance` leaf (in my example) for every next-hop-group that is specified.

If there's no use case for this, then I suggest omitting it. Generally, I'd push users towards using gRIBI for these more complicated programming use cases.

dplore: Got it and I agree on omitting, but wanted to explore the idea and document why we didn't include a network-instance name leaf at this time.

release/models/local-routing/openconfig-local-routing.yang
robshakir: Typo still looks to be there.

dplore: fixed!

release/models/local-routing/openconfig-local-routing.yang
```diff
@@ -43,7 +43,14 @@ module openconfig-local-routing {
     protocol-specific policy after importing the route into the
     protocol for distribution (again via routing policy).";
 
-  oc-ext:openconfig-version "3.0.0";
+  oc-ext:openconfig-version "3.1.0";
```
ghananigans: Should we be bumping the major version since this is technically a "breaking" change as we are removing a node?

dplore: Done

ghananigans (approved):

ghananigans:
To confirm, is this a breaking change for the overall OC public model?

dplore:
> To confirm, is this a breaking change for the overall OC model?
> 
> I see the [previous](https://github.com/openconfig/public/pull/1324#issuecomment-3024788353) change wasn't considered a breaking change, but this change had the breaking label.

This change is in the same category.  We make a breaking change in the OC models and reflect that in the yang module versions.  But the OC Release will not have breaking changes because the leaves we broke were not yet present in any [OpenConfig public release](https://github.com/openconfig/public/tags).  The forthcoming OC public release will only increment[ v5.2.0](https://github.com/openconfig/public/releases/tag/v5.2.0) to v5.3.0.

PR #1329, opened 2025-07-04
Add fragment-offsets as a match criteria for Policy Forwarding

amitarista:
### Change Scope

- Add a new match condition `fragment-offsets` to the policy-forwarding ipv4 container.
   Match packet if the Fragment Offset field value is included in the `fragment-offsets` leaf-list. Each element of this leaf list may represent a single value, an inclusive range `<lower>..<higher>` or `ANY` ( wildcard that matches any fragments ).
- `state` i.e. the read path will return the configured values.

### Tree view

```diff
   +--rw policy-forwarding
   |  +--rw policies
   |  |  +--rw policy* [policy-id]
   |  |     +--rw policy-id    -> ../config/policy-id
   |  |     +--rw config
   |  |     +--ro state
   |  |     +--rw rules
   |  |        +--rw rule* [sequence-id]
   |  |           +--rw sequence-id    -> ../config/sequence-id
   |  |           +--rw config
   |  |           +--ro state
   |  |           +--rw ipv4
   |  |           |  +--rw config
   |  |           |  |  +--rw source-address?                   oc-inet:ipv4-prefix
   |  |           |  |  +--rw source-address-prefix-set?        -> /oc-sets:defined-sets/ipv4-prefix-sets/ipv4-prefix-set/name
   |  |           |  |  +--rw destination-address?              oc-inet:ipv4-prefix
   |  |           |  |  +--rw destination-address-prefix-set?   -> /oc-sets:defined-sets/ipv4-prefix-sets/ipv4-prefix-set/name
+  |  |           |  |  +--rw fragment-offsets*                 oc-pkt-match-types:fragment-offset-range
   |  |           |  |  +--rw dscp?                             oc-inet:dscp
   |  |           |  |  +--rw dscp-set*                         oc-inet:dscp
   |  |           |  |  +--rw length?                           uint16
   |  |           |  |  +--rw protocol?                         oc-pkt-match-types:ip-protocol-type
   |  |           |  |  +--rw hop-limit?                        uint8
   |  |           |  +--ro state
   |  |           |  |  +--ro source-address?                   oc-inet:ipv4-prefix
   |  |           |  |  +--ro source-address-prefix-set?        -> /oc-sets:defined-sets/ipv4-prefix-sets/ipv4-prefix-set/name
   |  |           |  |  +--ro destination-address?              oc-inet:ipv4-prefix
   |  |           |  |  +--ro destination-address-prefix-set?   -> /oc-sets:defined-sets/ipv4-prefix-sets/ipv4-prefix-set/name
+  |  |           |  |  +--ro fragment-offsets*                 oc-pkt-match-types:fragment-offset-range
   |  |           |  |  +--ro dscp?                             oc-inet:dscp
   |  |           |  |  +--ro dscp-set*                         oc-inet:dscp
   |  |           |  |  +--ro length?                           uint16
   |  |           |  |  +--ro protocol?                         oc-pkt-match-types:ip-protocol-type
   |  |           |  |  +--ro hop-limit?                        uint8
   |  |           |  +--rw icmpv4
```

### Platform Implementations

In Arista EOS, keyword `fragment` is used to add match condition for fragments.
Following is an example on how this is configured via EOS CLI.

```
traffic-policies
   traffic-policy foo
      match ipv4 ipv4
         fragment offset 100, 200-300
```

### More info on the Inputs

- Fragment Offset is a 13-bit field in the IPv4 header. Hence, the valid value range is 0 to 8191. This range is already restricted for the single value as well as the range.
- Single values and ranges can be overlapping. Union of all the values must be considered for matching.

### DIFF
--- a/release/models/acl/openconfig-packet-match-types.yang
+++ b/release/models/acl/openconfig-packet-match-types.yang
@@ -23,7 +23,13 @@ module openconfig-packet-match-types {
     "This module defines common types for use in models requiring
     data definitions related to packet matches.";
 
-  oc-ext:openconfig-version "1.3.3";
+  oc-ext:openconfig-version "1.3.4";
+
+  revision "2025-06-10" {
+    description
+      "Add pattern regex for Fragment Offset range";
+    reference "1.3.4";
+  }
 
   revision "2023-01-29" {
     description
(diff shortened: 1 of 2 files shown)


dplore:
Additonal vendor references:

[JunOS firewall filters have syntax for fragments](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/example/routing-stateless-firewall-filter-security-handle-fragment-configuring.html)

[Cisco IOS XR implements access lists with fragment support (for ipv4)](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/ip-addresses/710x/b-ip-addresses-cg-8k-710x/implementing-access-lists.html)

dplore:
Reviewed in Sep 2, 2025 OC Operators meeting without objections.

PR #1334, opened 2025-07-09
Add FPGA, GRUB, ONIE platform types

mrevang:
### Change Scope

* Adds a new hardware component type: `FPGA`. This type will compliment `INTEGRATED_CIRCUIT`, which is widely used to represent a switching ASIC.
* Adds a `boot-loader` container with a new state leaf `type`. `type` can be either `BOOT_LOADER_ONIE` or `BOOT_LOADER_GRUB`.

```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--rw boot-loader <<<<<<<<<<<<<<<<
           +--rw config
           +--ro state
              +--ro oc-boot-loader:type?   identityref
```

### Platform Implementations

 * Implementation A: https://opencomputeproject.github.io/onie/overview/
 * Implementation B: https://www.gnu.org/software/grub/

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,8 +6,8 @@
 # the release/models directory (all YANG content)
 # and doc directory is maintained by the public-writers 
 # OpenConfig team. 
-/doc @openconfig/public-writers
-/release/models/ @openconfig/public-writers
+/doc @openconfig/public-approvers
+/release/models/ @openconfig/public-approvers
 
 # subfolders may have additional codeowners who
 # may merge content
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,13 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "3.0.0";
+  oc-ext:openconfig-version "3.1.0";
+
+  revision "2025-07-17" {
+    description
+      "Add GRE and MPLS to encap-headers.";
+    reference "3.1.0";
+  }
 
   revision "2025-03-12" {
     description
@@ -270,7 +276,7 @@ submodule openconfig-aft-common {
             description
               "State parameters relating to GRE encapsulation.";
 
-            uses aft-common-entry-nexthop-gre-state;
+            uses aft-common-entry-nexthop-encap-gre-config;
           }
         }
 
@@ -323,7 +329,7 @@ submodule openconfig-aft-common {
                 description
                   "State parameters relating to GRE encapsulation headers.";
 
-                uses aft-common-entry-nexthop-gre-state;
+                uses aft-common-entry-nexthop-encap-gre-config;
               }
             }
 
@@ -627,16 +633,14 @@ submodule openconfig-aft-common {
     }
   }
 
-  grouping aft-common-entry-nexthop-gre-state {
+  grouping aft-common-entry-nexthop-encap-gre-config {
     description
       "GRE encapsulation applied on a IPv4 and IPv6 next-hop.";
 
     leaf src-ip {
       type oc-inet:ip-address;
       description
-        "The source IP address for the GRE encapsulation may be expressed
-        using this leaf (src-ip) or if may be derived from
-        '../../interface-ref/state/subinterface'";
+        "The source IP address for the GRE encapsulation.";
     }
 
     leaf dst-ip {
@@ -648,10 +652,9 @@ submodule openconfig-aft-common {
     leaf ttl {
       type uint8;
       description
-        "This leaf reflects the configured/default TTL value that is used in the
-         outer header during packet encapsulation. When this leaf is not set,
-         the TTL value of the inner packet is copied over as the outer packet's
-         TTL value during encapsulation.";
+        "If specified, this TTL value is used on the GRE header during
+        encapsulation. If unspecified, the TTL value of the inner packet is
+        copied to the GRE header TTL value during encapsulation.";
     }
   }
 
@@ -684,6 +687,22 @@ submodule openconfig-aft-common {
     }
   }
 
+  grouping aft-common-entry-nexthop-encap-mpls-config {
+    description
+      "MPLS encapsulation of a packet.";
+
+    leaf traffic-class {
+      type oc-mplst:mpls-tc;
+      description
+        "The value of the MPLS traffic class (TC) bits, formerly known as the
+         EXP bits.";
+    }
+    leaf label {
+      type oc-mplst:mpls-label;
+      description
+        "A MPLS label value.";
+    }
+  }
 
   grouping aft-common-entry-nexthop-encap-udp-v4-state {
     description
--- a/release/models/aft/openconfig-aft-ethernet.yang
+++ b/release/models/aft/openconfig-aft-ethernet.yang
@@ -20,7 +20,13 @@ submodule openconfig-aft-ethernet {
     "Submodule containing definitions of groupings for the abstract
     forwarding tables for Ethernet.";
 
-  oc-ext:openconfig-version "3.0.0";
+  oc-ext:openconfig-version "3.1.0";
+
+  revision "2025-05-15" {
+    description
+      "Add GRE and MPLS to encap-headers.";
+    reference "3.1.0";
+  }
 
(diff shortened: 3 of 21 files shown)


release/models/platform/openconfig-platform-types.yang
```diff
@@ -417,6 +422,14 @@ module openconfig-platform-types {
        local area network";
   }
 
+  identity FPGA {
```
LimeHat, line 425: Can you provide a couple of examples where this is used/expected to be used?

release/models/platform/openconfig-platform-types.yang
```diff
@@ -450,6 +463,20 @@ module openconfig-platform-types {
       device OS or network OS.";
   }
 
+  identity BOOT_LOADER_GRUB {
```
LimeHat: What is the motivation for creating granular bootloader identities?

While I understand that the user may want to get more information about the bootloader used in a component, doing this via identities seems questionable.
Components don't have fixed names; identities are often used in component discovery procedures on the client side. Expanding this list implies additional complexity, and the benefits are not obvious. 

This change has implication for other component types as well. There are different BIOS flavors (ami, phoenix, and then there's UEFI..), different OS kinds, etc.
Where do we draw a line?

release/models/platform/openconfig-platform-types.yang
```diff
@@ -450,6 +463,20 @@ module openconfig-platform-types {
...
+    base OPENCONFIG_SOFTWARE_COMPONENT;
```
LimeHat: perhaps `base BOOT_LOADER`?

LimeHat (changes requested):

release/models/platform/openconfig-platform-types.yang
dplore: > This change has implication for other component types as well. There are different BIOS flavors (ami, phoenix, and then there's UEFI..), different OS kinds, etc. Where do we draw a line?

Operational use cases may provide a defensible use case; please do provide some information here @mrevang

release/models/platform/openconfig-platform-types.yang
mrevang: For our use case, this enables us to report firmware versions of these programmable devices.

release/models/platform/openconfig-platform-types.yang
mrevang: Knowing the boot loader type can inform recovery strategies. For some devices in our ecosystem, the boot loader type can vary.

release/models/platform/openconfig-platform-types.yang
mrevang: Done.

release/models/platform/openconfig-platform-types.yang
earies: +1 to @LimeHat comments here - I'd suggest keeping the identity as-is defined today and if necessary conveying any additional data as distinct leafs.  If we think is warranted then you can have bootloader specific data under a new anchor point at `/components/component/bootloader/...` to be conveyed only for components of `type=BOOT_LOADER`

mrevang: Thanks for the feedback. I'm fine with that direction too and am happy to make that pivot.

If that's an agreeable direction, I can prep the following: `/components/component/bootloader/state/type`.

Thoughts @LimeHat @earies ?

mrevang: I can also split this into its own PR and keep this one for adding the `FPGA` HARDWARE_COMPONENT.

dplore:
With only one leaf for component/bootloader (type), it's redundant to [/components/component/state/type](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-state-type).    If we had bootloader specific leaves, then I would strongly favor adding /components/component/bootloader.   The only other leaf I can think to add is also covered as generic component leaves ([/components/component/state/firmware-version](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-state-firmware-version))

What if we add the additional/more specific bootloader identities as children of `OPENCONFIG_SOFTWARE_COMPONENT` ?
ie:
```
identity BOOT_LOADER_GRUB {
    base OPENCONFIG_SOFTWARE_COMPONENT;
```

I think this is slightly simpler as it maintains the existing, single, flat identity tree for `OPENCONFIG_SOFTWARE_COMPONENT`.  (I acknowledge @mrevang originally suggested this)

If we expected a large number (10+ ?) of bootloaders then a new, nested bootloader identity seems better.  But there just aren't that many bootloaders for network devices and they are not rapidly changing.  

On the whole, I don't see any of these suggestions as wrong.  But I would like to not spend too much energy on this issue as I think this is a low impact change.    😉

earies:
> With only one leaf for component/bootloader (type), it's redundant to [/components/component/state/type](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-state-type). If we had bootloader specific leaves, then I would strongly favor adding /components/component/bootloader. The only other leaf I can think to add is also covered as generic component leaves ([/components/component/state/firmware-version](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-state-firmware-version))
> 
> What if we add the additional/more specific bootloader identities as children of `OPENCONFIG_SOFTWARE_COMPONENT` ? ie:
> 
> ```
> identity BOOT_LOADER_GRUB {
>     base OPENCONFIG_SOFTWARE_COMPONENT;
> ```
> 
> I think this is slightly simpler as it maintains the existing, single, flat identity tree for `OPENCONFIG_SOFTWARE_COMPONENT`. (I acknowledge @mrevang originally suggested this)
> 
> If we expected a large number (10+ ?) of bootloaders then a new, nested bootloader identity seems better. But there just aren't that many bootloaders for network devices and they are not rapidly changing.
> 
> On the whole, I don't see any of these suggestions as wrong. But I would like to not spend too much energy on this issue as I think this is a low impact change. 😉

uboot, syslinux, lilo.... all additionals so now we are getting into flavors - so far identities have been mostly generic so if we choose to take this approach vs. put the detail within distinct leafs under a new container such as `bootloader` we will get into the semantics of potentially overloading the generic `BOOT_LOADER` identity until trying to propose another identity into the model set.

To draw another correlation, we could have all sorts of "flavors" of controller cards or linecards but we would find another means (another leaf) to describe the flavor of the generic types of `CONTROLLER_CARD` or `LINECARD`

mrevang:
> > With only one leaf for component/bootloader (type), it's redundant to [/components/component/state/type](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-state-type). If we had bootloader specific leaves, then I would strongly favor adding /components/component/bootloader. The only other leaf I can think to add is also covered as generic component leaves ([/components/component/state/firmware-version](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-platform.html#components-component-state-firmware-version))
> > What if we add the additional/more specific bootloader identities as children of `OPENCONFIG_SOFTWARE_COMPONENT` ? ie:
> > ```
> > identity BOOT_LOADER_GRUB {
> >     base OPENCONFIG_SOFTWARE_COMPONENT;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this is slightly simpler as it maintains the existing, single, flat identity tree for `OPENCONFIG_SOFTWARE_COMPONENT`. (I acknowledge @mrevang originally suggested this)
> > If we expected a large number (10+ ?) of bootloaders then a new, nested bootloader identity seems better. But there just aren't that many bootloaders for network devices and they are not rapidly changing.
> > On the whole, I don't see any of these suggestions as wrong. But I would like to not spend too much energy on this issue as I think this is a low impact change. 😉
> 
> uboot, syslinux, lilo.... all additionals so now we are getting into flavors - so far identities have been mostly generic so if we choose to take this approach vs. put the detail within distinct leafs under a new container such as `bootloader` we will get into the semantics of potentially overloading the generic `BOOT_LOADER` identity until trying to propose another identity into the model set.
> 
> To draw another correlation, we could have all sorts of "flavors" of controller cards or linecards but we would find another means (another leaf) to describe the flavor of the generic types of `CONTROLLER_CARD` or `LINECARD`

I've introduced `/components/component/boot-loader/state/type`, to keep the component/type leaf generic.

dplore (approved):

PR #1335, opened 2025-07-14
Add NAT model

nleiva:
### Change Scope

* This PR introduces a new YANG data model for Network Address Translation (NAT).
* This is a new model, so the change is backward compatible.

We use NAT instances to stay in line with [RFC 8512](https://datatracker.ietf.org/doc/rfc8512/).
NAT rules can match a `routing-policy` or `acl-set` to accommodate different vendor implementations.

```bash
module: openconfig-nat
  +--rw nat
     +--rw config
     |  +--rw translation-timeout?   uint32
     |  +--rw max-translations?      uint32
     |  +--rw log-level?             enumeration
     +--ro state
     |  +--ro enabled?                      boolean
     |  +--ro total-instances?              uint32
     |  +--ro active-instances?             uint32
     |  +--ro total-active-translations?    yang:counter64
     |  +--ro total-translation-failures?   yang:counter64
     |  +--ro memory-usage?                 uint32
     |  +--ro uptime?                       yang:timeticks
     +--rw instances
        +--rw instance* [name]
           +--rw name          -> ../config/name
           +--rw config
           |  +--rw name?          string
           |  +--rw type?          identityref
           |  +--rw enabled?       boolean
           |  +--rw description?   string
           +--ro state
           |  +--ro name?               string
           |  +--ro type?               identityref
           |  +--ro enabled?            boolean
           |  +--ro description?        string
           |  +--ro active-mappings?    yang:counter64
           |  +--ro total-mappings?     yang:counter64
           |  +--ro mapping-failures?   yang:counter64
           +--rw interfaces
           |  +--rw interface* [interface]
           |     +--rw interface    -> ../config/interface
           |     +--rw config
           |     |  +--rw interface?   oc-if:base-interface-ref
           |     |  +--rw direction?   nat-direction
           |     +--ro state
           |        +--ro interface?            oc-if:base-interface-ref
           |        +--ro direction?            nat-direction
           |        +--ro packets-translated?   yang:counter64
           |        +--ro bytes-translated?     yang:counter64
           |        +--ro translation-errors?   yang:counter64
           +--rw pools
           |  +--rw pool* [name]
           |     +--rw config
           |     |  +--rw name?                           string
           |     |  +--rw start-address?                  inet:ipv4-address
           |     |  +--rw end-address?                    inet:ipv4-address
           |     |  +--rw port-range-start?               inet:port-number
           |     |  +--rw port-range-end?                 inet:port-number
           |     |  +--rw port-block-size?                uint16
           |     |  +--rw (traffic-matcher)?
           |     |  |  +--:(routing-policy)
           |     |  |  |  +--rw policy-definition?        -> /oc-rpol:routing-policy/policy-definitions/policy-definition/name
           |     |  |  +--:(access-list)
           |     |  |     +--rw acl-set?                  -> /oc-acl:acl/acl-sets/acl-set/config/name
           |     |  +--rw persistent?                     boolean
           |     |  +--rw max-sessions?                   uint32
           |     |  +--rw address-only?                   boolean
           |     |  +--rw timeout?                        uint32
           |     |  +--rw tcp-timeout?                    uint32
           |     |  +--rw udp-timeout?                    uint32
           |     |  +--rw icmp-timeout?                   uint32
           |     |  +--rw overload?                       boolean
           |     |  +--rw log-translations?               boolean
           |     |  +--rw max-translations-per-address?   uint32
           |     +--ro state
           |        +--ro name?                           string
           |        +--ro start-address?                  inet:ipv4-address
           |        +--ro end-address?                    inet:ipv4-address
           |        +--ro port-range-start?               inet:port-number
           |        +--ro port-range-end?                 inet:port-number
           |        +--ro port-block-size?                uint16
           |        +--ro (traffic-matcher)?
           |        |  +--:(routing-policy)
           |        |  |  +--ro policy-definition?        -> /oc-rpol:routing-policy/policy-definitions/policy-definition/name
           |        |  +--:(access-list)
           |        |     +--ro acl-set?                  -> /oc-acl:acl/acl-sets/acl-set/config/name
           |        +--ro persistent?                     boolean
           |        +--ro max-sessions?                   uint32
           |        +--ro address-only?                   boolean
           |        +--ro timeout?                        uint32
           |        +--ro tcp-timeout?                    uint32
           |        +--ro udp-timeout?                    uint32
           |        +--ro icmp-timeout?                   uint32
           |        +--ro overload?                       boolean
           |        +--ro log-translations?               boolean
           |        +--ro max-translations-per-address?   uint32
           |        +--ro allocated-addresses?            uint32
           |        +--ro available-addresses?            uint32
           |        +--ro hit-count?                      yang:counter64
           |        +--ro active-sessions?                uint32
           +--rw mappings
              +--rw mapping* [name]
                 +--rw config
                 |  +--rw internal-address?          inet:ipv4-address
                 |  +--rw external-address?          inet:ipv4-address
                 |  +--rw internal-port?             inet:port-number
                 |  +--rw external-port?             inet:port-number
                 |  +--rw protocol?                  protocol-type
                 |  +--rw (traffic-matcher)?
                 |  |  +--:(routing-policy)
                 |  |  |  +--rw policy-definition?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/name
                 |  |  +--:(access-list)
                 |  |     +--rw acl-set?             -> /oc-acl:acl/acl-sets/acl-set/config/name
                 |  +--rw persistent?                boolean
                 |  +--rw max-sessions?              uint32
                 |  +--rw bidirectional?             boolean
                 |  +--rw enabled?                   boolean
                 +--ro state
                    +--ro name?                      string
                    +--ro internal-address?          inet:ipv4-address
                    +--ro external-address?          inet:ipv4-address
                    +--ro internal-port?             inet:port-number
                    +--ro external-port?             inet:port-number
                    +--ro protocol?                  protocol-type
                    +--ro (traffic-matcher)?
                    |  +--:(routing-policy)
                    |  |  +--ro policy-definition?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/name
                    |  +--:(access-list)
                    |     +--ro acl-set?             -> /oc-acl:acl/acl-sets/acl-set/config/name
                    +--ro persistent?                boolean
                    +--ro max-sessions?              uint32
                    +--ro bidirectional?             boolean
                    +--ro enabled?                   boolean
                    +--ro creation-time?             yang:date-and-time
                    +--ro last-used?                 yang:date-and-time
                    +--ro packet-count?              yang:counter64
                    +--ro byte-count?                yang:counter64
                    +--ro hit-count?                 yang:counter64
                    +--ro active-sessions?           uint32
```

### Platform Implementations

 * Implementation A: [Arista static](https://arista.my.site.com/AristaCommunity/s/article/7150s-nat-practical-guide-source-nat-static), [Arista dynamic](https://arista.my.site.com/AristaCommunity/s/article/7150s-nat-practical-guide-source-nat-dynamic)
 * Implementation J: [Juniper](https://www.juniper.net/documentation/us/en/software/junos/nat/topics/topic-map/security-nat-configuration-overview.html)
 * Implementation C: [Cisco](https://www.cisco.com/c/en/us/support/docs/ip/network-address-translation-nat/13772-12.html)

### DIFF
--- a/release/models/nat/openconfig-nat.yang
+++ b/release/models/nat/openconfig-nat.yang
@@ -1,0 +1,802 @@
+module openconfig-nat {
+
+  yang-version "1";
+
+  // namespace
+  namespace "http://openconfig.net/yang/nat";
+
+  prefix "oc-nat";
+
+  // import some basic types
+  import ietf-inet-types { prefix inet; }
+  import ietf-yang-types { prefix yang; }
+  import openconfig-routing-policy { prefix oc-rpol; }
+  import openconfig-interfaces { prefix oc-if; }
+  import openconfig-extensions { prefix oc-ext; }
+  import openconfig-acl { prefix oc-acl; }
+
+  // meta
+  organization "OpenConfig working group";
+
+  contact
+    "OpenConfig working group
+    www.openconfig.net";
+
+  description
+    "This module defines configuration and operational state data
+    for Network Address Translation (NAT) functionality.
+
+    The model supports various NAT types including:
+    - Basic NAT44 (Network Address Translation)
+    - NAPT44 (Network Address Port Translation)  
+    - Static mappings
+    - Dynamic address pools
+    - Policy-based NAT using routing policies and ACLs";
+
+  oc-ext:openconfig-version "0.1.0";
+
+  revision "2025-07-14" {
+    description
+      "Initial revision";
+    reference "0.1.0";
+  }
+
+  // Identity statements
+
+  identity NAT_TYPE {
+    description
+      "Base identity for NAT types";
+  }
+
+  identity BASIC_NAT44 {
+    base NAT_TYPE;
+    description
+      "Basic NAT44 - translates IPv4 addresses only";
+  }
+
+  identity NAPT44 {
+    base NAT_TYPE;
+    description
+      "Network Address Port Translation - translates both
+      IPv4 addresses and port numbers";
+  }
+
+  identity STATIC_NAT {
+    base NAT_TYPE;
+    description
+      "Static NAT - fixed one-to-one address mappings";
+  }
+
+  // Typedef statements
+
+  typedef nat-direction {
+    type enumeration {
+      enum INSIDE_TO_OUTSIDE {
+        description
+          "Translation from inside (private) to outside (public)";
+      }
+      enum OUTSIDE_TO_INSIDE {
+        description
+          "Translation from outside (public) to inside (private)";
+      }
+      enum BIDIRECTIONAL {
+        description
+          "Translation in both directions";
+      }
+    }
+    description
+      "Direction of NAT translation";
+  }
+
+  typedef protocol-type {
+    type enumeration {
+      enum TCP {
+        description "Transmission Control Protocol";
+      }
+      enum UDP {
+        description "User Datagram Protocol";
+      }
+      enum ICMP {
+        description "Internet Control Message Protocol";
+      }
+      enum ALL {
+        description "All protocols";
+      }
+    }
+    description
+      "Protocol types supported for NAT translation";
+  }
+
+  typedef nat-action {
+    type enumeration {
+      enum TRANSLATE {
+        description
+          "Apply NAT translation";
+      }
+      enum BYPASS {
+        description
+          "Bypass NAT translation";
+      }
+    }
+    description
+      "Action to take for matching traffic";
+  }
+
+  // Grouping statements
+
+  grouping nat-instance-config {
+    description
+      "Configuration parameters for NAT instances";
+
+    leaf name {
+      type string;
+      description
+        "Unique name for the NAT instance";
+    }
+
+    leaf type {
+      type identityref {
+        base NAT_TYPE;
+      }
+      description
+        "Type of NAT translation to be performed";
+    }
+
+    leaf enabled {
+      type boolean;
+      default false;
+      description
+        "Administrative state of the NAT instance";
+    }
+
+    leaf description {
+      type string;
+      description
+        "Textual description of the NAT instance";
+    }
+  }
+
+  grouping nat-instance-state {
+    description
+      "Operational state data for NAT instances";
+
+    leaf active-mappings {
+      type yang:counter64;
+      description
+        "Number of currently active NAT mappings";
+    }
+
+    leaf total-mappings {
+      type yang:counter64;
+      description
+        "Total number of NAT mappings created since instance startup";
+    }
+
+    leaf mapping-failures {
+      type yang:counter64;
+      description
+        "Number of mapping creation failures";
+    }
+  }
+
+  grouping nat-interface-config {
+    description
+      "Configuration parameters for NAT interfaces";
+
+    leaf interface {
+      type oc-if:base-interface-ref;
+      description
+        "Reference to the interface";
+    }
+
+    leaf direction {
+      type nat-direction;
+      description
+        "NAT translation direction for this interface";
+    }
+  }
+
+  grouping nat-interface-state {
+    description
+      "Operational state data for NAT interfaces";
+
+    leaf packets-translated {
+      type yang:counter64;
+      description
+        "Number of packets that have been translated";
+    }
+
+    leaf bytes-translated {
+      type yang:counter64;
+      description
+        "Number of bytes that have been translated";
+    }
+
+    leaf translation-errors {
+      type yang:counter64;
+      description
+        "Number of translation errors encountered";
+    }
+  }
+
+  grouping nat-interface-top {
+    description
+      "Top-level grouping for NAT interface data";
+
+    container interfaces {
+      description
+        "Interface configuration and state for NAT";
+
+      list interface {
+        key "interface";
+        description
+          "List of interfaces participating in NAT";
+
+        leaf interface {
+          type leafref {
+            path "../config/interface";
+          }
+          description
+            "Reference to interface list key";
+        }
+
+        container config {
+          description
+            "Configuration parameters for the NAT interface";
+
+          uses nat-interface-config;
+        }
+
+        container state {
+          config false;
+          description
+            "Operational state data for the NAT interface";
+
+          uses nat-interface-config;
+          uses nat-interface-state;
+        }
+      }
+    }
+  }
+
+  grouping nat-pool-config {
+    description
+      "Configuration parameters for NAT address pools";
+
+    leaf name {
+      type string;
+      description
+        "Unique name for the NAT pool";
+    }
+
+    leaf start-address {
+      type inet:ipv4-address;
+      description
+        "Starting IPv4 address of the pool range";
+    }
+
+    leaf end-address {
+      type inet:ipv4-address;
+      description
+        "Ending IPv4 address of the pool range";
+    }
+
+    leaf port-range-start {
+      type inet:port-number;
+      description
+        "Starting port number for NAPT translations";
+    }
+
+    leaf port-range-end {
+      type inet:port-number;
+      description
+        "Ending port number for NAPT translations";
+    }
+
+    leaf port-block-size {
+      type uint16;
+      description
+        "Number of ports allocated per mapping";
+    }
+
+    choice traffic-matcher {
+      description
+        "Traffic matching criteria for pool usage";
+      
+      case routing-policy {
+        leaf policy-definition {
+          type leafref {
+            path "/oc-rpol:routing-policy/oc-rpol:policy-definitions/" +
+                 "oc-rpol:policy-definition/oc-rpol:name";
+          }
+          description
+            "Reference to routing policy that defines traffic 
+            eligible for this pool";
+        }
+      }
+
+      case access-list {
+        leaf acl-set {
+          type leafref {
+            path "/oc-acl:acl/oc-acl:acl-sets/oc-acl:acl-set/" +
+                 "oc-acl:config/oc-acl:name";
+          }
+          description
+            "Reference to ACL set that defines traffic 
+            eligible for this pool";
+        }
+      }
+    }
+
+    leaf persistent {
+      type boolean;
+      default false;
+      description
+        "Enable persistent NAT mapping";
+    }
+
+    leaf max-sessions {
+      type uint32;
+      description
+        "Maximum number of concurrent sessions";
+    }
+
+    leaf address-only {
+      type boolean;
+      default false;
+      description
+        "Enable address-only translation without port translation";
+    }
+
+    leaf timeout {
+      type uint32;
+      units "seconds";
+      description
+        "Default NAT mapping timeout";
+    }
+
+    leaf tcp-timeout {
+      type uint32;
+      units "seconds";
+      description
+        "TCP-specific mapping timeout";
+    }
+
+    leaf udp-timeout {
+      type uint32;
+      units "seconds";
+      description
+        "UDP-specific mapping timeout";
+    }
+
+    leaf icmp-timeout {
+      type uint32;
+      units "seconds";
+      description
+        "ICMP-specific mapping timeout";
+    }
+
+    leaf overload {
+      type boolean;
+      default true;
+      description
+        "Enable port overloading (PAT) allowing multiple internal
+        addresses to share the same external address with different ports";
+    }
+
+    leaf log-translations {
+      type boolean;
+      default false;
+      description
+        "Enable logging of NAT translations";
+    }
+
+    leaf max-translations-per-address {
+      type uint32;
+      description
+        "Maximum number of concurrent translations per external address";
+    }
+  }
+
+  grouping nat-pool-state {
+    description
+      "Operational state data for NAT address pools";
+
+    leaf allocated-addresses {
+      type uint32;
+      description
+        "Number of addresses currently allocated from this pool";
+    }
+
+    leaf available-addresses {
+      type uint32;
+      description
+        "Number of addresses available in this pool";
+    }
+
+    leaf hit-count {
+      type yang:counter64;
+      description
+        "Number of packets matching this pool's traffic criteria";
+    }
+
+    leaf active-sessions {
+      type uint32;
+      description
+        "Number of currently active sessions using this pool";
+    }
+  }
+
+  grouping nat-mapping-config {
+    description
+      "Configuration parameters for static NAT mappings";
+
+    leaf name {
+      type string;
+      description
+        "Unique name for the static mapping";
+    }
+
+    leaf internal-address {
+      type inet:ipv4-address;
+      description
+        "Internal (private) IPv4 address";
+    }
+
+    leaf external-address {
+      type inet:ipv4-address;
+      description
+        "External (public) IPv4 address";
+    }
+
+    leaf internal-port {
+      type inet:port-number;
+      description
+        "Internal port number for NAPT mappings";
+    }
+
+    leaf external-port {
+      type inet:port-number;
+      description
+        "External port number for NAPT mappings";
+    }
+
+    leaf protocol {
+      type protocol-type;
+      description
+        "Protocol for the mapping";
+    }
+
+    choice traffic-matcher {
+      description
+        "Traffic matching criteria for static mapping";
+      
+      case routing-policy {
+        leaf policy-definition {
+          type leafref {
+            path "/oc-rpol:routing-policy/oc-rpol:policy-definitions/" +
+                 "oc-rpol:policy-definition/oc-rpol:name";
+          }
+          description
+            "Reference to routing policy that defines traffic 
+            matching criteria for this static mapping";
+        }
+      }
+
+      case access-list {
+        leaf acl-set {
+          type leafref {
+            path "/oc-acl:acl/oc-acl:acl-sets/oc-acl:acl-set/" +
+                 "oc-acl:config/oc-acl:name";
+          }
+          description
+            "Reference to ACL set that defines traffic 
+            matching criteria for this static mapping";
+        }
+      }
+    }
+
+    leaf persistent {
+      type boolean;
+      default false;
+      description
+        "Enable persistent NAT mapping";
+    }
+
+    leaf max-sessions {
+      type uint32;
+      description
+        "Maximum number of concurrent sessions";
+    }
+
+    leaf bidirectional {
+      type boolean;
+      default false;
+      description
+        "Enable bidirectional translation for inside-to-outside
+        and outside-to-inside directions";
+    }
+
+    leaf enabled {
+      type boolean;
+      default true;
+      description
+        "Administrative state of the static mapping";
+    }
+  }
+
+  grouping nat-mapping-state {
+    description
+      "Operational state data for static NAT mappings";
+
+    leaf creation-time {
+      type yang:date-and-time;
+      description
+        "Timestamp when the mapping was created";
+    }
+
+    leaf last-used {
+      type yang:date-and-time;
+      description
+        "Timestamp when the mapping was last used";
+    }
+
+    leaf packet-count {
+      type yang:counter64;
+      description
+        "Number of packets translated using this mapping";
+    }
+
+    leaf byte-count {
+      type yang:counter64;
+      description
+        "Number of bytes translated using this mapping";
+    }
+
+    leaf hit-count {
+      type yang:counter64;
+      description
+        "Number of packets matching this mapping's traffic criteria";
+    }
+
+    leaf active-sessions {
+      type uint32;
+      description
+        "Number of currently active sessions for this mapping";
+    }
+  }
+
+  grouping nat-instance-top {
+    description
+      "Top-level grouping for NAT instances";
+
+    container instances {
+      description
+        "NAT instance configuration and state";
+
+      list instance {
+        key "name";
+        description
+          "List of NAT instances";
+
+        leaf name {
+          type leafref {
+            path "../config/name";
+          }
+          description
+            "Reference to NAT instance name";
+        }
+
+        container config {
+          description
+            "Configuration parameters for the NAT instance";
+
+          uses nat-instance-config;
+        }
+
+        container state {
+          config false;
+          description
+            "Operational state data for the NAT instance";
+
+          uses nat-instance-config;
+          uses nat-instance-state;
+        }
+
+        uses nat-interface-top;
+        uses nat-pool-top;
+        uses nat-mapping-top;
+      }
+    }
+  }
+
+  grouping nat-pool-top {
+    description
+      "Top-level grouping for NAT pools";
+
+    container pools {
+      description
+        "NAT address pool configuration and state";
+
+      list pool {
+        key "name";
+        description
+          "List of NAT address pools";
+
+        leaf name {
+          type leafref {
+            path "../config/name";
+          }
+          description
+            "Reference to pool name";
+        }
+
+        container config {
+          description
+            "Configuration parameters for the NAT pool";
+
+          uses nat-pool-config;
+        }
+
+        container state {
+          config false;
+          description
+            "Operational state data for the NAT pool";
+
+          uses nat-pool-config;
+          uses nat-pool-state;
+        }
+      }
+    }
+  }
+
+  grouping nat-mapping-top {
+    description
+      "Top-level grouping for static NAT mappings";
+
+    container mappings {
+      description
+        "Static NAT mapping configuration and state";
+
+      list mapping {
+        key "name";
+        description
+          "List of static NAT mappings";
+
+        leaf name {
+          type leafref {
+            path "../config/name";
+          }
+          description
+            "Reference to mapping name";
+        }
+
+        container config {
+          description
+            "Configuration parameters for the static mapping";
+
+          uses nat-mapping-config;
+        }
+
+        container state {
+          config false;
+          description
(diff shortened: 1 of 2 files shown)


release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,802 @@
...
+        "Ending port number for NAPT translations";
+    }
+
+    leaf port-block-size {
+      type uint16;
+      description
+        "Number of ports allocated per mapping";
+    }
+
+    choice traffic-matcher {
```
dplore: [OC style](https://github.com/openconfig/public/blob/692101403054d60763cc252df54ad0812e250169/doc/openconfig_style_guide.md#choice) strongly prefers to avoid choice statements, but you could use when for enforcing some validations in the model.  (So far we have not used `choice` in any OC model)

dplore (commented):
Thank your @nleiva for this contribution!  I've only done a very shallow review so far of the general tree structure so far, but please consider using when statements and or descriptions instead of the choice statement.

release/models/nat/openconfig-nat.yang
nleiva: Thanks @dplore, that was an oversight on our end. We replaced `choice` with `when` statements.

nleiva:
I removed the `routing-policy` option and updated the data-tree.

nleiva:
I addressed the comments in the community call.
1. Added a state translation-table.
2. Put any counter in a counters container.
3. Removed uptime.

ghananigans:
From today's OC weekly meeting: @ghananigans to do a pass to address some ambiguity in naming and expected behaviour

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+          this translation";
+      }
+    }
+  }
+
+  grouping nat-translation-table-top {
+    description
+      "Top-level grouping for NAT translation table";
+
+    container translation-table {
```
dplore: Style guide says this should be translations/translation for example

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+        different ports";
+    }
+
+    leaf log-translations {
+      type boolean;
+      description
+        "Enable logging of NAT translations";
+    }
+
+    leaf max-translations-per-address {
```
ghananigans, line 465: I see above we use the term "sessions". Does "translations" mean the same thing as "sessions?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+      type boolean;
+      description
+        "Enable address-only translation without port translation";
+    }
+
+    leaf timeout {
+      type uint32;
+      units "seconds";
+      description
+        "Default NAT mapping timeout";
```
ghananigans: When does this default apply? 
* Does setting this modify the other timeouts? 
* Does setting this control NAT timeouts for non-TCP/UDP/ICMP traffic?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+      }
+      enum DEBUG {
+        description "Log debug messages";
+      }
+    }
+    description
+      "Logging level for NAT operations";
+  }
+
+  typedef interface-type {
```
ghananigans, line 140: As discussed in the OC weekly meeting -- the terms "inside" and "outside" are ambiguous and may cause confusion.

Based on our discussion, the model seems to be focused on performing source NAT on traffic moving from a private to a public network. However, NAT may also be performed in the opposite direction (public -> private) or between private/private or public/public networks. 

To reduce iterations in the future, can we update the model to be general/flexible enough to support those other cases in the future, without having to actively support them today?

release/models/nat/openconfig-nat.yang
ghananigans: When does the timeout apply? Is it when no traffic is observed going through an active translation/session?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+      }
+      description
+        "Reference to ACL set that defines traffic
+        matching criteria for this static mapping";
+    }
+
+    leaf persistent {
+      type boolean;
+      description
+        "Enable persistent NAT mapping";
```
ghananigans: What is a persistent/non-persistent mapping?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+      }
+
+      leaf available-addresses {
+        type uint32;
+        description
+          "Number of addresses available in this pool";
+      }
+
+      leaf active-sessions {
+        type uint32;
```
ghananigans: any reason this isn't `yang:counter64`?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+    description
+      "Configuration parameters for static NAT mappings";
+
+    leaf name {
+      type string;
+      description
+        "Unique name for the static mapping";
+    }
+
+    leaf internal-address {
```
ghananigans, line 527: Since this model is focused on NAT44, should we update the model prefix from `nat` to `nat44`? It may help reduce confusion if in the future OC is updated to support other forms of NAT

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+
+        leaf translation-timeout {
+          type uint32;
+          units "seconds";
+          description
+            "Default translation timeout for all NAT instances";
+        }
+
+        leaf max-translations {
+          type uint32;
```
ghananigans: what happens when a new traffic flow is presented to the NAT subsystem that would normally be translated but we are already at this limit?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+          type boolean;
+          description
+            "Global administrative state for NAT functionality";
+        }
+
+        leaf translation-timeout {
+          type uint32;
+          units "seconds";
+          description
+            "Default translation timeout for all NAT instances";
```
ghananigans: When does this default apply? Is it a global max for NAT instance timeouts? If so, does configuring something like `tcp-timeout` to something larger than this cause an error or is only the MIN across both values used?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,965 @@
...
+
+        uses nat-interface-top;
+        uses nat-pool-top;
+        uses nat-mapping-top;
+        uses nat-translation-table-top;
+      }
+    }
+  }
+
+  grouping nat-pool-top {
```
ghananigans: IIUC, a NAT pool is for dynamic NAT (from/to some pool of addresses) and NAT mapping is for staticaly defined NAT (1:1).

If my understanding is correct, why do we need two separate groups for this? Can a static mapping be represented with a pool but with a pool size of 1 and `overload` set to false?

Please let me know if I am misunderstanding the intent!

ghananigans:
Thank you @nleiva! For my comments, I'm more than happy to discuss things over a video call if you'd like :)!

release/models/nat/openconfig-nat.yang
nleiva: Good catch. I'm fixing this in the next commit.

release/models/nat/openconfig-nat.yang
nleiva: You are absolutely right, I'm replacing sessions with translations everywhere for consistency.

release/models/nat/openconfig-nat.yang
nleiva: Good point, I'll update the default timeout description to clarify that it applies to all traffic unless specific protocol timeouts are configured.

release/models/nat/openconfig-nat.yang
nleiva: Absolutely. I'm removing any references to private or public addressing, as well as adding a NAT type to configure NAT to translate source or destination addresses.

release/models/nat/openconfig-nat.yang
nleiva: Correct. I'm updating the description to reflect this.
```
 "Default time a NAT translation entry remains active
  without traffic for all NAT instances. Applies to all
  traffic unless specific protocol timeouts are
  configured."
```

release/models/nat/openconfig-nat.yang
nleiva: I'll remove this, is probably redundant (static mappings are persistent).

release/models/nat/openconfig-nat.yang
nleiva: I'll add an `address-family` leaf and remove any IPv4 references, so the model covers NAT44 and NAT66

release/models/nat/openconfig-nat.yang
nleiva: I'll make it a 64-bit to potentially support CGNAT.

release/models/nat/openconfig-nat.yang
nleiva: I think that's up to the vendor implementation. Arista, for example, states:
```
The ip nat translation max-entries command specifies maximum number of NAT translation connections. After this threshold is reached, new connections are dropped until the number of programmed connections is reduced below the level specified by the ip nat translation low-mark command.
```

SOURCE: https://www.arista.com/en/um-eos/eos-data-plane-security#xx1149232

release/models/nat/openconfig-nat.yang
nleiva: I updated the description to this, the more specific timeout takes precedence.

```
            "Default time a NAT translation entry remains active
            without traffic for all NAT instances. Applies to all
            traffic unless specific protocol timeouts are
            configured.";
```

release/models/nat/openconfig-nat.yang
nleiva: Your observation is correct, we could probably optimize this. However, two separate groups is closer to how most vendors implement this.

ghananigans: Opening for discussion: If we agree that there is an opportunity to simplify this OC model by combining static/dynamic NAT modules and leveraging configurations to reasonably differentiate between them, should we prefer that even though vendors may require different native configuration paths to support the same? The idea being that based on a mapping's configuration (e.g., `overload`, pool sizes), an OC implementation can redirect the request to the native dynamic or static NAT subsystem.

cc: @ElodinLaarz @dplore @robshakir -- has OC hit a similar design question previously, and if so, what approach did we lean towards? From the [contribution guide](https://github.com/openconfig/public/blob/master/doc/contributions-guide.md#rationale), my interpretation of "minimum operationally viable" leans towards unifying the model, but I'd love to hear everyone's thoughts!

ghananigans: Clarification (I think this question applies beyond this discussion): What is the relative priority between static and dynamic mappings? 

Say an interface has both static and dynamic, are static mappings checked first or is dynamic? Is this prioritization configurable?

nleiva:
# Static

## CLI (Arista)

```c
interface ethernet3
  ip nat source static 192.0.2.114 acl 2-stat-nat-policy 198.51.100.97
  ip nat source static 192.0.2.116 acl 2-stat-nat-policy 198.51.100.99
  ...

```

## JSON

```json
{
  "openconfig-nat:instances": {
    "instance": [
      {
        "config": {
          "address-family": "openconfig-nat:IPV4",
          "name": "default",
          "type":"openconfig-nat:SOURCE"
        },
        "mappings": {
          "mapping": [
            {
              "config": {
                "acl-set": "2-stat-nat-policy",
                "external-address": "198.51.100.97",
                "internal-address": "192.0.2.114",
                "name": "192.0.2.114"
              },
              "name": "192.0.2.114"
            },
            {
              "config": {
                "acl-set": "2-stat-nat-policy",
                "external-address": "198.51.100.99",
                "internal-address": "192.0.2.116",
                "name": "192.0.2.116"
              },
              "name": "192.0.2.116"
            },
          ...
          ]
        },
        "name": "default"
      }
    ]
  }
}
```


!

# Dynamic

## CLI (Arista)

```c
interface ethernet3
  ip nat source dynamic access-list 1-dyna-nat-policy pool customer-nat address-only
!
ip nat pool customer-nat prefix-length 16
  range 203.0.113.1 137.19.175.255
!
```


## JSON

```json
{
  "openconfig-nat:instances": {
    "instance": [
      {
        "config": {
          "address-family": "openconfig-nat:IPV4",
          "name": "default",
          "type":"openconfig-nat:SOURCE"
        },
        "name": "default",
        "pools": {
          "pool": [
            {
              "config": {
                "acl-set": "1-dyna-nat-policy",
                "end-address": "137.19.175.255",
                "name": "customer-nat",
                "prefix-length":16,
                "start-address": "203.0.113.1"
              },
              "name": "customer-nat"
            }
          ]
        }
      }
    ]
  }
} 
```

release/models/nat/openconfig-nat.yang
ghananigans: Can we introduced a configuration and enum that enumerates this scenario? The behavior should always be explicitly defined.

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,1008 @@
...
+      "Base identity for NAT address families.";
+  }
+
+  identity IPV4 {
+    base ADDRESS_FAMILY;
+    description
+      "IPv4 address family for NAT44 translations.";
+  }
+
+  identity IPV6 {
```
ghananigans, line 79: Since this model originally started off as exclusively NAT44, is there a need for the IPv6 case? If not, lets defer adding this identity until it is needed

release/models/nat/openconfig-nat.yang
ghananigans: can we also avoid the terms "inside" and "outside"?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,1008 @@
...
+        "Reference to ACL set for this static mapping.";
+    }
+
+    leaf max-translations {
+      type uint64;
+      description
+        "Maximum number of concurrent translations.";
+    }
+
+    leaf enabled {
```
ghananigans: What happens when this is disabled _after_ a translation has already happened? Does the existing translation continue to remain active, or are the existing active static mappings cleared?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,1008 @@
...
+        uses nat-translation-table-top;
+      }
+    }
+  }
+
+  grouping nat-pool-top {
+    description
+      "Top-level grouping for NAT pools.";
+
+    container pools {
```
ghananigans: instead of "pools" for dynamic NAT and "mappings" for static NAT (which I find to be ambiguous), can we use the terms "dynamic" and "static" which seems to match the terminology used across vendors?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,1008 @@
...
+        "Top-level container for NAT.";
+
+      container config {
+        description
+          "Global NAT configuration parameters.";
+
+        leaf enabled {
+          type boolean;
+          description
+            "Global administrative state for NAT functionality.";
```
ghananigans: same comment as elsewhere -- what happens when this is disabled after some flow has already been NAT-ed and there is an active translation?

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,1008 @@
...
+
+      container state {
+        config false;
+        description
+          "Global NAT operational state data.";
+
+        leaf enabled {
+          type boolean;
+          description
+            "Global administrative state for NAT functionality.";
```
ghananigans: same comment as elsewhere -- what happens when this is disabled after some flow has already been NAT-ed and there is an active translation?

release/models/nat/openconfig-nat.yang
ghananigans: To clarify: Im not saying we should revert the recent changes made (e.g., `ip-address` in place of `ipv4-address`). I'm only suggesting we defer the use/configuration of IPv6 here (by excluding the IPv6 nat-type/address family variants) until there is an ask.

We can leave a TODO to support NAT66 if there is an ask.

release/models/nat/openconfig-nat.yang
ghananigans: Dug into some details and I think the dynamic & static NAT _are_ actually different enough that we can't merge the trees, so that question is unblocked.

However, what is the relative priority between static and dynamic mappings?

nleiva:
> I spent some time looking at some documentation and believe that dynamic NAT configurations cannot absorb static.
> 
> The key thing is that for static, the device seems to always require pre (for packet matching) and post translation addresses with an optional ACL; dynamic NAT on the other hand only requires the post-translation and relies exclusively on the ACL for packet matching.
> 
> Please feel free to double-check my understanding!

I believe that's correct. Should be good to merge then? Thanks

release/models/nat/openconfig-nat.yang
nleiva: I can remove `IPv6` from the identity `ADDRESS_FAMILY` if that's what you prefer. I'd also have to shorten `prefix-length`. Does this work?

```c
    leaf prefix-length {
      type uint8 {
        range "1..128";
      }
      description
        "Prefix length for the address pool. For IPv4, valid range
        is 1-32. For IPv6, valid range is 1-128.";
    }
```

release/models/nat/openconfig-nat.yang
nleiva: I'm not sure. Probably the same behavior as removing that map config entry altogether. Do you have a preference to include in the model?

release/models/nat/openconfig-nat.yang
nleiva: Good call, I just changed this.

nleiva:
> There are some open comments remaining on this PR that should be addressed first

I'm not sure why I missed that there were new ones. I just addressed most of them.

release/models/nat/openconfig-nat.yang
nleiva: I'm removing all `enabled` leaves to avoid ambiguity.

release/models/nat/openconfig-nat.yang
nleiva: Do you prefer internal, external for the `interface-type?`
```c
  typedef interface-type {
    type enumeration {
      enum INSIDE {
        description
          "Source interface for NAT address translation.";
      }
      enum OUTSIDE {
        description
          "Destination interface for translated addresses.";
      }
    }
    description
      "NAT interface role: INSIDE or OUTSIDE.";
  }
```

ElodinLaarz:
Discussed at the OC Operators Meeting on August 26th:
A couple outstanding comments to be resolving
  * [on ipv6](https://github.com/openconfig/public/pull/1335#discussion_r2297868154)
  * [static v. dynamic configuration precedence](https://github.com/openconfig/public/pull/1335#discussion_r2271076246)
  
  (and a couple minor changes / comment threads to be closed out).
  
  Pushing out last-call date to next week.

release/models/nat/openconfig-nat.yang
nleiva: > However, what is the relative priority between static and dynamic mappings?

Vendors documentation suggests static is preferred over dynamic in their public documentation, as static is a permanent table entry and will match its tuple. This behavior doesn't seem to be configurable, so we can go with non-configurable static over dynamic in the model as well. What do you think?

[Cisco](https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/ipaddr_nat/configuration/15-mt/nat-15-mt-book/iadnat-addr-consv.html?utm_source=chatgpt.com#GUID-64306C21-506F-4085-8587-654BB69C890C)

```
When static NAT and dynamic NAT are configured with route maps that share the same name, static NAT is given precedence over dynamic NAT.
```

[Juniper](https://www.juniper.net/documentation/us/en/software/junos/nat/topics/topic-map/security-nat-overview.html#id-understanding-nat-rule-sets-and-rules__d20837e309)

```
Rule Processing
The NAT type determines the order in which NAT rules are processed. During the first packet processing for a flow, NAT rules are applied in the following order:
1. Static NAT rules
2. Destination NAT rules
3. Route lookup
4. Security policy lookup
5. Reverse mapping of static NAT rules
6. Source NAT rules
```

[Arista](https://www.arista.com/en/um-eos/eos-data-plane-security)

```
Note: the same IP address can't be used for the NAT static configuration and in the pool for dynamic NAT configurations. Public IP addresses must be unique. the global addresses used in static translations aren't excluded from dynamic pools containing the same global addresses.
```

release/models/nat/openconfig-nat.yang
ghananigans: Discussed in OC operator meeting, existing terms are OK

release/models/nat/openconfig-nat.yang
ghananigans: kind ping on this

release/models/nat/openconfig-nat.yang
ghananigans: Thanks! Let's document that behaviour on this model -- that static takes precedence over dynamic.

release/models/nat/openconfig-nat.yang
ghananigans: Discussed during OC operator meeting, keeping IPV6 is fine

release/models/nat/openconfig-nat.yang
ghananigans: https://github.com/openconfig/public/pull/1335#discussion_r2279750626

release/models/nat/openconfig-nat.yang
ghananigans: https://github.com/openconfig/public/pull/1335#discussion_r2279750626

ghananigans:
LGTM once the following threads are closed:
* https://github.com/openconfig/public/pull/1335#issuecomment-3184863319
* https://github.com/openconfig/public/pull/1335#discussion_r2301631898

release/models/nat/openconfig-nat.yang
nleiva: Sure. I'm adding the following:
```c
  grouping nat-instance-config {
    description
      "Configuration parameters for NAT instances.

      When both static and dynamic NAT are configured within the same
      NAT instance, static NAT takes precedence over dynamic NAT.
      Static NAT entries are permanent mappings that will always match
      their configured address/port tuples before any dynamic NAT
      rules are evaluated.";

```

release/models/nat/openconfig-nat.yang
nleiva: To keep it simple, I'm documenting the behavior:

**Global**:
```c
    container nat {
    ...
        leaf max-translations {
          type uint64;
          description
            "Maximum number of concurrent translations globally across
            all NAT instances. When this threshold is reached, new
            translation requests are dropped until the total number
            of active translations falls below this limit.";
        }
```

**Dynamic**:
```c
  grouping nat-dynamic-config {
    ...
    leaf max-translations {
      type uint64;
      description
        "Maximum number of concurrent translations for this dynamic
        NAT pool. When this threshold is reached, new translation
        requests are dropped until the number of active translations
        falls below this limit.";
    }
```

**Static**:
```c
  grouping nat-static-config {
    leaf max-translations {
      type uint64;
      description
        "Maximum number of concurrent translations for this static
        NAT mapping. When this threshold is reached, new translation
        requests using this mapping are dropped until the number of
        active translations falls below this limit.";
    }
  }
```

nleiva:
> LGTM once the following threads are closed:
> 
> * [Add NAT model #1335 (comment)](https://github.com/openconfig/public/pull/1335#discussion_r2301633857)
> * [Add NAT model #1335 (comment)](https://github.com/openconfig/public/pull/1335#discussion_r2301631898)

Thanks @ghananigans for all your comments, the model looks so much better now. I believe I addressed all the remaining items, let me know otherwise.

@ElodinLaarz @dplore would you run `gcbrun` again for me to make sure I didn't break anything. Thanks.

release/models/nat/openconfig-nat.yang
```diff
@@ -0,0 +1,1006 @@
...
+      type inet:ip-address;
+      description
+        "External (destination) IP address. Must match the address
+        family configured for the NAT instance.";
+    }
+
+    leaf internal-port {
+      type inet:port-number;
+      description
+        "Internal port number for static NAPT mappings.";
```
dplore: Please clarify that the `internal-port` and `external-port` leaves are optional and what happens if they are not configured (port mappings are expected to be one to one)

nleiva: Thanks @dplore, I added:

```c
    leaf internal-port {
      type inet:port-number;
      description
        "Internal port number for static NAPT mappings. When
        specified, this option enables port forwarding, where both
        the IP address and the port are translated. If omitted,
        only the IP address is translated.";
    }

    leaf external-port {
      type inet:port-number;
      description
        "External port number for static NAPT mappings. When
        specified, this option enables port forwarding, where both
        the IP address and the port are translated. If omitted,
        only the IP address is translated.";
    }
```

dplore (approved):

PR #1337, opened 2025-07-17
Add alarm thresholds to fan container

mrevang:
### Change Scope

* Similar to transceivers, this change introduces a container of `thresholds`, indexed by `severity`, to the `fan` container. This allows the device to export upper/lower speed thresholds, which enables monitoring of the fan's instantaneous speed. 

```
        +--rw fan
        |  +--rw config
        |  +--ro state
        |  |  +--ro oc-fan:speed?   uint32
        |  +--rw oc-fan:thresholds <<<<<<<<<<<<<<<<<<<<<<<<<<<<
        |     +--ro oc-fan:threshold* [severity]
        |        +--ro oc-fan:severity    -> ../state/severity
        |        +--ro oc-fan:state
        |           +--ro oc-fan:severity?   identityref
        |           +--ro oc-fan:upper?      uint32
        |           +--ro oc-fan:lower?      uint32
```

### Platform Implementations

 * Implementation A: https://github.com/sonic-net/sonic-platform-daemons/blob/master/sonic-thermalctld/scripts/thermalctld
 * Implementation B: https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst6500/ios/15-4SY/config_guide/sup6T/15_3_sy_swcg_6T/environment.html

### DIFF
--- a/release/models/platform/openconfig-platform-fan.yang
+++ b/release/models/platform/openconfig-platform-fan.yang
@@ -9,7 +9,7 @@ module openconfig-platform-fan {
 
   import openconfig-platform { prefix oc-platform; }
   import openconfig-extensions { prefix oc-ext; }
-
+  import openconfig-alarm-types { prefix oc-alarm-types; }
 
   // meta
   organization "OpenConfig working group";
@@ -22,8 +22,14 @@ module openconfig-platform-fan {
     "This module defines data related to FAN components in the
     OpenConfig platform model.";
 
-  oc-ext:openconfig-version "0.1.1";
+  oc-ext:openconfig-version "0.2.0";
 
+  revision "2025-07-16" {
+    description
+      "Add fan alarm thresholds.";
+    reference "0.2.0";
+  }
+  
   revision "2018-11-21" {
     description
       "Add OpenConfig module metadata extensions.";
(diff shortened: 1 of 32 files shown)


release/models/platform/openconfig-platform-fan.yang
```diff
@@ -47,6 +53,68 @@ module openconfig-platform-fan {
...
+      "Grouping for all alarm thresholds for a particular severity level.";
+    leaf severity {
+      type identityref {
+        base oc-alarm-types:OPENCONFIG_ALARM_SEVERITY;
+      }
+      description
+        "The type of alarm to which the thresholds apply.";
+    }
+
+    leaf upper {
```
dplore:
```suggestion
    leaf speed-upper {
```

release/models/platform/openconfig-platform-fan.yang
```diff
@@ -47,6 +53,68 @@ module openconfig-platform-fan {
...
+      type uint32;
+      units rpm;
+      description
+        "Upper fan speed alarm threshold";
+    }
+
+    leaf lower {
```
dplore:
```suggestion
    leaf speed-lower {
```

ElodinLaarz:
Discussed at OC Operators meeting on August 5th.

(1) For consistency with how it's used in the `/transceiver/...` subtree, the name of the threshold should include the value captured. (In this case, `speed-upper` an `speed-lower`, rather than just `upper`/`lower`.)
(2) Since we have no `config` leaves here, we should include in the description the expectation for where these values are sourced from. (Manufacturer should be populating the expected thresholds here-- same as the expectation for transceivers.)

release/models/platform/openconfig-platform-fan.yang
mrevang: Applied this suggestion manually.

mrevang:
> Discussed at OC Operators meeting on August 5th.
> 
> (1) For consistency with how it's used in the `/transceiver/...` subtree, the name of the threshold should include the value captured. (In this case, `speed-upper` an `speed-lower`, rather than just `upper`/`lower`.) (2) Since we have no `config` leaves here, we should include in the description the expectation for where these values are sourced from. (Manufacturer should be populating the expected thresholds here-- same as the expectation for transceivers.)

Updated the state container's description to capture this expectation about sourcing.

dplore (approved):

PR #1339, opened 2025-07-18
Adding platform model for USB port

migueberumenjr:
### Change Scope

Adding new platform model for USB ports, enable/disable functionality.

### Platform Implementations

- Mobility Controller-based Aruba Access Points: https://arubanetworking.hpe.com/techdocs/CLI-Bank/Content/aos8/ap-system-pro.htm (`ap-usb-power-mode [enable|disable]`)

- Cisco Catalyst switches: https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst9400/software/release/17-17/command_reference/b_1717_9400_cr/interface_and_hardware_commands.html?dtid=osscdc000283&linkclickid=srch#wp3775305635  (`[no] platform usb disable`)

### DIFF
--- a/release/models/platform/openconfig-platform-types.yang
+++ b/release/models/platform/openconfig-platform-types.yang
@@ -393,7 +393,7 @@ module openconfig-platform-types {
   identity USB_PORT {
     base OPENCONFIG_HARDWARE_COMPONENT;
     description
-      "USB port, typically USB-A.";
+      "Non-serial USB ports of any type.";
   }
 
   identity TRANSCEIVER {
--- a/release/models/platform/openconfig-platform-types.yang
+++ b/release/models/platform/openconfig-platform-types.yang
@@ -22,8 +22,14 @@ module openconfig-platform-types {
     "This module defines data types (e.g., YANG identities)
     to support the OpenConfig component inventory model.";
 
-  oc-ext:openconfig-version "1.9.0";
+  oc-ext:openconfig-version "1.10.0";
 
+
+  revision "2025-07-18" {
+    description
+      "Add USB_PORT";
+    reference "1.10.0";
+  }
 
   revision "2024-11-04" {
     description
@@ -382,6 +388,12 @@ module openconfig-platform-types {
     description
       "Physical port, e.g., for attaching pluggables and networking
       cables";
+  }
+
+  identity USB_PORT {
+    base OPENCONFIG_HARDWARE_COMPONENT;
+    description
+      "USB port, typically USB-A.";
   }
 
   identity TRANSCEIVER {
--- a/release/models/platform/openconfig-platform-usb-port.yang
+++ b/release/models/platform/openconfig-platform-usb-port.yang
@@ -24,7 +24,7 @@ module openconfig-platform-usb-port {
 
   oc-ext:openconfig-version "0.1.0";
 
-  revision "2025-07-12" {
+  revision "2025-08-18" {
     description
       "Initial revision";
     reference "0.1.0";
@@ -37,49 +37,34 @@ module openconfig-platform-usb-port {
 
   // grouping statements
 
-  grouping platform-usb-port-config {
+  grouping usb-port-config {
     description
-      "Configuration data for USB port";
+      "Configuration data for USB ports";
 
-    leaf enabled {
-      type boolean;
-      description
-        "The configured state (enabled/disabled) of this USB port.";
-    }
+    uses oc-platform:component-power-management;
   }
 
-  grouping platform-usb-port-state {
+  // data definition statements
+
+  // augment statements
+
+  augment "/oc-platform:components/oc-platform:component/" +
+    "oc-platform:usb-port/oc-platform:config" {
     description
-      "Operational state data for USB port";
+      "Configuration data for USB ports.
+      A USB port can be configured for persistent powered-off
+      mode using the config/power-admin-state leaf.";
+
+    uses usb-port-config;
   }
 
-  grouping usb-port-top {
+  augment "/oc-platform:components/oc-platform:component/" +
+    "oc-platform:usb-port/oc-platform:state" {
     description
-      "Top-level grouping for USB port data";
+      "Adding USB port data to physical inventory. This subtree
+      is only valid when the type of the component is USB_PORT.";
 
-    container usb-port {
-      description
-        "Top-level container for USB port";
-
-      container config {
-        description
-          "Configuration data for USB port";
-
-        uses platform-usb-port-config;
-      }
-
-      container state {
-
-        config false;
-
-        description
-          "Operational state data for USB port";
-
-        uses platform-usb-port-config;
-        uses platform-usb-port-state;
-      }
-    }
+    uses usb-port-config;
   }
-  uses usb-port-top;
 
 }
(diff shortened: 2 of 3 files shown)


release/models/platform/openconfig-platform-usb-port.yang
```diff
@@ -0,0 +1,86 @@
...
+
+  // grouping statements
+
+  grouping platform-usb-port-config {
+    description
+      "Configuration data for USB port";
+
+    leaf enabled {
+      type boolean;
+      default "false";
```
dplore: OC [prefers to not define a default value](https://github.com/openconfig/public/blob/692101403054d60763cc252df54ad0812e250169/doc/openconfig_style_guide.md?plain=1#L265-L274), unless there is some industry standard we can refer to that requires this and is widely implemented.  This means any configuration needs to specify what the desire is and that there is no openconfig specification of what a 'factory default' setting must be (which is often problematic).

release/models/platform/openconfig-platform-types.yang
```diff
@@ -384,6 +390,12 @@ module openconfig-platform-types {
       cables";
   }
 
+  identity USB_PORT {
+    base OPENCONFIG_HARDWARE_COMPONENT;
+    description
+      "USB port, typically USB-A.";
```
dplore: We should make this more strict.  Is this intended to be used for any and all USB port types?  Or if we're expecting only USB-A, then we'll need something more explicit to identify the types you want.

release/models/platform/openconfig-platform-usb-port.yang
```diff
@@ -0,0 +1,86 @@
+module openconfig-platform-usb-port {
```
earies, line 1: Few general comments:

* Can you articulate what your specific intentions are with respect to conveying a port-type such as USB?  Will we start to define specific types for everything else?  Is this predominantly for serial (console, aux) communications, block storage devices, etc..?  How about non-USB serial, etc..?
* Why would this need to be a self-contained model vs. expansion of existing?
* If for serial purposes, then what about all other relevant settings?
* If for storage devices, then how might this component relate to a storage component?

release/models/platform/openconfig-platform-types.yang
earies: Why would we need to mention connector types at all here?  This seems irrelevant and could come in many flavors - in fact we already have many connector types across network element platforms when it comes to USB for console, storage devices, etc..

release/models/platform/openconfig-platform-usb-port.yang
migueberumenjr: Ack. I will remove the default value.

release/models/platform/openconfig-platform-types.yang
migueberumenjr: The intended use is all non-serial USB port types; therefore, I will update the description accordingly.

release/models/platform/openconfig-platform-usb-port.yang
migueberumenjr: Per July 22nd meeting:
- This would apply to non-serial USB ports
- The model's simplicity and self-containment is based on a security requirement to enable/disable USB ports, regardless of what kind of device is connected or its functionality

earies: I don't believe everything should be modeled as a "component" as this is already turning into an infinte growth issue within OC modeling that has no guidelines or model constraints to indicate what structures/nodes apply in correspondance to component types.  It is also complexity for no good reason that will only see more and more divergence across implementations imo.

My recommendation is this is one of many cases where it should likely rather be represented as a structure on the component in which it resides (e.g. For the `CONTROLLER_CARD` component).

If we consider the above component to be the one that represents a "system" these USB ports can likely be indexed as hubs (e.g. `lsusb`) which have bus and and device identifiers which need to be understood from a mapping perspective.

As-is, this would mean we are adding another set of components to the component tree that would have no rules around what child nodes apply or do not apply.

xavier-contreras: CONROLLER_CARD is defined as "A type of linecard whose primary role is management or control rather than data forwarding." which IMO sounds limiting for including USB_PORT under that. There is no "chassis" component for example, but that is potentially where USB_PORT could be (and others like linecard, fan, etc) but these models are already not structured that way.

On the topic of USB hub/child nodes, yeah a bit vague there but looking at vendor implementations linked and others, that distinction/level of detail is not really specified. Would a more verbose description here help?

release/models/platform/openconfig-platform-usb-port.yang
```diff
@@ -0,0 +1,86 @@
...
+        config false;
+
+        description
+          "Operational state data for USB port";
+
+        uses platform-usb-port-config;
+        uses platform-usb-port-state;
+      }
+    }
+  }
```
xavier-contreras, line 68: All other components augment "/oc-platform:components/oc-platform:component" -- seems like that should be the same here?

migueberumenjr: Updated to utilize an augment.

ElodinLaarz:
Discussed in the OC Operators Meeting on August 26th:

@dplore can you take another look to make sure things seem good and if we can move it out of Ready to Discuss?

Thanks~

nkitchen-arista:
The name `power-admin-state` looks like it is potentially misleading -- are the ports necessarily disabled by being powered off? Or can they be disabled in software while remaining powered?

release/models/platform/openconfig-platform-usb-port.yang
dplore: OC intentionally avoids defining any specific hardware design.  The component tree allows a system to expose a tree of hardware and software components, each optionally with a `parent` to allow a tree of components to be defined.  

A system should be free to construct a tree of components that nested a `USB_PORT` under whatever other components it choses to represent itself.    A USB port could have a parent of CHASSIS, LINECARD or other, to be introduced in the future, component.    I realize this flexibility leaves open a lot of room for different implementations.  This is intentional as OC avoids modeling hardware, even though we want to monitor and control it.  

In this area, expectations of operators are best represented with a [featureprofiles](https://github.com/openconfig/featureprofiles) test that targets a specific device and use case.

release/models/platform/openconfig-platform-usb-port.yang
```diff
@@ -0,0 +1,70 @@
...
+  oc-ext:catalog-organization "openconfig";
+  oc-ext:origin "openconfig";
+
+  // grouping statements
+
+  grouping usb-port-config {
+    description
+      "Configuration data for USB ports";
+
+    uses oc-platform:component-power-management;
```
dplore, line 44: I recommend defining an "enabled" leaf as a boolean, following the [precedent for interfaces](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-interfaces.html#interfaces-interface-config-enabled)

release/models/platform/openconfig-platform-usb-port.yang
dplore: Follow the example pattern here for transceivers:

https://github.com/openconfig/public/blob/7777c94ff390b6c86296328d6ea830396a79891d/release/models/platform/openconfig-platform-transceiver.yang#L1058-L1064

dplore:
closed in favor of #1361

PR #1340, opened 2025-07-18
mark transceiver serial-no and vendor-part as deprecated

SydneyCaulfeild:
### Change Scope

This change is marking the `components/component/transceiver/state/serial-no` and `components/component/transceiver/state/vendor-part` leaves as deprecated. The `components/component/state/serial-no` and `components/component/state/part-no` leaves should be used instead, respectively, going forward.

These duplicate serial and part number `components/component/transceiver/state/*` leaves are redundant and the top-level component ones (`components/component/state/*`) are preferred.

### Platform Implementations
Nvidia has hardware (GB200) that is using the `components/component/transceiver/state/serial-no` and `components/component/transceiver/state/vendor-part` instead of `components/component/state/serial-no` and `components/component/state/part-no`. Example:

```
openconfig/components/component/sw11/transceiver/state/serial-no, MT2503TDN3E6A
openconfig/components/component/sw11/transceiver/state/vendor-part, 980-9IAM2-00X001
```

Arista also has hardware (DCS-7808-CH) that uses `openconfig/components/component/transceiver/state/vendor-part` and `openconfig/components/component/transceiver/state/serial-no`. Note that this model of Arista is populating both serial-no leaves. Example:

```
openconfig/components/component/Ethernet4/1/state/serial-no, V8O25120026-B
openconfig/components/component/Ethernet4/1/transceiver/state/serial-no, V8O25120026-B
openconfig/components/component/Ethernet4/1/transceiver/state/vendor-part, LDO-080-DZ-GO1
```

### DIFF
--- a/release/models/platform/openconfig-platform-transceiver.yang
+++ b/release/models/platform/openconfig-platform-transceiver.yang
@@ -740,7 +740,7 @@ module openconfig-platform-transceiver {
         right with ASCII spaces (20h). If part number is undefined,
         all 16 octets = 0h. This leaf is deprecated and the
         components/component/state/part-no leaf should be used
-        instead.";
+        to hold this value instead (these leaves are duplicates).";
     }
 
     leaf vendor-rev {
@@ -792,7 +792,7 @@ module openconfig-platform-transceiver {
         ASCII spaces (20h). If part serial number is undefined, all
         16 octets = 0h. This leaf is deprecated and the
         components/component/state/serial-no leaf should be used
-        instead.";
+        to hold this value instead (these leaves are duplicates).";
     }
 
     leaf date-code {
--- a/release/models/platform/openconfig-platform-transceiver.yang
+++ b/release/models/platform/openconfig-platform-transceiver.yang
@@ -733,6 +733,7 @@ module openconfig-platform-transceiver {
       type string {
         length 1..16;
       }
+      status deprecated;
       description
         "Transceiver vendor's part number. 16-octet field that
         contains ASCII characters, left-aligned and padded on the
@@ -784,6 +785,7 @@ module openconfig-platform-transceiver {
       type string {
         length 1..16;
       }
+      status deprecated;
       description
         "Transceiver serial number. 16-octet field that contains
         ASCII characters, left-aligned and padded on the right with
--- a/release/models/platform/openconfig-platform-transceiver.yang
+++ b/release/models/platform/openconfig-platform-transceiver.yang
@@ -741,7 +741,7 @@ module openconfig-platform-transceiver {
         all 16 octets = 0h. This deprecated leaf should continue
         to be supported by implementations but clients should stop
         using this leaf and instead use
-        '/components/component/state/part-no'.";  
+        '/components/component/state/part-no'.";
     }
 
     leaf vendor-rev {
--- a/release/models/platform/openconfig-platform-transceiver.yang
+++ b/release/models/platform/openconfig-platform-transceiver.yang
@@ -731,7 +731,9 @@ module openconfig-platform-transceiver {
         "Transceiver vendor's part number. 16-octet field that
         contains ASCII characters, left-aligned and padded on the
         right with ASCII spaces (20h). If part number is undefined,
-        all 16 octets = 0h";
+        all 16 octets = 0h. This leaf is deprecated and the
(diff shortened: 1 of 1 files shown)


release/models/platform/openconfig-platform-transceiver.yang
```diff
@@ -731,7 +737,9 @@ module openconfig-platform-transceiver {
         "Transceiver vendor's part number. 16-octet field that
         contains ASCII characters, left-aligned and padded on the
         right with ASCII spaces (20h). If part number is undefined,
-        all 16 octets = 0h";
+        all 16 octets = 0h. This leaf is deprecated and the
```
ElodinLaarz: You can also mark the leaf as deprecated, explicitly using `status deprecated;` in the `leaf` container.

(Same for the other leaf.)

dplore: +1, this should be marked deprecated.  

As a note for the community, deprecated nodes should still be supported by devices as per [OC style guide](https://github.com/openconfig/public/blob/692101403054d60763cc252df54ad0812e250169/doc/releases.md?plain=1#L81) and the Yang standard. 

Operators should not create any new dependencies on deprecated nodes and should add support for the alternate nodes.  Eventually (many years) the deprecated node could be deleted from the yang models.  This would only happen at some point in the distant future when no devices using the node are present in networks.  There is no hard limit on this timeline, but 6+ years is a possibility.    OC introduced the use of deprecated node status about 3 years ago and so far, none of these deprecated nodes has been deleted.

SydneyCaulfeild: Thanks for pointing that out! I have added `status deprecated` to both.

ElodinLaarz:
> > You included a single vendor in your initial comment, and usually it helps to have more than one vendor to show neutrality? In this case, it's probably fine since the leaves exist in the model.
> > You mention a different leaf in your initial comment than the one you list in the description, though? `components/component/state/serial-no` vs. `transceiver/state/serial-no` (which I assume is `/components/component/transceiver/state/serial-no`.)
> > Do we want to distinguish / add a comment about this vs. `/components/component/state/serial-no`?
> 
> I'm only aware of this one vendor, but I figured since this involves marking something as deprecating and not actually yet removing anything that this would be ok.
> 
> @ElodinLaarz Could you please elaborate on your concern on distinguishing the two serial-no leaves? Do you mean add a note to `/components/component/transceiver/state/serial-no` in the .yang to explain that it is just a duplicate of `components/component/state/serial-no`? Or cleaning up the PR description?

Sorry, my description was a bit confusing upon reading it again-- yes, I was just asking that the description be updated to be clear on the leaf to be updated.

We also discussed this in the OC Operators Meeting today (July 22nd); you mentioned a single vendor in the initial comment, for sake of completeness, could you include two vendors that provide the data? Otherwise, the change seems good, overall.

SydneyCaulfeild:
I found Arista devices that currently export these leaves. I also found some old Juniper examples, but have not found a current one yet. So for now I have listed Nvidia and Arista in the description.

release/models/platform/openconfig-platform-transceiver.yang
```diff
@@ -727,11 +733,14 @@ module openconfig-platform-transceiver {
...
       }
+      status deprecated;
       description
         "Transceiver vendor's part number. 16-octet field that
         contains ASCII characters, left-aligned and padded on the
         right with ASCII spaces (20h). If part number is undefined,
-        all 16 octets = 0h";
+        all 16 octets = 0h. This leaf is deprecated and the
+        components/component/state/part-no leaf should be used
+        to hold this value instead (these leaves are duplicates).";
```
dplore: We should say something that emphasizes device should continue to support this leaf until it is removed.

Here's a suggestion:
```
This deprecated leaf should continue to be supported by implementations but clients should 
stop using this leaf and instead use '/components/component/state/part-no'.  
```

SydneyCaulfeild: Done, thanks for the suggestion!

ElodinLaarz (approved):
Generally LGTM.

Can you add links to some documentation from the vendors you listed that shows the claimed support? (You can look at some of the other PRs to see where folks normally add links from.)

PR #1341, opened 2025-07-21
Adding a host-certificate string to allow clients to authenticate the host.

sourcequench:
SSH host certificates are digital documents that verify the identity of an SSH server to a client. They are signed by a trusted Certificate Authority (CA) and include information like the server's public key, its hostname, and an expiration date. This allows clients to securely connect to SSH servers without needing to rely on static host keys, streamlining authentication and enhancing security.

### DIFF
--- a/release/models/system/openconfig-system-terminal.yang
+++ b/release/models/system/openconfig-system-terminal.yang
@@ -27,7 +27,7 @@ module openconfig-system-terminal {
   revision "2025-11-10" {
     description
       "Add a host-certificate leaf.";
-    reference "0.3.2";
+    reference "0.4.0";
   }
 
   revision "2018-11-21" {


dplore:
Should this follow our pattern elsewhere where we point to a certificate id?  ([example](https://github.com/openconfig/public/blob/692101403054d60763cc252df54ad0812e250169/release/models/system/openconfig-system-grpc.yang#L170))

earies:
> Do we need these other files/leafs?
> 
> Host Certificate - the signature made for the Host Public Key using the Certificate Authority. (presumably this is the `host-certificate` leaf added in this PR so far)
> 
> Certificate public key - the public component of the certificate authority
> 
> Host Public Key - the actual key that the SSH daemon uses to identify itself to the clients

How about authorized principles and mappings to users?  There are already implementations that can be referenced here (Note: this also would need to not assume gNSI)

release/models/system/openconfig-system-terminal.yang
```diff
@@ -154,6 +154,12 @@ module openconfig-system-terminal {
         "Set the protocol version for SSH connections to the system";
     }
 
+    leaf host-certificate {
+      type string;
+      description
+        "SSH Host Certificate string.";
```
dplore: Is this the signature or the full content of the host cert?

Should we reference https://datatracker.ietf.org/doc/draft-miller-ssh-cert/ (or other doc of your choice) so there is no ambiguity?

sourcequench: All ssh parameters, whether keys or certificates, whether the yang models or in credentialz are for openssh. Save for some niche, nobody uses the x509 ietf standard for certificates.

This is the full content of an openssh host certificate.

dplore: SGTM, can you update the description with `This is the full content of an openssh host certificate`

sourcequench: Done.

release/models/system/openconfig-system-terminal.yang
```diff
@@ -22,7 +22,13 @@ module openconfig-system-terminal {
     "This module defines configuration and operational state data
     related to remote terminal services such as ssh and telnet.";
 
-  oc-ext:openconfig-version "0.3.1";
+  oc-ext:openconfig-version "0.3.2";
```
ElodinLaarz: Should be upgraded to `0.4.0` since the "patch version" should only be for corrections.

ElodinLaarz:
Discussed at the OC Operators Meeting Dec 2nd - Seems the leaves are being added under `/system/ssh-server/config`, can we add a diff of the tree before/after this PR in the first comment?

Additionally, it seems important to have the certificate expiration leaf exposed here, as well? I know that we're trying to keep it minimal here, but  it seems beneficial to add here. (Similar to [this leaf](https://github.com/openconfig/public/blob/834db2b62f8ee66ee6c8f8b567569ad81caac34e/release/models/gnsi/openconfig-gnsi-certz.yang#L149).)

(The information would also be in the text, but parsing text seems like an unpleasant requirement to get the expiration information.)

release/models/system/openconfig-system-terminal.yang
```diff
@@ -22,7 +22,13 @@ module openconfig-system-terminal {
     "This module defines configuration and operational state data
     related to remote terminal services such as ssh and telnet.";
 
-  oc-ext:openconfig-version "0.3.1";
+  oc-ext:openconfig-version "0.4.0";
+
+  revision "2025-11-10" {
+    description
+      "Add a host-certificate leaf.";
+    reference "0.3.2";
```
dplore:
```suggestion
    reference "0.4.0";
```

PR #1342, opened 2025-07-23
Connections

ThatOneSix:
### Change Scope

* This change adds a new container, 'sockets', under /network-instance/. It is loosely based on the `netstat` command. There is room to expand the container at a later date to include other sub-containers as needed.

### Platform Implementations

 * [Ubuntu `netstat` manual](https://manpages.ubuntu.com/manpages/jammy/en/man8/netstat.8.html) 
 * [Aruba `show datapath session table`](https://arubanetworking.hpe.com/techdocs/CLI-Bank/Content/sd-branch/sh-dp-sestbl.htm)
 
```
augment /oc-ni:network-instances/oc-ni:network-instance:
    +--rw sockets
       +--ro socket* [local-address local-port remote-address remote-port]
          +--ro local-address     -> ../state/local-address
          +--ro local-port        -> ../state/local-port
          +--ro remote-address    -> ../state/remote-address
          +--ro remote-port       -> ../state/remote-port
          +--ro state
             +--ro socket-type?      union
             +--ro local-address?    oc-inet:host
             +--ro local-port?       oc-inet:port-number
             +--ro remote-address?   oc-inet:host
             +--ro remote-port?      oc-inet:port-number
             +--ro tcp-state?        enumeration
             +--ro process-name?     string
             +--ro pid?              uint64
```

### DIFF
--- a/release/models/devices-manifest/openconfig-terminal-device-properties.yang
+++ b/release/models/devices-manifest/openconfig-terminal-device-properties.yang
@@ -35,10 +35,14 @@ module openconfig-terminal-device-properties {
       the OTSi (OTSiMC). It also includes (optional) aspects such as
       filter characterization, CD and DGD tolerance.";
 
-  oc-ext:openconfig-version "0.2.0";
-
+  oc-ext:openconfig-version "0.2.1";
 
   // Revisions
+  revision "2025-08-06" {
+    description "Correct order of string type in unions.";
+    reference "0.2.1";
+  }
+
   revision "2024-05-28" {
       description "Comprehensive model update to undertake the limitations
       of the first version of the model, such the introduction of modes
@@ -111,8 +115,8 @@ module openconfig-terminal-device-properties {
 
     leaf fec-coding {
       type union {
-        type string;
         type oc-opt-term-prop-types:fec-coding;
+        type string;
       }
       description
         "Forward error correction (FEC) coding schema used in the
@@ -148,8 +152,8 @@ module openconfig-terminal-device-properties {
 
     leaf pulse-shaping-type {
       type union {
-        type string;
         type oc-opt-term-prop-types:pulse-shaping-type;
+        type string;
       }
       description
        "Pulse/spectral shaping type such as Raised-cosine (RC),
@@ -398,13 +402,13 @@ module openconfig-terminal-device-properties {
 
     leaf value {
       type union {
-        type string;
         type boolean;
         type int64;
         type uint64;
         type decimal64 {
           fraction-digits 2;
         }
+        type string;
       }
       description
         "Property values can take on a variety of types.  Signed and
@@ -425,8 +429,8 @@ module openconfig-terminal-device-properties {
 
     leaf publisher-organization {
       type union {
-        type string;
         type oc-opt-term-prop-types:interoperability-modes-organization;
+        type string;
       }
       description
         "Name of the organization, standard body, Multi-Source Agreement, or
@@ -533,8 +537,8 @@ module openconfig-terminal-device-properties {
 
     leaf modulation-format {
       type union {
-        type string;
         type oc-opt-term-prop-types:modulation-format;
+        type string;
       }
       description
         "Optical modulation format associated to the mode. The
--- a/release/models/gnpsi/openconfig-gnpsi-types.yang
+++ b/release/models/gnpsi/openconfig-gnpsi-types.yang
@@ -7,8 +7,8 @@ module openconfig-gnpsi-types {
     prefix oc-ext;
   }
 
-  import openconfig-grpc-types {
-    prefix oc-grpct;
+  import openconfig-system-grpc {
+    prefix oc-grpc;
   }
 
   organization
@@ -20,9 +20,16 @@ module openconfig-gnpsi-types {
     "This module adds gRPC server type relating to gNPSI running on
     a network device.";
 
-  oc-ext:openconfig-version "0.1.0";
+  oc-ext:openconfig-version "0.1.1";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
+
+  revision 2025-10-29 {
+    description
+      "Update description and move the GNPSI identity to be a GRPC_SERVICE
+      under openconfig-system-grpc.";
+    reference "0.1.1";
+  }
 
   revision 2024-05-29 {
     description
@@ -31,8 +38,8 @@ module openconfig-gnpsi-types {
   }
 
   identity GNPSI {
-    base oc-grpct:GRPC_SERVICE;
+    base "oc-grpc:GRPC_SERVICE";
     description
-      "gNMI: gRPC Network Management Interface";
+      "gNPSI: gRPC Network Packet Sampling Interface.";
   }
 }
--- a/release/models/gnsi/openconfig-gnsi-certz.yang
+++ b/release/models/gnsi/openconfig-gnsi-certz.yang
@@ -30,7 +30,13 @@ module openconfig-gnsi-certz {
     "This module provides a data model for the metadata of gRPC credentials
     installed on a networking device.";
 
-  oc-ext:openconfig-version "0.6.0";
+  oc-ext:openconfig-version "0.7.0";
+
+  revision 2025-09-26 {
+    description
+      "Add certificate-expires-on leaf";
+    reference "0.7.0";
+  }
 
   revision 2024-03-05 {
     description
@@ -78,14 +84,6 @@ module openconfig-gnsi-certz {
       a reboot.";
   }
 
-  typedef created-on {
-    type oc-types:timeticks64;
-    description
-      "The creation time of the credential as reported by the credential
-      manager when the credential was pushed to the device. This value is
-      reported as nanoseconds since epoch (January 1st, 1970 00:00:00 GMT).
-      This leaf persists through a reboot.";
-  }
   // gRPC server related definitions.
   // Success/failure counters.
   grouping grpc-server-certz-counters  {
@@ -139,11 +137,24 @@ module openconfig-gnsi-certz {
         private key) that is used by this gRPC server.";
     }
     leaf certificate-created-on {
-      type created-on;
+      type oc-types:timeticks64;
       description
         "The timestamp of the moment when the certificate
         (and associated private key) that is currently used
-        by this gRPC server was created.";
+        by this gRPC server was created.
+
+        The value is the timestamp in nanoseconds relative to
+        the Unix Epoch (Jan 1, 1970 00:00:00 UTC).";
+    }
+    leaf certificate-expires-on {
+      type oc-types:timeticks64;
+      description
+        "The timestamp of the moment when the certificate
+        (and associated private key) that is currently used
+        by this gRPC server will expire.
+
+        The value is the timestamp in nanoseconds relative to
+        the Unix Epoch (Jan 1, 1970 00:00:00 UTC).";
     }
     leaf ca-trust-bundle-version {
       type version;
@@ -153,11 +164,15 @@ module openconfig-gnsi-certz {
         this gRPC server.";
     }
     leaf ca-trust-bundle-created-on {
-      type created-on;
+      type oc-types:timeticks64;
       description
         "The timestamp of the moment when the bundle of
         the Certificate Authority certificates (a.k.a.
-            trust bundle) was created.";
+        trust bundle) was created.
+
+        This leaf persists through a reboot.  The value is the
+        timestamp in nanoseconds relative to the Unix Epoch
+        (Jan 1, 1970 00:00:00 UTC).";
     }
     leaf certificate-revocation-list-bundle-version {
       type version;
@@ -166,10 +181,14 @@ module openconfig-gnsi-certz {
         this gRPC server.";
     }
     leaf certificate-revocation-list-bundle-created-on {
-      type created-on;
+      type oc-types:timeticks64;
       description
         "The timestamp of the moment when the Certificate Revocation
-        List bundle was created.";
+        List bundle was created.
+
+        This leaf persists through a reboot.  The value is the
+        timestamp in nanoseconds relative to the Unix Epoch
+        (Jan 1, 1970 00:00:00 UTC).";
     }
     leaf authentication-policy-version {
       type version;
@@ -178,10 +197,14 @@ module openconfig-gnsi-certz {
         this gRPC server.";
     }
     leaf authentication-policy-created-on {
-      type created-on;
+      type oc-types:timeticks64;
       description
         "The timestamp of the moment when the authentication policy
-        that is currently used by this gRPC server was created.";
+        that is currently used by this gRPC server was created.
+
+        This leaf persists through a reboot.  The value is the
+        timestamp in nanoseconds relative to the Unix Epoch
+        (Jan 1, 1970 00:00:00 UTC).";
     }
     // TODO(earies): Consider aligning this with grpc-server key after moving
     // to a system-wide subtree:
--- a/release/models/gnsi/openconfig-gnsi-credentialz.yang
+++ b/release/models/gnsi/openconfig-gnsi-credentialz.yang
@@ -39,7 +39,13 @@ module openconfig-gnsi-credentialz {
       /system/aaa/authentication/users/user/config/password-hashed
       /system/aaa/authentication/users/user/state/password-hashed";
 
-  oc-ext:openconfig-version "0.7.0";
+  oc-ext:openconfig-version "0.8.0";
+
+  revision 2025-10-23 {
+    description
+      "Adding active-host-certificate-expires-on.";
+    reference "0.8.0";
+  }
 
   revision 2024-09-10 {
     description
@@ -93,15 +99,6 @@ module openconfig-gnsi-credentialz {
       a reboot.";
   }
 
-  typedef created-on {
-    type oc-types:timeticks64;
-    description
-      "The creation time of the credential as reported by the credential
-      manager when the credential was pushed to the device. This value is
-      reported as nanoseconds since epoch (January 1st, 1970 00:00:00 GMT).
-      This leaf persists through a reboot.";
-  }
-
   // SSH server related definitions.
 
   grouping ssh-server-credentials-version {
@@ -115,10 +112,14 @@ module openconfig-gnsi-credentialz {
     }
 
     leaf active-trusted-user-ca-keys-created-on {
-      type created-on;
+      type oc-types:timeticks64;
       description
         "The timestamp of the moment when the trusted user CA keys
-        were created.";
+        were created.
+
+        This leaf persists through a reboot.  The value is the
+        timestamp in nanoseconds relative to the Unix Epoch
+        (Jan 1, 1970 00:00:00 UTC).";
     }
 
     leaf active-host-certificate-version {
@@ -128,10 +129,25 @@ module openconfig-gnsi-credentialz {
     }
 
     leaf active-host-certificate-created-on {
-      type created-on;
+      type oc-types:timeticks64;
       description
         "The timestamp of the moment when the host certificate
-        was created.";
+        was created.
+
+        This leaf persists through a reboot.  The value is the
+        timestamp in nanoseconds relative to the Unix Epoch
+        (Jan 1, 1970 00:00:00 UTC).";
+    }
+
+    leaf active-host-certificate-expires-on {
+      type oc-types:timeticks64;
+      description
+        "The timestamp of the moment when the host certificate
+        expires.
+
+        This leaf persists through a reboot.  The value is the
+        timestamp in nanoseconds relative to the Unix Epoch
+        (Jan 1, 1970 00:00:00 UTC).";
     }
 
     leaf active-host-key-version {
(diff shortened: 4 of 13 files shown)


ThatOneSix:
Changed to use `augment` in the sub-module and removed all changes to openconfig-system.yang.

jsterne:
Don't we likely need a network-instance dimension to this?

ThatOneSix:
To clarify, do you mean adding something like...

```
import openconfig-network-instance { prefix oc-ni; }

leaf network-instance {
      type oc-ni:network-instance-ref;
      description
        "Lorem ipsum";
```

So that, for example, L3 devices can report which VRF the connection is on?

ThatOneSix:
Oh, I didn't consider that. Good point. So something like...

+--rw system
|  +--rw connections
|  |  +--rw network-instance [network-instance]
|  |  |  +--rw network-sockets
|  |  |      +--rw network-socket [socket]

Where socket is source IP + port.

Or

+--rw system
|  +--rw connections
|  |  +--rw internet-sockets
|  |     +--rw internet-socket [socket]

Where socket is IP + port + network instance.

Please correct me if I've misunderstood.

Looking at the two, I think the first makes more sense.

jsterne:
I somewhat prefer the first one as well.

Another question: aren't sockets sometimes identified by a 4-part tuple (local ip, local port, remote ip, remote port)?  [listening ports are maybe just local ip + local port]

ThatOneSix:
I've reworked the file to the following:

module: openconfig-connections

  augment /oc-sys:system:
    +--rw connections
       +--rw network-instances
          +--ro network-instance* [instance]
             +--ro instance           -> ../state/instance
             +--ro state
             |  +--ro instance?   oc-ni:network-instance-ref
             +--ro network-sockets
                +--ro network-socket* [socket]
                   +--ro socket    -> ../state/socket
                   +--ro state
                      +--ro socket?           string
                      +--ro protocol?         union
                      +--ro local-address?    oc-inet:host
                      +--ro local-port?       oc-inet:port-number
                      +--ro remote-address?   oc-inet:host
                      +--ro remote-port?      oc-inet:port-number
                      +--ro tcp-state?        enumeration
                      +--ro process-name?     string
                      +--ro pid?              uint64

Ran through pyang lint successfully.

As for the socket, I was mostly using local address + local port to make it more readable. I am open to changing the key for the socket list to something else.

earies:
RE: `netstat` - note netstat is obsolete, suggest referring to iproute2 `ss` and `ip` utils rather as an alternative

Structural comment based off prior comments....

If we are going to enumerate network-instances, I'd propose we not create yet another list but rather this could be anchored under `/network-instances`

But when it comes to `/system` we often have the design decision to choose to treat the system as a single entity which then often means "active controller-card" only or consider the possibility of a distributed system where connections, pids, storage, etc... could all be a subset of a component thus possibly warranting this to be a subset of `/components` (We already see this w/ prior published system related items)

Another quick comment as to the naming of the module and top-level node - imo "connections" is too vague here so suggest narrowing this in rather to the categories that correspond to the above utilities (if we lean towards such which I think is mostly ok for this purpose)

release/models/network-instance/openconfig-sockets.yang
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    leaf process-name {
+      type string;
+      description
+        "The name of the process associated with the connection.";
+    }
+
+    leaf pid {
+      type uint64;
```
earies, line 222: While `openconfig-procmon` defines a pid as `uint64`, I think this is too generous.  Linux/BSD as well as VxWorks based systems use `uint32` for PIDs/tasks often w/ a much lower max/default

```
$ cat /proc/sys/kernel/pid_max 
32768
```

dplore:
I suspect to resolve the issue about network-instances, adding it as part of the "connections" list key simplifies the model and matches the linux implementation well.  


```
module: openconfig-connections

  augment /oc-sys:system:
    +--rw connections
       +--rw connection* [network-instance,socket]
             +--ro network-instance -> ../state/network-instance
             +--ro socket           -> ../state/socker
             +--ro state
             |  +--ro network-instance?   oc-ni:network-instance-ref
             |  +--ro socket?           string
             |  +--ro protocol?         union
                +--ro local-address?    oc-inet:host
                +--ro local-port?       oc-inet:port-number
                +--ro remote-address?   oc-inet:host
                +--ro remote-port?      oc-inet:port-number
                +--ro tcp-state?        enumeration
                +--ro process-name?     string
                +--ro pid?              uint64
```


Regarding multiple CONTROLLER_CARDs, we can choose (in a future PR) to replicate this as a subtree underneath /components/component when type = CONTROLLER_CARD.  ie:
`/components/component/controller-card/connections/connection`

This follows the style that /system is a copy of the current state of the "whole device" and /components provides a "drill down" view into individual component state.

ThatOneSix:
Modified the openconfig-connections.yang file to incorporate suggestions about simplifying the list. Ran through pyang lint and received no errors. The tree it output is:

```
module: openconfig-system
  +--rw system
     +--rw oc-conn:connections
        +--ro oc-conn:connection* [network-instance socket]
           +--ro oc-conn:network-instance    -> ../state/network-instance
           +--ro oc-conn:socket              -> ../state/socket
           +--ro oc-conn:state
              +--ro oc-conn:socket?             string
              +--ro oc-conn:network-instance?   oc-ni:network-instance-ref
              +--ro oc-conn:protocol?           union
              +--ro oc-conn:local-address?      oc-inet:host
              +--ro oc-conn:local-port?         oc-inet:port-number
              +--ro oc-conn:remote-address?     oc-inet:host
              +--ro oc-conn:remote-port?        oc-inet:port-number
              +--ro oc-conn:tcp-state?          enumeration
              +--ro oc-conn:process-name?       string
              +--ro oc-conn:pid?                uint64
```

LimeHat (changes requested):
1. A major issue with the proposal is the definition of what we are trying to expose.
If we are talking about _connections_, then this applies only to TCP. Most of the other standard protocols (IP, UDP, ICMP) are connectionless by definition.
I'm not sure what is the intention: to list the sockets, or to list the connections, but the current combination of the proposed models (that lists many protocols) and the terminology/description (connections, active connections) doesn't make much sense to me.
2. I would question the placement of this in /system/ tree. IMO a better home for this is /network-instances/ tree (where each NI can expose its own state) -> I don't know if this was discussed in earlier calls, but to me this would be a much cleaner model.

release/models/system/openconfig-connections.yang
```diff
@@ -0,0 +1,247 @@
...
+          }
+          enum ICMP {
+            description
+              "Index value indicating ICMP";
+          }
+          enum TCP {
+            description
+              "Index value indicating TCP";
+          }
+          enum TCP6 {
```
LimeHat: Does this mean that the other TCP (above) is TCP over IPv4? (the description doesn't mention that)
What's the motivation for this split?

ThatOneSix: As mentioned in the description, this is based on the `netstat` command, which separates TCPv4 and TCPv6 traffic into `TCP` and `TCP6`. Depending on the vendor implementation, they could choose to use TCP for both IPv4 and IPv6 traffic, or split them into the separate enums. I also included `type string;` at the end of the enums so that vendors could use additional/different descriptions as they prefer.

LimeHat: I don't think we should take netstat output as the ultimate authority, or try to model netstat in openconfig.
While it is a good idea to look at Linux when modeling such abstractions, we shouldn't copy random things blindly. 
netstat "proto" field just represents names from /proc/net/ files. They may or may not be protocols. tcp6 is not a protocol. 

Moreover, using `/proc/net/` and `netstat` is generally not considered best practices. Netlink with `sock_diag`/`tcp_diag` and `ss` tool are the modern replacements. 

>they could choose to use TCP for both IPv4 and IPv6 traffic, or split them into the separate enums

That's unnecessary ambiguity.  
Another fun fact: ipv6 sockets can also accept ipv4 connections, therefore even if you read the data from /proc/net/tcp6 you can still (sometimes!) see ipv4 connections there :-)

tl;dr i don't think that's the right way. this field needs to be simplified to remove non-protocols, or eliminated altogether (depending on the answer to my other questions).
>I also included type string; at the end of the enums so that vendors could use additional/different descriptions as they prefer.

not a fan of that either, i'd eliminate that if you don't have a real use case in mind. 
if you want to allow extensibility just do an identityref instead of enums (and if we stick to the notion of protocols, maybe we can even do that outside of this module)

release/models/system/openconfig-connections.yang
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    leaf network-instance {
+      type oc-ni:network-instance-ref;
+      description
+        "The network instance used to for the connection.  If no
+        instance is specified, DEFAULT_INSTANCE is used.";
+    }
+
+    leaf protocol {
```
ghananigans: This seems to read like less like a protocol and more like a socket ~_type_~_kind_.

release/models/system/openconfig-connections.yang
ghananigans: +1 to what @LimeHat said.

What problem are we trying to solve with this model change? 

I don't think there is an issue using existing solutions like `netstat`/`/proc/net`/`sock_diag` for inspiration when solving our problems, but we should also confirm how our problems align with those of existing solutions before doing so.

release/models/system/openconfig-connections.yang
```diff
@@ -0,0 +1,247 @@
...
+      description
+        "Parameters related to all active network connections";
+
+      list connection {
+        key "network-instance socket";
+        config false;
+        description
+          "Parameters related to individual active network connections";
+
+        leaf network-instance {
```
ghananigans: What does this leaf point to? In the context of Ubuntu/Linux (since that is 1 of 2 implementation listed in the PR description) does it point to the network interface that a socket is _explicitly_ bound to, or the interface a route for the socket is cached to egress through? 

What if the socket is bound to a specific local address but no remote, what should show up here? Similarly, what if the socket is bound all addresses?

release/models/system/openconfig-connections.yang
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  // OpenConfig specific extensions for module metadata.
+  oc-ext:regexp-posix;
+  oc-ext:catalog-organization "openconfig";
+  oc-ext:origin "openconfig";
+
+  // grouping statements
+
+  grouping connections-top {
```
ghananigans: As @LimeHat suggested, I don't think "connections" is an appropriate name for a container that includes connection-_less_ protocols/socket.

ThatOneSix: Both the `ss` and `netstat` manuals describe connectionless protocols as having established a connection. I am following the language used there.

"When no option is used ss displays a list of open non-listening sockets (e.g. TCP/UNIX/UDP) that have established connection."

https://man7.org/linux/man-pages/man8/ss.8.html

"Active Internet connections (TCP, UDP, UDPLite, raw)
   Proto
       The protocol (tcp, udp, udpl, raw) used by the socket."
       
https://man7.org/linux/man-pages/man8/netstat.8.html

release/models/system/openconfig-connections.yang
ghananigans: Or in linux terms, an identifier for the union of `{SOCKET_ADDRESS_FAMILY, SOCKET_TYPE, SOCKET_PROTOCOL}`

where the `SOCKET_ADDRESS_FAMILY` does not necessarily align with the address family of the underlying connection itse.f

release/models/system/openconfig-connections.yang
ghananigans: I agree that existing Linux-focused tools use those terms, but as OC is a vendor(platform)-neutral network configuration model, we should lean towards terms that better represent the _intent_ of the configuration or state.

The intent here seems to be to represent sockets and not connections per se. For example, a TCPv6 _socket_ may be used to perform an IPv4 TCP _connection_. 

```
[hi on] ghanan@ghanan-cloudtop:~$ netstat -n -a | grep tcp6 | head -5
tcp6       0      0 ::1:9557                :::*                    LISTEN     
tcp6       0      0 ::1:9558                :::*                    LISTEN     
tcp6       0      0 ::1:631                 :::*                    LISTEN     
tcp6       0      0 ::1:6670                :::*                    LISTEN     
tcp6       0      0 127.0.0.1:53195         :::*                    LISTEN     
[hi on] ghanan@ghanan-cloudtop:~$ netstat -n -a | grep tcp6 | tail -5
tcp6       0      0 127.0.0.1:32888         127.0.0.1:9557          ESTABLISHED
tcp6       0      0 127.0.0.1:40870         127.0.0.1:9557          ESTABLISHED
tcp6       0      0 127.0.0.1:43366         127.0.0.1:9557          ESTABLISHED
tcp6       0      0 127.0.0.1:38710         127.0.0.1:9557          ESTABLISHED
tcp6       0      0 127.0.0.1:43120         127.0.0.1:9557          ESTABLISHED
```

release/models/system/openconfig-connections.yang
dplore: In OC,  network-instance is a VRF / [Virtual Routing a Forwarding](https://en.wikipedia.org/wiki/Virtual_routing_and_forwarding).

release/models/system/openconfig-connections.yang
dplore: Predecent in OC is for v4 vs. v6 things is to either

1. Regarding ip-address, this can be a string that contains v4 and v6 formatted strings
2.  Or an address family container is introduced, with leaves underneath it.  ie:  
`foo/v4/leaf1`
`foo/v6/leaf2`

In this case I suggest we model the IP enum as v4 and v6 and TCP also as a single enum value for v4 and v6 "connections".

release/models/system/openconfig-connections.yang
dplore: This could be socket-type. (we don't have precedent to use "kind" in OC)

LimeHat:
> 1. A major issue with the proposal is the definition of what we are trying to expose.
>    If we are talking about _connections_, then this applies only to TCP. Most of the other standard protocols (IP, UDP, ICMP) are connectionless by definition.
>    I'm not sure what is the intention: to list the sockets, or to list the connections, but the current combination of the proposed models (that lists many protocols) and the terminology/description (connections, active connections) doesn't make much sense to me.
> 2. I would question the placement of this in /system/ tree. IMO a better home for this is /network-instances/ tree (where each NI can expose its own state) -> I don't know if this was discussed in earlier calls, but to me this would be a much cleaner model.

I would like to get my comments addressed as well.

release/models/system/openconfig-connections.yang
LimeHat: >Both the ss and netstat manuals describe connectionless protocols as having established a connection. I am following the language used there.

Linux tools do use the term, but even in that case the term and the interpretation can have multiple meanings.

For example, in `ss` output, you'll see `estab` state for UDP when an app uses `connect()` syscall. But that's the socket state. Then you have other means such as `conntrack`, which is another mechanism to track "connections" (pseudo-connections) in linux, including ICMP and UDP.. but it works differently and the output will not match the socket state.

I think we need to be more precise in what we trying to define and why.
Also, if you're considering a security angle, wouldn't it be important to list the listening sockets?

dplore:
> > 1. A major issue with the proposal is the definition of what we are trying to expose.
> >    If we are talking about _connections_, then this applies only to TCP. Most of the other standard protocols (IP, UDP, ICMP) are connectionless by definition.
> >    I'm not sure what is the intention: to list the sockets, or to list the connections, but the current combination of the proposed models (that lists many protocols) and the terminology/description (connections, active connections) doesn't make much sense to me.
> > 2. I would question the placement of this in /system/ tree. IMO a better home for this is /network-instances/ tree (where each NI can expose its own state) -> I don't know if this was discussed in earlier calls, but to me this would be a much cleaner model.
> 
> I would like to get my comments addressed as well.

`/system` is a good location for this.  `/system` is intended to represent the state of the device's currently active running operating system and is mostly independent of all the "networking" configuration.  

Because the intent of `/system` is to model state of the NOS and Linux is the dominant kernel used and is open, it's precedent is strong.   While the UDP protocol is connectionless and does not track a connection state, the "system" does. This PR models system state, not the packets on the wire.

Regarding `connections`, your comments are very insightful and appreciated.  Maybe this clarification of modeling the "system" helps explain why `connections` is defensible.  If you still feel otherwise, then in order to affect this PR please provide some alternative which can be reviewed.  The suggestion will need stronger and more relevant precedent than what is already provided.  The final decision on the naming is up to the rough consensus of the OC Operators. 

Expanding the scope to include listening ports is at the discretion of PR author.  It sounds like a good idea.  But if there isn't an operational need, modeling listening sockets doesn't have to be included in this PR at this time.

release/models/system/openconfig-connections.yang
ghananigans: "socket-type" sgtm!

LimeHat:
> This PR models system state, not the packets on the wire.

I totally agree with that, but that doesn't mean it belongs to the `/system/` :-)

> /system is intended to represent the state of the device's currently active running operating system and is mostly independent of all the "networking" configuration.

Great, then this is exactly why the connections should be moved to `/network-instance`.

Connections and sockets are not independent from the networking configuration. Network instances are (typically) mapped to [network namespaces](https://man7.org/linux/man-pages/man7/network_namespaces.7.html) in Linux! Each network namespace operates its own set of interfaces and sockets. They are totally configuration dependent. If i create network-instances `mgmt1` and `mgmt2`, I will create two new network namespaces in linux. Each namespace will have its own socket table (`/proc/net/udp`, `/proc/net/tcp` that are consumed by netstat), etc. There's nothing global about this :-) And sockets and connections will be opened according to configuration of apps that attached to these network instances.

ghananigans:
> > This PR models system state, not the packets on the wire.
> 
> I totally agree with that, but that doesn't mean it belongs to the `/system/` :-)
> 
> > /system is intended to represent the state of the device's currently active running operating system and is mostly independent of all the "networking" configuration.
> 
> Great, then this is exactly why the connections should be moved to `/network-instance`.
> 
> Connections and sockets are not independent from the networking configuration. Network instances are (typically) mapped to [network namespaces](https://man7.org/linux/man-pages/man7/network_namespaces.7.html) in Linux! Each network namespace operates its own set of interfaces and sockets. They are totally configuration dependent. If i create network-instances `mgmt1` and `mgmt2`, I will create two new network namespaces in linux. Each namespace will have its own socket table (`/proc/net/udp`, `/proc/net/tcp` that are consumed by netstat), etc. There's nothing global about this :-) And sockets and connections will be opened according to the configuration of apps that attached to these network instances.

Thats a good point, @LimeHat. Thank you for pointing that out. I hadn't considered the fact that, at least on Linux, each network namespace maintains its own distinct set of socket tables and to support a full query of sockets on a system, we would have to check each namespace. 

It's unclear to me if _all_ vendors implement sockets & network-instances this way, but I think expressing this with the model is fine, so long as a socket cannot easily/non-disruptively move across network-instances throughout their lifetime (I believe this to be true on most platforms, but just explicitly pointing it out as an assumption I have).

```
# Terminal 1
$ sudo ip netns exec testing nc -u -l :: 12346 

# Terminal 2
$ nc -u -l :: 12345

# Terminal 3
$ netstat -tunlp | grep "12345\|12346"
udp6       0      0 :::12345                :::*                                796206/nc           
$ sudo ip netns exec testing netstat -tunlp | grep "12345\|12346"
udp6       0      0 :::12346                :::*                                796592/nc       
```
Given the above, I now _do_ believe that sockets may be better placed within `network-instance` which itself may be part a wider platform/system alongside other `network-instances`.

dplore:
@ThatOneSix I'd like to see a response regarding modeling socket connections and namespaces.  Are sockets in linux contained within namespaces?  How do the vendor implemenations you want to support this compare?  Do they support vrfs/network-instances?

ThatOneSix:
@dplore Apologies for the delay; I was out of the country. I will review and respond ASAP.

ThatOneSix:
@dplore @LimeHat @ghananigans

I have updated the sockets config to now be in network-instances. No issues when run through pyang lint. This is the tree:

```
+--rw oc-sockets:sockets
           +--ro oc-sockets:socket* [socket]
              +--ro oc-sockets:socket    -> ../state/socket
              +--ro oc-sockets:state
                 +--ro oc-sockets:socket?           string
                 +--ro oc-sockets:socket-type?      union
                 +--ro oc-sockets:local-address?    oc-inet:host
                 +--ro oc-sockets:local-port?       oc-inet:port-number
                 +--ro oc-sockets:remote-address?   oc-inet:host
                 +--ro oc-sockets:remote-port?      oc-inet:port-number
                 +--ro oc-sockets:tcp-state?        enumeration
                 +--ro oc-sockets:process-name?     string
                 +--ro oc-sockets:pid?              uint64
```

I have modified it to fit other suggestions as well, such as removing TCP6 and UDP6 and making the language focus on sockets instead of connections (other than TCP state info).

ThatOneSix:
Added the new line characters at the ends of the YAML files to fix the lint issue.

release/models/network-instance/openconfig-sockets.yang
ElodinLaarz: This matches `/system/processes/process/state/pid` being a uint64, as well. Seems reasonable (and protbuf will be efficient if the value doesn't use the excess room, anyway).

release/models/system/openconfig-connections.yang
ElodinLaarz: Now `TCP` (without an address family indicator) is the only socket type.

PR #1343, opened 2025-07-27
Create openconfig-bgp-bmp.yang

vvlakshmanamurthy:
Creating BGP BMP yang OC path definitions

### Change Scope

### Platform Implementations

   implementation output.
   implementation output.

### Tree view

```diff
         |     +--rw bgp
+        |     |  +--rw oc-bgp-bmp:bmp
+        |     |     +--rw oc-bgp-bmp:stations
+        |     |        +--rw oc-bgp-bmp:station* [name]
+        |     |           +--rw oc-bgp-bmp:name                -> ../config/name
+        |     |           +--rw oc-bgp-bmp:config
+        |     |           |  +--rw oc-bgp-bmp:name?                       string
+        |     |           |  +--rw oc-bgp-bmp:enabled?                    boolean
+        |     |           |  +--rw oc-bgp-bmp:description?                string
+        |     |           |  +--rw oc-bgp-bmp:station-address?            oc-inet:host
+        |     |           |  +--rw oc-bgp-bmp:station-port?               oc-inet:port-number
+        |     |           |  +--rw oc-bgp-bmp:station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
+        |     |           |  +--rw oc-bgp-bmp:connection-mode?            connection-mode-type
+        |     |           |  +--rw oc-bgp-bmp:source-address?             union
+        |     |           |  +--rw oc-bgp-bmp:reconnect-interval?         uint32
+        |     |           |  +--rw oc-bgp-bmp:statistics-enabled?         boolean
+        |     |           +--ro oc-bgp-bmp:state
+        |     |           |  +--ro oc-bgp-bmp:name?                       string
+        |     |           |  +--ro oc-bgp-bmp:enabled?                    boolean
+        |     |           |  +--ro oc-bgp-bmp:description?                string
+        |     |           |  +--ro oc-bgp-bmp:station-address?            oc-inet:host
+        |     |           |  +--ro oc-bgp-bmp:station-port?               oc-inet:port-number
+        |     |           |  +--ro oc-bgp-bmp:station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
+        |     |           |  +--ro oc-bgp-bmp:connection-mode?            connection-mode-type
+        |     |           |  +--ro oc-bgp-bmp:source-address?             union
+        |     |           |  +--ro oc-bgp-bmp:reconnect-interval?         uint32
+        |     |           |  +--ro oc-bgp-bmp:statistics-enabled?         boolean
+        |     |           +--rw oc-bgp-bmp:tcp-keepalive
+        |     |           |  +--rw oc-bgp-bmp:config
+        |     |           |  |  +--rw oc-bgp-bmp:idle-time?            uint32
+        |     |           |  |  +--rw oc-bgp-bmp:probe-count?          uint16
+        |     |           |  |  +--rw oc-bgp-bmp:probe-interval?       uint16
+        |     |           |  |  +--rw oc-bgp-bmp:keepalive-interval?   uint16
+        |     |           |  +--ro oc-bgp-bmp:state
+        |     |           |     +--ro oc-bgp-bmp:idle-time?            uint32
+        |     |           |     +--ro oc-bgp-bmp:probe-count?          uint16
+        |     |           |     +--ro oc-bgp-bmp:probe-interval?       uint16
+        |     |           |     +--ro oc-bgp-bmp:keepalive-interval?   uint16
+        |     |           +--rw oc-bgp-bmp:route-monitoring
+        |     |              +--rw oc-bgp-bmp:config
+        |     |              |  +--rw oc-bgp-bmp:exclude-non-eligible?   boolean
+        |     |              |  +--rw oc-bgp-bmp:exclude-non-feasible?   boolean
+        |     |              +--rw oc-bgp-bmp:ribs
+        |     |                 +--rw oc-bgp-bmp:rib* [rib-type]
+        |     |                    +--rw oc-bgp-bmp:rib-type     -> ../config/rib-type
+        |     |                    +--rw oc-bgp-bmp:config
+        |     |                    |  +--rw oc-bgp-bmp:rib-type?   identityref
+        |     |                    |  +--rw oc-bgp-bmp:enabled?    boolean
+        |     |                    +--ro oc-bgp-bmp:state
+        |     |                    |  +--ro oc-bgp-bmp:rib-type?   identityref
+        |     |                    |  +--ro oc-bgp-bmp:enabled?    boolean
+        |     |                    +--rw oc-bgp-bmp:afi-safis
+        |     |                       +--rw oc-bgp-bmp:afi-safi* [name]
+        |     |                          +--rw oc-bgp-bmp:name      -> ../config/name
+        |     |                          +--rw oc-bgp-bmp:config
+        |     |                          |  +--rw oc-bgp-bmp:name?      identityref
+        |     |                          |  +--rw oc-bgp-bmp:enabled?   boolean
+        |     |                          +--ro oc-bgp-bmp:state
+        |     |                             +--ro oc-bgp-bmp:name?      identityref
+        |     |                             +--ro oc-bgp-bmp:enabled?   boolean

[ ... ]

         |     |  +--rw global

[...]

+        |     |  |  +--rw oc-bgp-bmp:bmp
+        |     |  |     +--rw oc-bgp-bmp:config
+        |     |  |     |  +--rw oc-bgp-bmp:enabled?                    boolean
+        |     |  |     |  +--rw oc-bgp-bmp:statistics-interval?        uint32
+        |     |  |     |  +--rw oc-bgp-bmp:dscp?                       uint8
+        |     |  |     |  +--rw oc-bgp-bmp:station-port?               oc-inet:port-number
+        |     |  |     |  +--rw oc-bgp-bmp:station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
+        |     |  |     +--ro oc-bgp-bmp:state
+        |     |  |     |  +--ro oc-bgp-bmp:enabled?                    boolean
+        |     |  |     |  +--ro oc-bgp-bmp:statistics-interval?        uint32
+        |     |  |     |  +--ro oc-bgp-bmp:dscp?                       uint8
+        |     |  |     |  +--ro oc-bgp-bmp:station-port?               oc-inet:port-number
+        |     |  |     |  +--ro oc-bgp-bmp:station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
+        |     |  |     +--rw oc-bgp-bmp:route-monitoring
+        |     |  |     |  +--rw oc-bgp-bmp:ribs
+        |     |  |     |     +--rw oc-bgp-bmp:rib* [rib-type]
+        |     |  |     |        +--rw oc-bgp-bmp:rib-type     -> ../config/rib-type
+        |     |  |     |        +--rw oc-bgp-bmp:config
+        |     |  |     |        |  +--rw oc-bgp-bmp:rib-type?   identityref
+        |     |  |     |        |  +--rw oc-bgp-bmp:enabled?    boolean
+        |     |  |     |        +--ro oc-bgp-bmp:state
+        |     |  |     |        |  +--ro oc-bgp-bmp:rib-type?   identityref
+        |     |  |     |        |  +--ro oc-bgp-bmp:enabled?    boolean
+        |     |  |     |        +--rw oc-bgp-bmp:afi-safis
+        |     |  |     |           +--rw oc-bgp-bmp:afi-safi* [name]
+        |     |  |     |              +--rw oc-bgp-bmp:name      -> ../config/name
+        |     |  |     |              +--rw oc-bgp-bmp:config
+        |     |  |     |              |  +--rw oc-bgp-bmp:name?      identityref
+        |     |  |     |              |  +--rw oc-bgp-bmp:enabled?   boolean
+        |     |  |     |              +--ro oc-bgp-bmp:state
+        |     |  |     |                 +--ro oc-bgp-bmp:name?      identityref
+        |     |  |     |                 +--ro oc-bgp-bmp:enabled?   boolean
         |     |  +--rw neighbors
         |     |  |  +--rw neighbor* [neighbor-address]

[ ... ]

+        |     |  |     +--rw oc-bgp-bmp:bmp
+        |     |  |        +--rw oc-bgp-bmp:config
+        |     |  |        |  +--rw oc-bgp-bmp:enabled?                    boolean
+        |     |  |        |  +--rw oc-bgp-bmp:statistics-interval?        uint32
+        |     |  |        |  +--rw oc-bgp-bmp:dscp?                       uint8
+        |     |  |        |  +--rw oc-bgp-bmp:station-port?               oc-inet:port-number
+        |     |  |        |  +--rw oc-bgp-bmp:station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
+        |     |  |        +--ro oc-bgp-bmp:state
+        |     |  |        |  +--ro oc-bgp-bmp:enabled?                    boolean
+        |     |  |        |  +--ro oc-bgp-bmp:statistics-interval?        uint32
+        |     |  |        |  +--ro oc-bgp-bmp:dscp?                       uint8
+        |     |  |        |  +--ro oc-bgp-bmp:station-port?               oc-inet:port-number
+        |     |  |        |  +--ro oc-bgp-bmp:station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
+        |     |  |        +--rw oc-bgp-bmp:route-monitoring
+        |     |  |        |  +--rw oc-bgp-bmp:ribs
+        |     |  |        |     +--rw oc-bgp-bmp:rib* [rib-type]
+        |     |  |        |        +--rw oc-bgp-bmp:rib-type     -> ../config/rib-type
+        |     |  |        |        +--rw oc-bgp-bmp:config
+        |     |  |        |        |  +--rw oc-bgp-bmp:rib-type?   identityref
+        |     |  |        |        |  +--rw oc-bgp-bmp:enabled?    boolean
+        |     |  |        |        +--ro oc-bgp-bmp:state
+        |     |  |        |        |  +--ro oc-bgp-bmp:rib-type?   identityref
+        |     |  |        |        |  +--ro oc-bgp-bmp:enabled?    boolean
+        |     |  |        |        +--rw oc-bgp-bmp:afi-safis
+        |     |  |        |           +--rw oc-bgp-bmp:afi-safi* [name]
+        |     |  |        |              +--rw oc-bgp-bmp:name      -> ../config/name
+        |     |  |        |              +--rw oc-bgp-bmp:config
+        |     |  |        |              |  +--rw oc-bgp-bmp:name?      identityref
+        |     |  |        |              |  +--rw oc-bgp-bmp:enabled?   boolean
+        |     |  |        |              +--ro oc-bgp-bmp:state
+        |     |  |        |                 +--ro oc-bgp-bmp:name?      identityref
+        |     |  |        |                 +--ro oc-bgp-bmp:enabled?   boolean
         |     |  +--rw peer-groups
         |     |  |  +--rw peer-group* [peer-group-name]

[ ... ]

+        |     |  |     |  +--rw config
+        |     |  |     |  |  +--rw enabled?                       boolean
+        |     |  |     |  |  +--rw desired-minimum-tx-interval?   uint32
+        |     |  |     |  |  +--rw required-minimum-receive?      uint32
+        |     |  |     |  |  +--rw detection-multiplier?          uint8
+        |     |  |     |  +--ro state
+        |     |  |     |     +--ro enabled?                       boolean
+        |     |  |     |     +--ro desired-minimum-tx-interval?   uint32
+        |     |  |     |     +--ro required-minimum-receive?      uint32
+        |     |  |     |     +--ro detection-multiplier?          uint8
+        |     |  |     +--rw oc-bgp-bmp:bmp
+        |     |  |        +--rw oc-bgp-bmp:config
+        |     |  |        |  +--rw oc-bgp-bmp:enabled?                    boolean
+        |     |  |        |  +--rw oc-bgp-bmp:statistics-interval?        uint32
+        |     |  |        |  +--rw oc-bgp-bmp:dscp?                       uint8
+        |     |  |        |  +--rw oc-bgp-bmp:station-port?               oc-inet:port-number
+        |     |  |        |  +--rw oc-bgp-bmp:station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
+        |     |  |        +--ro oc-bgp-bmp:state
+        |     |  |        |  +--ro oc-bgp-bmp:enabled?                    boolean
+        |     |  |        |  +--ro oc-bgp-bmp:statistics-interval?        uint32
+        |     |  |        |  +--ro oc-bgp-bmp:dscp?                       uint8
+        |     |  |        |  +--ro oc-bgp-bmp:station-port?               oc-inet:port-number
+        |     |  |        |  +--ro oc-bgp-bmp:station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
+        |     |  |        +--rw oc-bgp-bmp:route-monitoring
+        |     |  |        |  +--rw oc-bgp-bmp:ribs
+        |     |  |        |     +--rw oc-bgp-bmp:rib* [rib-type]
+        |     |  |        |        +--rw oc-bgp-bmp:rib-type     -> ../config/rib-type
+        |     |  |        |        +--rw oc-bgp-bmp:config
+        |     |  |        |        |  +--rw oc-bgp-bmp:rib-type?   identityref
+        |     |  |        |        |  +--rw oc-bgp-bmp:enabled?    boolean
+        |     |  |        |        +--ro oc-bgp-bmp:state
+        |     |  |        |        |  +--ro oc-bgp-bmp:rib-type?   identityref
+        |     |  |        |        |  +--ro oc-bgp-bmp:enabled?    boolean
+        |     |  |        |        +--rw oc-bgp-bmp:afi-safis
+        |     |  |        |           +--rw oc-bgp-bmp:afi-safi* [name]
+        |     |  |        |              +--rw oc-bgp-bmp:name      -> ../config/name
+        |     |  |        |              +--rw oc-bgp-bmp:config
+        |     |  |        |              |  +--rw oc-bgp-bmp:name?      identityref
+        |     |  |        |              |  +--rw oc-bgp-bmp:enabled?   boolean
+        |     |  |        |              +--ro oc-bgp-bmp:state
+        |     |  |        |                 +--ro oc-bgp-bmp:name?      identityref
+        |     |  |        |                 +--ro oc-bgp-bmp:enabled?   boolean
```


release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+    "This module defines configuration and operational state data for
+    BGP Monitoring Protocol (BMP) as defined in RFC 7854.";
+
+  oc-ext:openconfig-version "0.1.0";
+
+  revision "2025-04-21" {
+    description
+      "Initial revision";
+    reference
+      "RFC 7854: BGP Monitoring Protocol";
```
dplore: You can move the RFC reference to the module description:
https://github.com/openconfig/public/pull/1343/files#diff-cfba4613951c6f0bc189f9dcdf9adfa231718e7665adce8dc8a15fb80b2fe01cR23-R25

And change the reference here to the version # of the module

```suggestion
    reference
      "0.1.0";
```

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+  namespace "http://openconfig.net/yang/bgp-bmp";
+
+  prefix "oc-bgp-bmp";
+
+  import openconfig-bgp { prefix oc-bgp; }
+  import openconfig-bgp-types { prefix oc-bgp-types; }
+  import openconfig-network-instance { prefix oc-netinst; }
+  import openconfig-inet-types { prefix oc-inet; }
+  import openconfig-extensions { prefix oc-ext; }
+  import openconfig-yang-types { prefix oc-yang; }
```
dplore: Run pyang to build the tree and you'll see a series of errors you can correct:

```
 dloher@nettool:~$ python3 -m venv ~/venv
 dloher@nettool:~$ source ~/venv/bin/activate
(venv) dloher@nettool:~$ pip install pyang
(venv) dloher@nettool:~$ pyang -f tree -p release/models/*/* > ~/bmp.tree
```
When the tree looks like what you want, you can add it to the comment in the description for everyone to see for easy review.

Here are the errors so far:
```
release/models/bgp/openconfig-bgp-bmp.yang:9: warning: imported module "openconfig-bgp" not used
release/models/bgp/openconfig-bgp-bmp.yang:10: warning: imported module "openconfig-bgp-types" not used
release/models/bgp/openconfig-bgp-bmp.yang:97: error: prefix "oc-types" is not defined (reported only once)
release/models/bgp/openconfig-bgp-bmp.yang:281: error: unexpected keyword "type"
release/models/bgp/openconfig-bgp-bmp.yang:330: error: "openconfig-bgp-bmp:stations" in the path for station-names at release/models/bgp/openconfig-bgp-bmp.yang:556 (at release/models/bgp/openconfig-bgp-bmp.yang:328) is not found
release/models/bgp/openconfig-bgp-bmp.yang:343: error: "openconfig-bgp-bmp:stations" in the path for station-names at release/models/bgp/openconfig-bgp-bmp.yang:556 (at release/models/bgp/openconfig-bgp-bmp.yang:341) is not found
release/models/bgp/openconfig-bgp-bmp.yang:381: error: grouping "bmp-afi-safis-top" not found in module "openconfig-bgp-bmp"
release/models/bgp/openconfig-bgp-bmp.yang:446: error: grouping name "bmp-config" is already defined at release/models/bgp/openconfig-bgp-bmp.yang:413
release/models/bgp/openconfig-bgp-bmp.yang:496: error: grouping "bmp-state" not found in module "openconfig-bgp-bmp"
release/models/bgp/openconfig-bgp-bmp.yang:501: error: grouping "bmp-afi-safis-top" not found in module "openconfig-bgp-bmp"
release/models/bgp/openconfig-bgp-bmp.yang:548: error: grouping "bmp-afi-safis-top" not found in module "openconfig-bgp-bmp"
```

dplore (commented):
Please add a `.spec.yml` file to ensure our CI tools and website will check and build this model.  For example:
https://github.com/openconfig/public/blob/master/release/models/acl/.spec.yml

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+    }
+  }
+
+  grouping bmp-route-monitoring-config {
+    description
+      "Configuration parameters for BMP route monitoring";
+
+    leaf policy-type {
+      type route-monitoring-policy-type;
+      default "BOTH";
```
dplore: Recommend removing this default.  We should avoid setting defaults unless required by RFC.

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+        "The maximum number of TCP keepalive probes to send before
+        declaring the connection dead";
+    }
+
+    leaf probe-interval {
+      type oc-types:timeticks64;
+      description
+        "The time in seconds between TCP keepalive probes";
+    }
+  }
```
dplore: This grouping is redundant to the -config grouping and should be deleted.  OC commonly reuses the config to represent state.

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+        description
+          "Configuration parameters for BMP TCP keepalive";
+        uses bmp-tcp-keepalive-config;
+      }
+
+      container state {
+        config false;
+        description
+          "Operational state parameters for BMP TCP keepalive";
+        uses bmp-tcp-keepalive-state;
```
dplore: Let's reuse the config grouping here to follow common OC style

```suggestion
        uses bmp-tcp-keepalive-config;
```

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+      type route-monitoring-policy-type;
+      default "BOTH";
+      description
+        "Specifies whether to send pre-policy, post-policy, or both
+        types of route monitoring messages";
+    }
+
+    leaf exclude-non-eligible {
+      type boolean;
+      default false;
```
dplore, line 117: please remove the default

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    leaf exclude-non-eligible {
+      type boolean;
+      description
+        "When true, exclude routes that are not eligible for
+        installation in the routing table from route monitoring
+        messages";
+    }
+  }
```
dplore: as above, this grouping is redundant to -config and can be deleted.

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+        description
+          "Configuration parameters for BMP route monitoring";
+        uses bmp-route-monitoring-config;
+      }
+
+      container state {
+        config false;
+        description
+          "Operational state parameters for BMP route monitoring";
+        uses bmp-route-monitoring-state;
```
dplore:
```suggestion
        uses bmp-route-monitoring-config;
```

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+      }
+
+      leaf route-mirroring {
+        type oc-yang:counter64;
+        description
+          "Number of route mirroring messages sent";
+      }
+    }
+
+    uses bmp-route-monitoring-state;
```
dplore:
```suggestion
    uses bmp-route-monitoring-config;
```

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    leaf statistics-timeout {
+      type oc-types:timeticks64;
+      description
+        "Time in seconds between periodic statistics messages";
+    }
+
+    uses bmp-tcp-keepalive-state;
+  }
```
dplore: this is duplicate and can be deleted

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+        description
+          "Configuration parameters for BGP Monitoring Protocol";
+        uses bmp-config;
+      }
+
+      container state {
+        config false;
+        description
+          "Operational state parameters for BGP Monitoring Protocol";
+        uses bmp-state;
```
dplore:
```suggestion
        uses bmp-config;
```

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+
+    leaf connection-mode {
+      type connection-mode-type;
+      default "ACTIVE";
+      description
+        "Specifies whether the BMP client actively initiates connections
+        to BMP servers or passively listens for connections";
+    }
+
+    leaf source-address {
```
dplore:
```suggestion
    leaf local-address {
```

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+    }
+  }
+
+  grouping bmp-neighbor-config {
+    description
+      "Configuration parameters for BMP at the neighbor level";
+
+    leaf enabled {
+      type boolean;
+      default false;
```
dplore, line 485: remove default

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+  grouping bmp-neighbor-state {
+    description
+      "Operational state parameters for BMP at the neighbor level";
+
+    leaf enabled {
+      type boolean;
+      description
+        "When true, BMP monitoring is enabled for this BGP neighbor";
+    }
+  }
```
dplore: remove duplicate grouping

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+        description
+          "Configuration parameters for BMP at the neighbor level";
+        uses bmp-neighbor-config;
+      }
+
+      container state {
+        config false;
+        description
+          "Operational state parameters for BMP at the neighbor level";
+        uses bmp-neighbor-state;
```
dplore:
```suggestion
        uses bmp-neighbor-config;
```

dplore (changes requested):

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+  }
+
+  // Groupings
+
+  grouping bmp-tcp-keepalive-config {
+    description
+      "Configuration parameters for BMP TCP keepalive settings";
+
+    leaf idle-time {
+      type oc-types:timeticks64;
```
earies: wrong datatype which is measured as nanoseconds delta

suggest: proper primitive uint type with a `units seconds;`

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+
+    leaf probe-count {
+      type uint8;
+      description
+        "The maximum number of TCP keepalive probes to send before
+        declaring the connection dead";
+    }
+
+    leaf probe-interval {
+      type oc-types:timeticks64;
```
earies: wrong datatype which is measured as nanoseconds delta

suggest: proper primitive uint type with a `units seconds;`

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+      description
+        "Configuration and state parameters for BMP monitoring stations";
+
+      list station {
+        key "name";
+        description
+          "List of BMP monitoring stations";
+
+        leaf name {
+          type string;
```
earies: List keys in OC design are a leafref to their child `./config/...` node

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+    leaf address {
+      type oc-inet:ip-address;
+      mandatory true;
+      description
+        "IP address of the BMP monitoring station";
+    }
+
+    leaf port {
+      type oc-inet:port-number;
+      default 7039;
```
earies: Where is this default coming from?  There is no IANA registered port - recommend removing - also sounds like you want this to be mandatory as well then

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    leaf connection-status {
+      type connection-status-type;
+      description
+        "Operational status of the BMP session";
+    }
+
+    leaf uptime {
+      type oc-types:timeticks64;
```
earies: wrong datatype which is measured as nanoseconds delta

suggest: proper primitive uint type with a units seconds;

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+      description
+        "Time in seconds that the BMP session has been established";
+    }
+
+    leaf flap-count {
+      type uint32;
```
earies: 2 types defined - suggest colocating counters under a proper `counters` container as you'll see in other domains

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+          "Operational state parameters for BMP at the neighbor level";
+        uses bmp-neighbor-state;
+      }
+
+      uses bmp-afi-safis-top;
+    }
+  }
+
+  // Augmentations
+  augment "/oc-netinst:network-instances/oc-netinst:network-instance/oc-netinst:protocols/oc-netinst:protocol/oc-netinst:bgp/oc-netinst:global" {
```
earies: While obviously tied to BGP, this would be tying it to a BGP protocol instance

Suggest decoupling the BMP global parameters away and also determining the network-instance aspects of stations

earies (commented):
Provided quick first pass - some overall comments

* Suggest looking at and providing reference implementations.  There are various parameters not covered in this and would need to expand to accomodate. e.g. https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/statement/bmp-edit-routing-options.html
* Consider proper anchor points and NI relationships.  Much of the OC modeling is often very loose allowing for combinations that are unrealistic or not supported so best to narrow that in

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+      enum PASSIVE {
+        description
+          "BMP client passively listens for connections from BMP servers";
+      }
+    }
+    description
+      "Type to define the connection mode for BMP sessions";
+  }
+
+  typedef route-monitoring-policy-type {
```
mukulksrivastava: how do we map route-monitoring-policy-type to rib-in and rib-out? In other words, how do we support all 5 RIB views - rib-in-pre-policy, rib-in-post-policy, rib-out-pre-policy, rib-out-post-policy and loc-rib?

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+      enum BOTH {
+        description
+          "Send both pre-policy and post-policy route monitoring messages";
+      }
+    }
+    description
+      "Type to define the route monitoring policy for BMP";
+  }
+
+  typedef connection-status-type {
```
mukulksrivastava: Can we add other state:
1. listening (this would be state if the station is in the passive mode).
2. cleanup (this would be when station goes into cleanup phase due to configuration change or socket disconnection)
3. holddown

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+      "Operational state parameters for BMP route monitoring";
+
+    leaf policy-type {
+      type route-monitoring-policy-type;
+      description
+        "Specifies whether to send pre-policy, post-policy, or both
+        types of route monitoring messages";
+    }
+
+    leaf exclude-non-eligible {
```
mukulksrivastava: This seems to be applicable for rib-in-post-policy case only. Should there be some conditional statement in this yang model to enforce that?

We need to add below clarification here  for this knob:

Routes that have post-policy protocol next-hops that are unresolved or are otherwise unusable are considered to be non-eligible. If using post-policy with exclude-non-eligible, routes that are accepted by import policy but considered non-eligible will be advertised as unreachable.

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+      type boolean;
+      default false;
+      description
+        "When true, exclude routes that are not eligible for
+        installation in the routing table from route monitoring
+        messages";
+    }
+  }
+
+  grouping bmp-route-monitoring-state {
```
mukulksrivastava: Can we consider another leaf node - "exclude-non-feasible". This is applicable only for rib-in-pre-policy. We should add this condition in yang.

Routes that are received but are hidden are considered to be non-feasible.  If using pre-policy exclude-non-feasible, such routes will be advertised by BMP as unreachable.   A route could be hidden because one of the following reason:
•	suppressed by damping
•	cluster loop
•	AS confed path loop
•	AS path loop
•	originator loop
•	invalid optional transitive partial attribute
•	empty path on external peer
•	fails enforce-first-as check
•	invalid confed segment
•	path received out of the domain
•	AS of value zero in AS path
•	AS restricted value (4294967295) in AS path
•	protocol nexthop is local
•	protocol nexthop is not on the interface
•	protocol nexthop is IPv6 link local
•	protocol nexthop is multicast
•	protocol nexthop is martian
•	prefix is martian
•	protocol nexthop is an unsupported address family
•	AS path size overfflow
•	Flow-route fails validation
•	unsupported label stack or label block
•	MPLS not enabled on interface

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+          config false;
+          description
+            "Operational state parameters for the BMP monitoring station";
+          uses bmp-station-state;
+        }
+        uses bmp-afi-safis-top;
+      }
+    }
+
+    container station-groups {
```
mukulksrivastava: What is the use case for this container?

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+      "Configuration parameters for a BMP monitoring station";
+
+    leaf address {
+      type oc-inet:ip-address;
+      mandatory true;
+      description
+        "IP address of the BMP monitoring station";
+    }
+
+    leaf port {
```
mukulksrivastava: This should be mandatory if connection mode is active.

release/models/bgp/openconfig-bgp-bmp.yang
mukulksrivastava: yes we don't need to have a default value for this.

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+    leaf enabled {
+      type boolean;
+      default false;
+      description
+        "When true, BGP Monitoring Protocol is enabled";
+    }
+
+    leaf connection-mode {
+      type connection-mode-type;
+      default "ACTIVE";
```
mukulksrivastava: why do we need default mode as active? We can make this as mandatory configuration.

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+          config false;
+          description
+            "Operational state parameters for the station group";
+          uses bmp-station-group-state;
+        }
+      }
+    }
+  }
+
+  grouping bmp-config {
```
mukulksrivastava: we need "local-port" as well. This shouldn't be present if the connection mode is active. This should be mandatory for passive connection mode.

We also need station-address and station-port here.

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+    leaf statistics-timeout {
+      type oc-types:timeticks64;
+      description
+        "Time in seconds between periodic statistics messages";
+    }
+
+    uses bmp-tcp-keepalive-config;
+  }
+
+  grouping bmp-config {
```
mukulksrivastava: duplicate

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+        uses bmp-route-monitoring-state;
+      }
+    }
+  }
+
+  grouping bmp-station-config {
+    description
+      "Configuration parameters for a BMP monitoring station";
+
+    leaf address {
```
mukulksrivastava: "address" is very generic word. Should be renamed to station-address. This can be part of "bmp-config".

release/models/bgp/openconfig-bgp-bmp.yang
mukulksrivastava: "port" is very generic word. Should be renamed to station-port. This can be part of "bmp-config".

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,570 @@
...
+        uses bmp-state;
+      }
+
+      uses bmp-tcp-keepalive-top;
+      uses bmp-stations-top;
+      uses bmp-afi-safis-top;
+    }
+  }
+
+  grouping bmp-neighbor-config {
```
mukulksrivastava: we need to be able configure monitoring type (rib-in-pre-policy or/and rib-in-post-policy  or/and rib-out-pre-policy  or/and rib-out-post-policy) per neighbor or bgp group level.

Ankur19:
Hi @vvlakshmanamurthy! Looking at the config, `system/` or `system/logging` looks like a better fit for the collector information.

vvlakshmanamurthy (commented):
updating a few comments. few more comments to resolve.

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: Done

vvlakshmanamurthy (commented):
update comments

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: If you have 50 routers all streaming data to a cluster of 3 BMP collectors, you don't want to manually define the route-monitoring-policy-type or idle-time for each one.

Without Groups: You must repeat all 10+ leaves for every station.

With Groups: You define the policy in the group, and the stations simply "join" that group to inherit those settings.

vvlakshmanamurthy (commented):
updated comments

vvlakshmanamurthy (commented):
Only 2 comments left , will need to check with Darren.

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: @dplore To suggest on this

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: I ran this but i dont see any output , still checking.

:~$ python3 -m venv oc-env
:~$ source oc-env/bin/activate
:~$ pip install pyang
:~$ source oc-env/bin/activate
:~$ pyang -f tree -p release/models/*/* > ~/bmp.tree

vvlakshmanamurthy (commented):
updating the config

vvlakshmanamurthy:
**For reference:**  @dplore 

module: openconfig-bgp-bmp
  +--rw bmp
     +--rw config
     |  +--rw enabled?              boolean
     |  +--rw connection-mode       connection-mode-type
     |  +--rw local-address?        oc-inet:ip-address
     |  +--rw local-port?           oc-inet:port-number
     |  +--rw statistics-timeout?   uint32
     +--ro state
     |  +--ro enabled?              boolean
     |  +--ro connection-mode       connection-mode-type
     |  +--ro local-address?        oc-inet:ip-address
     |  +--ro local-port?           oc-inet:port-number
     |  +--ro statistics-timeout?   uint32
     +--rw tcp-keepalive
     |  +--rw config
     |  |  +--rw idle-time?        uint32
     |  |  +--rw probe-count?      uint8
     |  |  +--rw probe-interval?   uint32
     |  +--ro state
     |     +--ro idle-time?        uint32
     |     +--ro probe-count?      uint8
     |     +--ro probe-interval?   uint32
     +--rw stations
     |  +--rw station* [name]
     |     +--rw name         -> ../config/name
     |     +--rw config
     |     |  +--rw name?                     string
     |     |  +--rw station-address           oc-inet:ip-address
     |     |  +--rw station-port?             oc-inet:port-number
     |     |  +--rw policy-type?              route-monitoring-policy-type
     |     |  +--rw exclude-non-eligible?     boolean
     |     |  +--rw exclude-non-feasible?     boolean
     |     +--ro state
     |     |  +--ro name?                     string
     |     |  +--ro station-address           oc-inet:ip-address
     |     |  +--ro station-port?             oc-inet:port-number
     |     |  +--ro policy-type?              route-monitoring-policy-type
     |     |  +--ro exclude-non-eligible?     boolean
     |     |  +--ro exclude-non-feasible?     boolean
     |     |  +--ro connection-status?        connection-status-type
     |     |  +--ro uptime?                   uint64
     |     |  +--ro flap-count?               oc-yang:counter64
     |     |  +--ro message-counters
     |     |     +--ro total?              oc-yang:counter64
     |     |     +--ro tx-statistics?      oc-yang:counter64
     |     |     +--ro route-monitoring?   oc-yang:counter64
     |     |     +--ro peer-monitoring?    oc-yang:counter64
     |     |     +--ro route-mirroring?    oc-yang:counter64
     |     +--rw afi-safis
     |        +--rw afi-safi* [afi-safi-name]
     |           +--rw afi-safi-name    -> ../config/afi-safi-name
     |           +--rw config
     |           |  +--rw afi-safi-name?   identityref
     |           |  +--rw enabled?         boolean
     |           +--ro state
     |              +--ro afi-safi-name?   identityref
     |              +--ro enabled?         boolean
     +--rw station-groups
     |  +--rw station-group* [name]
     |     +--rw name      string
     |     +--rw config
     |     |  +--rw station-names*   -> ../../../stations/station/name
     |     +--ro state
     |        +--ro station-names*   -> ../../../stations/station/name
     +--rw afi-safis
           +--rw afi-safi-name    -> ../config/afi-safi-name
           +--rw config
           |  +--rw afi-safi-name?   identityref
           |  +--rw enabled?         boolean
           +--ro state
              +--ro afi-safi-name?   identityref
              +--ro enabled?         boolean

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: Latest: 

module: openconfig-bgp-bmp
  +--rw bmp
     +--rw config
     |  +--rw enabled?              boolean
     |  +--rw connection-mode       connection-mode-type
     |  +--rw local-address?        oc-inet:ip-address
     |  +--rw local-port?           oc-inet:port-number
     |  +--rw statistics-timeout?   uint32
     +--ro state
     |  +--ro enabled?              boolean
     |  +--ro connection-mode       connection-mode-type
     |  +--ro local-address?        oc-inet:ip-address
     |  +--ro local-port?           oc-inet:port-number
     |  +--ro statistics-timeout?   uint32
     +--rw tcp-keepalive
     |  +--rw config
     |  |  +--rw idle-time?        uint32
     |  |  +--rw probe-count?      uint8
     |  |  +--rw probe-interval?   uint32
     |  +--ro state
     |     +--ro idle-time?        uint32
     |     +--ro probe-count?      uint8
     |     +--ro probe-interval?   uint32
     +--rw stations
     |  +--rw station* [name]
     |     +--rw name         -> ../config/name
     |     +--rw config
     |     |  +--rw name?                     string
     |     |  +--rw station-address           oc-inet:ip-address
     |     |  +--rw station-port?             oc-inet:port-number
     |     |  +--rw policy-type?              route-monitoring-policy-type
     |     |  +--rw exclude-non-eligible?     boolean
     |     |  +--rw exclude-non-feasible?     boolean
     |     +--ro state
     |     |  +--ro name?                     string
     |     |  +--ro station-address           oc-inet:ip-address
     |     |  +--ro station-port?             oc-inet:port-number
     |     |  +--ro policy-type?              route-monitoring-policy-type
     |     |  +--ro exclude-non-eligible?     boolean
     |     |  +--ro exclude-non-feasible?     boolean
     |     |  +--ro connection-status?        connection-status-type
     |     |  +--ro uptime?                   uint64
     |     |  +--ro flap-count?               oc-yang:counter64
     |     |  +--ro message-counters
     |     |     +--ro total?              oc-yang:counter64
     |     |     +--ro tx-statistics?      oc-yang:counter64
     |     |     +--ro route-monitoring?   oc-yang:counter64
     |     |     +--ro peer-monitoring?    oc-yang:counter64
     |     |     +--ro route-mirroring?    oc-yang:counter64
     |     +--rw afi-safis
     |        +--rw afi-safi* [afi-safi-name]
     |           +--rw afi-safi-name    -> ../config/afi-safi-name
     |           +--rw config
     |           |  +--rw afi-safi-name?   identityref
     |           |  +--rw enabled?         boolean
     |           +--ro state
     |              +--ro afi-safi-name?   identityref
     |              +--ro enabled?         boolean
     +--rw station-groups
     |  +--rw station-group* [name]
     |     +--rw name      string
     |     +--rw config
     |     |  +--rw station-names*   -> ../../../stations/station/name
     |     +--ro state
     |        +--ro station-names*   -> ../../../stations/station/name
     +--rw afi-safis
           +--rw afi-safi-name    -> ../config/afi-safi-name
           +--rw config
           |  +--rw afi-safi-name?   identityref
           |  +--rw enabled?         boolean
           +--ro state
              +--ro afi-safi-name?   identityref
              +--ro enabled?         boolean

vvlakshmanamurthy (commented):
Updated bgp.spec.yml

vvlakshmanamurthy (commented):
added .spec.yml

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,275 @@
...
+          }
+        }
+        container state {
+          config false;
+          leaf-list station-names {
+            type leafref { path "../../../stations/station/name"; }
+          }
+        }
+      }
+    }
```
dplore:
```suggestion
      list station-group {
        key "name";
        leaf name {
          type leafref {
            path "../config/name";
          }
          description
            "references the station-group name key";
        }
        container config {
          leaf name {
             type string;
             description "Name of the station-group."
          }
          leaf-list station-names {
            type leafref { path "../../../stations/station/name"; }
          }
        }
        container state {
          config false;
          leaf-list station-names {
            type leafref { path "../../../stations/station/name"; }
          }
        }
      }
    }
```

vvlakshmanamurthy (commented):
Updated the BMP oc yang as per discussing with the various vollaborators.

vvlakshmanamurthy (commented):
correction

vvlakshmanamurthy (commented):
corrections

vvlakshmanamurthy (commented):
corrections

vvlakshmanamurthy (commented):
correction

vvlakshmanamurthy (commented):
correction

vvlakshmanamurthy (commented):
corrections

vvlakshmanamurthy (commented):
corrections

vvlakshmanamurthy (commented):
corrected trailing white spaces

vvlakshmanamurthy (commented):
updated

vvlakshmanamurthy (commented):
commit

vvlakshmanamurthy (commented):
corrected

vvlakshmanamurthy (commented):
updated few more corrections.

dplore:
Fixed imports, removed duplicate nodes, fixed augments, created new tree view in PR description.

dplore:
Will review at May 12, 2026 OC Operators meeting

vvlakshmanamurthy (commented):
Updated copy

dplore:
Moved stations to only appear at the top level for bgp:

ie: `/network-instances/network-instance/protocols/protocol/bgp/bmp/stations`

release/models/bgp/openconfig-bgp-bmp.yang
dplore: pushed a bunch of updates for us.

vvlakshmanamurthy (commented):
updates

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,420 @@
...
+        "This is optional config at global/system level,the same can be defined at station level as well";
+    }
+    leaf dscp {
+      type uint8;
+    }
+    leaf station-port {
+      type oc-inet:port-number;
+      description
+        "Thi is optional config at global/system level, the same can be defined at station level as well";
+    }
```
dplore, line 370: This is redundant, but seems intentional?    What do you do if the station has a port defined and then a different port is also defined at the global/peer/neighbor level?   Which port do you use?   Do we really need this flexibility and do vendors already support something like this?

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,420 @@
...
+    leaf station-network-instance {
+      type leafref {
+        path "/oc-netinst:network-instances/oc-netinst:network-instance/oc-netinst:name";
+      }
+      description
+        "Thi is optional config at global/system level, the same can be defined at station level as well";
+    }
+  }
```
dplore, line 378: This is redundant, but seems intentional?    What do you do if the station has a network-instance defined and then a different network-instance is also defined at the global/peer/neighbor level?   Which network-instance do you use?   Do we really need this flexibility and do vendors already support something like this?

release/models/bgp/openconfig-bgp-bmp.yang
dplore, line 1: How are the stations that are defined, linked to a global/peer/neighbor definition?  Is it implied that all stations get streamed to?  Or do we need a list of station names to be used in each global/peer/neighbor definition?

dplore (commented):
@vvlakshmanamurthy few comments to respond to please.  Thanks!

dplore:
> ### Platform Implementations
> * Implementation A: [link to documentation](http://foo.com) and/or
>   implementation output.
> * Implementation B: [link to documentation](http://foo.com) and/or
>   implementation output.

Please add at least two references to implementations of BMP

ElodinLaarz:
@vvlakshmanamurthy -- Some open comments to address, could you take a look?

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,420 @@
...
+    leaf rib-type {
+      type identityref {
+        base oc-bgp-types:AFI_SAFI_TYPE;
+      }
+      description
+        "The AFI-SAFI name for this RIB.";
+    }
+    leaf enabled {
+      type boolean;
+      default false;
```
rolandphung, line 147: Please remove the default value here too.

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,420 @@
...
+    }
+  }
+
+// Augmentations
+  augment "/oc-netinst:network-instances/oc-netinst:network-instance/" +
+        "oc-netinst:protocols/oc-netinst:protocol/oc-netinst:bgp" {
+    description
+      "Augments the global BGP configuration with BMP parameters";
+    uses station-top;
+  }
```
rolandphung, line 402: Please consider to add this in the root level as it doesn't make sense to put the station config in the network-instance level. If keyed by NI means we are intentionally separating the configuration into multiple lists of stations all the while it must maintain uniqueness across these lists. If the NI key is meaningless then it would end up in the default instance NI which could be be a top level container instead.

release/models/bgp/openconfig-bgp-bmp.yang
```diff
@@ -0,0 +1,420 @@
...
+    }
+  }
+
+  grouping bmp-top {
+    description
+      "Top-level container for BMP configuration and operational state augmented into openconfig-system.";
+    container bmp {
+
+      container config {
+        uses bmp-system-config;
```
rolandphung, line 386: I think `bmp-system-config` belongs in `station-top` grouping under `container bmp` because these are configurations for the stations themselves, not how the stations interact with BGP.

Separately, this config container should contain `leaf enabled` to enable/disable BMP per network-instance/peer-group/neighbor.

dplore:
@vvlakshmanamurthy there are still a lot of open comments on this PR. I suggest answering these two broader questions first:

1.   I looks like  `/bmp` or `/system/bmp`  should be a better fit for the collector information.  This is because collectors are not defined per network instance.  

2. repetition of the bmp-station parameters per global,peer-group and neighbor seem unnecessary.  Is this how implementions expose this configuration today?  If yes, please provide at least two references for this.

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: All stations should be streamed

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: default is false

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: If port is defined the port should be honored and the same should be honored for respective hierarchies. 
When a session is initiated this is the destination port to be used. and yes we need it in all levels.

Example  config:
 
connection mode active port 1234

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: This was suggested by a vendor and hence we appended , can you please suggest if you want this to be changed ?

release/models/bgp/openconfig-bgp-bmp.yang
vvlakshmanamurthy: done , updated the leaf

vvlakshmanamurthy (commented):
updated the comments

vvlakshmanamurthy (commented):
updated

vvlakshmanamurthy:
@dplore 

1. It looks like /bmp or /system/bmp should be a better fit for the collector information. This is because collectors are not defined per network instance.

**Reply:** As per the discussion we had , was told to use /bgp as the BMP is part of it. hence created the below:
  
       release/models/bgp/.spec.yml
       release/models/bgp/openconfig-bgp-bmp.yang

2. repetition of the bmp-station parameters per global,peer-group and neighbor seem unnecessary. Is this how implementions expose this configuration today? 
If yes, please provide at least two references for this.

**Reply:** I don not have an example for bmp but have for bgp if needed, but we did discuss this in previous calls , the recommendation i got was that more specific will take precendence. This is similar to the route-map or routing policies under BGP.
       Where neigbor level is most preferred , next comes group level and next global level. This is good to have feature and shouldnt break any.  Vendors are already doing this in BGP for routing policies.

release/models/bgp/openconfig-bgp-bmp.yang
rolandphung: This is required if we follow the logic of moving the stations out of the network-instance tree.

rolandphung: Sorry, I spoke too soon, this is part of the `/bmp/config/` and `bmp-system-config` grouping which is not part of the station configuration. This is not supposed to be here.

vvlakshmanamurthy:
New OC Tree @dplore 
```json
module: openconfig-bgp-bmp
  +--rw bmp
     +--rw config
     |  +--rw statistics-interval?   uint32
     |  +--rw dscp?                  uint8
     |  +--rw local-port?            oc-inet:port-number
     +--ro state
     |  +--ro enabled?               boolean
     |  +--ro statistics-interval?   uint32
     |  +--ro dscp?                  uint8
     |  +--ro local-port?            oc-inet:port-number
     +--rw stations
        +--rw station* [name]
           +--rw name                -> ../config/name
           +--rw config
           |  +--rw name?                       string
           |  +--rw enabled?                    boolean
           |  +--rw description?                string
           |  +--rw station-address?            oc-inet:host
           |  +--rw station-port?               oc-inet:port-number
           |  +--rw station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
           |  +--rw connection-mode?            connection-mode-type
           |  +--rw source-address?             union
           |  +--rw reconnect-interval?         uint32
           |  +--rw statistics-enabled?         boolean
           +--ro state
           |  +--ro name?                       string
           |  +--ro enabled?                    boolean
           |  +--ro description?                string
           |  +--ro station-address?            oc-inet:host
           |  +--ro station-port?               oc-inet:port-number
           |  +--ro station-network-instance?   -> /oc-netinst:network-instances/network-instance/name
           |  +--ro connection-mode?            connection-mode-type
           |  +--ro source-address?             union
           |  +--ro reconnect-interval?         uint32
           |  +--ro statistics-enabled?         boolean
           +--rw tcp-keepalive
           |  +--rw config
           |  |  +--rw idle-time?            uint32
           |  |  +--rw probe-count?          uint16
           |  |  +--rw probe-interval?       uint16
           |  |  +--rw keepalive-interval?   uint16
           |  +--ro state
           |     +--ro idle-time?            uint32
           |     +--ro probe-count?          uint16
           |     +--ro probe-interval?       uint16
           |     +--ro keepalive-interval?   uint16
           +--rw route-monitoring
              +--rw config
              |  +--rw exclude-non-eligible?   boolean
              |  +--rw exclude-non-feasible?   boolean
              +--ro state
              |  +--ro exclude-non-eligible?   boolean
              |  +--ro exclude-non-feasible?   boolean
              +--rw ribs
                 +--rw rib* [rib-type]
                    +--rw rib-type     -> ../config/rib-type
                    +--rw config
                    |  +--rw rib-type?   identityref
                    |  +--rw enabled?    boolean
                    +--ro state
                    |  +--ro rib-type?   identityref
                    |  +--ro enabled?    boolean
                    +--rw afi-safis
                       +--rw afi-safi* [name]
                          +--rw config
                          |  +--rw name?      identityref
                          |  +--rw enabled?   boolean
                          +--ro state
                             +--ro name?      identityref
                             +--ro enabled?   boolean

  augment /oc-netinst:network-instances/oc-netinst:network-instance/oc-netinst:protocols/oc-netinst:protocol/oc-netinst:bgp/oc-netinst:global:
    +--rw bmp
       +--rw config
       |  +--rw enabled?   boolean
       +--ro state
       |  +--ro enabled?   boolean
       +--rw route-monitoring
          +--rw ribs
             +--rw rib* [rib-type]
                +--rw rib-type     -> ../config/rib-type
                +--rw config
                |  +--rw rib-type?   identityref
                |  +--rw enabled?    boolean
                +--ro state
                |  +--ro rib-type?   identityref
                |  +--ro enabled?    boolean
                +--rw afi-safis
                   +--rw afi-safi* [name]
                      +--rw config
                      |  +--rw name?      identityref
                      |  +--rw enabled?   boolean
                      +--ro state
                         +--ro name?      identityref
                         +--ro enabled?   boolean
  augment /oc-netinst:network-instances/oc-netinst:network-instance/oc-netinst:protocols/oc-netinst:protocol/oc-netinst:bgp/oc-netinst:neighbors/oc-netinst:neighbor:
    +--rw bmp
       +--rw config
       |  +--rw enabled?   boolean
       +--ro state
       |  +--ro enabled?   boolean
       +--rw route-monitoring
          +--rw ribs
             +--rw rib* [rib-type]
                +--rw rib-type     -> ../config/rib-type
                +--rw config
                |  +--rw rib-type?   identityref
                |  +--rw enabled?    boolean
                +--ro state
                |  +--ro rib-type?   identityref
                |  +--ro enabled?    boolean
                +--rw afi-safis
                   +--rw afi-safi* [name]
                      +--rw config
                      |  +--rw name?      identityref
                      |  +--rw enabled?   boolean
                      +--ro state
                         +--ro name?      identityref
                         +--ro enabled?   boolean
  augment /oc-netinst:network-instances/oc-netinst:network-instance/oc-netinst:protocols/oc-netinst:protocol/oc-netinst:bgp/oc-netinst:peer-groups/oc-netinst:peer-group:
    +--rw bmp
       +--rw config
       |  +--rw enabled?   boolean
       +--ro state
       |  +--ro enabled?   boolean
       +--rw route-monitoring
          +--rw ribs
             +--rw rib* [rib-type]
                +--rw rib-type     -> ../config/rib-type
                +--rw config
                |  +--rw rib-type?   identityref
                |  +--rw enabled?    boolean
                +--ro state
                |  +--ro rib-type?   identityref
                |  +--ro enabled?    boolean
                +--rw afi-safis
                   +--rw afi-safi* [name]
                      +--rw config
                      |  +--rw name?      identityref
                      |  +--rw enabled?   boolean
                      +--ro state
                         +--ro name?      identityref
                         +--ro enabled?   boolean

```

vvlakshmanamurthy (commented):
@darren to take a look as we want to keep the BMP OC simple at this time and then later append as needed as we are getting too many comments on changes. 

I also updated the OC tree in the comment section of this PR#.

dplore:
After discussion with @vvlakshmanamurthy , I understand the desire is to only have the monitoring station inside the same network instance as the bgp session being monitored.  The question to the community is, is this a common, or even defacto standard for deployment?   

If we model this way, and we want to expand to allow other instances in the future, we should implement the model in a way that allows that without a breaking change.

rolandphung:
> After discussion with @vvlakshmanamurthy , I understand the desire is to only have the monitoring station inside the same network instance as the bgp session being monitored. The question to the community is, is this a common, or even defacto standard for deployment?
> 
> If we model this way, and we want to expand to allow other instances in the future, we should implement the model in a way that allows that without a breaking change.

My input is that there are scenarios where the NI scale is very high and it's not very practical (and expensive) to create 1k BMP stations to map to each NI. Most deployments I've seen connects to the MGMT NI in order to monitor the device, if you need inband connections just for monitoring, wouldn't it complicate things?

Also from https://www.rfc-editor.org/info/rfc7854/#section-3.3
```
3.3.  Lifecycle of a BMP Session
   A router is configured to speak BMP to one or more monitoring
   stations.  It MAY be configured to send monitoring information for
   only a subset of its BGP peers.  Otherwise, all BGP peers are assumed
   to be monitored.
```

rolandphung:
That being said, I looked over the model and see that we don't have a way to filter based on the network instance per station or globally. Could we consider to have NI selection/filter be added to the station itself? Probably as a filter though because of https://www.rfc-editor.org/info/rfc7854/#section-3.3 `Otherwise, all BGP peers are assumed to be monitored.`

This could be done in a separate PR and it should be backwards compatible.

PR #1346, opened 2025-08-05
Faults

mrevang:
### Change Scope

This proposal introduces a list of faults under a component's healthz container.

A fault indicates that a component has recently or is currently experiencing a problem as expressed by a symptom. These symptoms have underlying conditions that are platform-specific. The fault should not be deleted as soon as the underlying condition is no longer asserted to avoid excessive fault creation and deletion. Deletion of the fault should occur after a suitable period outside of the triggering condition

Nested under each fault is a list of actions recommended to remediate the fault condition.

```
        +--rw oc-platform-healthz:healthz
        |  +--rw oc-platform-healthz-fault:faults <<<<<<<<<<<<<<<<<<<<<<<
        |     +--ro oc-platform-healthz-fault:fault* [symptom]
        |        +--ro oc-platform-healthz-fault:symptom         -> ../state/symptom
        |        +--ro oc-platform-healthz-fault:state
        |        |  +--ro oc-platform-healthz-fault:symptom?               identityref
        |        |  +--ro oc-platform-healthz-fault:origin-time?           oc-types:timeticks64
        |        |  +--ro oc-platform-healthz-fault:last-detection-time?   oc-types:timeticks64
        |        |  +--ro oc-platform-healthz-fault:description?           string
        |        |  +--ro oc-platform-healthz-fault:status?                enumeration
        |        |  +--ro oc-platform-healthz-fault:counters
        |        |     +--ro oc-platform-healthz-fault:occurrences?   oc-yang:counter64
        |        +--ro oc-platform-healthz-fault:remediations
        |           +--ro oc-platform-healthz-fault:remediation* [index]
        |              +--ro oc-platform-healthz-fault:index    -> ../state/index
        |              +--ro oc-platform-healthz-fault:state
        |                 +--ro oc-platform-healthz-fault:index?    uint64
        |                 +--ro oc-platform-healthz-fault:action?   identityref
        |                 +--ro oc-platform-healthz-fault:target?   -> /oc-platform:components/component/name

```

### Platform Implementations

We are working with a few vendors offline to support this model directly. Here are some publicly available references that show devices reporting faults:
* https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/bsm/configuration/15-2mt/bsm-troubleshooting.html ?
* https://www.juniper.net/documentation/us/en/software/junos/cli-reference/topics/ref/command/show-system-errors.html

### DIFF
--- a/doc/component_fault_overview.md
+++ b/doc/component_fault_overview.md
@@ -10,7 +10,7 @@
 
 This proposal introduces a list of faults under a component's healthz container.
 
-A fault indicates that a component has recently or is currently experiencing a problem as expressed by a symptom. These symptoms have underlying conditions that are platform-specific. The fault should not be deleted as soon as the underlying condition is no longer asserted to avoid excessive fault creation and deletion. Deletion of the fault should occur after a suitable period outside of the triggering condition
+A fault indicates that a component has recently or is currently experiencing a problem as expressed by a symptom. These symptoms have underlying conditions that are platform-specific. The fault should not be deleted as soon as the underlying condition is no longer asserted to avoid excessive fault creation and deletion. Deletion of the fault should occur after a suitable period outside of the triggering condition. The `status` leaf should be updated to `INACTIVE` to indicate that the fault has entered this cooldown period.
 
 Nested under each fault is a list of actions recommended to remediate the fault condition.
 


release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,301 @@
...
+  identity SYMPTOM_PHYSICALLY_MISSING_COMPONENT {
+    base SYMPTOM_BASE;
+    description
+      "Component is physically missing";
+  }
+
+  identity SYMPTOM_MISWIRED {
+    base SYMPTOM_BASE;
+    description
+      "Component is miswired";
```
mrevang: Let's add examples.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,301 @@
...
+    description
+      "An unknown symptom";
+  }
+
+  identity ACTION_BASE {
+    description
+      "Base identity for fault symptoms";
+  }
+
+  identity ACTION_NON_IMPACTING_RESEAT {
```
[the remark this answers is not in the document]
mrevang: "Reseat component"

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,301 @@
...
+    base ACTION_BASE;
+    description
+      "The process of turning a device off completely and then turning it back
+      on.";
+  }
+
+  identity ACTION_FACTORY_RESET {
+    base ACTION_BASE;
+    description
+      "Restores a device to its original factory settings.";
```
mrevang, line 133: device->component

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,301 @@
...
+  identity ACTION_IMPACTING_RESEAT {
+    base ACTION_BASE;
+    description
+      "A reseat that does impact operation of the overal device.";
+  }
+
+  identity ACTION_POWER_CYCLE {
+    base ACTION_BASE;
+    description
+      "The process of turning a device off completely and then turning it back
```
mrevang, line 126: device->component

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,301 @@
...
+      description
+        "Enclosing container for a component's active faults.";
+
+      list fault {
+        key "symptom";
+        config false;
+        description
+          "List of active faults, indexed by symptom.
+
+          A fault indicates that a component has recently exhibited a symptom.
```
mrevang: A fault indicates that a component has recently or is currently experiencing a problem as expressed by a symptom. These symptoms have underlying conditions that are platform-specific. The fault should not be deleted as soon as the underlying condition is no longer asserted to avoid excessive fault creation and deletion. Deletion of the fault should occur after a suitable period outside of the triggering condition

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,301 @@
...
+      description
+        "The time at which the fault first became active.
+        This value should not change until the fault is deleted";
+    }
+
+    leaf last-detection-time {
+      type oc-types:timeticks64;
+      description
+        "The time at which the fault's underlying condition was last sampled
+        and confirmed asserted.";
```
mrevang, line 204: Add an MD file to provide examples.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,301 @@
...
+        }
+        description
+          "List of recommended actions to respond to the fault";
+
+        container state {
+          config false;
+          leaf index {
+            type uint64;
+            description
+            "This index indicates the order in which the actions should be
```
mrevang, line 274: The order should progress from least to most disruptive.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,299 @@
...
+      "A fatal fault.";
+  }
+
+  identity SYMPTOM_MEMORY_ERRORS {
+    base SYMPTOM_BASE;
+    description
+      "Memory errors";
+  }
+
+  identity SYMPTOM_TOTAL_MEMORY_ERRORS {
```
mrevang: rm

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  identity SYMPTOM_UNDER_THRESHOLD {
+    base SYMPTOM_BASE;
+    description
+      "Indicates that a sensor is measuring below its lower operating
+      threshold.";
+  }
+
+  identity SYMPTOM_FATAL_FAULT {
```
mrevang: unknown can serve as the kitchen sink

release/models/platform/openconfig-platform-healthz-fault.yang
mrevang: Add bus error?

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,293 @@
...
+      type oc-types:timeticks64;
+      description
+        "The time at which the fault's underlying condition was last sampled
+        and confirmed asserted.";
+    }
+
+    leaf occurrences {
+      type uint64;
+      description
+        "The number of times this fault transitioned into the active state.";
```
ghananigans: is this for the duration of the device's uptime, or since the fault was last surfaced (maybe after previously deasserted and delated?

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,293 @@
...
+              disruptive.";
+          }
+
+          leaf action {
+            description
+              "The specific action to be performed";
+            type identityref {
+              base ACTION_BASE;
+            }
+          }
```
ghananigans, line 304: Should this only be one action? Are there scenarios when multiple actions may be performed to address the fault? e.g. `ACTION_WARM_REBOOT` _or_ `ACTION_COLD_REBOOT`

Alternatively, is there any ordering between actions where performing action A also covers remediations performed by action B?

ghananigans: Disregard, i see that remediation is a list so that covers the multiple-action case.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,293 @@
...
+        and confirmed asserted.";
+    }
+
+    leaf occurrences {
+      type uint64;
+      description
+        "The number of times this fault transitioned into the active state.";
+    }
+
+    leaf description {
```
ghananigans, line 218: IIUC, this leaf is not expected to be "vendor-neutral" for a given issue and may even change for each occurrence. 

Can we document that this is purely for informational purposes and consuming systems should not expect a common behaviour/pattern, even across software releases or hardware platforms.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,293 @@
...
+        base SYMPTOM_BASE;
+      }
+      description
+        "The symptom that raised this fault";
+    }
+
+    leaf origin-time {
+      type oc-types:timeticks64;
+      description
+        "The time at which the fault first became active.
```
ghananigans, line 201: Is this time since boot-time?

ghananigans:
Apologies, i didn't realize this was a draft. I only noticed that `public-approvers` was added as a reviewer.

release/models/platform/openconfig-platform-healthz-fault.yang
mrevang: When the fault was added to the container, the origin-time should never change for a fault. In order for a fault with that same symptom to have a different origin time, the preceding fault would need to be deleted, and a new fault added.

mrevang: I will update the language to disambiguate.

release/models/platform/openconfig-platform-healthz-fault.yang
mrevang: During the life of a fault, its status can oscillate in/out of active. This counter tracks those transitions to active. If the fault is deleted, occurrence counting would start back at zero should the fault be reintroduced.

I'll update the language to clarify the intent.

release/models/platform/openconfig-platform-healthz-fault.yang
mrevang: Will do, thanks for the suggestion.

mrevang:
> Discussed in the OC Operators meeting on August 26th:
> 
> Healthz is _meant_ to simplify the health of the device to remove some of the if / then conditionals on actions that the operator needs to worry about. In its current form, I could see cases where this would 'encourage' a status of "unhealthy" + "minor non-actionable symptom", which seems antithetical to the healthz container.
> 
> The authors will see about changing the language / allowable severities in the container to avoid these types of situations.

I've introduced a new identityref for severity instead of reusing the identityref from alarms. I've also updated the language to indicate that faults should only be reported if the underlying condition rises to the existing "healthz-unhealthy" bar.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,318 @@
...
+    base OPENCONFIG_FAULT_SEVERITY;
+    description
+      "Indicates that a fault requires urgent remediation.";
+  }
+
+  identity NON_URGENT {
+    base OPENCONFIG_FAULT_SEVERITY;
+    description
+      "Indicates that a fault does not require urgent remediation.";
+  }
```
dplore: My thinking is that it is very likely that the network management system will need to determine this according to the an operator's policies for running their network.  So my bias is this should not be included in the model.

If there are some operational scenarios that could be  quantified / measured that can be used to know the urgency then maybe these are justified.  I think we'd need to see some examples and see if there is some pattern that can be measured and modeled.

dplore:
Reviewed in Sep 9, 2025 OC operators meeting.  Main feedback is the need for clearer definition of severities and how these are set, or possibly removing severity from the model.

dplore (changes requested):
Some comments on severity

release/models/platform/openconfig-platform-healthz-fault.yang
mrevang: I've gone with your bias and removed the severity leaf/corresponding indentityref.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,293 @@
...
+    description
+      "The process of turning a component off completely and then turning it back
+      on.";
+  }
+
+  identity ACTION_FACTORY_RESET {
+    base ACTION_BASE;
+    description
+      "Restores a component to its original factory settings.";
+  }
```
dplore, line 121: If would be helpful to map these actions to the relevant gnoi RPC's

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,293 @@
...
+
+  grouping fault-remediations {
+    description
+      "Grouping containing a faults recommended remediations";
+
+    container remediations {
+      description
+        "Enclosing container for a fault's recommended remediations.";
+
+      list remediation {
```
dplore, line 273: I think a concrete use case should be described to show how this should be used.  My thinking is that defining the symptoms could be done such that each symptom has one action versus a more complex list of ordered remediations.     If this model is "just in case people find it useful", we should leave it out.

mrevang: We have observed that repair strategies often contain progressive actions; "try this then that". Consider reseating a component. That can sometimes clear fault state and bring a component back into working order, but if that doesn't work, you might eventually need to replace that component or parent component. Enabling the device to express multiple actions mirrors what we have in our operator space, so I'd argue it's more than hypothetically useful.

Regarding order, the operator knows the cost of executing a recommended action and actions have inherent cost associated with them (ie: reboot cheaper than replace). So I'd be open to removing the `index` and just indexing based on action. Thoughts?

mrevang: Updated the remediation description to help clarify the intention/expectation around its use.

dplore: My bias given your workflow description of "try this then that" also makes me think an ordered list is probably more explicit and better.  The operator can still chose to execute in a different order or skip repair actions.

dplore:
Moved back to ready to discuss for the next OC operators meeting on Sep 30, 2025

ElodinLaarz:
Discussed in OC Operators Meet on Oct 07: Once any remaining changes are included here, we can move it to last-call and get any further input from folks. (Just ping me or @dplore to move it to last-call once it's ready.)

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,304 @@
...
+      "Reboot that interrupts device operation, ie a gNOI.System Reboot with
+      COLD RebootMethod";
+  }
+
+  identity ACTION_POWER_CYCLE {
+    base ACTION_BASE;
+    description
+      "The process of turning a component off completely and then turning it
+      back on.";
+  }
```
dplore, line 113: Can we differentiate this a bit from ACTION_COLD_REBOOT?  Could this be implemented using gNOI.System.RebootRequest Method= POWERDOWN followed by POWERUP?  Or is this expected to be some other method?  (pulling the plug?  using a remote controlled power strip?)

mrevang: Done! I think all of the above fit the bill, so I've included them as examples.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,311 @@
...
+    leaf origin-time {
+      type oc-types:timeticks64;
+      description
+        "The time at which the fault first became active.
+        This value is static for the life of the fault.";
+    }
+
+    leaf last-detection-time {
+      type oc-types:timeticks64;
+      description
```
earies, line 210: Per `oc-types:timeticks64` definition, this leaf must define the epoch referenced (Can see other uses for examples)

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,311 @@
...
+        "The time at which the fault's underlying condition was last sampled
+        and confirmed asserted.";
+    }
+
+    leaf occurrences {
+      type uint64;
```
earies: `oc-yang:counter64`?  worthy of a counters sub-container?

dplore: @mrevang this is worth refactoring IMO

release/models/platform/openconfig-platform-healthz-fault.yang
dplore: @mrevang please update the description with the intended epoch.

release/models/platform/openconfig-platform-healthz-fault.yang
mrevang: Done.

release/models/platform/openconfig-platform-healthz-fault.yang
mrevang: Done.

mrevang:
Thanks for your comments and thoughts. I appreciate the feedback. I will try to address some of your points inline below.

> A few questions/comments:
> 
> Reading through this proposal, the desire is well understood.
> 
> However, what I would like to see is when proposals as such come across that encode what are likely to be new, altered or similar behavior, that we start with concrete problem statements, use-cases and examples including example instance-data payloads according to the modeling.
> 
> While there are vendor implementation references, those tend to merely be similarities and/or may not apply to implementation targets for such feature-sets.

* Problem statement: Alarms provide a way to express that the device is in some kind of distress. However, alarms lack component attribution or a suggested response to the alarm. This requires operators to build up this knowledge over time. The ambition of this feature is that vendors have this context (via FMEA) and are able to provide this additional context along with the distress signal. 
* Use-case: Operators will use this information to mitigate fault conditions that affect operation. I can provide concrete fault examples in a later update.

> 
> That aside:
> 
> * Components modeling decided to group in software components some time back but the verbiage and constructs used within here lean towards hardware only.  Since there are no gating statements, the possibilities are wide atm.  What if any thoughts are there around how this should be applied or handled with software components?  It is fine to say it's not but this lack of clarity or accompanying documents leaves the industry lost and diverged (not unique to this proposal but want to point this out)

This proposal is definitely motivated by hardware component failures and has that bias. I don’t see a reason to prevent fault reporting on software components should they rise to healthz’s “UNHEALTHY” standard.

> * TMK, we have not yet discussed fault handling yet (much like alarms).  Sure, we can reference prior art to similar approaches but we have a tree where items pop up, remidiation actions may assist but are these to be cleared automatically by the system?  Does a user have to ack/clear?  If so, how?

There is no expectation that the operator should ack/clear the event. If the underlying fault condition is no longer asserted, after a vendor-defined cool down period (in the interest of fault dampening), the fault is expected to be removed by the device.

> * How about fault categories or severities?

Severity was in the early proposal, but was ultimately removed. I believe the consensus was that if “component is no longer able to perform the function expected of it and requires remediation/intervention” that we’d always be in a “severe” place.

> * Historical and/or clear counts?  e.g. the fault is not active now but there was prior record

A historical counter was also discussed. Many of the remediations (reboot, reimage, ect) would require this history to persist across these actions, which was deemed expensive to support. The prevailing thought was that the operator receiving these faults could monitor for recidivism and escalate as necessary.
 
> I see this is tagged for last call w/o any implementor comments yet. Can last call be extended post the next OC community call (and maybe a topic to bring to the table)?

I was too slow adding this topic to the agenda and the time was already claimed. We can try for the next meeting if needed.

I will prepare an accompanying document that captures some of these questions and design decisions for future guidance.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,330 @@
...
+      with COLD RebootMethod";
+  }
+
+  identity ACTION_POWER_CYCLE {
+    base ACTION_BASE;
+    description
+      "The process of turning a component off completely and then turning it
+      back on.
+
+      This could be acheived with two gNOI.System.RebootRequest calls:
```
nkitchen-arista: Spelling: achieved

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,330 @@
...
+      description
+        "A collection of fault counters.";
+
+      uses fault-state-counters;
+    }
+  }
+
+  grouping fault-remediations {
+    description
+      "Grouping containing a faults recommended remediations";
```
nkitchen-arista: faults -> fault's?

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,330 @@
...
+      "Memory errors";
+  }
+
+  identity SYMPTOM_MISSING_COMPONENT {
+    base SYMPTOM_BASE;
+    description
+      "Component is missing";
+  }
+
+  identity SYMPTOM_MISWIRED {
```
nkitchen-arista: The state of being miswired seems more like a fault cause than the visible symptom of a fault.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,330 @@
...
+
+  identity SYMPTOM_UNKNOWN {
+    base SYMPTOM_BASE;
+    description
+      "An unknown symptom";
+  }
+
+  identity ACTION_BASE {
+    description
+      "Base identity for fault symptoms";
```
nkitchen-arista: Description copied from SYMPTOM_BASE and not updated

doc/component_fault_overview.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Problem Statement
+
+`/system/alarms` provide a way to express that the device is in some kind of distress. However, alarms lack component attribution or a suggested response to the alarm. This requires operators to build up this knowledge over time. The ambition of this proposal is that since vendors already have this context (via FMEA) they can provide it along with the distress signal.
+
+## Proposal
+
+This proposal introduces a list of faults under a component's healthz container.
+
+A fault indicates that a component has recently or is currently experiencing a problem as expressed by a symptom. These symptoms have underlying conditions that are platform-specific. The fault should not be deleted as soon as the underlying condition is no longer asserted to avoid excessive fault creation and deletion. Deletion of the fault should occur after a suitable period outside of the triggering condition
```
brianneville: >The fault should not be deleted as soon as the underlying condition is no longer asserted to avoid excessive fault creation and deletion

Could this lead to confusion? An operator may attempt to resolve some fault (e.g. by rebooting the component), and then not have any kind of feedback in the YANG that their efforts were successful, until some indeterminate point in the future where the device decides to clean up the yang.

> Deletion of the fault should occur after a suitable period outside of the triggering condition

is this period going to be vendor-defined? would there be value in specifying it here so that all vendors are consistent and operators know what the YANG is expected to show?

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,324 @@
...
+  grouping platform-health-faults {
+    description
+      "Grouping containing a component's active faults";
+
+    container faults {
+      description
+        "Enclosing container for a component's active faults.";
+
+      list fault {
+        key "symptom";
```
brianneville, line 138: As this `symptom` list key is an identitref, it means that we cannot represent multiple faults of the same identity on the same component. E.g. if we had a component with two sensors operating over a threshold (`SYMPTOM_OVER_THRESHOLD`), we can only store the state of one of them in this `fault` list.

I'm also not sure how the `status` enum fits in here. E.g. imagine two sensors on the component report faults, but then only one of those faults is active in the next sampling period. would the status still report ACTIVE?

Or is the intention to use the `occurrences` leaf to track how many occurances of a particular type of fault there was? 
if this is the case, you may lose details about the specific faults.

release/models/platform/openconfig-platform-healthz-fault.yang
```diff
@@ -0,0 +1,324 @@
...
+        This value is static for the life of the fault.
+
+        The value is a Unix Epoch timestamp (nanoseconds since
+        Jan 1, 1970 00:00:00 UTC)";
+    }
+
+    leaf last-detection-time {
+      type oc-types:timeticks64;
+      description
+        "The time at which the fault's underlying condition was last sampled
```
brianneville, line 211: is the expectation here that the device is doing some sampling of the fault to validate its presence, and the sampling is not something that is driven by the client?
if so can you note this in the `component_fault_overview.md`?

doc/component_fault_overview.md
mrevang: I don't think so. Ultimately if the component is performing its intended function, the operator should be happy. Providing feedback about why a condition was alleviated (operator intervention or otherwise) seems low value IMO.

> is this period going to be vendor-defined?

Yes. The concern with providing a value, is that there isn't a universally appropriate one. It may change from device, component, or even symptom. Ultimately the bar for a fault is set pretty high, ie that it's not performing the function expected of it. One could expect that once it reaches that bar, it's not likely to recover. However, with threshold breaches (temp high), instantaneous values can oscillate in and out of that condition. This proposal would like to prevent fault oscillation.

release/models/platform/openconfig-platform-healthz-fault.yang
mrevang: I'd argue this is desirable because it bounds concurrent faults. SENSOR is a supported component type, so a component could be spun off for each sensor in your example, that increased granularity is still expressible.

If one or both of the sensors on your component was out of range to the point that it prevented the component from performing its function, the status would be active.

Occurrences are meant to track transitions into the active state. In the scenario above, if the component continues to experience loss of functionality due to a rotating roster of sub-components being out of range, that would be a single occurrence.

release/models/platform/openconfig-platform-healthz-fault.yang
mrevang: How these faults are found is up to the vendor implementation, whether that's sampling or event driven. Per gNMI, the client could send a `gnmi.Subscribe` for this leaf with a mode of `SAMPLE`, `ON_CHANGE` or `TARGET_DEFINED`.

doc/component_fault_overview.md
mrevang: While we don't have a "why", the `status` leaf does provide some feedback that the underlying condition is no longer satisfied. I'll add a note to the markdown doc to highlight this point.

PR #1347, opened 2025-08-05
Fix typo in openconfig-relay-agent.yang

sumedhak27:
### Change Scope

* DHCPv6 advertise message is spelled incorrectly; correct it.
* The change won't be backward compatible, i.e., the incorrectly
    spelled leaf won't be available further.


ghananigans (approved):

dplore (approved):

dplore (commented):
Module version number needs to be updated (to `0.2.0`)

PR #1349, opened 2025-08-05
openconfig-terminal-device-properties.yang: Fix order of string type in unions

jstobnicki:
### Change Scope

Reorder union types that have `string` type, moving this type to the very end. This allow proper recognizing of other `union` types.
In my opinion, the change is backwards compatible (no changes in the types, only reordering).

### DIFF
--- a/release/models/devices-manifest/openconfig-terminal-device-properties.yang
+++ b/release/models/devices-manifest/openconfig-terminal-device-properties.yang
@@ -37,8 +37,11 @@ module openconfig-terminal-device-properties {
 
   oc-ext:openconfig-version "0.2.0";
 
-
   // Revisions
+  revision "2025-08-06" {
+    description "Correct order of string type in unions.";
+  }
+
   revision "2024-05-28" {
       description "Comprehensive model update to undertake the limitations
       of the first version of the model, such the introduction of modes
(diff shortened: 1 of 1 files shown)


dplore:
This looks like a reasonable change to make as the types cannot be effectively used without this change.

We could use some comment from the community users of the devices manifest models if this breaking change is desirable to merge.  Since the models are v0.x.x models, [by policy we do expect to merge breaking changes](https://github.com/openconfig/public/blob/master/doc/releases.md).  But I would like to give some time for community members to speak up.

earies:
LGTM and to quote RFC 6020 Section 9.12

```
   When a string representing a union data type is validated, the string
   is validated against each member type, in the order they are
   specified in the "type" statement, until a match is found.
```

Note: this is an issue across other models where `type union` is used.  Do you want to expand the scope of this PR to pass through the entire model set to adjust ordering?

rgwilton:
As per Ebben's comment, having an issue to track fixing these generically would be a good idea.  Even if it does cause issues it would at least flush them out ...

dplore (approved):

ElodinLaarz:
@jstobnicki  -- There's another lint error `Revision statement should contain reference substatement  corresponding to semantic version 0.2.1`. Once it's fixed we can merge.

jstobnicki:
@ElodinLaarz / @dplore : Versions aligned.

PR #1351, opened 2025-08-10
openconfig-ospf-area.yang removed duplicate import and updated affected prefixes

slinkish:
### Change Scope

* duplicate import with no revision statements should be considered an error per RFC
*  no schema updates so should not break compatibility

### Platform Implementations

* bug fix with no schema updates


release/models/ospf/openconfig-ospf-area.yang
```diff
@@ -23,8 +22,13 @@ submodule openconfig-ospf-area {
     "This submodule provides common OSPF configuration and operational
     state parameters that are specific to the area context";
 
-  oc-ext:openconfig-version "0.0.1";
+  oc-ext:openconfig-version "0.0.2";
 
+  revision "2025-08-10" {
+    description
+      "removed duplicate import and updated prefixes";
+    reference "0.0.2";
```
earies: LGTM - initial revision should have never been `0.0.1` but rather `0.1.0`.  Suggest making this `0.1.1` (also capitalize "Removed")

dplore: +1

slinkish: I have just brought in recommendations, thanks!

slinkish:
I totally missed that this was a submodule, bumped up parent revision due to linter errors.

slinkish:
I updated remaining child submodules, and rebased on latest master.

PR #1361, opened 2025-09-03
Platform model for USB port

ThatOneSix:
### Change Scope

* This change adds a new component to the openconfig-platforms container for managing USB ports on devices. It is based on the logic used in openconfig-platform-transceiver.yang and augments /oc-platform:components/oc-platform:component. This is a continuation of [PR #1339](https://github.com/openconfig/public/pull/1339). It has been run through pyang lint successfully, with the following tree:

```
module: openconfig-platform-usb-port

  augment /oc-platform:components/oc-platform:component:
    +--rw usb-port
       +--rw config
       |  +--rw enabled?   boolean
       +--ro state
          +--ro enabled?   boolean
```

Also viewed as:

```
+--rw components
     +--rw component* [name]
         +--rw oc-usb-port:usb-port
               +--rw oc-usb-port:config
               |  +--rw oc-usb-port:enabled?   boolean
               +--ro oc-usb-port:state
                  +--ro oc-usb-port:enabled?   boolean
```

### Platform Implementations

* Mobility Controller-based Aruba Access Points: https://arubanetworking.hpe.com/techdocs/CLI-Bank/Content/aos8/ap-system-pro.htm (ap-usb-power-mode [enable|disable])

* Cisco Catalyst switches: https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst9400/software/release/17-17/command_reference/b_1717_9400_cr/interface_and_hardware_commands.html?dtid=osscdc000283&linkclickid=srch#wp3775305635 ([no] platform usb disable)

### DIFF
--- a/release/models/bgp/openconfig-bgp-policy.yang
+++ b/release/models/bgp/openconfig-bgp-policy.yang
@@ -28,7 +28,13 @@ module openconfig-bgp-policy {
     It augments the base routing-policy module with BGP-specific
     options for conditions and actions.";
 
-  oc-ext:openconfig-version "8.1.0";
+  oc-ext:openconfig-version "8.2.0";
+
+  revision "2025-05-23" {
+    description
+      "Refactor must restriction behavior for BGP MED values/actions.";
+    reference "8.2.0";
+  }
 
   revision "2024-11-13" {
     description
@@ -1385,24 +1391,19 @@ module openconfig-bgp-policy {
 
     leaf set-med {
       type bgp-set-med-type;
-      must "../set-med-action" {
-        error-message
-          "set-med cannot be specified without a valid set-med-action";
-      }
       description
-        "Set the MED metric attribute in the route update.  When set, a
-        valid `set-med-action` must be specified.";
+        "Set the MED metric attribute in the route update.  When set to
+        an integer metric, all `set-med-action` options apply.  When set
+        to 'IGP', the `set-med-action` must be set to 'SET'.";
     }
 
     leaf set-med-action {
       type bgp-set-med-action;
-      must "../set-med" {
-        error-message
-          "set-med-action cannot be specified without a set-med value";
-      }
       description
-        "When set-med is specified, this leaf is mandatory to set the
-        appropriate action on the MED metric value.";
+        "This leaf is mandatory when `set-med` is specified.  When
+        `set-med` is specified as a integer metric, all actions are
+        valid options.  When set to 'IGP', this action must be set to
+        'SET'.";
     }
   }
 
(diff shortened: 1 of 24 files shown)


release/models/platform/openconfig-platform-usb-port.yang
```diff
@@ -0,0 +1,95 @@
...
+
+  grouping usb-port-config {
+    description
+      "Configuration data for USB ports.";
+
+    leaf enabled {
+      type boolean;
+      default "true";
+      description
+        "Turns power on / off to the USB port.";
```
nkitchen-arista: Can you generalize this so that it can apply to disabling by software and not just power on/off?

release/models/platform/openconfig-platform-usb-port.yang
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  // data definition statements
+
+  // augment statements
+
+  augment "/oc-platform:components/oc-platform:component" {
+    description
+      "Configuration data for USB ports.
+      A USB port can be configured for persistent powered-off
```
nkitchen-arista: This description looks like it has not been updated from the earlier version that used `power-admin-state`.

I recommend you drop the second sentence and just keep "Configuration data for USB ports".

release/models/platform/openconfig-platform-usb-port.yang
ThatOneSix: Changed to "Enables or disables USB port functionality."

release/models/platform/openconfig-platform-usb-port.yang
ThatOneSix: Removed the second line. I missed that in my pass through, thanks for catching it.

dplore:
Reviewed in September 4, 2025 OC Community meeting without objections.  I'll wait to move this to last call until @ThatOneSix has reviewed this with more implementors.  @ThatOneSix let me know when you have internal sign off on this approach and we can move it to a last call status and then merge.

LimeHat:
>My intent is to model USB ports as their own components.

Is there a specific reason behind this choice?
I would argue that inlining usb ports to the parent components is an easier approach to work with, considering that they have only one property.

> but it doesn't mean that we expect to have containers named usb-port as children of arbitrary components.

Only the ones that have such ports :-). There's nothing in OC community that says this is invalid, it is a choice vendors can make.

LimeHat:
>Can you give an example of what you are suggesting regarding "inlining usb ports to the parent components"?

Just placing usb-port container directly under the parent component, e.g.
`/components/component[name=Supervisor1]/usb-port/{config,state}/enabled`

instead of creating a new component with a new identity, such as `/components/component[name=USBPort-Supervisor1]/usb-port/{config,state}/enabled`.

dplore (approved):

ElodinLaarz (approved):

PR #1362, opened 2025-09-04
Reset initial version of openconfig-system-bootz to 0.1.0

earies:
* (M) release/models/system/openconfig-system-bootz.yang
    - Reset initial version from 1.0.0 to 0.1.0

### Change Scope

Initial model publish of openconfig-system-bootz was published with 1.0.0 which
conflicts with standard OC publish policy.  All initial versions are to be
0.1.0 to accomodate the possible backwards-incompatible iterations < 1.0.0.

Slightly obscure case that should have been caught in initial review thus we
are not versioning backwards in revision statements but rather resetting the
initial publish version and adjusting the date.

### Platform Implementations

N/A: Version reset for implementors to handle

### DIFF
--- a/release/models/system/openconfig-system-bootz.yang
+++ b/release/models/system/openconfig-system-bootz.yang
@@ -19,14 +19,14 @@ module openconfig-system-bootz {
     service running on a network device.";
 
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "0.1.0";
   oc-ext:catalog-organization "openconfig";
   oc-ext:origin "openconfig";
 
-  revision "2023-06-16" {
+  revision "2025-09-03" {
     description
-      "Creation of bootz state paths needed for the service.";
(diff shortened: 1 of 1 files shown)


earies:
And general comment out of this direct patch intent:

```
    leaf checksum {
      type string;
      default "";
      description
```

Is a default really necessary here?

robshakir:
I'm not sure I understand why this revision needs to go backwards, given that this has been published for some time. Is there a reason it can't just increment going forwards - including going to 2.x.x if a backwards-incompatible change is introduced?

rgwilton:
I'm not convinced that having the version going backwards is a good idea - there is a risk that this could cause problems for various tooling or need to be handled as an exception case.  Why not just move to 2.0.0/etc, as Rob Shakir pointed out.  After all, it is only a version label.  I presume the underlying concern is that it makes the model appear more mature than it may actually be?

PR #1364, opened 2025-09-05
Add enable-echo leaf to BFD interface config.

AnilKR123:
### Change Scope
Add the enable-echo leaf in bfd-interface-config to allow BFD echo mode to be enabled per interface.

### DIFF
--- a/release/models/bfd/openconfig-bfd.yang
+++ b/release/models/bfd/openconfig-bfd.yang
@@ -26,7 +26,14 @@ module openconfig-bfd {
     "An OpenConfig model of Bi-Directional Forwarding Detection (BFD)
     configuration and operational state.";
 
-  oc-ext:openconfig-version "0.4.1";
+  oc-ext:openconfig-version "0.4.2";
+
+  revision "2025-09-03" {
+    description
+      "Add the enable-echo leaf in bfd-interface-config to allow
+      BFD echo mode to be enabled per interface.";
+    reference "0.4.2";
+ }
 
   revision "2025-03-18" {
     description
@@ -269,6 +276,14 @@ module openconfig-bfd {
       description
         "When this leaf is set to true - BFD will be enabled on
         each member interface of the aggregated Ethernet bundle.";
+    }
+
+    leaf enable-echo {
+      type boolean;
+      default false;
+      description
+        "When this leaf is set to true, the BFD echo mode is enabled
+        for all sessions on the interface.";
     }
   }
 


AnilKR123:
Bfd echo mode is defined in RFC 5880:

https://datatracker.ietf.org/doc/html/rfc5880

release/models/bfd/openconfig-bfd.yang
```diff
@@ -270,6 +277,14 @@ module openconfig-bfd {
         "When this leaf is set to true - BFD will be enabled on
         each member interface of the aggregated Ethernet bundle.";
     }
+
+    leaf enable-echo {
+      type boolean;
+      default false;
```
dplore: The RFC doesn't mention any default value for echo mode.  (I didn't see it here https://datatracker.ietf.org/doc/html/rfc5880#section-6.8.13).  Please remove the default as OC prefers to not define defaults.

dplore (changes requested):
Please also include at least two references to implementations of BFD which include enabling and disabling echo mode support

LimeHat:
>BFD Echo mode has its own timer considerations. See "Required Min Echo RX Interval" in RFC 5880.

I'd second that comment.  
We have an implementation that can respond to echo requests and the interval is configurable [where 0 means disabled per the RFC].
https://yang.srlinux.dev/v25.7.2?path=%2Fbfd%2Fsubinterface%5Bid%3D*%5D%2Fminimum-echo-receive-interval

Verma-Anukul:
@AnilKR123 
Please get a closure on this

release/models/bfd/openconfig-bfd.yang
AnilKR123: I have removed the default value, thanks.

jhaas-pfrc:
Note that the timer considerations issue still hasn't been addressed.

AnilKR123:
> > As suggested, we will follow RFC 9127 for Echo config.
> > I will update the PR to:
> > 
> > * Remove the `enable-echo` leaf
> > * Add the `desired-min-echo-tx-interval` and `required-min-echo-rx-interval` leaves.
> > 
> > This gives the OpenConfig model RFC parity. Please confirm and I'll push the changes, thanks.
> 
> This sounds like positive progress.
> 
> You'll need to describe desired vs. required min and justify why both are needed? My thought is a `required-minimum-echo-receive` would be sufficient?

The desired-min-echo-tx-interval and required-min-echo-rx-interval are defined config parameters in RFC 9127.

The desired-min-echo-tx-interval has two purposes:

- It acts as the “enable” knob for Echo when non-zero.

- It specifies how fast the local device wants to transmit Echo packets.


The required-min-echo-rx-interval has two purposes:

- It acts as a knob for the peer device to opt out of participating in Echo if set to 0.

- It specifies the local device's capability regarding how fast it can loop back packets.

RFC 5880 Section 6.8.9 mandates that the interval between transmitted BFD Echo packets must not be less than the value advertised by the remote system, but explicitly leaves the actual transmission logic "outside the scope of this specification."

Without this local TX leaf, the system has no variable to define its operational rate subject to that constraint. The system would be forced to default to the minimum interval permitted by the peer, removing the operator's ability to throttle outbound traffic.

release/models/bfd/openconfig-bfd.yang
```diff
@@ -26,7 +26,15 @@ module openconfig-bfd {
     "An OpenConfig model of Bi-Directional Forwarding Detection (BFD)
     configuration and operational state.";
 
-  oc-ext:openconfig-version "0.4.1";
+  oc-ext:openconfig-version "0.4.2";
```
dplore: Need to increment minor version since leaves are being added.  (patch version is only for correcting cosmetic typos or very minor bugs in description.

```suggestion
  oc-ext:openconfig-version "0.5.0";
```

release/models/bfd/openconfig-bfd.yang
```diff
@@ -26,7 +26,15 @@ module openconfig-bfd {
...
 
-  oc-ext:openconfig-version "0.4.1";
+  oc-ext:openconfig-version "0.5.0";
+
+  revision "2025-09-03" {
+    description
+      "Add desired-min-echo-tx-interval and required-min-echo-rx-interval
+      leaves in bfd-interface-config to allow RFC 9127-compliant BFD echo
+      mode configuration per interface.";
+    reference "0.4.2";
```
dplore:
```suggestion
    reference "0.5.0";
```

PR #1365, opened 2025-09-05
Add stats-update-time leaf to BFD session state for both async and echo sessions.

AnilKR123:
### Change Scope

The stats-update-time leaf represents the timestamp of the most recent refresh of BFD session statistics, such as last-packet-transmitted, last-packet-received, and packet counters. Although the session statistics themselves have nanosecond granularity, updates may not be reflected immediately. This leaf indicates the actual time the statistics were last updated, providing an accurate view of their freshness.


rgwilton:
> > Although the session statistics themselves have nanosecond granularity, updates may not be reflected immediately.
> 
> That's how stats work in general. Is there a reason we need to reflect the timestamp in BFD stats specifically?

Agreed.  This feels quite implementation specific.

PR #1369, opened 2025-09-17
Add sensor component container

mrevang:
### Change Scope

* This change adds a generic `sensor` container to `/components/component[name=]`.  This container defines an instantaneous value along with operating thresholds.

It's worth highlighting, the units of the sensor's instantaneous value are defined by a parallel leaf, allowing this generic sensor component to represent multiple sensor types (current, voltage, ect). 
```
module: openconfig-platform
  +--rw components
     +--rw component* [name]
        +--ro state
        |  +--ro sensor
        |  |  +--ro state
        |  |  |  +--ro oc-sensor:instant-precision2?   decimal64
        |  |  |  +--ro oc-sensor:units?                identityref
        |  |  +--ro oc-sensor:thresholds
        |  |     +--ro oc-sensor:threshold* [severity]
        |  |        +--ro oc-sensor:severity    -> ../state/severity
        |  |        +--ro oc-sensor:state
        |  |           +--ro oc-sensor:severity?   identityref
        |  |           +--ro oc-sensor:upper?      decimal64
        |  |           +--ro oc-sensor:lower?      decimal64
```

### Platform Implementations

* Openconfig already defines specific sensor types among other component types. This change provides a more generic way to model a sensor that doesn't align with the fixed sensors already available.
 * Sonic [temp](https://github.com/sonic-net/SONiC/blob/master/thermal-control-design.md#1-overview) [voltage](https://github.com/sonic-net/sonic-platform-daemons/blob/44666a0015c45f00cf8e883730411a550d189295/sonic-sensormond/scripts/sensormond#L181)
 * We are working offline with a couple of vendors on implementing this model.


release/models/platform/openconfig-platform-sensor.yang
```diff
@@ -0,0 +1,191 @@
...
+  identity UNITS_RELATIVE_HUMIDITY {
+    base UNITS_BASE;
+    description
+      "Relative humidity expressed in percentage; where 1=1%";
+  }
+
+  identity UNITS_ABSOLUTE_HUMIDITY {
+    base UNITS_BASE;
+    description
+      "Absoltue humidity expressed in grams per cubic meter (g/m^3)";
```
earies:
```suggestion
      "Absolute humidity expressed in grams per cubic meter (g/m^3)";
```

release/models/platform/openconfig-platform-sensor.yang
```diff
@@ -0,0 +1,191 @@
...
+          config false;
+          description
+           "The severity applied to the group of thresholds.
+            An implementation's highest severity threshold
+            should be mapped to OpenConfig's `CRITICAL`
+            severity level.";
+        }
+
+        container state {
+          config false;
```
earies, line 162: config false is inherited from parents.  Is there ever intention for configurable thresholds (there are other variants of this in OC modeling as well in platform/system)

release/models/platform/openconfig-platform-sensor.yang
```diff
@@ -0,0 +1,191 @@
...
+
+
+  // data definition statements
+
+  // augment statements
+
+  augment "/oc-platform:components/oc-platform:component/" +
+    "oc-platform:sensor" {
+    description
+      "Adding fan data to component model";
```
earies: Miscopy from another model?

release/models/platform/openconfig-platform-sensor.yang
```diff
@@ -0,0 +1,191 @@
...
+
+    uses sensor-threshold-top ;
+  }
+
+  augment "/oc-platform:components/oc-platform:component/" +
+    "oc-platform:sensor/oc-platform:state" {
+    description
+      "Adding fan state data to component model";
```
earies: Miscopy?

release/models/platform/openconfig-platform-sensor.yang
```diff
@@ -0,0 +1,191 @@
...
+      description
+        "Current (instantaneous) sensor value with 2 fraction digits";
+    }
+
+    leaf units {
+      type identityref {
+        base UNITS_BASE;
+      }
+      description
+        "The unit type for this sensor (volts, amps, celsius degree, ect)";
```
earies:
```suggestion
        "The unit type for this sensor (volts, amps, celsius degree, etc.)";
```

mrevang:
> As more and more structures are proposed under the `/components` subtree, it has been observed that we are in a state where there are N# of potential ways to represent like data all without true modeling restrictions. Can you elaborate when this structure should be represented under components of which type? Is this meant to be for `type=SENSOR` components only? If so, subtrees should ideally carry when restrictions indicating such to firm up consistent behavior.
> 
> There are also `vendor` anchor points and other thresholding that could also be leveraged for such uses so I think some guidance is necessary for implementors otherwise it is likely to see divergence across implementations.

Great comment. The intention is for this container to only be used when type=SENSOR. From a quick search, looks like this can be accomplished with a `must` , let me add this. Thanks for the suggestion!

release/models/platform/openconfig-platform-sensor.yang
mrevang: I can imagine a scenario where one might want to configure these thresholds. Consider if the management system wanted to tune when an alarm might fire due to the sensor being out of range. I don't have this usage in mind. The use case I'm modeling for is to have these thresholds be vendor supplied, so `config false` seems appropriate. Thoughts?

release/models/platform/openconfig-platform-sensor.yang
mrevang: I can swear I addressed this at some point... losing my mind. Good catch, thanks!

release/models/platform/openconfig-platform-sensor.yang
mrevang: Good catch, thanks.

release/models/platform/openconfig-platform-sensor.yang
dplore: the state container should be config false always.  But don't label the parent container config false.  Allow for a config container to be added later which would allow user configurable thresholds.

Today all the operational use cases I am aware of do not require a user override of factory defined thresholds for hardware components.

mrevang:
> > > As more and more structures are proposed under the `/components` subtree, it has been observed that we are in a state where there are N# of potential ways to represent like data all without true modeling restrictions. Can you elaborate when this structure should be represented under components of which type? Is this meant to be for `type=SENSOR` components only? If so, subtrees should ideally carry when restrictions indicating such to firm up consistent behavior.
> > > There are also `vendor` anchor points and other thresholding that could also be leveraged for such uses so I think some guidance is necessary for implementors otherwise it is likely to see divergence across implementations.
> > 
> > 
> > Great comment. The intention is for this container to only be used when type=SENSOR. From a quick search, looks like this can be accomplished with a `must` , let me add this. Thanks for the suggestion!
> 
> Similar example here: https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform.yang#L1350-L1355

In order to access `/components/component/state/type` for use by `when`, it looks like the sensor will need to be relocated:
* /components/component/sensor -> /components/component/state/sensor

The following snippet...
```
container sensor {
      when "../state/type = 'oc-platform-types:SENSOR'";
```
... resulted in this warning:
```
warning: node "openconfig-platform::state" is config false and is not part of the accessible tree
```
The relocation worked... so I'm going with it. Updating the description to capture this change.

release/models/platform/openconfig-platform-sensor.yang
mrevang: Removing this
```
list threshold {
        key "severity";
        config false; <<<<<<<<
```
... requires adding the key into a config container, just FYI, thanks.

mrevang: Ignore that comment... transceiver provides an example where this isn't true.. sorting out this error:
> The list's keys must have the same `config` value as the list itself.(invalid-config)

mrevang: Maybe I'm not understanding the comment. The thresholds container in the `transceiver` container also has `config false`:
https://github.com/openconfig/public/blob/e1a6ae01bbd868990407899605eb778c4cd10100/release/models/platform/openconfig-platform-transceiver.yang#L581

earies:
> > > > As more and more structures are proposed under the `/components` subtree, it has been observed that we are in a state where there are N# of potential ways to represent like data all without true modeling restrictions. Can you elaborate when this structure should be represented under components of which type? Is this meant to be for `type=SENSOR` components only? If so, subtrees should ideally carry when restrictions indicating such to firm up consistent behavior.
> > > > There are also `vendor` anchor points and other thresholding that could also be leveraged for such uses so I think some guidance is necessary for implementors otherwise it is likely to see divergence across implementations.
> > > 
> > > 
> > > Great comment. The intention is for this container to only be used when type=SENSOR. From a quick search, looks like this can be accomplished with a `must` , let me add this. Thanks for the suggestion!
> > 
> > 
> > Similar example here: https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform.yang#L1350-L1355
> 
> In order to access `/components/component/state/type` for use by `when`, it looks like the sensor will need to be relocated:
> 
>     * /components/component/sensor -> /components/component/state/sensor
> 
> 
> The following snippet...
> 
> ```
> container sensor {
>       when "../state/type = 'oc-platform-types:SENSOR'";
> ```
> 
> ... resulted in this warning:
> 
> ```
> warning: node "openconfig-platform::state" is config false and is not part of the accessible tree
> ```
> 
> The relocation worked... so I'm going with it. Updating the description to capture this change.



> > > > As more and more structures are proposed under the `/components` subtree, it has been observed that we are in a state where there are N# of potential ways to represent like data all without true modeling restrictions. Can you elaborate when this structure should be represented under components of which type? Is this meant to be for `type=SENSOR` components only? If so, subtrees should ideally carry when restrictions indicating such to firm up consistent behavior.
> > > > There are also `vendor` anchor points and other thresholding that could also be leveraged for such uses so I think some guidance is necessary for implementors otherwise it is likely to see divergence across implementations.
> > > 
> > > 
> > > Great comment. The intention is for this container to only be used when type=SENSOR. From a quick search, looks like this can be accomplished with a `must` , let me add this. Thanks for the suggestion!
> > 
> > 
> > Similar example here: https://github.com/openconfig/public/blob/master/release/models/platform/openconfig-platform.yang#L1350-L1355
> 
> In order to access `/components/component/state/type` for use by `when`, it looks like the sensor will need to be relocated:
> 
>     * /components/component/sensor -> /components/component/state/sensor
> 
> 
> The following snippet...
> 
> ```
> container sensor {
>       when "../state/type = 'oc-platform-types:SENSOR'";
> ```
> 
> ... resulted in this warning:
> 
> ```
> warning: node "openconfig-platform::state" is config false and is not part of the accessible tree
> ```
> 
> The relocation worked... so I'm going with it. Updating the description to capture this change.

I think your prior anchor point was the more correct location as this follows the same pattern for other "type" hierarchies... just that many of these containers today have no schema rules of when they should or should not exist (left up to interpretation)

Your compiler issue stems from interleaving a config=true container to attempt to depend on config=false.  One option is leave as-is w/o `when` restrictions (which I think is a topic of discussion that should be had as I'm finding we are revisiting this subject over and over) or align the `when` stmt to a config=false hierarchy.  Some are pure state but if there is ever any chance of configuration in the future, we cannot be that strict.

release/models/platform/openconfig-platform-sensor.yang
mrevang: The latest change applies a `when` on `sensor/state` at L177.

release/models/platform/openconfig-platform-sensor.yang
```diff
@@ -0,0 +1,197 @@
...
+
+  // typedef statements
+
+  // grouping statements
+
+  grouping sensor-state {
+    description
+      "Operational state data for sensor components";
+
+    leaf instant-precision2 {
```
nkitchen-arista, line 90: What's the rationale for including the precision in the name of the leaf? Is the idea that a different name would be used if a different precision were needed? If so, wouldn't that be contrary to the aim of a generic sensor model?

mrevang: Exactly. I wanted to accommodate different precision requirements should they arise in the future. From a survey of existing sensor values, "2 decimal places" covers the current use cases. Should another use case come by requiring another level of precision, I wanted to avoid the scenario where "instant" carried some implicit precision and exists among "instant-precisionX".

If other precisions are required, the expectation should be that only a single "instant" should be populated.

PR #1371, opened 2025-09-25
OC policy-forwarding match action reference documentation

dplore:
### Change Scope

* This PR introduces a new policy-forwarding policy type named MATCH_ACTION_POLICY.
* The documentation includes the motivation, use cases and a high level design and references to similar features implemented in current NOS's.
* After an initial review, yang model file changes will be introduced in follow up pull requests

### DIFF
--- a/doc/tmp.md
+++ b/doc/tmp.md
@@ -1,1 +1,0 @@
-https://github.com/openconfig/public/pull/1371/changes


doc/ocpf-match-action.md
```diff
@@ -0,0 +1,386 @@
...
+                       +--rw ipv6
+                          +--rw hop-limit?
+                          +--rw destination-address?
+                          +--rw destination-address-prefix-set?  # exclude use case public#1314
+                          +--rw fragment-offset?
+                          +--rw protocol?
+                          +--rw source-address?
+                          +--rw source-address-prefix-set?
+                       +--rw action
+                          +--rw config
```
sugrimov: We also need support for "interface" or "next-interface" action option to support egress interface redirect.

https://www.arista.com/en/support/toi/eos-4-32-2f/14550-support-for-traffic-policy-on-interfaces#redirect-interface

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
+OpenConfig policy-forwarding match/action
+=============
+
+Authors:
```
robshakir: (nit) can you wrap this document to 80ch?

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+There are operational use cases to create packet “match / action”  policies which can match on L2, L3, and L4 packet fields and execute actions like next-hop selection, encapsulation/decapsulation, egress forwarding class/queue assignment, and packet field manipulation.
+
+Currently, OpenConfig data models for performing these actions are split into ACL, QoS, and OC Policy-Forwarding (OCPF) trees, each with unique match and action paths and slightly different schemas.
+Instead of writing 3 groups of rules, we can express them all in one policy:
+
+![OCPF Summary](img/ocpf-match-action-summary.png)
+
+This document proposes extending the OCPF schema with rules and actions and a new OCPF policy type which allows more flexible rules and actions and provides references to existing vendor implementations.  This provides benefits of simplified rule expression, reducing the number of matches to be configured and providing more readable policies.
+
+[insert drawing here]
```
robshakir: Stale TODO?

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+
+# 2. Background
+
+Existing OpenConfig (OC) models for Access Control Lists (ACL), Quality of Service (QoS), and OpenConfig Policy-Forwarding (OCPF) all define match-action rules, with actions grouped by type. ACL rules primarily involve drop and accept actions. OC QoS focuses on classifying packets and setting specific fields, which must be combined with scheduling, policing, and queuing definitions. OCPF introduces a third schema for packet matching, with expected actions including next-hop setting, encapsulation, and decapsulation.
+
+Modern hardware and Network Operating Systems (NOS) offer greater flexibility in defining match-action rules, extending beyond the categories currently defined by OC (ACL, OCPF, and QoS). Examples include JunOS's filter-based forwarding, Arista EOS's traffic policy features and Cisco IOS XR ACL based forwarding.
+
+This proposal defines an order of operations which allow match action rules to coexist with OC’s traditional ACL, QoS classifiers, policing, queuing, TCAM-based match and forward rules, and LPM-based match and forward rules.
```
robshakir: This document references multiple times that there are existing models. I think it would be prudent to have a clear statement that says whether we expect that these models will continue to exist. My view is that since this model only does a subset of what those models can do, then we expect that this is an *additional* model. Can you confirm this, and add something to the text?

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+
+# 3. OC Model Design
+
+This design expands the policy-forwarding tree with more rules and actions and a new policy type, MATCH_ACTION_POLICY.
```
robshakir:
```suggestion
This design expands the policy-forwarding tree with more rules and actions and a new policy type, `MATCH_ACTION_POLICY`.
```

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+        |  +--rw policies
+        |  |  +--rw policy* [policy-id]
+        |  |     +--rw config
++                     +--rw type?    # Add enum MATCH_ACTION
+        |  |     +--rw rules
+        |  |        +--rw rule* [sequence-id]
++                      +--rw type?              -> oc-pf:rule-type # identity ref, similar to /acl/type
++                      +--rw name?              string  # user defined name for a rule
+                       +--rw l2?
++                      +--rw mpls-traffic-class?     oc-mplst:mpls-tc
```
robshakir: This seems like it should be in an MPLS container to match the other types?

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+                          +--rw hop-limit?
+                          +--rw destination-address?
+                          +--rw destination-address-prefix-set?  # exclude use case public#1314
+                          +--rw fragment-offset?
+                          +--rw protocol?
+                          +--rw source-address?
+                          +--rw source-address-prefix-set?
+                       +--rw action
+                          +--rw config
++                            +--rw count?              string  #user defined counter
```
robshakir: Should this not be a `leafref` to one of the defined counters that are referenced below?

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
++    |        +--rw name?           string
++    |  |  +--rw state
++    |  |     +--rw name?           string
++    |  |     +--rw counters
++    |  |        +--rw count?       oc-yang:counter64
+```
+
+# 4. Multiple Rules and Actions
+
+OC does not explicitly state what logic is used when an OCPF policy has multiple rules, multiple criteria within a rule and if there are multiple actions on a rule.  OC is not clear on what should happen after an OCPF rule is matched or not matched.  The following text is proposed to be added to the OCPF model to describe how to handle multiple rules and actions.
```
robshakir: Does not __currently__?

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+
+Multiple match criteria in a single rule is treated logically as “AND”.
```
robshakir: For formatting, can these be made into a bulleted list to ease readability?

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+
+When a packet matches a rule, all actions associated with a rule are performed.
+
+When a rule is matched, processing stops, unless the `next-rule` leaf is populated.
+
+# 5. Order of operations and mapping match/action rules into device implementations
+
+## 5.1. OC defined order of operations
+
+Today OC defines the following order of operations between ACL, QOS and OCPF.  “Routing” is included as the step expected to follow policy forwarding.  Note that In today’s OC model, the sequence of ACL and QOS is ambiguous; OCPF is defined to happen after ACL and QOS.
```
robshakir:
```suggestion
Today OC defines the following order of operations between ACL, QOS and OCPF.  “Routing” is included as the step expected to follow policy forwarding.  Note that in today’s OC model, the sequence of ACL and QOS is ambiguous; OCPF is defined to happen after ACL and QOS.
```

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+
+![OCPF Summary](img/ocpf-match-action-old-order.png)
+
+In practice, implementations using ASICs for their packet processing compile their “match action rule” (ACL, QoS classifier,  policy forwarding, PBR, firewall, traffic policy, etc…) configurations into banks of TCAM memories.   The rules are compiled as bitfield masks which extract key data from packets.  The keys are looked up in TCAM memory to map to an action.    The TCAM memories are often rather limited in scale and the implementation details of the hardware and programming vary widely between ASICs and NOSs.
```
robshakir: It's not clear to me that this is wholly correct -- not all of these things must use TCAM. Do we need this level of detail? What is the take-home message that one should be taking away here?

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+
+Abstractly the programming of these rules can be organized as follows:
```
robshakir: I would suggest to make it clear that this is an example mapping for a theoretical chip. It might be worth demonstrating that this model is independent of chip architecture.

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+
+# 6. Use cases
+
+Below is a survey of use cases where match and action criteria can be written today using Juniper filter based forwarding and/or Arista traffic policy.  These use cases include fields that are not supported in OC today and/or implement expressions which mix what OC defines across ACL, QoS and OCPF.
+
+## 6.1. Summary of use cases
+
+A quick enumeration of the operational use cases is:
+
+Match by src and/or dst ip, count, sample and drop
```
robshakir: Can we make this a bulleted list?

doc/ocpf-match-action.md
```diff
@@ -0,0 +1,385 @@
...
+Match by ip fragment and ip protocol icmp, tcp, udp. Action: drop
+Match from “configured bgp neighbors” , Action: count, ratelimit, set qos, accept
+Match icmp to “locally configured interfaces”, Action: count, set qos, accept
+Match ttl=1, Action: count, police, set next-hop-group
+Match all src ip except some prefixes, and list of dest prefixes. Action: count, encap, queue, sample
+Match a prefix list, set-dscp, update a user defined counter and set the next-hop
+
+Using today’s OC equivalent for these policy terms would require writing
+A series of ACL’s
```
robshakir: Again, can we use a bulleted list?

doc/ocpf-match-action.md
dplore: Yes, this is correct.  I've updated the text to clarify.

doc/ocpf-match-action.md
dplore: Ah agreed, upon re-reading the statement here is much more absolute than intended.  I've updated to use "some ASICs"

doc/ocpf-match-action.md
dplore: I've reduced the detail in this text.  It is intended to provide hints on how to map this configuration into hardware and to achieve high scalability.

doc/tmp.md
robshakir, line 1: I presume this can be removed, or shouldn't be in the submitted PR?

PR #1374, opened 2025-09-26
Add certificate-expires-on leaf

dplore:
### Change Scope

* Add `certificate-expires-on` leaf which provides a timestamp when a grpc server's cert will expire.

### Platform Implementations

gNSI is a new protocol with several implementations in progress.

### DIFF
--- a/release/models/gnsi/openconfig-gnsi-certz.yang
+++ b/release/models/gnsi/openconfig-gnsi-certz.yang
@@ -156,7 +156,7 @@ module openconfig-gnsi-certz {
       description
         "The timestamp of the moment when the certificate
         (and associated private key) that is currently used
-        by this gRPC server was created.";
+        by this gRPC server will expire.";
     }
     leaf ca-trust-bundle-version {
       type version;
--- a/release/models/gnsi/openconfig-gnsi-certz.yang
+++ b/release/models/gnsi/openconfig-gnsi-certz.yang
@@ -30,7 +30,13 @@ module openconfig-gnsi-certz {
     "This module provides a data model for the metadata of gRPC credentials
     installed on a networking device.";
 
-  oc-ext:openconfig-version "0.6.0";
+  oc-ext:openconfig-version "0.7.0";
+
+  revision 2025-09-26 {
+    description
+      "Add certificate-expires-on leaf";
+    reference "0.7.0";
+  }
 
   revision 2024-03-05 {
     description
@@ -140,6 +146,13 @@ module openconfig-gnsi-certz {
     }
     leaf certificate-created-on {
       type created-on;
+      description
+        "The timestamp of the moment when the certificate
+        (and associated private key) that is currently used
+        by this gRPC server was created.";
+    }
+    leaf certificate-expires-on {
+      type oc-types:timeticks64;
       description
         "The timestamp of the moment when the certificate
         (and associated private key) that is currently used


dplore:
@mihirpitale-googler

release/models/gnsi/openconfig-gnsi-certz.yang
```diff
@@ -145,6 +151,13 @@ module openconfig-gnsi-certz {
         (and associated private key) that is currently used
         by this gRPC server was created.";
     }
+    leaf certificate-expires-on {
+      type oc-types:timeticks64;
+      description
+        "The timestamp of the moment when the certificate
+        (and associated private key) that is currently used
+        by this gRPC server will expire.";
```
earies: For use of `timeticks64`, the leaf has to define the epochs used.

e.g. `The value is the timestamp in nanoseconds relative to the Unix Epoch (Jan 1, 1970 00:00:00 UTC)`

dplore:
Agreed on the typedef for created-on.  It's not warranted.

I also agree that moving certificates to a system wide tree and then have gnsi reference it makes more sense from a model robustness point of view.

@morrowc and @marcushines do you have an opinion regarding moving the anchor point for certificates data into something like `/system/certificates` versus the currently modeled `/system/grpc-servers`?  Is that feasible given the current state of deployment?

release/models/gnsi/openconfig-gnsi-certz.yang
```diff
@@ -78,14 +84,6 @@ module openconfig-gnsi-certz {
       a reboot.";
   }
 
-  typedef created-on {
-    type oc-types:timeticks64;
-    description
-      "The creation time of the credential as reported by the credential
-      manager when the credential was pushed to the device. This value is
-      reported as nanoseconds since epoch (January 1st, 1970 00:00:00 GMT).
-      This leaf persists through a reboot.";
```
brianneville, line 87: With this definition removed, are the leafs still expected to persist through a reboot?

earies: This is just removal of the typedef that has no reuse but I think your point is that this last sentence was not carried into the leaf nodes.  Does this need to be explicitly called out?  Seems to me this doesn't need to be called out esp. in correspondance to prior comments as a "certificate store" is expected to be persistent

dplore: I also thought it was obvious that the `certificate-expires-on` leaf is not a date that is relative to uptime or reboots.  The date is related to the certificate, a non-volatile entity. 

However, if you could point out why there might be some ambiguity,  I am happy to add the text back in.

brianneville: for the `certificate-expires-on` field i agree that its obvious this is static information which is extracted from the cert, and not supplied by some other credential store (i.e. its the `notAfter` field)

For the other `<thing>-created-on` leafs that were using this `created-on` type, i'd say its best to retain the explicit clarification that these are both reported by the credential manager and expected to persist across reboot. 
This will keep the YANG file consistent with the proto:
https://github.com/openconfig/gnsi/blob/9b4ec7445b14b773940f1fc3b51351a6081c9c72/certz/certz.proto#L570-L571

Additionally (though this is more pedantic and less likely),its possible the `<thing>-created-on` fields are not necessarily something that we can pull out of the certificate, as the cert may be created with a `notBefore` time set to some timestamp in the past.

dplore: SGTM,  done.

brianneville: Can we re-add the 
"The creation time of the credential as reported by the credential manager when the credential was pushed to the device." line as well? 

unless there are some other references to this value being pushed via the credential manager, or you feel that it would be obvious to readers that this value comes from a credential manager?

dplore: How about I align it closer to the proto definition?
https://github.com/openconfig/gnsi/blob/11f299940854ca24e6f06f46253f36017efb0a31/certz/certz.proto#L564-L573

The proto text I think is more clear than the old text in the yang.   WDYT?

ManikBhasin: I believe this update should be applied consistently across all gNSI verticals, for example:

* [[openconfig-gnsi-pathz.yang](https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-pathz.yang#L66)](https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-pathz.yang#L66)
* [[openconfig-gnsi-credentialz.yang](https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-credentialz.yang#L96)](https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-credentialz.yang#L96)
* [[openconfig-gnsi-authz.yang](https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-authz.yang#L69)](https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-authz.yang#L69)

However, I don’t see a need to remove the existing `created_on` typedef
A simple update specifying that the timestamp is represented in **nanoseconds** should be sufficient. The corresponding proto definitions should also reflect this for consistency.

Some relevant examples from the proto files:

* [[certz.proto#L572](https://github.com/openconfig/gnsi/blob/main/certz/certz.proto#L572)](https://github.com/openconfig/gnsi/blob/main/certz/certz.proto#L572)
* [[pathz.proto#L169](https://github.com/openconfig/gnsi/blob/main/pathz/pathz.proto#L169)](https://github.com/openconfig/gnsi/blob/main/pathz/pathz.proto#L169)

PR #1378, opened 2025-10-08
Add GLOME YANG paths

zasweq:
Add GLOME YANG paths for Telemetry. The descriptions and leaves were already defined, just not linked to YANG paths.

The paths are:
/system/aaa/authentication/glome/state/enabled
/system/aaa/authentication/glome/state/active-glome-key-version
/system/aaa/authentication/glome/state/active-glome-key-created-on

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,4 +11,4 @@
 
 # subfolders may have additional codeowners who
 # may merge content
-/release/models/wifi @mike-albano
+/release/models/wifi @jsnyder81 @zpereyo
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,5 @@
-[Note: Please fill out the following template for your pull request. lines
-tagged with "Note" can be removed from the template.]
+[Note: Please fill out the following template for your pull request. Replace
+all the text in `[]` with your own content.]
 
 [Note: Before this PR can be reviewed please agree to the CLA covering this
 repo. Please also review the contribution guide -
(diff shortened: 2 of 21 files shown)


zasweq:
More than three vendors are implementing GLOME and this model.

zasweq:
Sounds good, updated branch :).

release/models/gnsi/openconfig-gnsi-credentialz.yang
```diff
@@ -326,7 +337,15 @@ module openconfig-gnsi-credentialz {
   }
 
   // Augments section.
-
+  augment "/oc-sys:system/oc-sys:aaa/oc-sys:authentication/oc-sys:glome/" +
```
dplore, line 340: This augment looks good.  It includes the grouping into the /system/aaa/authentication/glome/state` tree.

However, If you want this to ALSO appear in the 	`/access-points/access-point/system/aaa/authentication/glome/state` tree, then you will need an additional augment statement for that  I think.

zasweq:
The PYANG tree looks like:
image

release/models/gnsi/openconfig-gnsi-credentialz.yang
zasweq: The design that our team wrote and you approved only mentions "/system/aaa/authentication/glome/state" this path. Do we need the access-points path? I had to add the /system/aaa/authentication/glome/state to augment into.

dplore: Yes, please also add an augment statement for the access-points, so we don't have an empty container there.

Something like this should do it:

```
augment "/oc-ap:access-points/oc-ap:access-point/oc-sys:system/oc-sys:aaa/oc-sys:authentication/oc-sys:glome/" 
        "oc-sys:state" {
    when "../../../../oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server" +
        "/oc-sys-grpc:config[contains(oc-sys-grpc:services, 'oc-gnsi:GNSI')]" +
        "/oc-sys-grpc:enable = 'true'";
    description
        "glome freshness information.";
    uses glome-key-version;
  }
```

PR #1380, opened 2025-10-09
Generalize storage counters in OpenConfig for broader support.

AnilKR123:
### Change Scope
Generalizes the storage counter descriptions to support a wider range of storage technologies beyond SMART specific metrics.

### DIFF
--- a/release/models/qos/openconfig-qos-elements.yang
+++ b/release/models/qos/openconfig-qos-elements.yang
@@ -35,7 +35,14 @@ submodule openconfig-qos-elements {
       packets for transmission, including policer and shaper
       functions";
 
-  oc-ext:openconfig-version "0.11.2";
+  oc-ext:openconfig-version "0.12.0";
+
+  revision "2025-10-13" {
+    description
+      "Correct percentage types for min-threshold-percent and
+      max-threshold-percent leafs";
+    reference "0.12.0";
+  }
 
   revision "2023-10-13" {
     description
--- a/release/models/qos/openconfig-qos-interfaces.yang
+++ b/release/models/qos/openconfig-qos-interfaces.yang
@@ -25,7 +25,14 @@ submodule openconfig-qos-interfaces {
     configuration and operational state associated with
     interfaces.";
 
-  oc-ext:openconfig-version "0.11.2";
+  oc-ext:openconfig-version "0.12.0";
+
+  revision "2025-10-13" {
+    description
+      "Correct percentage types for min-threshold-percent and
+      max-threshold-percent leafs";
+    reference "0.12.0";
+  }
 
   revision "2023-10-13" {
     description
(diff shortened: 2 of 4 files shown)


release/models/platform/openconfig-platform-storage.yang
```diff
@@ -110,25 +123,32 @@ oc-ext:origin "openconfig";
     leaf life-left {
       type uint8;
```
rgwilton: Should we change this to type oc-types:percentage?

release/models/platform/openconfig-platform-storage.yang
```diff
@@ -110,25 +123,32 @@ oc-ext:origin "openconfig";
...
-        "SMART attribute details,
-        https://media.kingston.com/support/downloads/MKP_306_SMART_attribute.pdf";
+        "SMART attribute details:
+        https://media.kingston.com/support/downloads/MKP_306_SMART_attribute.pdf
+        eMMC lifecycle details:
+        https://www.kingston.com/en/blog/embedded-and-industrial/emmc-lifecycle";
     }
 
     leaf percentage-used {
       type uint8;
```
rgwilton, line 147: Should we change this to type oc-types:percentage?

earies: The only gotcha here appears in the description which indicates values > 100 are possible and have meaning.  What that meaning is is not clear to me.

dplore: I think we can deduce that values greater than 100 show that the device has been used more than the manufacturer's estimated lifespan.

```
percentage of device life used based on actual usage and the manufacturer’s prediction of device life.
``

rgwilton: Okay, then perhaps change the units to percentage, but leave the type as uint8?

earies: > I think we can deduce that values greater than 100 show that the device has been used more than the manufacturer's estimated lifespan.
> 
> ```
> percentage of device life used based on actual usage and the manufacturer’s prediction of device life.
> ``
> ```

Yea - the mitigation/action i'd imagine is still the same where one is going to ideally address/threshold prior to reaching 100... reporting > 100 is more truthful marshaling of the raw data (up to 255 at least) but I'd also question if it is necessary.  It's about a proper signal and we could abstract this a bit if we want..... also something that could contribute to boolean signals such as healthz.

As I see it, anything > 100 increases risk and we can just leave this the raw value, keep the type, change the units as @rgwilton suggests as well.

dplore: @AnilKR123 adding  units percentage sounds like a good idea (but leaving this at uint8)

AnilKR123: Agreed. I added units percentage to life-left and percentage-used as suggested.

release/models/platform/openconfig-platform-storage.yang
dplore: It would be better to use oc-types:percentage here, as the SMART standard does seem to strictly specify a range of 0-100.

https://media.kingston.com/support/downloads/MKP_306_SMART_attribute.pdf

dplore: @AnilKR123 can you make this change please?

AnilKR123: Updated life-left to use oc-types:percentage as suggested.

PR #1385, opened 2025-10-29
Add gNPSI module

ElodinLaarz:
The main changes are focused on updating module dependencies, versioning, and clarifying the GNPSI identity.

**Module dependency updates:**

* Changed the imported module from `openconfig-grpc-types` (prefix `oc-grpct`) to `openconfig-system-grpc` (prefix `oc-grpc`) to reflect the new location of the `GRPC_SERVICE` identity.

**Versioning and documentation:**

* Bumped the `openconfig-version` to "0.1.1" and added a new revision entry for 2025-10-29, describing the update and referencing the new version.

**GNPSI identity changes:**

* Updated the `GNPSI` identity to use the new base (`oc-grpc:GRPC_SERVICE`) and clarified its description to "gNPSI: gRPC Network Packet Sampling Interface."


release/models/gnpsi/openconfig-gnpsi-types.yang
```diff
@@ -33,6 +39,6 @@ module openconfig-gnpsi-types {
   identity GNPSI {
     base oc-grpct:GRPC_SERVICE;
```
earies: This was called out some time back as well but never fixed but this identity is derived from an unused base identity so there is a disconnect that breaks usage today.

The plan is that there would be a transition to the types model but there is no use of this in any identityref's today.

Not sure how this could be functional for anyone today

dplore:
```suggestion
    base "oc-grpc:GRPC_SERVICE";
```

release/models/gnpsi/openconfig-gnpsi-types.yang
dplore:
```suggestion
  import openconfig-system-grpc { prefix oc-grpc; }
```

release/models/gnpsi/openconfig-gnpsi-types.yang
ElodinLaarz: Yeah, I found this out when I was reviewing https://github.com/openconfig/featureprofiles/pull/4445.

Doesn't seem like it does anything, ATM? Was discussing whether we should fix it. But the description change was innocuous enough that I was gonna submit a PR.

dplore: Yes, this is a bug in the model.  I created some  suggested changes just now

ElodinLaarz: Do we still want the file name to be `-types`? Added a file to match the other `g*` protocols.

earies: I know it's not related to the initial intent of your PR but yea we have to make a decision here if we want to fix this now....

The thoughts were as part of gRPC service restructuring (which is a very old PR that needs revisiting) that services be broken out from a modeling perspective, carry augmentations, define their own types (inline or separate modules for re-use) and overall identities for those services and the gRPC base modeling be readjusted to accomodate.

One way is to modify `openconfig-system-grpc` for where the identityref is defined today to that of a union (This doesn't break payloads but could affect tooling) then we take a transitional approach while eventually peeling off the prior definitions.

We could also leave this for another PR and readdress the gRPC service restructuring as a whole.

release/models/gnpsi/openconfig-gnpsi-types.yang
earies: By doing this, we are reversing the intent of the approach that has yet to come to fruition rendering `openconfig-grpc-types` then unused.  As I mentioned in the other comment, we can choose to fix this now or later as part of the more holistic approach.

release/models/gnpsi/openconfig-gnpsi.yang
dplore, line 1: this file is not needed based on other suggested changes.

release/models/gnpsi/openconfig-gnpsi-types.yang
dplore: I understand.  We should follow the existing pattern (suggested in the review comments) for this PR and do gRPC service restructuring in a separate PR.

release/models/gnpsi/openconfig-gnpsi-types.yang
dplore: I'm also uncertain of the tools impact there.  I'd prefer to follow the existing pattern and do the gprc services refactor in it's own, standalone PR.      The current PR does not make that effort materially more difficult.

dplore: > Do we still want the file name to be `-types`? Added a file to match the other `g*` protocols.

It is more clean to rename the file to openconfig-gnpsi.yang.  You'll need to update the spec file in this PR as well.  There is also a downstream impact to tools which are importing this -types file.  (ondatra)

dplore (approved):

PR #1389, opened 2025-11-09
Adding initial draft of the Containerz yang model

alshabib:
### Change Scope

This is a yang model for the gNOI.Containerz service.


release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,139 @@
...
+    when "../config[contains(services, 'oc-containerz:CONTAINERZ')]/enable = 'true'";
+    description
+      "A gRPC server containerz freshness information.";
+
+    uses grpc-server-containerz-state;
+  }
+
+  augment "/oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server/" +
+          "oc-sys-grpc:state" {
+    when "../config[contains(services, 'oc-containerz:CONTAINERZ')]/enable = 'true'";
```
earies: This identity does not exist.  For new gRPC services, there is current way and potential future for expansion but you'll want to at a min follow what you should see for P4RT and GRIBI

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,139 @@
...
+
+
+  // gRPC server related definitions.
+  // Success/failure counters.
+  grouping grpc-server-containerz-counters  {
+    description
+      "A collection of counters that were collected while evaluating
+      access to the gRPC server.";
+
+    container counters {
```
earies: This appears to try to add an overlapping container under `/system/grpc-servers/grpc-server/state`.  Suggest consideration of bucketizing "containerz" specific state/counters under it's own hierarchy specific to this set of services

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,139 @@
...
+      config false;
+      description
+      "A collection of counters that were collected by the gRPC during
+      the authentication process.";
+
+      leaf connection-rejects {
```
earies: What are these connection-* related leafs meant to convey or were they copied from other model?

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,139 @@
...
+        "The number of containers currently running on the system.";
+      }
+    }
+  }
+
+  grouping grpc-server-containerz-state {
+    description
+      "gRPC server containerz freshness-related data.";
+
+    container containers {
```
earies, line 82: Similar suggestion as above - consider grouping under `containerz` hierarchy first.  If this is meant to be similar to `docker ps` outputs then can align as such.  Any counters should go under child `counters` container at relevant hierarchy.

release/models/containerz/openconfig-containerz.yang
robshakir: We don't allow for this kind of `when` statement using `contains` AFAIK.

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,139 @@
...
+      list container {
```
robshakir, line 83: I see this as duplicative to the `ListContainer` RPC, which is fine in principle, but how do we expect the two to be used? I presume the thinking here is that some clients will just subscribe through gNMI to understand this state?

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,139 @@
...
+        "The total number of times that gRPC clients have succeeded
+        in establishing a connection to the server.";
+      }
+      leaf last-connection-accept {
+        type oc-types:timeticks64;
+        description
+        "A timestamp of the last time a gRPC client succeeded
+        in establishing a connection to the server.";
+      }
+      leaf running-containers {
```
robshakir: Is this a useful metric if we also have the list of containers?

release/models/containerz/openconfig-containerz.yang
earies: This concept was introduced in the gNSI related models

https://github.com/openconfig/public/blob/master/release/models/gnsi/openconfig-gnsi-authz.yang#L205

But this identity does not exist and each gRPC service domain should really bring in a distinct related container and group everything related underneath (commented above)

release/models/containerz/openconfig-containerz.yang
dplore: They seem to be a duplicate of the counters in 	`/system/grpc-servers/grpc-server/connections/connection/state/counters`.   I think they are not needed because they are already present.

`/system/grpc-servers/grpc-server/containerz/state/counters` could be created if there are containerz specific counters.  But all generic "grpc-servers" counters regarding client connections should appear in:
`/system/grpc-servers/grpc-server/connections/connection/state/counters`

release/models/containerz/openconfig-containerz.yang
dplore: @alshabib see:

[/system/grpc-servers/grpc-server/connections/connection/state/counters](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-grpc-servers-grpc-server-connections-connection-state-counters)

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,139 @@
...
+
+  grouping grpc-server-containerz-state {
+    description
+      "gRPC server containerz freshness-related data.";
+
+    container containers {
+      list container {
+        key "name";
+
+        leaf name {
```
dplore, line 85: For openconfig style compliance, this needs to be a leafref to a leaf of the same name in a  "state container"

for example:
https://github.com/openconfig/public/blob/1463f23ad71310b4f297db82cf737e75b51c48f8/release/models/system/openconfig-system-grpc.yang#L75-L129

release/models/containerz/openconfig-containerz.yang
dplore, line 82: Consider the same fields from a `docker ps` (or `docker ls` if the intent is only to show running containers)?

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,139 @@
...
+            "The version of the of the container. This corresponds to the
+             tag of of the image associated to this container.";
+        }
+      }
+    }
+  }
+
+  // Augments section.
+
+  augment "/oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server/" +
```
brianneville: I'm not sure if we really should be augmenting the grpc-servers list with information about something which is system-level (i.e. containers are not tied to a particular gRPC server in any way).
maybe something like `/system/containers` would be more appropriate?

brianneville: (likewise for the augment below)

dplore:
FYI: set project review status to "waiting for author".

release/models/containerz/openconfig-containerz.yang
alshabib: I am dropping these counters - they don't bring much value tbh.

release/models/containerz/openconfig-containerz.yang
alshabib: Added the identity

release/models/containerz/openconfig-containerz.yang
alshabib: Yes - the idea is to export base container telemetry to such that we do not depend on containers to export base telemetry that may cause a circular dependency.

release/models/containerz/openconfig-containerz.yang
alshabib: It's gone ;)

release/models/containerz/openconfig-containerz.yang
alshabib: I remove the other container here.

release/models/containerz/openconfig-containerz.yang
alshabib: That is not entirely possible as restarts or build-label are not directly returned from ps

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        leaf build-label {
+          type string;
+          description
+            "This build-label of the container. This corresponds to the
+            build label associated to the container if any.";
+        }
+
+        leaf version {
```
brianneville: should this be named something like `tag` instead of `version`?

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,149 @@
...
+      uses containerz-config;
+    }
+  }
+  identity CONTAINERZ {
+    base "oc-sys-grpc:GRPC_SERVICE";
+    description
+      "Containerz: gRPC Container Lifecycle Management";
+  }
+
+  augment "/oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server/" +
```
brianneville: im not sure if we should be augmenting a particular grpc-server with this config/state - this config/state is applicable moreso to the underlying container runtime (docker/podman), which is singular and independent of any particular grpc-server.

E.g. if you had two grpc-servers and each one had a containerz service running to satisfy the `when` constraint, then each grpc-server was configured with a different `vrf`, i think it would be unclear what VRF we should run the containers in, as both containerz services would be communicating with the same runtime.

Would it make sense to have this be augmented underneath `/oc-sys:system/` instead? 
e.g. you could have `/system/containerz/config` and `/system/containerz/state`.

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,149 @@
...
+          "oc-sys-grpc:state" {
```
brianneville: this model is currently putting everything underneath `/system/grpc-server/grpc-server/state`, including the config for containerz. Usually we wouldnt have config under state like this. 

I'd say these new paths that you're adding should be underneath `/system/grpc-server/grpc-server` directly, if you do want to keep them on a per-grpc-server basis. 
Or they should be underneath `/system` (see my other coment above https://github.com/openconfig/public/pull/1389/changes#r2832985674)

You can see this with pyang, shown below:
```sh
bash-3.2$ v=$(pwd);echo github${v#*github}
github.com/openconfig/public/release/models
bash-3.2$ pyang -p .. -f tree containerz/openconfig-containerz.yang
containerz/openconfig-containerz.yang:143: warning: "oc-containerz" looks like a prefix but is not defined
module: openconfig-containerz

    +--ro state
    |  +--ro containers
    |     +--ro container* [name]
    |        +--ro name           string
    |        +--ro created-on?    oc-types:timeticks64
    |        +--ro restarts?      oc-yang:counter64
    |        +--ro build-label?   string
    |        +--ro version?       string
    +--ro config
       +--ro vrf?                 string
       +--ro soft-memory-limit?   uint64
       +--ro hard-memory-limit?   uint64
       +--ro cpu-limit?           oc-types:percentage
```

release/models/containerz/openconfig-containerz.yang
brianneville: nit: if you're using this when statement, `oc-containerz:CONTAINERZ` should be `oc-gnoi-containerz:CONTAINERZ` to match the module prefix `oc-gnoi-containerz` you defined at the top of the file here

release/models/containerz/openconfig-containerz.yang
alshabib: done - thanks!

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,149 @@
...
+
+      uses containerz-state;
+    }
+
+    container config {
+    config true;
+    description
+      "Configuration data for Containerz";
+
+      uses containerz-config;
```
brianneville: this config is going to be directly underneath the `/oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server` path - should we put it in a container so that it doesnt appear to be config for the gprc-server specifically? 

e.g. right now the `vrf` leaf is at `/system/grpc-servers/grpc-server[name=X]/config/vrf`, and based on the path people could be misled into thinking that this should be the VRF for the grpc-server itself, rather than for the container runtime. 
something like:
`/system/grpc-servers/grpc-server[name=X]/containerz/config/vrf`
would make it more clear  that this is not for the grpc-server specifically

brianneville: I still dont think this should be on a per-grpc-server basis though - could you take a look at https://github.com/openconfig/public/pull/1389#discussion_r2832985674 ? 
having this config be under a global `/system` path (e.g. `/system/containerz/config` and `/system/containerz/state`) would make a lot more sense imo, and remove the ability for the client to push config that is conflicting across each grpc-server.

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,158 @@
...
+    description
+      "Initial revision.";
+    reference "0.1.0";
+  }
+
+  grouping containerz-config {
+    description
+      "Containerz gRPC server configuration data.";
+
+    leaf enable {
```
LimeHat, line 42: I don't think this is needed since you introduce a new gRPC service identity (which enables/disables the service for a given grpc-server). Unless the implication here is that you want to decouple containerz from the grpc  (which is not a bad idea in principle), but in that case it needs to be described in more detail

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,158 @@
...
+
+    leaf hard-memory-limit {
+      type uint64;
+      units megabytes;
+      description
+        "The amount of memory to limit the entire container workload.";
+    }
+
+    leaf cpu-limit {
+      type oc-types:percentage;
```
LimeHat, line 71: I would suggest changing this to `cpu.weight` and `cpu.max` (in cgroupv2 terms and units).

While the current cpu-limit (expressed in percentage points) can be translated by the implementation into the cpu.max limit, this creates ambiguity/platform dependency (cpu-limit on a 4-core system would be vastly different from the cpu-limit on a 32-core system)

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,158 @@
...
+          type string;
+          description
+            "This build-label of the container. This corresponds to the
+             build label associated to the container if any.";
+        }
+
+        leaf tag {
+          type string;
+          description
+            "The version of the container. This corresponds to the
```
LimeHat, line 113: Hmm. So we're exposing the image tag, but not the image name?

LimeHat (changes requested):

release/models/containerz/openconfig-containerz.yang
robshakir: +1, I don't think that this is needed -- if we want to enable containerz on a particular gRPC server then it can just be added to `/system/grpc-servers/grpc-server/config/services`.

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,158 @@
...
+      "Containerz gRPC server configuration data.";
+
+    leaf enable {
+      type boolean;
+      default false;
+      description
+        "Enable or disable the Containerz service on the system.";
+    }
+
+    leaf vrf {
```
robshakir, line 49: This should be for the gRPC-server -- particularly, we have `/system/grpc-services/grpc-server/config` which specifies the port. This should probably have (at this level) an argument as to what `network-instance` this is enabled for. 

The naming here should be `network-instance` rather than `vrf`.

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,158 @@
...
+      type uint64;
+      units megabytes;
+      description
+        "The amount of memory to limit the entire container workload.";
+    }
+
+    leaf cpu-limit {
+      type oc-types:percentage;
+      description
+        "The percentage of system cpu to allocate to the entire container
```
robshakir, line 73:
```suggestion
        "The percentage of system CPU to allocate to the entire container
```

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,158 @@
...
+         workload.";
+    }
+  }
+
+  grouping containerz-state {
+    description
+      "Containerz base container operational data.";
+
+    container containers {
```
robshakir, line 82: Is there ever a case where we have anything that is configurable per container through this API? If so, then we probably should have it as `/system/grpc-servers/grpc-server/containerz/containers/container` (the list) and then `config` and `state` underneath it.

If not -- should we move this away from being under `/system/grpc-servers`. It seems like we might just want to have `/system/contaners/container` (list) and use this to list any containers that are running (which might have been started via some API other than containerz). WDYT?

release/models/containerz/openconfig-containerz.yang
```diff
@@ -0,0 +1,158 @@
...
+      list container {
+        key "name";
```
brianneville, line 84: the system may have containers running on the active and/or backup supervisor (https://github.com/openconfig/gnoi/blob/b2cfe7eb561014631a4f4878fe78bfe7b6a0ab83/containerz/containerz.proto#L558) 
I think we should use this yang model to report the containers across both supervisors. 

However, if a container is created with location = L_ALL, then it will have the same name on both supervisors, so just keying this list with name wouldnt be sufficient to distinguish between instances. 
We could make this list have multiple keys, to differentiate here.
The keys could be "name" (as-is currently, i.e. the name of the container), and "location" which would be an identityref for which the L_ALL, L_PRIMARY, L_BACKUP identities would be defined.

PR #1390, opened 2025-11-10
Refactor fec-uncorrectable-blocks relationship to PM interval

earies:
* (M) release/models/optical-transport/openconfig-terminal-device.yang
    - Change `fec-uncorrectable-blocks` to appropriate type and adjust
      description to clarify relation to PM interval sampling
    - Increment version to 1.10.0

### Change Scope

It was brought to our intention that various implementations have associated the
leaf node `fec-uncorrectable-blocks` to be a value that is reset according to an
internal PM interval (influenced via other means today outside of OC modeling).
This means that the current type and description are incorrect.

This change, while by nature is backwards incompatible is currently flagged as a
semver minor change `1.9.2` -> `1.10.0`.  The underlying type is still a
`uint64` but the description and type are changed to reflect current
implementation/intent.

This leaf is used under the following paths today:

```
/openconfig-platform:components/component/openconfig-platform-transceiver:transceiver/state/fec-uncorrectable-blocks
/openconfig-platform:components/component/openconfig-terminal-device:optical-channel/state/fec-uncorrectable-blocks
/openconfig-terminal-device:terminal-device/logical-channels/channel/ethernet/state/fec-uncorrectable-blocks
/openconfig-terminal-device:terminal-device/logical-channels/channel/otn/state/fec-uncorrectable-blocks
```

### Platform Implementations

Please comment if your implemention implements a cumulative counter value here
and not one that is associated with being reset at defined PM intervals.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,4 +11,4 @@
 
 # subfolders may have additional codeowners who
 # may merge content
-/release/models/wifi @mike-albano
+/release/models/wifi @jsnyder81 @zpereyo
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,5 @@
-[Note: Please fill out the following template for your pull request. lines
-tagged with "Note" can be removed from the template.]
+[Note: Please fill out the following template for your pull request. Replace
+all the text in `[]` with your own content.]
 
 [Note: Before this PR can be reviewed please agree to the CLA covering this
 repo. Please also review the contribution guide -
@@ -9,6 +9,7 @@ https://github.com/openconfig/public/blob/master/doc/contributions-guide.md]
 
 * [Please briefly describe the change that is being made to the models.]
 * [Please indicate whether this change is backwards compatible.]
+
 ### Platform Implementations
 
  * Implementation A: [link to documentation](http://foo.com) and/or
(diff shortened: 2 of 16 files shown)


release/models/optical-transport/openconfig-terminal-device.yang
```diff
@@ -607,10 +614,11 @@ module openconfig-terminal-device {
...
     leaf fec-uncorrectable-blocks {
-      type yang:counter64;
+      type yang:gauge64;
       description
-        "The number of blocks or frames that were uncorrectable by
-        the FEC";
+        "The number of blocks or frames that were uncorrectable by the
+        FEC. The reported value is the cumulative count over the prior
+        Performance Monitoring (PM) interval defined by the host.";
     }
```
dplore, line 628: we could add a reference here:

```
    reference "OIF-CMIS-05.3  section 7.1.5 
    https://www.oiforum.com/wp-content/uploads/OIF-CMIS-05.3.pdf";
```

earies: Thx - added in https://github.com/openconfig/public/pull/1390/commits/6f62c11ae0a5026bc9d18d93e716d9a644cf1dc6

ElodinLaarz:
Reviewed at the OC Operators Meeting on Nov 11: seems reasonable based upon the spec. It also seems prudent to do a review of other leaves in this container (outside the scope of this PR) to see if this is happening elsewhere.

release/models/optical-transport/openconfig-terminal-device.yang
yiwenhu-g: Please update this CMIS reference as only "for pluggable or on-board modules", as this OC leaf was initially introduced for DCI

earies: Thx - this is updated in https://github.com/openconfig/public/pull/1390/commits/7eb8c2599042d9987a70c4d07376aed449dca2a7 - Please confirm

dplore (approved):

PR #1395, opened 2025-11-14
Add MPLS static LSP  nh-network-instance

dplore:
### Change Scope

*  Add a reference for next-hop network-instance for MPLS static LSP's.
* Add configuration and state parameters for a MPLS static LSP to point to a next-hop-group.
*  This change is backwards compatible.

### Platform Implementations

 * [Arista EOS static LSP](https://www.arista.com/en/um-eos/eos-mpls-commands#xx1141474) supports configuration of a destination network instance
 * [Cisco XROS](https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/asr9k_r5-3/mpls/configuration/guide/b-mpls-cg53x-asr9k/b-mpls-cg53x-asr9k_chapter_011.html#task_80068627902047E994E38FE60C4E6441) supports this feature but with a different configuration by mapping a static MPLS label to an IP prefix for a specific VRF
 

### Reference operational use case 
See [featureprofiles#4538](https://github.com/openconfig/featureprofiles/pull/4538/files/2321db45e640f685825e3a4874e0c85f7a317fa7..e3a4b8321400029a0d7c382bfaa3b84cda4cb598)

### Tree view
```
pyang -f tree -p release/models/*/* > ~/new-tree.txt 
diff -bU 100 ~/old-tree.txt ~/new-tree.txt
```

```diff
         |           +--rw egress
         |              +--rw config
         |              |  +--rw incoming-label?              oc-mplst:mpls-label
         |              |  x--rw next-hop?                    inet:ip-address
         |              |  x--rw push-label?                  oc-mplst:mpls-label
         |              |  +--rw interface?                   -> /oc-if:interfaces/interface/name
         |              |  +--rw subinterface?                -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/
index
         |              |  +--rw metric?                      uint8
+        |              |  +--rw oc-mpls-ni:next-hop-group
+        |              |     +--rw oc-mpls-ni:config
+        |              |     |  +--rw oc-mpls-ni:name?   -> ../../../../../../../../../oc-ni:static/next-hop-groups/next-hop-group/config/name
+        |              |     +--ro oc-mpls-ni:state
+        |              |        +--ro oc-mpls-ni:name?   -> ../../../../../../../../../oc-ni:static/next-hop-groups/next-hop-group/config/name
         |              +--ro state
         |              |  +--ro incoming-label?   oc-mplst:mpls-label
         |              |  x--ro next-hop?         inet:ip-address
         |              |  x--ro push-label?       oc-mplst:mpls-label
         |              |  +--ro interface?        -> /oc-if:interfaces/interface/name
         |              |  +--ro subinterface?     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
         |              |  +--ro metric?           uint8
         |              +--rw lsp-next-hops
         |                 +--rw lsp-next-hop* [index]
         |                    +--rw index     -> ../config/index
         |                    +--rw config
         |                    |  +--rw index?                            uint32
         |                    |  +--rw ip-address?                       inet:ip-address
         |                    |  +--rw push-label?                       oc-mplst:mpls-label
         |                    |  +--rw interface?                        -> /oc-if:interfaces/interface/name
         |                    |  +--rw subinterface?                     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/su
binterface/index
+        |                    |  +--rw oc-mpls-ni:nh-network-instance?   -> /oc-ni:network-instances/network-instance/config/name
         |                    +--ro state
         |                       +--ro index?                            uint32
         |                       +--ro ip-address?                       inet:ip-address
         |                       +--ro push-label?                       oc-mplst:mpls-label
         |                       +--ro interface?                        -> /oc-if:interfaces/interface/name
         |                       +--ro subinterface?                     -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/su
binterface/index
+        |                       +--ro oc-mpls-ni:nh-network-instance?   -> /oc-ni:network-instances/network-instance/config/name
```


trlongth (approved):

release/models/mpls/openconfig-mpls-network-instance.yang
```diff
@@ -0,0 +1,132 @@
...
+
+        For example, let 'nh-network-instance' = 'DEFAULT'.  If a
+        packet arrives on the 'BLUE' network-instance and the
+        destination address matches this route, look up the IP
+        configured as next-hop in the 'DEFAULT' network-instance.
+        This leaf is mutually exclusive with next-network-instance
+        leaf";
+
+      type leafref {
+        path "/oc-ni:network-instances/oc-ni:network-instance/oc-ni:config/oc-ni:name";
```
earies, line 61: We have a typedef of `network-instance-ref` already for this however it sits directly in `openconfig-network-instance` which I suspect would cause potential recursive import issues in some cases.

Since this is augmenting into that model and already imported, this typedef could be used here instead

release/models/mpls/openconfig-mpls-network-instance.yang
```diff
@@ -0,0 +1,132 @@
...
+    description
+      "Grouping of configuration parameters for a reference to a
+      static next-hop-group.";
+
+    leaf name {
+      description
+        "A user defined name to reference a static next-hop-group.";
+      // we are at /network-instances/network-instance/mpls/lsps/static-lsps/static-lsp/egress/next-hop-group/config/name
+      type leafref {
+        path "../../../../../../../../../oc-ni:static/oc-ni:next-hop-groups/oc-ni:next-hop-group/oc-ni:config/oc-ni:name";
```
earies: This relative path reference seems to negate reusable groupings unless it can be ensured its reuse is always at the same depth of a tree.  Maybe a absolute path reference rather and if targeting is necessary then maybe XPath functions like `current()` are in order? (much like in the reusable `interface-ref-common`)

release/models/mpls/openconfig-mpls-network-instance.yang
dplore: Yes, our pattern for preventing recursive imports when referencing network-instance within a network-instance ancestor is to create a <entity>-network-instance.yang and augment.   So yes, we can't import `openconfig-network-instance` where that typedef resides.

Are you recommending creating another `typedef network-instance-ref` within this model file versus what is here?

release/models/mpls/openconfig-mpls-network-instance.yang
dplore: Are you proposing a path like this?
```
      type leafref {
        path "/oc-ni:network-instances/oc-ni:network-instance[name=current()]" +
        "/oc-ni:static/oc-ni:next-hop-groups/oc-ni:next-hop-group/oc-ni:config/oc-ni:name
      }
```
I'm not sure I see a reuse case on the horizon, but this is easier to read and works.  Let me know if this is what you were looking for.

release/models/mpls/openconfig-mpls-network-instance.yang
earies: In this case, we could just directly use the typedef that is in `openconfig-network-instance` but going fwd, I think the better solution to not run into recursive import issues is to put all such definitions in domain "types" modules - any domain can then import another domains type module that doesn't contain direct data trees.

So we wouldn't create another typedef here but we could in `openconfig-network-instance-types`

release/models/mpls/openconfig-mpls-network-instance.yang
earies: yep that is precisely what I was thinking and agree, it's always hard to determine how much reuse there will be in reality up front.  Some design patterns however will negate reuse even though the intention might have been there to collocate in groupings

Issue #1376, opened 2025-09-29
YANG Atomic extensions and behavioral expectations


earies:
While there is still unmerged work and discussion surrounding YANG atomic extensions here: https://github.com/openconfig/public/pull/1330

As well as prior discussion on the relationship between the YANG atomic extension and gNMI atomic field here: https://github.com/openconfig/gnmi/issues/180

If we use the OC AFT modeling as an example of atomic datasets

* There is 0 mention of atomic behavior expectations from a gNMI specification perspective
* These datasets are not marked as `oc-ext:telemetry-on-change` (Similar to what is described in https://github.com/openconfig/public/issues/1375)
* There is no description of expectations should one choose to query/subscribe to a subset of data under an atomic marked container/list

The intention of this issue is to spark discussion to come to consensus on expectations when

* Subscriptions target subsets of data under atomic flagged content (deliver subset, error, or entire atomic flagged content)
* YANG containers/lists are marked as atomic, agreement that this corresponds to a case when gNMI marks `atomic = true`

ncorran:
Cisco-XR has implemented OC-AFT on-change, with support for 'atomic' (in at least some of the meanings of atomic and ignoring the telemetry-on-chage gap, as per comments on the linked PR).

I'll cross-check our interpretations and see how that fits with this and linked discussions and update back here.

PR #1400, opened 2025-11-18
Mark /system/alarms list entries as atomic

earies:
* (M) release/models/system/openconfig-alarms.yang
    - Add telemetry-atomic to alarm list
    - Increment version to 0.4.0

### Change Scope

For reference, the current `/system/alarms` subtree:

```
module: openconfig-system
  +--rw system
     +--ro alarms
        +--ro alarm* [id]
           +--ro id        -> ../state/id
           +--ro config
           +--ro state
              +--ro id?             string
              +--ro resource?       string
              +--ro text?           string
              +--ro time-created?   oc-types:timeticks64
              +--ro severity?       identityref
              +--ro type-id?        union
```

This PR does not propose any structural or type changes but rather introduces
the `oc-ext:telemetry-atomic` annotation to the `alarm` list structure to align
w/ current expectations and implementations.

### Platform Implementations

* Juniper JUNOS/EVO: Current behavior

### DIFF
--- a/release/models/system/openconfig-alarms.yang
+++ b/release/models/system/openconfig-alarms.yang
@@ -42,7 +42,13 @@ module openconfig-alarms {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "0.3.2";
+  oc-ext:openconfig-version "0.4.0";
+
+  revision "2025-11-18" {
+    description
+      "Mark alarm list entries as telemetry-atomic";
+    reference "0.4.0";
+  }
 
   revision "2019-07-09" {
     description
@@ -176,6 +182,7 @@ module openconfig-alarms {
       config false;
 
       list alarm {
+        oc-ext:telemetry-atomic;
         key "id";
         description
           "List of alarms, keyed by a unique id";


release/models/system/openconfig-alarms.yang
```diff
@@ -176,6 +182,7 @@ module openconfig-alarms {
       config false;
 
       list alarm {
+        oc-ext:telemetry-atomic;
```
rgwilton, line 185: Have we reached a conclusion on what the definition & behaviour is?

rgwilton: And in addition, we need to ensure that when applied to a list that the context is consistent.  I.e., it is individual entries of the list that are atomic, not the entire list itself.

jsterne: definition/behavior is in a separate issue (it does need to be clarified, but maybe it is OK to go ahead and identify this as a reasonable place to use atomic in parallel?).

Good point about the list. I can't see a way for this extension to distinguish between:
1) please return the entire list atomically (every list entry and their contents), vs
2) please return any individual entry of the list atomically

In this case (and I think most cases) we'd want behavior #2. Maybe this is something to also raise back in the issue around generically clarifying that atomic extension.

earies: Agreed - good points and yes, this PR is to at least consistently tag the obvious structs we all likely already treat as atomic (There may be more but this was an obvious one similar to that of syslog messages).

For issue: https://github.com/openconfig/public/issues/1376 and PR: https://github.com/openconfig/reference/pull/229, let me know if you agree and will add verbiage around how data is intended to be packed based on where the extension is annotated in the tree:

* If at the list, the expectation would be the entirety of a single list entry is to be returned atomically
* If at the container (whether or not there are child lists, etc..), all children are to be returned atomically

I believe this aligns w/ (2) above.

And another thing that came to mind is I believe most discussions to date of atomic bundling have been centered around ON_CHANGE (e.g. if any node changes, bundle and send) but another clarification that may be needed is intentions around SAMPLE (e.g. subscribe to single node, list or parent container) - should a Notification always pack that data atomically (and thus set `atomic=true`)

ElodinLaarz: This will make the requirement that system alarm notifications (for each alarm id) must match the following form exactly:

```textproto
update: {
  timestamp: 1234567890
  prefix: {
    origin: "openconfig"
    elem: {
      name: "system"
    }
    elem: {
      name: "alarms"
    }
    elem: {
      name: "alarm"
      key: {
        key: "id"
        value: "alarm-id"
      }
    }
    target: "target"
  }

  ...

  ### ... // all the updates for the paths under system/alarms/alarm[alarm-id]/... e.g. ###
  update: {
    path: {
      elem: {
        name: "state"
      }
      elem: {
        name: "severity"
      }
    }
    val: {
      string_val: "openconfig-alarm-types:MAJOR"
    }
  }

  ...

  atomic: true
}
```

Presumably this is what we're requesting here?

earies: > Presumably this is what we're requesting here?

Exactly

cfyo:
As an operator we agree that this container is not very useful with only a subset of the leaves streaming out for a given alarm event.  It would be surprising if other operators do not concur with that sentiment.  If other operators can chime in and say if they agree or if not, if they can explain how they find a subset useful.

dplore:
I'm not hearing a strong case (no operational use case where lack of this annotation is causing a break) that this change needs to be made.      And since it is formally a breaking change I will close this PR and leave the container as is.  

If a use case comes up where atomic is really required for  operational functionality, please do reopen.

PR #1413, opened 2025-12-10
Add explicit bypass tunnel assignment to RSVP-TE interface protection

KyoyaTeramae:
### Change Scope
*   This PR enhances to support an RSVP-TE Fast Reroute (FRR) capability by introducing a mechanism to explicitly assign pre-configured LSPs as bypass tunnels. This is a common requirement for service providers who want deterministic failover paths for protected interfaces. When an interface running RSVP-TE with FRR enabled fails, traffic can be redirected onto specified bypass LSPs, which have their own explicit paths.
 
    To support this, the following changes are made:
    1.  A new `leaf-list` named `bypass-tunnel` is added to the RSVP interface protection configuration (`openconfig-mpls-rsvp.yang`). This allows one or more pre-configured TE tunnels to be designated as bypass LSPs.
    2.  To differentiate these bypass LSPs from regular LSPs, a new `BYPASS_P2P` identity is introduced under the `TUNNEL_TYPE` base (`openconfig-mpls-types.yang`).
    3.  The `description` for the tunnel `type` leaf has been updated accordingly (`openconfig-mpls-te.yang`).
 
*   This change is **backwards compatible** as it only introduces new optional data nodes and a new identity.
 
### Platform Implementations
*   **Cisco IOS-XR:** The `backup-path` command under the protected interface allows specifying one or more tunnel interfaces to act as bypass LSPs. This directly corresponds to the proposed `bypass-tunnel` leaf-list.
 
    https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/mpls/command/reference/b-mpls-cr-8k/b-mpls-te-commands.html#:~:text=backup%2Dbw-,backup%2Dpath%20tunnel%2Dte,-bandwidth%2Daccounting
 
```
mpls traffic-eng
 interface GigabitEthernet0/0/0/0
  backup-path tunnel-te 20020
  backup-path tunnel-te 30040
 !
 interface tunnel-te20020
  description NHOP_xxxxx
  ipv4 unnumbered Loopback0
  signalled-name NHOP_20020_xxxx
  priority 7 7
  destination xx.xx.xx.xx
  record-route
  path-option 50 explicit name NHOP_20020-EPath_50
 !
 interface tunnel-te30040
  description NNHOP_30040_xxxx
  ipv4 unnumbered Loopback0
  signalled-name NNHOP_30040_xxxx
  priority 7 7
  destination xx.xx.xx.xx
  record-route
  path-option 50 explicit name NNHOP_30040-EPath_50
```

 
*   **Drivenets DNOS:** The `bypass-tunnel` command under `protocols rsvp interface <interface-name> protection` is used to assign bypass LSPs. The bypass tunnel itself is also explicitly configured with `bypass` differentiated from a regular tunnel. This implicitly requires the proposed `BYASS_P2P` tunnel type.
 
```
protocols rsvp interface ge100-0/0/0 protection bypass-tunnel NHOP_XXXX
protocols rsvp interface ge100-0/0/0 protection bypass-tunnel NNHOP_XXXX

protocols rsvp tunnel NHOP_XXXX bypass admin-state enabled
protocols rsvp tunnel NHOP_XXXX bypass description NHOP_XXXX
protocols rsvp tunnel NHOP_XXXX bypass destination-address xx.xx.xx.xx
protocols rsvp tunnel NHOP_XXXX bypass primary path-options 5 path explicit NHOP_xxxx

protocols rsvp tunnel NNHOP_XXXX bypass admin-state enabled
protocols rsvp tunnel NNHOP_XXXX bypass description NNHOP_XXXX
protocols rsvp tunnel NNHOP_XXXX bypass destination-address xx.xx.xx.xx
protocols rsvp tunnel NNHOP_XXXX bypass primary path-options 5 path explicit NNHOP_xxxx
```
 
*   **Nokia SR OS:** The `bypass-only` command option for the configure router mpls lsp context explicitly defines the LSP as a dedicated manual configured bypass tunnel. This creates an exclusive LSP type that is differentiated from carrying regular service traffic, directly supporting the proposed BYPASS_P2P identity. When a manual path (ERO) is applied to this LSP type, it functions as a deterministic bypass tunnel.

 https://documentation.nokia.com/sr/24-7/7x50-shared/mpls/mpls-rsvp.html#ai9emdyo7f

```
configure
 router
  mpls
   path "bypass-path"
    hop 1 xx.xx.xx.xx strict
    hop 2 xx.xx.xx.xx strict
    no shutdown
   exit
   lsp "bypass-lsp" bypass-only
    to xx.xx.xx.xx
    primary "bypass-path"
    exit
    no shutdown
   exit
  exit
 exit
```

### Tree View
*   The following diff shows the addition of the `bypass-tunnel` leaf-list to the RSVP-TE interface protection configuration.
 
```diff
 module: openconfig-mpls
   +--rw network-instances
      +--rw network-instance* [name]
         +--rw mpls
            +--rw signaling-protocols
               +--rw rsvp-te
                  +--rw interface-attributes
                     +--rw interface* [interface-id]
                        +--rw protection
                           +--rw config
                              +--rw link-protection-style-requested?   identityref
                              +--rw bypass-optimize-interval?          uint16
+                             +--rw bypass-tunnel*                     -> ../../../../../../../lsps/constrained-path/tunnels/tunnel/name
```

### DIFF
--- a/release/models/mpls/openconfig-mpls-rsvp.yang
+++ b/release/models/mpls/openconfig-mpls-rsvp.yang
@@ -443,8 +443,12 @@ module openconfig-mpls-rsvp {
           "lsps/constrained-path/tunnels/tunnel/name";
       }
       description
-        "list of references to named tunnels to be used
-        as bypass LSPs.";
+        "List of references to named tunnels to be used as bypass
+        LSPs. This allows associating multiple bypass tunnels with
+        a single protected interface to provide different Merge
+        Points (Link/Node protection). In cases where multiple
+        valid bypass tunnels exist, the selection priority is
+        implementation-dependent.";
     }
     // to be completed, things like enabling link protection,
     // optimization times, etc.


release/models/mpls/openconfig-mpls-rsvp.yang
```diff
@@ -28,7 +28,14 @@ module openconfig-mpls-rsvp {
      parameters and LSP-specific configuration for constrained-path
      LSPs";
 
-  oc-ext:openconfig-version "4.0.1";
+  oc-ext:openconfig-version "4.0.2";
```
earies: For new backwards compat definitions, increment the minor version rather

e.g. `4.1.0`

release/models/mpls/openconfig-mpls-rsvp.yang
```diff
@@ -429,6 +436,16 @@ module openconfig-mpls-rsvp {
       // note: this is interface specific on juniper
       // on iox, this is global. need to resolve.
     }
+
+    leaf-list bypass-tunnel {
```
earies, line 440: This leafref currently can reference any tunnel but I believe what is desired here is to reference a pre-defined tunnel that is also of the new `type=BYPASS_P2P` - is this correct?

This can likely be accomplished by a `when` statement, otherwise each implementation will need to account for special logic - the prior is desirable imo

release/models/mpls/openconfig-mpls-rsvp.yang
ky-teramae: Thank you for the comment. The version has been updated (e.g., to 4.1.0) as per the suggestion.

release/models/mpls/openconfig-mpls-rsvp.yang
ky-teramae: Thank you for the comment.
You are correct that the current leafref is not explicitly constrained.
This design is intentional to support  inter-vendor compatibility:  Drivenets' implementation requires the BYPASS_P2P type to explicitly configure a bypass tunnel distinguished from a regular tunnel, but Cisco's does not do so with that specific type definition.
Therefore, to support both vendor implementations without enforcing a change in the Cisco's implementation, we want to keep the current proposal (relying on implementation-specific logic) for the time being, rather than applying the when constraint here.
Appreciate your understanding.

release/models/mpls/openconfig-mpls-types.yang
```diff
@@ -19,7 +19,13 @@ module openconfig-mpls-types {
   description
     "General types for MPLS / TE data model";
 
-  oc-ext:openconfig-version "3.5.0";
+  oc-ext:openconfig-version "3.5.1";
```
dplore: A minor version increment is needed because a leaf is being added.

For reference:
https://github.com/openconfig/public/blob/3f8a4a7e2c90b91db91f36222f55a22236d16006/release/models/openconfig-extensions.yang#L67

```suggestion
  oc-ext:openconfig-version "3.6.0";
```

release/models/mpls/openconfig-mpls-rsvp.yang
dplore: Hi @KyoyaTeramae, thank you for this contribution.  

To justify adding the `BYPASS_P2P` type, we need at least 2 references for implementation of this type.  Can you find more references for us?  If DriveNets DNOS is the only implementation, then one will instead need to use a yang augment that is not part of OC to add in the DNOS defined type.

ky-teramae: Thank you for the feedback. I have researched additional vendor implementations and added Nokia SR OS as the second reference

release/models/mpls/openconfig-mpls-types.yang
ky-teramae: Thank you for the comment. The version has been updated (e.g., to 3.6.0) as per the suggestion.

ElodinLaarz:
Discussed in the OC Operators Meeting Jan 6th 2026:

Seems reasonable; no objections raised. Setting this last-call to 2 weeks from now.

akalluru1:
@dplore Sorry for the late comments.

1. Newly introduced `bypass-tunnel` is of type leaf-list.
     a. What are the semantics behind having multiple bypass tunnels listed under one interface?
     b. There is no inherent order so how does the switch decide which bypass tunnel to use upon 
          link failure if there're multiple valid tunnels are available?
2. This comment is not completely related to this PR but relevant in this context:
     a. Is there an inherent assumption that the interfaces referred in the hierarchy are
         point-to-point? For LAN links multiple destination next hops are reachable through the
         same interface (i.e. a one-to-many link). In such a scenario, distinct bypasses should be 
         programmed for LSPs going through the same interface. This comment can be addressed
         in a different PR if it is relevant & addressable through description/documentation or some 
         other means.

PR #1417, opened 2025-12-16
Introduce MATCH_ACTION policy-forwarding

dplore:
### Change Scope

* Introduce MATCH_ACTION policy-forwarding type and associated actions.
* The PR fulfills a portion of the model and use cases described by #1371  for documentation.
* Additional PR's will define the `next-rule`, `count` and `sample` actions.

### Platform Implementations

See #1371 which includes references to implementation.

### Tree View

```diff
module: openconfig-network-instance
  +--rw network-instances
     +--rw network-instance* [name]

         +--rw policy-forwarding
         |  +--rw config
         |  |  +--rw global-decap-policy?   string
         |  +--ro state
         |  |  +--ro global-decap-policy?   string
         |  +--rw policies
         |  |  +--rw policy* [policy-id]
         |  |     +--rw policy-id    -> ../config/policy-id
         |  |     +--rw config
         |  |     |  +--rw policy-id?   string
         |  |     |  +--rw type?        enumeration
         |  |     +--ro state
         |  |     |  +--ro policy-id?   string
         |  |     |  +--ro type?        enumeration
         |  |     +--rw rules
         |  |        +--rw rule* [sequence-id]
         |  |           +--rw sequence-id        -> ../config/sequence-id
         |  |           +--rw config
         |  |           |  +--rw sequence-id?   uint32
+        |  |           |  +--rw address-family?   identityref
+        |  |           |  +--rw description?      string
         |  |           +--ro state
         |  |           |  +--ro sequence-id?      uint32
+        |  |           |  +--ro address-family?   identityref
+        |  |           |  +--ro description?      string
         |  |           |  +--ro matched-pkts?     oc-yang:counter64
         |  |           |  +--ro matched-octets?   oc-yang:counter64
         |  |           +--rw l2
         |  |           |  +--rw config
         |  |           |  |  +--rw source-mac?             oc-yang:mac-address
         |  |           |  |  +--rw source-mac-mask?        oc-yang:mac-address
         |  |           |  |  +--rw destination-mac?        oc-yang:mac-address
         |  |           |  |  +--rw destination-mac-mask?   oc-yang:mac-address
         |  |           |  |  +--rw ethertype?              oc-pkt-match-types:ethertype-type
         |  |           |  +--ro state
         |  |           |     +--ro source-mac?             oc-yang:mac-address
         |  |           |     +--ro source-mac-mask?        oc-yang:mac-address
         |  |           |     +--ro destination-mac?        oc-yang:mac-address
         |  |           |     +--ro destination-mac-mask?   oc-yang:mac-address
         |  |           |     +--ro ethertype?              oc-pkt-match-types:ethertype-type
         |  |           +--rw ipv4
         |  |           |  +--rw config
         |  |           |  |  +--rw source-address?                   oc-inet:ipv4-prefix
         |  |           |  |  +--rw source-address-prefix-set?        -> /oc-sets:defined-sets/ipv4-prefix-sets/ipv4-prefix-set/name
         |  |           |  |  +--rw destination-address?              oc-inet:ipv4-prefix
         |  |           |  |  +--rw destination-address-prefix-set?   -> /oc-sets:defined-sets/ipv4-prefix-sets/ipv4-prefix-set/name
         |  |           |  |  +--rw fragment-offsets*                 oc-pkt-match-types:fragment-offset-range
         |  |           |  |  +--rw dscp?                             oc-inet:dscp
         |  |           |  |  +--rw dscp-set*                         oc-inet:dscp
         |  |           |  |  +--rw length?                           uint16
         |  |           |  |  +--rw protocol?                         oc-pkt-match-types:ip-protocol-type
         |  |           |  |  +--rw hop-limit?                        uint8
         |  |           |  +--ro state
         |  |           |  |  +--ro source-address?                   oc-inet:ipv4-prefix
         |  |           |  |  +--ro source-address-prefix-set?        -> /oc-sets:defined-sets/ipv4-prefix-sets/ipv4-prefix-set/name
         |  |           |  |  +--ro destination-address?              oc-inet:ipv4-prefix
         |  |           |  |  +--ro destination-address-prefix-set?   -> /oc-sets:defined-sets/ipv4-prefix-sets/ipv4-prefix-set/name
         |  |           |  |  +--ro fragment-offsets*                 oc-pkt-match-types:fragment-offset-range
         |  |           |  |  +--ro dscp?                             oc-inet:dscp
         |  |           |  |  +--ro dscp-set*                         oc-inet:dscp
         |  |           |  |  +--ro length?                           uint16
         |  |           |  |  +--ro protocol?                         oc-pkt-match-types:ip-protocol-type
         |  |           |  |  +--ro hop-limit?                        uint8
         |  |           |  +--rw icmpv4
         |  |           |     +--rw config
         |  |           |     |  +--rw type?   identityref
         |  |           |     |  +--rw code?   identityref
         |  |           |     +--ro state
         |  |           |        +--ro type?   identityref
         |  |           |        +--ro code?   identityref
         |  |           +--rw ipv6
         |  |           |  +--rw config
         |  |           |  |  +--rw source-address?                   oc-inet:ipv6-prefix
         |  |           |  |  +--rw source-address-prefix-set?        -> /oc-sets:defined-sets/ipv6-prefix-sets/ipv6-prefix-set/name
         |  |           |  |  +--rw source-flow-label?                oc-inet:ipv6-flow-label
         |  |           |  |  +--rw destination-address?              oc-inet:ipv6-prefix
         |  |           |  |  +--rw destination-address-prefix-set?   -> /oc-sets:defined-sets/ipv6-prefix-sets/ipv6-prefix-set/name
         |  |           |  |  +--rw destination-flow-label?           oc-inet:ipv6-flow-label
         |  |           |  |  +--rw dscp?                             oc-inet:dscp
         |  |           |  |  +--rw dscp-set*                         oc-inet:dscp
         |  |           |  |  +--rw length?                           uint16
         |  |           |  |  +--rw protocol?                         oc-pkt-match-types:ip-protocol-type
         |  |           |  |  +--rw hop-limit?                        uint8
         |  |           |  +--ro state
         |  |           |  |  +--ro source-address?                   oc-inet:ipv6-prefix
         |  |           |  |  +--ro source-address-prefix-set?        -> /oc-sets:defined-sets/ipv6-prefix-sets/ipv6-prefix-set/name
         |  |           |  |  +--ro source-flow-label?                oc-inet:ipv6-flow-label
         |  |           |  |  +--ro destination-address?              oc-inet:ipv6-prefix
         |  |           |  |  +--ro destination-address-prefix-set?   -> /oc-sets:defined-sets/ipv6-prefix-sets/ipv6-prefix-set/name
         |  |           |  |  +--ro destination-flow-label?           oc-inet:ipv6-flow-label
         |  |           |  |  +--ro dscp?                             oc-inet:dscp
         |  |           |  |  +--ro dscp-set*                         oc-inet:dscp
         |  |           |  |  +--ro length?                           uint16
         |  |           |  |  +--ro protocol?                         oc-pkt-match-types:ip-protocol-type
         |  |           |  |  +--ro hop-limit?                        uint8
         |  |           |  +--rw icmpv6
         |  |           |     +--rw config
         |  |           |     |  +--rw type?   identityref
         |  |           |     |  +--rw code?   identityref
         |  |           |     +--ro state
         |  |           |        +--ro type?   identityref
         |  |           |        +--ro code?   identityref
         |  |           +--rw transport
         |  |           |  +--rw config
         |  |           |  |  +--rw source-port?                  oc-pkt-match-types:port-num-range
         |  |           |  |  +--rw source-port-set?              -> /oc-sets:defined-sets/port-sets/port-set/name
         |  |           |  |  +--rw destination-port?             oc-pkt-match-types:port-num-range
         |  |           |  |  +--rw destination-port-set?         -> /oc-sets:defined-sets/port-sets/port-set/name
         |  |           |  |  +--rw detail-mode?                  enumeration
         |  |           |  |  +--rw explicit-detail-match-mode?   enumeration
         |  |           |  |  +--rw explicit-tcp-flags*           identityref
         |  |           |  |  +--rw builtin-detail?               enumeration
         |  |           |  +--ro state
         |  |           |     +--ro source-port?                  oc-pkt-match-types:port-num-range
         |  |           |     +--ro source-port-set?              -> /oc-sets:defined-sets/port-sets/port-set/name
         |  |           |     +--ro destination-port?             oc-pkt-match-types:port-num-range
         |  |           |     +--ro destination-port-set?         -> /oc-sets:defined-sets/port-sets/port-set/name
         |  |           |     +--ro detail-mode?                  enumeration
         |  |           |     +--ro explicit-detail-match-mode?   enumeration
         |  |           |     +--ro explicit-tcp-flags*           identityref
         |  |           |     +--ro builtin-detail?               enumeration
         |  |           +--rw action
         |  |           |  +--rw config
-        |  |           |  |  +--rw discard?                           boolean
+        |  |           |  |  x--rw discard?                           boolean
         |  |           |  |  +--rw decapsulate-gre?                   boolean
         |  |           |  |  +--rw decap-network-instance?            -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw decap-fallback-network-instance?   -> /network-instances/network-instance/config/name
+        |  |           |  |  +--rw log?                               boolean
         |  |           |  |  +--rw post-decap-network-instance?       -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw network-instance?                  -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw path-selection-group?              -> ../../../../../../../path-selection-groups/path-selection-group/config/group-id
         |  |           |  |  +--rw next-hop?                          oc-inet:ip-address
         |  |           |  |  +--rw next-hop-group?                    -> ../../../../../../../../static/next-hop-groups/next-hop-group/config/name
         |  |           |  |  +--rw decapsulate-mpls-in-udp?           boolean
         |  |           |  |  +--rw decapsulate-gue?                   boolean
-        |  |           |  |  +--rw ip-ttl?                            uint8
+        |  |           |  |  x--rw ip-ttl?                            uint8
         |  |           |  +--ro state
-        |  |           |  |  +--ro discard?                           boolean
+        |  |           |  |  x--ro discard?                           boolean
         |  |           |  |  +--ro decapsulate-gre?                   boolean
         |  |           |  |  +--ro decap-network-instance?            -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro decap-fallback-network-instance?   -> /network-instances/network-instance/config/name
+        |  |           |  |  +--ro log?                               boolean
         |  |           |  |  +--ro post-decap-network-instance?       -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro network-instance?                  -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro path-selection-group?              -> ../../../../../../../path-selection-groups/path-selection-group/config/group-id
         |  |           |  |  +--ro next-hop?                          oc-inet:ip-address
         |  |           |  |  +--ro next-hop-group?                    -> ../../../../../../../../static/next-hop-groups/next-hop-group/config/name
         |  |           |  |  +--ro decapsulate-mpls-in-udp?           boolean
         |  |           |  |  +--ro decapsulate-gue?                   boolean
-        |  |           |  |  +--ro ip-ttl?                            uint8
+        |  |           |  |  x--ro ip-ttl?                            uint8
         |  |           |  +--rw encapsulate-gre
         |  |           |  |  +--rw config
         |  |           |  |  |  +--rw identifying-prefix?   oc-inet:ip-prefix
         |  |           |  |  +--ro state
         |  |           |  |  |  +--ro identifying-prefix?   oc-inet:ip-prefix
         |  |           |  |  +--rw targets
         |  |           |  |     +--rw target* [id]
         |  |           |  |        +--rw id        -> ../config/id
         |  |           |  |        +--rw config
         |  |           |  |        |  +--rw id?            string
         |  |           |  |        |  +--rw source?        oc-inet:ip-address
         |  |           |  |        |  +--rw destination?   oc-inet:ip-prefix
         |  |           |  |        |  +--rw ip-ttl?        uint8
         |  |           |  |        +--ro state
         |  |           |  |           +--ro id?            string
         |  |           |  |           +--ro source?        oc-inet:ip-address
         |  |           |  |           +--ro destination?   oc-inet:ip-prefix
         |  |           |  |           +--ro ip-ttl?        uint8
+        |  |           |  +--rw ipv4
+        |  |           |  |  +--rw config
+        |  |           |  |  |  +--rw dscp?   oc-inet-types:dscp
+        |  |           |  |  |  +--rw ttl?    uint8
+        |  |           |  |  +--ro state
+        |  |           |  |     +--ro dscp?   oc-inet-types:dscp
+        |  |           |  |     +--ro ttl?    uint8
+        |  |           |  +--rw ipv6
+        |  |           |  |  +--rw config
+        |  |           |  |  |  +--rw dscp?        oc-inet-types:dscp
+        |  |           |  |  |  +--rw hop-limit?   uint8
+        |  |           |  |  +--ro state
+        |  |           |  |     +--ro dscp?        oc-inet-types:dscp
+        |  |           |  |     +--ro hop-limit?   uint8
+        |  |           |  +--rw mpls
+        |  |           |  |  +--rw config
+        |  |           |  |  |  +--rw mpls-traffic-class?   oc-mplst:mpls-tc
+        |  |           |  |  +--ro state
+        |  |           |  |     +--ro mpls-traffic-class?   oc-mplst:mpls-tc
+        |  |           |  +--rw forwarding
+        |  |           |  |  +--rw config
+        |  |           |  |  |  +--rw forwarding-action?      identityref
+        |  |           |  |  |  +--rw qos-forwarding-group?   -> /oc-qos:qos/forwarding-groups/forwarding-group/config/name
+        |  |           |  |  +--ro state
+        |  |           |  |     +--ro forwarding-action?      identityref
+        |  |           |  |     +--ro qos-forwarding-group?   -> /oc-qos:qos/forwarding-groups/forwarding-group/config/name
         |  |           |  +--rw oc-pf-srte:segment-lists
         |  |           |     +--rw oc-pf-srte:segment-list* [index]
         |  |           |        +--rw oc-pf-srte:index     -> ../config/index
         |  |           |        +--rw oc-pf-srte:config
         |  |           |        |  +--rw oc-pf-srte:index?    uint64
         |  |           |        |  +--rw oc-pf-srte:weight?   uint32
         |  |           |        +--ro oc-pf-srte:state
         |  |           |        |  +--ro oc-pf-srte:index?    uint64
         |  |           |        |  +--ro oc-pf-srte:weight?   uint32
         |  |           |        +--rw oc-pf-srte:sids
         |  |           |           +--rw oc-pf-srte:sid* [index]
         |  |           |              +--rw oc-pf-srte:index     -> ../config/index
         |  |           |              +--rw oc-pf-srte:config
         |  |           |              |  +--rw oc-pf-srte:index?      uint64
         |  |           |              |  +--rw oc-pf-srte:value?      oc-srt:sr-sid-type
         |  |           |              |  +--rw oc-pf-srte:mpls-ttl?   uint8
         |  |           |              |  +--rw oc-pf-srte:mpls-tc?    uint8
         |  |           |              +--ro oc-pf-srte:state
         |  |           |                 +--ro oc-pf-srte:index?      uint64
         |  |           |                 +--ro oc-pf-srte:value?      oc-srt:sr-sid-type
         |  |           |                 +--ro oc-pf-srte:mpls-ttl?   uint8
         |  |           |                 +--ro oc-pf-srte:mpls-tc?    uint8
         |  |           +--rw oc-pf-srte:srte
         |  |              +--rw oc-pf-srte:config
         |  |              |  +--rw oc-pf-srte:mpls-bsid?         oc-mplst:mpls-label
         |  |              |  +--rw oc-pf-srte:srte-endpoint?     oc-inet:ip-address
         |  |              |  +--rw oc-pf-srte:srte-color?        uint32
         |  |              |  +--rw oc-pf-srte:srte-preference?   uint32
         |  |              +--ro oc-pf-srte:state
         |  |                 +--ro oc-pf-srte:mpls-bsid?         oc-mplst:mpls-label
         |  |                 +--ro oc-pf-srte:srte-endpoint?     oc-inet:ip-address
         |  |                 +--ro oc-pf-srte:srte-color?        uint32
         |  |                 +--ro oc-pf-srte:srte-preference?   uint32
```

### DIFF
--- a/release/models/bfd/openconfig-bfd.yang
+++ b/release/models/bfd/openconfig-bfd.yang
@@ -26,7 +26,15 @@ module openconfig-bfd {
     "An OpenConfig model of Bi-Directional Forwarding Detection (BFD)
     configuration and operational state.";
 
-  oc-ext:openconfig-version "0.4.1";
+  oc-ext:openconfig-version "0.5.0";
+
+  revision "2025-09-03" {
+    description
+      "Add desired-min-echo-tx-interval and required-min-echo-rx-interval
+      leaves in bfd-interface-config to allow RFC 9127-compliant BFD echo
+      mode configuration per interface.";
+    reference "0.5.0";
+ }
 
   revision "2025-03-18" {
     description
@@ -269,6 +277,32 @@ module openconfig-bfd {
       description
         "When this leaf is set to true - BFD will be enabled on
         each member interface of the aggregated Ethernet bundle.";
+    }
+
+    leaf desired-min-echo-tx-interval {
+      type uint32;
+      units microseconds;
+      description
+        "The minimum interval between transmission of BFD echo
+        packets that the operator desires. This value controls the
+        transmission rate of echo packets sent by the local system.
+        Setting this value to 0 disables BFD echo mode on this
+        interface. This value is specified as an integer number
+        of microseconds.";
+      reference "RFC 9127, Appendix A.1";
+    }
+
+    leaf required-min-echo-rx-interval {
+      type uint32;
+      units microseconds;
+      description
+        "The minimum interval between received BFD echo packets that
+        this system should support. This value is advertised to the
+        remote peer and allows the peer to opt-out of echo mode or
+        specify receive capabilities. Setting this value to 0 indicates
+        that the local system does not support receiving echo packets.
+        This value is specified as an integer number of microseconds.";
+      reference "RFC 9127, Appendix A.1";
     }
   }
 


release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -245,6 +279,18 @@ submodule openconfig-pf-forwarding-policies {
...
+          description
+            "A policy which defines one or more packet match criteria rules
+            and one or more actions to perform on matched packets.  While
+            PBR_POLICY generally supports setting the next-hop as an action,
+            MATCH_ACTION_POLICY allows a wide variety of actions to be combined
+            which include setting the next-hop, re-writing packet field and
+            setting qos forwarding groups.  Actions intended to be used with
+            MATCH_ACTION_POLICY are indicated by a where statement in the
+            action leaf.  Match action policies are defined in more detail at
+            https://github.com/openconfig/public/doc/ocpf-match-action.md.";
```
robshakir: I don't think this link will ever work -- you will likely need to include a URL of the form: `https://github.com/openconfig/public/blob/master/doc/component_fault_overview.md`

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,6 +310,22 @@ submodule openconfig-pf-forwarding-policies {
...
+    leaf address-family {
+      type identityref {
+        base oc-types:ADDRESS_FAMILY;
+      }
+      description
+        "The type determines the address family allowed in the
+        policy-fowarding match entries belonging to the rule.";
+    }
+
+    leaf name {
```
robshakir: Is this ever used as a key? I assume not -- and that we're adding it mainly because we are mapping to existing data structures (CLI or otherwise)?

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -351,6 +414,49 @@ submodule openconfig-pf-forwarding-policies {
...
 
+    leaf dscp {
+      when "../../../../../config/type = 'MATCH_ACTION_POLICY'";
+      type oc-inet-types:dscp;
+      description
+        "Set the matching packet DSCP value.";
+    }
+
+    leaf forwarding-action {
+      when "../../../../../config/type = 'MATCH_ACTION_POLICY'";
```
robshakir: There is a regression introduced in the model based on the combination of this `when` statement and `discard` being deprecated -- now one cannot specify an ACL that drops a packet. I don't think this is intentional. Can you confirm?

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -351,6 +414,49 @@ submodule openconfig-pf-forwarding-policies {
         etc).";
     }
 
+    leaf dscp {
```
robshakir: Should this be `set-dscp` to make it clear that we are setting a value?

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -351,6 +414,49 @@ submodule openconfig-pf-forwarding-policies {
...
+    leaf log {
+      type boolean;
+      description
+        "Create a system logging message when this rule is matched.  Note,
+        this is intended to be used for debugging purposes.  Enabling per
+        packet logging is expected to have significant performance
+        limitations.";
+    }
+
+    leaf mpls-traffic-class {
```
robshakir: Do you need to consider grouping the fields? Today, there's a subset of fields that are specified, but I wonder whether without some level of hierarchy for IPv4, IPv6, etc. then you will end up with a huge container here going forward.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: The `name` leaf here is not used as a key in the model.  The purpose is only for a human friendly name that describes the rule.   There is some precedence in current vendor implementation in CLI.  The name in some NOS do use it as a key (Cisco IOSXR)  and others do not (Arista EOS).    Regardless there is value to humans by giving a rule a semantic name (eg: very short description) :

Arista EOS
```
traffic-policies
  traffic-policy my-new-policy-forwarding
    match my1and2ipblocks ipv4
      destination prefix 1.1.0.0/24 2.1.0.0/16
   match my3and4blocks ipv6
     destination prefix 2001:db8:abcd::/48 2001:1234:5678:1234::/64

! note that my1and2ipblocks is not defined anywhere else, it's only
! purpose is a human friendly name
```

CIsco IOS XR
```
policy-map type pbr my-new-policy-forwarding
 10 class ingress-10-discard
  drop
 20 class ingress-20-encap
  set nexthop-group my-cloud-dest
 !

class-map type pbr match-any ingress-20-encap
 10 match ip access-group acl-in-encap
ip access-list acl-in-encap
 permit ip any 1.2.16.0 0.0.15.255
 permit ip any 2.2.8.0 0.0.7.255
```

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: Good question!  Elsewhere we in this subtree we have `action/ip-ttl`,  `action/next-hop`, `/action/network-instance`.  There is no `set-` prefix for these actions.  There might be some minor confusion where we have `rule/dscp-set` which is used to match against an array of dscp values.

That said, `/routing-policy` and `/qos/scheduler-policy` trees do use `set-<something>`.  So there's some inconsistency currently present in the model.

I'm ok with changing this to `set-dscp` if there's rough consensus that is more intuitive.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: Sure.  Added containers for mpls and forwarding.  

There is am existing (recently added)  `ip-ttl` leaf which is not in a subcontainer at this level.  I have left it where it is for now.

```diff
         +--rw policy-forwarding
         |  +--rw policies
         |  |  +--rw policy* [policy-id]
         |  |     +--rw rules
         |  |        +--rw rule* [sequence-id]
         |  |           +--rw action
         |  |           |  +--rw config
-        |  |           |  |  +--rw discard?                           boolean
+        |  |           |  |  x--rw discard?                           boolean
         |  |           |  |  +--rw decapsulate-gre?                   boolean
         |  |           |  |  +--rw decap-network-instance?            -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw decap-fallback-network-instance?   -> /network-instances/network-instance/config/name
+        |  |           |  |  +--rw dscp?                              oc-inet-types:dscp
+        |  |           |  |  +--rw log?                               boolean
         |  |           |  |  +--rw post-decap-network-instance?       -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw network-instance?                  -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw path-selection-group?              -> ../../../../../../../path-selection-groups/path-selection-group/config/group-id
         |  |           |  |  +--rw next-hop?                          oc-inet:ip-address
         |  |           |  |  +--rw next-hop-group?                    -> ../../../../../../../../static/next-hop-groups/next-hop-group/config/name
         |  |           |  |  +--rw decapsulate-mpls-in-udp?           boolean
         |  |           |  |  +--rw decapsulate-gue?                   boolean
         |  |           |  |  +--rw ip-ttl?                            uint8
         |  |           |  +--ro state
-        |  |           |  |  +--ro discard?                           boolean
+        |  |           |  |  x--ro discard?                           boolean
         |  |           |  |  +--ro decapsulate-gre?                   boolean
         |  |           |  |  +--ro decap-network-instance?            -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro decap-fallback-network-instance?   -> /network-instances/network-instance/config/name
+        |  |           |  |  +--ro dscp?                              oc-inet-types:dscp
+        |  |           |  |  +--ro log?                               boolean
         |  |           |  |  +--ro post-decap-network-instance?       -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro network-instance?                  -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro path-selection-group?              -> ../../../../../../../path-selection-groups/path-selection-group/config/group-id
         |  |           |  |  +--ro next-hop?                          oc-inet:ip-address
         |  |           |  |  +--ro next-hop-group?                    -> ../../../../../../../../static/next-hop-groups/next-hop-group/config/name
         |  |           |  |  +--ro decapsulate-mpls-in-udp?           boolean
         |  |           |  |  +--ro decapsulate-gue?                   boolean
         |  |           |  |  +--ro ip-ttl?                            uint8
         |  |           |  +--rw encapsulate-gre
         |  |           |  |  +--rw config
         |  |           |  |  |  +--rw identifying-prefix?   oc-inet:ip-prefix
         |  |           |  |  +--ro state
         |  |           |  |  |  +--ro identifying-prefix?   oc-inet:ip-prefix
         |  |           |  |  +--rw targets
         |  |           |  |     +--rw target* [id]
         |  |           |  |        +--rw id        -> ../config/id
         |  |           |  |        +--rw config
         |  |           |  |        |  +--rw id?            string
         |  |           |  |        |  +--rw source?        oc-inet:ip-address
         |  |           |  |        |  +--rw destination?   oc-inet:ip-prefix
         |  |           |  |        |  +--rw ip-ttl?        uint8
         |  |           |  |        +--ro state
         |  |           |  |           +--ro id?            string
         |  |           |  |           +--ro source?        oc-inet:ip-address
         |  |           |  |           +--ro destination?   oc-inet:ip-prefix
         |  |           |  |           +--ro ip-ttl?        uint8
+        |  |           |  +--rw mpls
+        |  |           |  |  +--rw config
+        |  |           |  |  |  +--rw mpls-traffic-class?   oc-mplst:mpls-tc
+        |  |           |  |  +--ro state
+        |  |           |  |     +--ro mpls-traffic-class?   oc-mplst:mpls-tc
+        |  |           |  +--rw forwarding
+        |  |           |  |  +--rw config
+        |  |           |  |  |  +--rw forwarding-action?     identityref
+        |  |           |  |  |  +--rw qos-forwarding-roup?   -> /oc-qos:qos/forwarding-groups/forwarding-group/config/name
+        |  |           |  |  +--ro state
+        |  |           |  |     +--ro forwarding-action?     identityref
+        |  |           |  |     +--ro qos-forwarding-roup?   -> /oc-qos:qos/forwarding-groups/forwarding-group/config/name
         |  |           |  +--rw oc-pf-srte:segment-lists
         |  |           |     +--rw oc-pf-srte:segment-list* [index]
         |  |           |        +--rw oc-pf-srte:index     -> ../config/index
         |  |           |        +--rw oc-pf-srte:config
         |  |           |        |  +--rw oc-pf-srte:index?    uint64
         |  |           |        |  +--rw oc-pf-srte:weight?   uint32
         |  |           |        +--ro oc-pf-srte:state
         |  |           |        |  +--ro oc-pf-srte:index?    uint64
         |  |           |        |  +--ro oc-pf-srte:weight?   uint32
         |  |           |        +--rw oc-pf-srte:sids
         |  |           |           +--rw oc-pf-srte:sid* [index]
         |  |           |              +--rw oc-pf-srte:index     -> ../config/index
         |  |           |              +--rw oc-pf-srte:config
         |  |           |              |  +--rw oc-pf-srte:index?      uint64
         |  |           |              |  +--rw oc-pf-srte:value?      oc-srt:sr-sid-type
         |  |           |              |  +--rw oc-pf-srte:mpls-ttl?   uint8
         |  |           |              |  +--rw oc-pf-srte:mpls-tc?    uint8
         |  |           |              +--ro oc-pf-srte:state
         |  |           |                 +--ro oc-pf-srte:index?      uint64
         |  |           |                 +--ro oc-pf-srte:value?      oc-srt:sr-sid-type
         |  |           |                 +--ro oc-pf-srte:mpls-ttl?   uint8
         |  |           |                 +--ro oc-pf-srte:mpls-tc?    uint8
         |  |           +--rw oc-pf-srte:srte
         |  |              +--rw oc-pf-srte:config
         |  |              |  +--rw oc-pf-srte:mpls-bsid?         oc-mplst:mpls-label
         |  |              |  +--rw oc-pf-srte:srte-endpoint?     oc-inet:ip-address
         |  |              |  +--rw oc-pf-srte:srte-color?        uint32
         |  |              |  +--rw oc-pf-srte:srte-preference?   uint32
         |  |              +--ro oc-pf-srte:state
         |  |                 +--ro oc-pf-srte:mpls-bsid?         oc-mplst:mpls-label
         |  |                 +--ro oc-pf-srte:srte-endpoint?     oc-inet:ip-address
         |  |                 +--ro oc-pf-srte:srte-color?        uint32
         |  |                 +--ro oc-pf-srte:srte-preference?   uint32

```

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: `action/forwarding/config/forwarding-action = DROP` should be used, but because discard is only deprecated (and not deleted), `discard` may be used, but is not recommended.  

(Yang requires deprecated leaves to be implemented, deprecation is merely an informational attribute)

Does this address your concern?

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
robshakir: Let's favour consistency within the model. I'm OK with no `set-` based on this.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
robshakir: No, this doesn't resolve my issue. It's based on the `when` statement here. You only allow `forwarding-action` to be used as a leaf when it is a `MATCH_ACTION_POLICY`.

Today one can have a type of `PBR_POLICY` which has `discard=TRUE`. `discard` is now deprecated, but this new `forwarding-action` leaf can't be used because of the `when` clause.

Essentially: I think this `when` clause is erroneous and should be removed.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
robshakir: Is there a reason not to also move the `ip-ttl` and `dscp` leaves? (which I assume apply both only to IPv4)

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: I see now.  forwarding-action needs to be usable in both `PBR_POLICY` and `MATCH_ACTION_POLICY` .  Removing this.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: `ip-ttl`is a pre-existing leaves, but very new.   I will deprecate `ip-ttl` and add  to the address family container.  I'll move `dscp` it is an all new leaf.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -283,12 +351,195 @@ submodule openconfig-pf-forwarding-policies {
...
+
+    leaf dscp {
+      when "../../../../../../config/type = 'MATCH_ACTION_POLICY'";
+      type oc-inet-types:dscp;
+      description
+        "Set the matching packet DSCP value.";
+    }
+
+    leaf hop-limit {
+      when "../../../../../../config/type='MATCH_ACTION_POLICY'";
```
robshakir: (very nitty nit) You have inconsistent spacing around the `=` in the `when` clause.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -283,12 +351,195 @@ submodule openconfig-pf-forwarding-policies {
...
+      "Configuration parameters for IPv4 actions.";
+
+    leaf dscp {
+      when "../../../../../../config/type = 'MATCH_ACTION_POLICY'";
+      type oc-inet-types:dscp;
+      description
+        "Set the matching packet DSCP value.";
+    }
+
+    leaf ip-ttl {
```
robshakir: You can probably just call this `ttl` given that we're in the `ipv4` container at this point?

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: Thanks!  Updated to use "` = `" as that is consistent with other models.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: Done!

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,6 +315,23 @@ submodule openconfig-pf-forwarding-policies {
...
+      }
+      description
+        "The type determines the address family allowed in the
+        policy-fowarding match entries belonging to the rule.";
+    }
+
+    leaf name {
+      type string;
+      description
+      "A user defined name for a rule, used to give a user friendly
```
amitarista: Are there any constraints for the `name` attribute? Is it allowed for multiple rules within a policy to share the same name? From the description of this leaf, more appropriate leaf name is `description`.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,6 +315,23 @@ submodule openconfig-pf-forwarding-policies {
       description
         "Unique sequence number for the policy rule.";
     }
+
+    leaf address-family {
+      type identityref {
+        base oc-types:ADDRESS_FAMILY;
+      }
+      description
+        "The type determines the address family allowed in the
```
amitarista, line 324: Please include information about transport layer match handling in the description, since the address family is limited to L2, IPV4, IPV6 and MPLS.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -283,12 +351,195 @@ submodule openconfig-pf-forwarding-policies {
...
+      "Configuration parameters for IPv4 actions.";
+
+    leaf dscp {
+      when "../../../../../../config/type = 'MATCH_ACTION_POLICY'";
+      type oc-inet-types:dscp;
+      description
+        "Set the matching packet DSCP value.";
+    }
+
+    leaf ttl {
```
amitarista, line 386: nit: leaf name here is `ttl` vs `ip-ttl` in the PR description

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -441,9 +701,15 @@ submodule openconfig-pf-forwarding-policies {
     }
 
     leaf ip-ttl {
+      when "../../../../../config/type = 'MATCH_ACTION_POLICY'";
       type uint8;
+      status deprecated;
       description
-        "If set, this leaf causes the local system to not decrement the TTL
+        "This is is deprecated and may be removed in a future OC model release.
```
amitarista: nit: is is

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: done

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: updated  tree view in PR description to show the latest commit

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
dplore: fixed

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,6 +315,21 @@ submodule openconfig-pf-forwarding-policies {
...
         "Unique sequence number for the policy rule.";
     }
+
+    leaf address-family {
+      type identityref {
+        base oc-types:ADDRESS_FAMILY;
+      }
+      description
+        "The type determines the address family allowed in the
+        policy-fowarding match entries belonging to the rule.";
```
dplore:
```suggestion
        policy-forwarding match entries belonging to the rule.";
```

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -264,6 +315,21 @@ submodule openconfig-pf-forwarding-policies {
       description
         "Unique sequence number for the policy rule.";
     }
+
+    leaf address-family {
```
earies, line 319: Since this is introduced as a non-key node to the `rule` list, the model can allow for mixing different address families within the same PF policy - this differs in design from the ACL modeling where there is an additional key at a parent level (often the same implementation)

And to the point raised on the OC community call 2026-02-05, there are a number of child containers that may only apply based on how this non-mandatory/non-key value is set.  If the modeling does not define these possibilities and restrictions, I feel we are going to see divergence amongst implementation behavior that may not align to use-cases, not to mention corner cases and logic that needs handling outside of the modeling.

dplore: Making this node mandatory or adding it as a key in the list are both breaking changes to the model which are not easily absorbed as the model is in production use.    

We chose the tradeoff of adding to the existing model for operational ease versus introducing a new model (where we could define a list with an address-family as a key or as a mandatory node).

dplore: I do have an open issue from [OC community](https://docs.google.com/document/d/1KYJsmSXTsu-cX2AOQe7cMhXmXYYJbt9hMBqWl_swr5w/edit?tab=t.0)  to clarify if the address family constrains both match and actions, or just the match rules.    I'll complete that investigation shortly.

dplore:
> > I will merge this as is, without the actions container constrained to an address family. Some actions are Address Family Neutral, meaning they don't care about the IP version. However, some are Address Family Specific. Which actions can be used are relevant to the match and can be rejected on a logical basis (such as ipv4 matching with an action for an ipv4 next-hop or implemention specific limitations).
> > Address family neutral actions:
> > 
> > * drop / discard
> > * set dscp / set cos
> > * count (Statistics)
> > * mirror
> > * encapsulation / decapsulation
> > 
> > Address family specific actions:
> > 
> > * set next-hop (Requires an IP of the same family)
> > * redirect to vrf (The destination VRF must have a route for that specific family)
> > * nat (Network Address Translation is strictly family-specific)
> > 
> > Exceptions exist for cross-family redirection. You can match an IPv4 packet and redirect it to a 6PE (IPv6 Provider Edge) label or an Recursive Next-hop that eventually resolves via an IPv6 transport, but the "Match" is still strictly IPv4. It would be complex and generally against OC style to attempt to model all these conditions with logic in the yang model.
> 
> set-dscp is mentioned under address family neutral actions, but in the merged code we have address-family specific dscp action. Is that intentional?

Yes, there are containers for address family in the match and action containers for readability.  The topic of this thread is whether there should be yang constraints modeled to formally restrict the address family for match rules and actions.   

In the (now merged) model, there is no use of yang `must` or `when` directives for address family.  But we do have an `/network-instances/network-instance/policy-forwarding/policies/policy/config/address-family` leaf with the description:   "The address-family allowed in the  policy-forwarding match entries belonging to the rule.";

PR #1418, opened 2025-12-16
Introduce policy-forwarding policy-counters

dplore:
### Change Scope

* Introduce  `user-counters` within the `/network-instances/network-instance/policy-forwarding/policy` subtree.
* Add a policy-forwarding action to reference the per-policy user-counters

### Platform Implementations
Arista EOS and Juniper JunOS each have packet counter actions for their traffic-policy and filter based forwarding respectively. 

Arista EOS traffic policies
```none
traffic-policies
  traffic-policy send-v6-to-my_nhg
    match v6-dscp3 ipv6
         dscp 3
         !
         actions
            count my_nhg
            redirect next-hop group v6_my_nhg
```

JunOS filter based forwarding

```none
      match my-af3-traffic ipv4
         source prefix field-set src-af3-prefixes
         destination prefix field-set dst-af3-prefixes
         actions
            count ip-af3-queue
            set dscp 24
            set traffic class 3
            redirect next-hop group ingress-vip
         !
```

 See #1371 for additional implementation references and example operational use cases.

The CLI semantics are different in that action to perform a count and the user named counter are both created with a single line of CLI.  In OC, we need a way to create the counter and stream the state of the counter.  The user-defined counters container is added at the `/network-instances/network-instance/policy-forwarding/policy` subtree, allowing counters to be defined per policy.  Each rule within the policy may define an `action` to `count` packets matching the rule. 

### Tree View
diff -wU 20 ~/old-tree.txt ~/pfcounter-tree.txt 
```diff
         +--rw policy-forwarding
         |  +--rw config
         |  |  +--rw global-decap-policy?   string
         |  +--ro state
         |  |  +--ro global-decap-policy?   string
         |  +--rw policies
         |  |  +--rw policy* [policy-id]
         |  |     +--rw policy-id    -> ../config/policy-id
+        |  |     +--rw policy-counters
+        |  |     |  +--rw policy-counter* [name]
+        |  |     |     +--rw name      -> ../config/name
+        |  |     |     +--rw config
+        |  |     |     |  +--rw name?   string
+        |  |     |     +--ro state
+        |  |     |        +--ro name?       string
+        |  |     |        +--ro counters
+        |  |     |           +--ro packets?   oc-yang:counter64
+        |  |     |           +--ro bytes?     oc-yang:counter64
         |  |     +--rw config
         |  |     |  +--rw policy-id?   string
         |  |     |  +--rw type?        enumeration
         |  |     +--ro state
         |  |     |  +--ro policy-id?   string
         |  |     |  +--ro type?        enumeration
         |  |     +--rw rules
         |  |        +--rw rule* [sequence-id]
         |  |           +--rw sequence-id        -> ../config/sequence-id
         |  |           +--rw config
         |  |           |  +--rw sequence-id?   uint32
         |  |           +--ro state
         |  |           |  +--ro sequence-id?      uint32
         |  |           |  +--ro matched-pkts?     oc-yang:counter64
         |  |           |  +--ro matched-octets?   oc-yang:counter64
         |  |           +--rw l2
         |  |           |  +--rw config
         |  |           |  |  +--rw source-mac?             oc-yang:mac-address
         |  |           |  |  +--rw source-mac-mask?        oc-yang:mac-address
         |  |           |  |  +--rw destination-mac?        oc-yang:mac-address
@@ -3653,53 +3663,55 @@
         |  |           |  +--rw config
         |  |           |  |  +--rw source-port?                  oc-pkt-match-types:port-num-range
         |  |           |  |  +--rw source-port-set?              -> /oc-sets:defined-sets/port-sets/port-set/name
         |  |           |  |  +--rw destination-port?             oc-pkt-match-types:port-num-range
         |  |           |  |  +--rw destination-port-set?         -> /oc-sets:defined-sets/port-sets/port-set/name
         |  |           |  |  +--rw detail-mode?                  enumeration
         |  |           |  |  +--rw explicit-detail-match-mode?   enumeration
         |  |           |  |  +--rw explicit-tcp-flags*           identityref
         |  |           |  |  +--rw builtin-detail?               enumeration
         |  |           |  +--ro state
         |  |           |     +--ro source-port?                  oc-pkt-match-types:port-num-range
         |  |           |     +--ro source-port-set?              -> /oc-sets:defined-sets/port-sets/port-set/name
         |  |           |     +--ro destination-port?             oc-pkt-match-types:port-num-range
         |  |           |     +--ro destination-port-set?         -> /oc-sets:defined-sets/port-sets/port-set/name
         |  |           |     +--ro detail-mode?                  enumeration
         |  |           |     +--ro explicit-detail-match-mode?   enumeration
         |  |           |     +--ro explicit-tcp-flags*           identityref
         |  |           |     +--ro builtin-detail?               enumeration
         |  |           +--rw action
         |  |           |  +--rw config
+        |  |           |  |  +--rw count?                             -> ../../../../../policy-counters/policy-counter/config/name
         |  |           |  |  +--rw discard?                           boolean
         |  |           |  |  +--rw decapsulate-gre?                   boolean
         |  |           |  |  +--rw decap-network-instance?            -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw decap-fallback-network-instance?   -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw post-decap-network-instance?       -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw network-instance?                  -> /network-instances/network-instance/config/name
         |  |           |  |  +--rw path-selection-group?              -> ../../../../../../../path-selection-groups/path-selection-group/config/group-id
         |  |           |  |  +--rw next-hop?                          oc-inet:ip-address
         |  |           |  |  +--rw next-hop-group?                    -> ../../../../../../../../static/next-hop-groups/next-hop-group/config/name
         |  |           |  |  +--rw decapsulate-mpls-in-udp?           boolean
         |  |           |  |  +--rw decapsulate-gue?                   boolean
         |  |           |  |  +--rw ip-ttl?                            uint8
         |  |           |  +--ro state
+        |  |           |  |  +--ro count?                             -> ../../../../../policy-counters/policy-counter/config/name
         |  |           |  |  +--ro discard?                           boolean
         |  |           |  |  +--ro decapsulate-gre?                   boolean
         |  |           |  |  +--ro decap-network-instance?            -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro decap-fallback-network-instance?   -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro post-decap-network-instance?       -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro network-instance?                  -> /network-instances/network-instance/config/name
         |  |           |  |  +--ro path-selection-group?              -> ../../../../../../../path-selection-groups/path-selection-group/config/group-id
         |  |           |  |  +--ro next-hop?                          oc-inet:ip-address
         |  |           |  |  +--ro next-hop-group?                    -> ../../../../../../../../static/next-hop-groups/next-hop-group/config/name
         |  |           |  |  +--ro decapsulate-mpls-in-udp?           boolean
         |  |           |  |  +--ro decapsulate-gue?                   boolean
         |  |           |  |  +--ro ip-ttl?                            uint8
         |  |           |  +--rw encapsulate-gre
         |  |           |  |  +--rw config
         |  |           |  |  |  +--rw identifying-prefix?   oc-inet:ip-prefix
         |  |           |  |  +--ro state
         |  |           |  |  |  +--ro identifying-prefix?   oc-inet:ip-prefix
         |  |           |  |  +--rw targets
         |  |           |  |     +--rw target* [id]
         |  |           |  |        +--rw id        -> ../config/id
```


earies:
Regarding the proposal to introduce a generic list of top-level configurable `user-counters` - are there any existing implementation references to such a feature?  The references provided do not indicate this.

I realize the desire here due to how such a named counter is referenced from other subtrees or accessed but speaking on behalf of JUNOS/EVO, this is not how this feature is implemented and named counters cannot be shared across references or domains and may also get derived names depending on if interface granularity is tagged.

As it stands in current implementation, this would result in phantom alien config nodes that have no real meaning and could give the perception of conformance for specific use on state only as I see it.

dplore:
@amitarista @earies @jsterne @rgwilton for your comments on the revised model

dplore:
In OC modeling we are looking for at least 2 examples of existing implementations for a feature (but not a particular native configuration style) or assertion from operator(s) that feature requests are in progress with multiple implementors.  

We clearly have at least three pre-existing implementations of counters for policy-forwarding rules, noted below.  This is not a vendor proprietary feature.   The configuration schemas are indeed all different.  Getting to a common schema is one of the key goals of OC and doing so means there's going to be some changes needed to adapt.  

Comments should be directed regarding the functionality of what is being proposed.  Secondarily, if we can make the single OC model easier for multiple implementations, that is a nice to have benefit.

Here are 4 references I've researched.  

[Arista EOS traffic policy counters](https://www.arista.com/en/support/toi/eos-4-24-2f/14550-support-for-traffic-policy-on-interfaces#named-counter)
```none
traffic-policies
   traffic-policy foo
      counter bar baz
      match rule1 ipv4
      actions
         count bar
      match rule2 ipv4
      actions
         count bar baz
```

[JunOS filter based forwarding](https://www.juniper.net/documentation/us/en/software/junos/routing-policy/topics/example/filter-based-forwarding-example.html)

Simplified example:
```none
      match my-af3-traffic ipv4
         source prefix field-set src-af3-prefixes
         destination prefix field-set dst-af3-prefixes
         actions
            count ip-af3-queue
            set dscp 24
            set traffic class 3
            redirect next-hop group ingress-vip
         !
```


[Cisco IOS XR  implements ACL rules for packet matching](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/routing/25xx/configuration/guide/b-routing-cg-cisco8000-25xx/m-policy-based-routing.html#configure-acls-with-policy-based-routing)  in policy based forwarding.  Those [ACL rules can expose counters](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/ip-addresses/b-ip-addresses-cr-8k/access-list-commands.html) in the policybased routing and ACL based forwarding features (which are most similar to OC's "policy-forwarding" model here.   So IOS XR does implement counters for matching rules, but the schema is rather different from other implementations and this model.  

#### JunOS filter based forwarding
```
      match my-af3-traffic ipv4
         source prefix field-set src-af3-prefixes
         destination prefix field-set dst-af3-prefixes
         actions
            count ip-af3-queue
            set dscp 24
            set traffic class 3
            redirect next-hop group ingress-vip
         !
```


**Nokia SR Linux**
I am not able to find any equivalent Nokia SR Linux feature support for policy-forwarding counters (including the current OC defined counters)  of any kind.  This [SR Linux doc documents support for policy-forwarding rules and actions](https://documentation.nokia.com/srlinux/25-7/books/acl-policy-based-routing/policy-based-forward.html) for a recent release, but has no mention of counters.

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -25,7 +25,13 @@ submodule openconfig-pf-forwarding-policies {
     "This submodule contains configuration and operational state
     relating to the definition of policy-forwarding policies.";
 
-  oc-ext:openconfig-version "1.1.0";
+  oc-ext:openconfig-version "1.2.0";
+
+  revision "2025-12-16" {
```
ElodinLaarz: Same date as last one? (2025-12-16)

Have to fix this (and then the other dates, as well cross-file.)

release/models/policy-forwarding/openconfig-pf-forwarding-policies.yang
```diff
@@ -535,6 +614,17 @@ grouping pf-forwarding-policy-action-forwarding-config {
     description
       "Forwarding policy action configuration parameters.";
 
+    leaf count {
+      type leafref {
+        path "../../../../../policy-counters/policy-counter/config/name";
+      }
+
+      // type string;
```
ElodinLaarz: Remove?

ElodinLaarz:
@dplore I think the failing CI checks are from https://github.com/openconfig/models-ci/pull/112 ?? (Nvm: seems to have been failing before then...)

New thought is https://pkg.go.dev/golang.org/x/tools 0.43 came out and needs Go 1.25 .

nokia1adam:
On the debate about whether the counters should be predefined or added as state..

It's important to think about a use case where the same policy is applied to multiple interfaces. In this situation you could want one shared counter X to be incremented by packets matching the rule on all the applied interfaces or you could want per-interface, per-rule counters X1, X2, .. Xn. Both approaches can provide this choice, but they will do it differently:

With pre-configured counter names, the policy-counter itself needs a configuration property to indicate if it is to be used as a single shared counter, or a template for creating per-interface counters.

With auto-created counter names, the count X action in the policy-forwarding rule needs to say whether to use the shared counter with name X or whether to create a new interface-specific counter like X-ethernet1/1.

Both proposals end up in a situation where you have at least some (if not all) counters only being in state and not config.

dplore:
I received feedback from our users and there is a desire to have options for  policy based counters and interface specific policy counters.  I will refactor to add "interface-specific" counters.

dplore:
Here is a pseudo tree I will model to handle per policy  and per policy + interface variants:

```diff
module: openconfig-network-instance
  +--rw network-instances
     +--rw network-instance* [name]
        +--rw policy-forwarding
           +--rw policies
              +--rw policy* [policy-id]
 +                +--rw policy-counters
 +                    +--rw policy-counter* [name]
 +                        +--ro config
 +                           +--ro name
 +                        +--ro state
 +                           +--ro name
 +                              +--ro counters
 +                                 +--ro packets
 +                                 +--ro octets
 +                        +--rw interfaces
 +                           +--rw interface* [interface-ref]
 +                                +--ro state
 +                                   +--ro interface-ref
 +                                   +--ro counters
 +                                      +--ro packets
 +                                      +--ro octets
                 +--rw rules
                    +--rw rule* [sequence-id]
                       +--ro state
                          +--ro sequence-id?      uint32
                          +--ro matched-pkts?     oc-yang:counter64
                          +--ro matched-octets?   oc-yang:counter64
```

amitarista:
> Here is a pseudo tree I will model to handle per policy and per policy + interface variants:
> 
> ```diff
> module: openconfig-network-instance
>   +--rw network-instances
>      +--rw network-instance* [name]
>         +--rw policy-forwarding
>            +--rw policies
>               +--rw policy* [policy-id]
>  +                +--rw policy-counters
>  +                    +--rw policy-counter* [name]
>  +                        +--ro state
>  +                           +--ro name
>  +                              +--ro counters
>  +                                 +--ro packets
>  +                                 +--ro octets
>  +                        +--rw interfaces
>  +                           +--rw interface* [interface-ref]
>  +                                +--ro state
>  +                                   +--ro interface-ref
>  +                                   +--ro counters
>  +                                      +--ro packets
>  +                                      +--ro octets
>                  +--rw rules
>                     +--rw rule* [sequence-id]
>                        +--ro state
>                           +--ro sequence-id?      uint32
>                           +--ro matched-pkts?     oc-yang:counter64
>                           +--ro matched-octets?   oc-yang:counter64
> ```

Thanks for the update Darren. The previously discussed `policy-counters/policy-counter/config` container is missing, is it intentional?

dplore:
> > Here is a pseudo tree I will model to handle per policy and per policy + interface variants:
> > ```diff
> > module: openconfig-network-instance
> > ```
> 
> Thanks for the update Darren. The previously discussed `policy-counters/policy-counter/config` container is missing, is it intentional?

Added the config container to the pseudo tree

navaneethyv (approved):

PR #1423, opened 2026-01-06
Fix typos

felix-kaestner:
### Change Scope

* Fixed some typos by making the following replacements in the YANG files:
  * "entitites" should be "entities"
  * "mininum" should be "minimum"
  * "specifc" should be "specific"
* These changes are backwards compatible as they only change the documentation of fields


release/models/devices-manifest/openconfig-terminal-device-property-types.yang
earies: Bump the patch version and add a new revision-date/reference to the new version for each file edited (pay attention to any submodules in the collection to increment in tandem)

felix-kaestner: Done in 8783bedfc2d04f50a8e7421350b747e93a4c172a. Let me know if there should be any adjustments. :)

PR #1424, opened 2026-01-10
Update style guide to reflect path compression rules.

robshakir:
```
 * (M) doc/openconfig_style_guide.md
   - Provide further documentation in the OpenConfig style guide
     on the rules required to allow programmatic schema transformation.
     These rules have been adhered to throughout the lifetime of
     OpenConfig, and are implemented by widely used code generation
     tools. This PR simply records these defacto style guide rules.

     This PR is accompanied by https://github.com/openconfig/oc-pyang/pull/59
     which implements the relevant linter checks for the rules
     described herein.
```


doc/openconfig_style_guide.md
```diff
@@ -476,6 +477,33 @@ reference for [regular
...
+requiring programmers to reference longer paths that they could. Tooling
+generating programmatic APIs around OpenConfig can improve usability by
+transforming the schema. Clearly, this is wholly reliant on consistency
+in the modelling approach.
+
+The following rules MUST be adherered to within the OpenConfig models to
+ensure that downstream tooling does not break.
+
+* **`list` nodes MUST NOT share their identifier with any grandparent
+  node.** It is not legal for `/foos/foo` and `/foo` to exist. This rule
```
jsterne: I'm not sure the term "grandparent node" is clear enough here. I'd generally interpret grandparent to mean direct parent of a direct parent. But in the example above /foo isn't really a grandparent of /foos/foo. /foos isn't the parent of /foo.  /foos is a sibling of /foo.
Maybe you mean a list node must not share their identifier with any ancestor node or sibling node of any ancestor?  (and is it node or just lists it can't share with?).

doc/openconfig_style_guide.md
```diff
@@ -476,6 +477,33 @@ reference for [regular
...
+  exists to allow the `foos` container to be removed during schema
+  transformation.
+* **A leaf-node may not share a name with a grandparent**. It is not legal
+  for `/a/config/leaf` and `/a/leaf` to both exist nor for `/b/state/leaf`
```
jsterne, line 502: This could use further clarification because it looks & feels very close to a convention of OpenConfig for list keys:

     |     +--rw acl-entries
     |        +--rw acl-entry* [sequence-id]
     |           +--rw sequence-id        -> ../config/sequence-id
     |           +--rw config
     |           |  +--rw sequence-id?   uint32

doc/openconfig_style_guide.md
```diff
@@ -476,6 +477,33 @@ reference for [regular
...
+consistent programmatic handling of the schema. For instance, the `config` and
+`state` containers used for operational state, and surrounding containers for
+YANG `list` statements. In some cases (e.g., programmatically generating
+configuration), this schema verbosity negatively impacts usability -- e.g.,
+requiring programmers to reference longer paths that they could. Tooling
+generating programmatic APIs around OpenConfig can improve usability by
+transforming the schema. Clearly, this is wholly reliant on consistency
+in the modelling approach.
+
+The following rules MUST be adherered to within the OpenConfig models to
```
ElodinLaarz: nit: (spl) "adhered"

doc/openconfig_style_guide.md
```diff
@@ -476,6 +477,33 @@ reference for [regular
...
 However, this is not a commonly used standard for implementors.
 
+## Schema Usability
+
+Some elements of the OpenConfig style guide are implemented in order to provide
+consistent programmatic handling of the schema. For instance, the `config` and
+`state` containers used for operational state, and surrounding containers for
+YANG `list` statements. In some cases (e.g., programmatically generating
+configuration), this schema verbosity negatively impacts usability -- e.g.,
+requiring programmers to reference longer paths that they could. Tooling
```
ElodinLaarz: Phrasing: `... requiring programmers to reference paths that are longer than necessary.` or similar.

doc/openconfig_style_guide.md
```diff
@@ -476,6 +477,33 @@ reference for [regular
...
+in the modelling approach.
+
+The following rules MUST be adherered to within the OpenConfig models to
+ensure that downstream tooling does not break.
+
+* **`list` nodes MUST NOT share their identifier with any grandparent
+  node.** It is not legal for `/foos/foo` and `/foo` to exist. This rule
+  exists to allow the `foos` container to be removed during schema
+  transformation.
+* **A leaf-node may not share a name with a grandparent**. It is not legal
```
ElodinLaarz: nit: Referred to as `leaf node` (without the hyphen) elsewhere in the doc. Make consistent.

doc/openconfig_style_guide.md
ElodinLaarz: +1 to update the phrasing.

In this case, the grandparent is the root (`openconfig` or whatever), which feels surprising since it isn't a path element.

The "grandparent" description is more clear in a case where they're in a container, e.g. `/grandparent/foo` and `/grandparent/foos/foo`, but even in such a case, it's a "parent node" of one of the `foo`, which could be confusing.

Perhaps we want to say "a node MUST NOT share its identifier with any of *the children* of its grandparent node."?

doc/openconfig_style_guide.md
```diff
@@ -476,6 +477,33 @@ reference for [regular
...
+  for `/a/config/leaf` and `/a/leaf` to both exist nor for `/b/state/leaf` 
+  and `/b/leaf` to both exist. This rule ensures that the `config` and `state`
+  containers can be removed during schema transformation.
+
+An example of programmatic compression is implemented for the generation of
+code in ygot -- both for Go and Protobuf artifact generation
+((reference)[https://github.com/openconfig/ygot/blob/master/docs/design.md#openconfig-path-compression]).
```
ElodinLaarz: Link formatting is backwards. Should be `[]()`

doc/openconfig_style_guide.md
ElodinLaarz: Yeah, perhaps it would be useful to distinguish leaf refs from leafs?

rolandphung: What happens when we have to deprecate the `/a/config/leaf` in favor for `/a/leaf/config`? Are we restricting those types of recategorization/refactoring to have unique names?

earies:
Just wanted to post an overall comment/observation - this style proposal is to accomodate patterns for collapsing schema nodes and codegen for ygot and just want to confirm that that then means ygot would not be able to accomodate patterns that fall outside of this boundary - is that correct?  (e.g. other model-sets)

I'll admit I haven't looked across various repos verbiage to this matter but the interoperability between OC projects and content outside of OC projects scope has been a growing concern that I'm not sure has been fully articulated other than YMMV.

If that is all correct, do we agree the following from ygot README is sufficient to articulate this?

_Whilst ygot is designed to work with any YANG module, for OpenConfig modules, it can provide transformations of
the schema to optimise the data structures that are produced for use in systems that generate data instances of the
models for configuration purposes. These helper methods require that the OpenConfig style guide patterns are implemented, a model can be verified to conform with these requirements using the OpenConfig linter._

robshakir:
> Just wanted to post an overall comment/observation - this style proposal is to accomodate patterns for collapsing schema nodes and codegen for ygot and just want to confirm that that then means ygot would not be able to accomodate patterns that fall outside of this boundary - is that correct? (e.g. other model-sets)

ygot has never done schema compression (what it refers to these transformations as) for any other type of model. It could be extended to do so, but it hasn't been thus far. 

> I'll admit I haven't looked across various repos verbiage to this matter but the interoperability between OC projects and content outside of OC projects scope has been a growing concern that I'm not sure has been fully articulated other than YMMV.

I'm not sure I understand this comment. Is your concern that ygot is outside of the OpenConfig project's scope? 

> If that is all correct, do we agree the following from ygot README is sufficient to articulate this?
> 
> _Whilst ygot is designed to work with any YANG module, for OpenConfig modules, it can provide transformations of the schema to optimise the data structures that are produced for use in systems that generate data instances of the models for configuration purposes. These helper methods require that the OpenConfig style guide patterns are implemented, a model can be verified to conform with these requirements using the OpenConfig linter._

robshakir:
> > ygot has never done schema compression (what it refers to these transformations as) for any other type of model. It could be extended to do so, but it hasn't been thus far.
> 
> I haven't dug into ygot sources but are you suggesting the schema compression and schema design assumptions are specific to OC model-sets (schema aware)? And if used against alternate model-sets where there is likelyhood to breach these rulesets, ygot can support (assuming coverage of the entirety of YANG 1.0/1.1 language) but not using these compression rules?

That's absolutely correct. There are active users of goyang and ygot that are using models other than OpenConfig. The compression rules are specifically to improve developer experience for models that comply with OpenConfig's style rules. Today, this is actively used for both OpenConfig and OpenTrafficGenerator models.

> > > I'll admit I haven't looked across various repos verbiage to this matter but the interoperability between OC projects and content outside of OC projects scope has been a growing concern that I'm not sure has been fully articulated other than YMMV.
> > 
> > 
> > I'm not sure I understand this comment. Is your concern that ygot is outside of the OpenConfig project's scope?
> 
> Was just trying to understand since this imposes a design protocol on the YANG modeling that if this means ygot and OC YANG modeling are only interoperable as a result and how ygot might handle schemas that cannot conform to this ruleset (e.g. native, IETF, etc...)

This change simply makes a rule that has been adhered to throughout the modelling explicit (i.e., converts it from a convention to a "rule"). ygot will always support schemas using the "uncompressed" mode (and has done for ~10 years in production network deployments :-)). Essentially, ygot is a generic toolkit for YANG users in Go, but has a "works with OpenConfig" badge since a number of its primary users use OpenConfig. :-)

doc/openconfig_style_guide.md
robshakir: ACK -- I fixed two things here:

* I adopted Caleb's proposed wording here.
* I expanded on the examples to show both an example where the grandparent is the root, and a case where it is a container.

doc/openconfig_style_guide.md
robshakir: Fixed, thanks.

doc/openconfig_style_guide.md
robshakir: Thanks, fixed.

doc/openconfig_style_guide.md
robshakir: Fixed, thanks.

doc/openconfig_style_guide.md
robshakir: Thanks, fixed.

doc/openconfig_style_guide.md
robshakir: Added a clarification of this case.

doc/openconfig_style_guide.md
```diff
@@ -476,6 +477,39 @@ reference for [regular
...
+The following rules MUST be adhered to within the OpenConfig models to
+ensure that downstream tooling does not break.
+
+* **`list` nodes MUST NOT share their identifier with any of the children
+  of its grandparent node.** It is not legal for `/a/foos/foo` and `/a/foo` 
+  to exist, or for `/interfaces/interface` and `/interface` to exist, where
+  `/a/foos/foo` and `/interfaces/interface` are `list` nodes. This rule
+  exists to allow the (style-guide-required) "surrounding" container of 
+  a list to be removed during schema transformation.
+* **A leaf node may not share a name with a grandparent**. It is not legal
```
ElodinLaarz, line 501: `**A leaf node may not share a name with a grandparent**.`

Do we want this here still? Seems unrelated to the explanation that follows. (And is also not a true statement -- presumably having been replaced with the previous bullet about children of its grandparent.)

doc/openconfig_style_guide.md
```diff
@@ -476,6 +477,39 @@ reference for [regular
...
+  for `/a/config/leaf` and `/a/leaf` to both exist nor for `/b/state/leaf` 
+  and `/b/leaf` to both exist. The single exception to this rule is the
+  special case of nodes which act as the `key` of a `list`, where the leaf
+  must be a direct child of the `list` node. Such leaves MUST be the `key`
```
ElodinLaarz, line 505: Optional way to rephrase (also fine as is):

"The single exception to this rule is the OpenConfig list key pattern. The leaf inside the `config` or `state` container is permitted to share its name with the `leafref` node acting as the `key` of the parent list."

doc/openconfig_style_guide.md
ElodinLaarz, line 501: Also, if you keep some requirement around you probably want to use must vs may, per [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) -- `MAY` (optional) vs `MUST` (required).

ElodinLaarz (approved):
Left some things to tidy up, but generally LGTM.

ElodinLaarz:
Reviewed at the OC Operators Meeting March 31, 2026 - @ElodinLaarz to fork the PR and submit with the suggested changes.

ElodinLaarz:
merged in https://github.com/openconfig/public/pull/1475

PR #1436, opened 2026-02-10
Fix yanglint schema warnings

rgwilton:
### Change Scope

* Fixing yanglint warnings (there were no warnings after this run).
* I've marked all of these as bugfixes.  In theory the changes are not backwards compatible, but generally the YANG would have been broken previously, so ...

### Platform Implementations

 * N/A.  No new nodes added or removed.

### Tree View

* N/A.  No changes to the tree have occurred.

The change involve:
- Adding the module prefix for when identities are referenced in when statements (YANG 1.1 requires this, but I suspect that these may be broken when evaluating when expressions for YANG 1).
- Removing module prefixes for enum values used in when expressions (these are just broken).
- Correcting some typos (e.g., TE_LINK_MAXIUMUM_RESERVABLE_BANDWIDTH => TE_LINK_MAXIMUM_RESERVABLE_BANDWIDTH)
- Deleting references to IPV4_ADDRESS_INDEX and IPV6_ADDRESS_INDEX which are not defined as identities.
- I've also fixed up some references in authz, pathz, certz, credz where the paths don't appear to be correct.  I've not convinced that the paths are necessarily correct after my changes (but they were more correct that before and I'm not an xpath expert).

### DIFF
--- a/release/models/acl/openconfig-acl.yang
+++ b/release/models/acl/openconfig-acl.yang
@@ -34,7 +34,12 @@ module openconfig-acl {
     packets should be handled. Entries have a type that indicates
     the type of match criteria, e.g., MAC layer, IPv4, IPv6, etc.";
 
-  oc-ext:openconfig-version "1.3.3";
+  oc-ext:openconfig-version "1.3.4";
+  revision "2026-02-10" {
+    description
+      "Fix yanglint schema warnings";
+    reference "1.3.4";
+  }
 
   revision "2023-02-06" {
     description
@@ -451,8 +456,8 @@ module openconfig-acl {
         }
 
         uses oc-match:ethernet-header-top {
-          when "../../config/type='ACL_L2' or " +
-            "../../config/type='ACL_MIXED'" {
+          when "../../config/type='oc-acl:ACL_L2' or " +
+            "../../config/type='oc-acl:ACL_MIXED'" {
             description
               "MAC-layer fields are valid when the ACL type is L2 or
               MIXED";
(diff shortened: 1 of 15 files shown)


release/models/acl/openconfig-acl.yang
```diff
@@ -34,7 +34,12 @@ module openconfig-acl {
     packets should be handled. Entries have a type that indicates
     the type of match criteria, e.g., MAC layer, IPv4, IPv6, etc.";
 
-  oc-ext:openconfig-version "1.3.3";
+  oc-ext:openconfig-version "1.3.4";
```
earies, line 37: nit: add blank line

earies (commented):
+1 for this cleanup - thx @rgwilton .... we did similar long back as we ran into them but don't believe the model set was swept and likely more of these were added over time.

I think patch version ++ is acceptable here as prior was likely just danced around and possibly not working as expected in various uses.

rgwilton:
> Should more validation be added to the CI pipeline of OC models to prevent these types of issues from creeping back in? (e.g. adding libyang/yanglint)

Yes, I think that would be great if that could be added.  Particularly because some operator tool chains are relying on these validating cleanly.

release/models/aft/openconfig-aft-common.yang
```diff
@@ -340,7 +346,7 @@ submodule openconfig-aft-common {
             }
 
             container ipv4 {
-              when "../state/type = 'oc-aftt:IPV4'";
+              when "../state/type = 'IPV4'";
```
dplore, line 349: Why is this prefix removed where it is being added elsewhere?  Is this intended?

release/models/optical-transport/openconfig-terminal-device.yang
```diff
@@ -76,13 +76,19 @@ module openconfig-terminal-device {
...
+    description
+      "Fix yanglint schema warnings";
+    reference "1.12.1";
+  }
 
   revision "2026-01-14" {
     description
       "Removal of references to unused/removed LLDP groupings";
     reference "1.12.0";
-  }
```
dplore, line 85: looks like a typo?  This should not be deleted.

dplore: This is probably the source of most if not all of the failing CI checks for this PR.

navaneethyv:
@rgwilton Hi Robert, do you mind taking care of the conflicts?

PR #1437, opened 2026-02-10
allow-config-mismatch extension

dplore:
It is common for hardware to be installed in a device which does not 100% match the applied or newly set configuration.  In this scenario it is desirable to still accept the configuration and return a warning.  

For example, as a network operator I want to allow a device to have unexpected interface transceivers installed and still be able to push configuration changes to a device.  Without this behavior, a single unexcepted transceiver installed in a device would block all configuration changes.  Blocking these configuration changes can be much more operationally impacting than having a single interface be unavailable.  

One way to address this might be with a yang extension to annotate containers or leaves which should allow a configuration mismatch with the underlying hardware.

### Change Scope
* Introduce a new OpenConfig yang extension called `allow-config-mismatch`
* Annotate interfaces and subinterfaces containers to include `allow-config-mismatch`

### DIFF
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -9,9 +9,9 @@ module openconfig-interfaces {
 
   // import some basic types
   import ietf-interfaces { prefix ietf-if; }
-  import openconfig-yang-types { prefix oc-yang; }
   import openconfig-types { prefix oc-types; }
   import openconfig-extensions { prefix oc-ext; }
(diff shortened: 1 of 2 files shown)


jsterne:
IMO we should not have a new extension for this. That should be the expected behavior. Accepting a config should not depend on state (e.g. what transceiver is plugged in, what linecards are plugged in, etc).
That config is perfectly valid (as a self contained config). Someone could plug in a different (correct) transceiver and it would all work.

PR #1441, opened 2026-03-11
Add initial prefix filter

ElodinLaarz:
### Change Scope

* We add a new container to filter the `afts` tree based on a named routing policy.
* The change is backwards compatible, though it marks `afts` as a `rw` container rather than a `ro` container since we're adding a configuration leaf.

### Platform Implementations

This request is being made without any existing implementations that I know of. We will be requesting this from the vendors. (and so we will need at least 2 approvals from different vendors)

### Tree View

```diff
-        +--ro afts
+        +--rw afts
         |  +--ro ipv4-unicast
         |  |  +--ro ipv4-entry* [prefix]
         |  |     +--ro prefix    -> ../state/prefix
         |  |     +--ro state
         |  |        +--ro prefix?               oc-inet:ipv4-prefix
         |  |        +--ro counters
         |  |        |  x--ro packets-forwarded?          oc-yang:counter64
         |  |        |  x--ro octets-forwarded?           oc-yang:counter64
         |  |        |  x--ro packets-forwarded-backup?   oc-yang:counter64
         |  |        |  x--ro octets-forwarded-backup?    oc-yang:counter64
         |  |        +--ro entry-metadata?       binary
         |  |        +--ro origin-protocol?      identityref
         |  |        +--ro decapsulate-header?   oc-aftt:encapsulation-header-type
         |  +--ro ipv6-unicast
         |  |  +--ro ipv6-entry* [prefix]
         |  |     +--ro prefix    -> ../state/prefix
         |  |     +--ro state
         |  |        +--ro prefix?               oc-inet:ipv6-prefix
         |  |        +--ro counters
         |  |        |  x--ro packets-forwarded?          oc-yang:counter64
         |  |        |  x--ro octets-forwarded?           oc-yang:counter64
         |  |        |  x--ro packets-forwarded-backup?   oc-yang:counter64
         |  |        |  x--ro octets-forwarded-backup?    oc-yang:counter64
         |  |        +--ro entry-metadata?       binary
         |  |        +--ro origin-protocol?      identityref
         |  |        +--ro decapsulate-header?   oc-aftt:encapsulation-header-type
         |  +--ro policy-forwarding
         |  |  +--ro policy-forwarding-entry* [index]
         |  |     +--ro index    -> ../state/index
         |  |     +--ro state
         |  |        +--ro index?            uint64
         |  |        +--ro ip-prefix?        oc-inet:ip-prefix
         |  |        +--ro mac-address?      oc-yang:mac-address
         |  |        +--ro mpls-label?       oc-mplst:mpls-label
         |  |        +--ro mpls-tc?          oc-mplst:mpls-tc
         |  |        +--ro ip-dscp?          oc-inet:dscp
         |  |        +--ro ip-protocol?      oc-pkt-match-types:ip-protocol-type
         |  |        +--ro l4-src-port?      oc-inet:port-number
         |  |        +--ro l4-dst-port?      oc-inet:port-number
         |  |        +--ro counters
         |  |        |  x--ro packets-forwarded?   oc-yang:counter64
         |  |        |  x--ro octets-forwarded?    oc-yang:counter64
         |  |        +--ro entry-metadata?   binary
         |  +--ro mpls
         |  |  +--ro label-entry* [label]
         |  |     +--ro label    -> ../state/label
         |  |     +--ro state
         |  |        +--ro label?                     oc-mplst:mpls-label
         |  |        +--ro counters
         |  |        |  x--ro packets-forwarded?   oc-yang:counter64
         |  |        |  x--ro octets-forwarded?    oc-yang:counter64
         |  |        +--ro entry-metadata?            binary
         |  |        +--ro popped-mpls-label-stack*   oc-mplst:mpls-label
         |  +--ro ethernet
         |  |  +--ro mac-entry* [mac-address]
         |  |     +--ro mac-address    -> ../state/mac-address
         |  |     +--ro state
         |  |        +--ro mac-address?      oc-yang:mac-address
         |  |        +--ro counters
         |  |        |  x--ro packets-forwarded?   oc-yang:counter64
         |  |        |  x--ro octets-forwarded?    oc-yang:counter64
         |  |        +--ro entry-metadata?   binary
         |  +--ro state-synced
         |  |  +--ro state
         |  |     +--ro ipv4-unicast?   boolean
         |  |     +--ro ipv6-unicast?   boolean
         |  +--ro next-hop-groups
         |  |  +--ro next-hop-group* [id]
         |  |     +--ro id             -> ../state/id
         |  |     +--ro state
         |  |     |  +--ro id?                      uint64
         |  |     |  +--ro next-hop-group-name?     string
         |  |     |  +--ro programmed-id?           uint64
         |  |     |  +--ro color?                   uint64
         |  |     |  +--ro backup-next-hop-group?   -> ../../../next-hop-group/state/id
         |  |     |  +--ro backup-active?           boolean
         |  |     +--ro next-hops
         |  |     |  +--ro next-hop* [index]
         |  |     |     +--ro index    -> ../state/index
         |  |     |     +--ro state
         |  |     |        +--ro index?    -> ../../../../../../next-hops/next-hop/state/index
         |  |     |        +--ro weight?   uint64
         |  |     +--ro conditional
         |  |        +--ro condition* [id]
         |  |           +--ro id                  -> ../state/id
         |  |           +--ro state
         |  |           |  +--ro id?               uint64
         |  |           |  +--ro dscp*             oc-inet:dscp
         |  |           |  +--ro next-hop-group?   -> ../../../../../next-hop-group/state/id
         |  |           +--ro input-interfaces
         |  |              +--ro input-interface* [id]
         |  |                 +--ro id       -> ../state/id
         |  |                 +--ro state
         |  |                    +--ro id?             string
         |  |                    +--ro interface?      -> /oc-if:interfaces/interface/name
         |  |                    +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
         |  +--ro next-hops
-        |     +--ro next-hop* [index]
-        |        +--ro index            -> ../state/index
-        |        +--ro state
-        |        |  +--ro index?                     uint64
-        |        |  +--ro programmed-index?          uint64
-        |        |  +--ro ip-address?                oc-inet:ip-address
-        |        |  +--ro mac-address?               oc-yang:mac-address
-        |        |  +--ro pop-top-label?             boolean
-        |        |  +--ro pushed-mpls-label-stack*   oc-mplst:mpls-label
-        |        |  +--ro encapsulate-header?        oc-aftt:encapsulation-header-type
-        |        |  +--ro decapsulate-header?        oc-aftt:encapsulation-header-type
-        |        |  +--ro origin-protocol?           identityref
-        |        |  +--ro lsp-name?                  string
-        |        |  +--ro counters
-        |        |  |  x--ro packets-forwarded?   oc-yang:counter64
-        |        |  |  x--ro octets-forwarded?    oc-yang:counter64
-        |        |  +--ro vni-label?                 oc-evpn-types:evi-id
-        |        |  +--ro tunnel-src-ip-address?     oc-inet:ip-address
-        |        +--ro ip-in-ip
-        |        |  +--ro state
-        |        |     +--ro src-ip?   oc-inet:ip-address
-        |        |     +--ro dst-ip?   oc-inet:ip-address
-        |        +--ro gre
-        |        |  +--ro state
-        |        |     +--ro src-ip?   oc-inet:ip-address
-        |        |     +--ro dst-ip?   oc-inet:ip-address
-        |        |     +--ro ttl?      uint8
-        |        +--ro encap-headers
-        |        |  +--ro encap-header* [index]
-        |        |     +--ro index     -> ../state/index
-        |        |     +--ro state
-        |        |     |  +--ro index?   uint8
-        |        |     |  +--ro type?    oc-aftt:encapsulation-header-type
-        |        |     +--ro gre
-        |        |     |  +--ro state
-        |        |     |     +--ro src-ip?   oc-inet:ip-address
-        |        |     |     +--ro dst-ip?   oc-inet:ip-address
-        |        |     |     +--ro ttl?      uint8
-        |        |     +--ro ipv4
-        |        |     |  +--ro state
-        |        |     |     +--ro src-ip?   oc-inet:ip-address
-        |        |     |     +--ro dst-ip?   oc-inet:ip-address
-        |        |     +--ro ipv6
-        |        |     |  +--ro state
-        |        |     |     +--ro src-ip?   oc-inet:ip-address
-        |        |     |     +--ro dst-ip?   oc-inet:ip-address
-        |        |     +--ro mpls
-        |        |     |  +--ro state
-        |        |     |     +--ro traffic-class?      oc-mplst:mpls-tc
-        |        |     |     +--ro mpls-label-stack*   oc-mplst:mpls-label
-        |        |     +--ro udp-v4
-        |        |     |  +--ro state
-        |        |     |     +--ro src-ip?         oc-inet:ipv4-address
-        |        |     |     +--ro dst-ip?         oc-inet:ipv4-address
-        |        |     |     +--ro dscp?           oc-inet:dscp
-        |        |     |     +--ro src-udp-port?   oc-inet:port-number
-        |        |     |     +--ro dst-udp-port?   oc-inet:port-number
-        |        |     |     +--ro ip-ttl?         uint8
-        |        |     +--ro udp-v6
-        |        |     |  +--ro state
-        |        |     |     +--ro src-ip?         oc-inet:ipv6-address
-        |        |     |     +--ro dst-ip?         oc-inet:ipv6-address
-        |        |     |     +--ro dscp?           oc-inet:dscp
-        |        |     |     +--ro src-udp-port?   oc-inet:port-number
-        |        |     |     +--ro dst-udp-port?   oc-inet:port-number
-        |        |     |     +--ro ip-ttl?         uint8
-        |        |     +--ro vxlan
-        |        |        +--ro state
-        |        |           +--ro vni-label?               oc-evpn-types:evi-id
-        |        |           +--ro tunnel-src-ip-address?   oc-inet:ip-address
-        |        +--ro interface-ref
-        |           +--ro state
-        |              +--ro interface?      -> /oc-if:interfaces/interface/name
-        |              +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
+        |  |  +--ro next-hop* [index]
+        |  |     +--ro index            -> ../state/index
+        |  |     +--ro state
+        |  |     |  +--ro index?                     uint64
+        |  |     |  +--ro programmed-index?          uint64
+        |  |     |  +--ro ip-address?                oc-inet:ip-address
+        |  |     |  +--ro mac-address?               oc-yang:mac-address
+        |  |     |  +--ro pop-top-label?             boolean
+        |  |     |  +--ro pushed-mpls-label-stack*   oc-mplst:mpls-label
+        |  |     |  +--ro encapsulate-header?        oc-aftt:encapsulation-header-type
+        |  |     |  +--ro decapsulate-header?        oc-aftt:encapsulation-header-type
+        |  |     |  +--ro origin-protocol?           identityref
+        |  |     |  +--ro lsp-name?                  string
+        |  |     |  +--ro counters
+        |  |     |  |  x--ro packets-forwarded?   oc-yang:counter64
+        |  |     |  |  x--ro octets-forwarded?    oc-yang:counter64
+        |  |     |  +--ro vni-label?                 oc-evpn-types:evi-id
+        |  |     |  +--ro tunnel-src-ip-address?     oc-inet:ip-address
+        |  |     +--ro ip-in-ip
+        |  |     |  +--ro state
+        |  |     |     +--ro src-ip?   oc-inet:ip-address
+        |  |     |     +--ro dst-ip?   oc-inet:ip-address
+        |  |     +--ro gre
+        |  |     |  +--ro state
+        |  |     |     +--ro src-ip?   oc-inet:ip-address
+        |  |     |     +--ro dst-ip?   oc-inet:ip-address
+        |  |     |     +--ro ttl?      uint8
+        |  |     +--ro encap-headers
+        |  |     |  +--ro encap-header* [index]
+        |  |     |     +--ro index     -> ../state/index
+        |  |     |     +--ro state
+        |  |     |     |  +--ro index?   uint8
+        |  |     |     |  +--ro type?    oc-aftt:encapsulation-header-type
+        |  |     |     +--ro gre
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?   oc-inet:ip-address
+        |  |     |     |     +--ro dst-ip?   oc-inet:ip-address
+        |  |     |     |     +--ro ttl?      uint8
+        |  |     |     +--ro ipv4
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?   oc-inet:ip-address
+        |  |     |     |     +--ro dst-ip?   oc-inet:ip-address
+        |  |     |     +--ro ipv6
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?   oc-inet:ip-address
+        |  |     |     |     +--ro dst-ip?   oc-inet:ip-address
+        |  |     |     +--ro mpls
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro traffic-class?      oc-mplst:mpls-tc
+        |  |     |     |     +--ro mpls-label-stack*   oc-mplst:mpls-label
+        |  |     |     +--ro udp-v4
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?         oc-inet:ipv4-address
+        |  |     |     |     +--ro dst-ip?         oc-inet:ipv4-address
+        |  |     |     |     +--ro dscp?           oc-inet:dscp
+        |  |     |     |     +--ro src-udp-port?   oc-inet:port-number
+        |  |     |     |     +--ro dst-udp-port?   oc-inet:port-number
+        |  |     |     |     +--ro ip-ttl?         uint8
+        |  |     |     +--ro udp-v6
+        |  |     |     |  +--ro state
+        |  |     |     |     +--ro src-ip?         oc-inet:ipv6-address
+        |  |     |     |     +--ro dst-ip?         oc-inet:ipv6-address
+        |  |     |     |     +--ro dscp?           oc-inet:dscp
+        |  |     |     |     +--ro src-udp-port?   oc-inet:port-number
+        |  |     |     |     +--ro dst-udp-port?   oc-inet:port-number
+        |  |     |     |     +--ro ip-ttl?         uint8
+        |  |     |     +--ro vxlan
+        |  |     |        +--ro state
+        |  |     |           +--ro vni-label?               oc-evpn-types:evi-id
+        |  |     |           +--ro tunnel-src-ip-address?   oc-inet:ip-address
+        |  |     +--ro interface-ref
+        |  |        +--ro state
+        |  |           +--ro interface?      -> /oc-if:interfaces/interface/name
+        |  |           +--ro subinterface?   -> /oc-if:interfaces/interface[oc-if:name=current()/../interface]/subinterfaces/subinterface/index
+        |  +--rw global-filter
+        |     +--rw config
+        |     |  +--rw ipv4-policy?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/config/name
+        |     |  +--rw ipv6-policy?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/config/name
+        |     +--ro state
+        |        +--ro ipv4-policy?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/config/name
+        |        +--ro ipv6-policy?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/config/name
```

Notably,

```
+        |  +--rw global-filter
+        |     +--rw config
+        |     |  +--rw ipv4-policy?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/config/name
+        |     |  +--rw ipv6-policy?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/config/name
+        |     +--ro state
+        |        +--ro ipv4-policy?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/config/name
+        |        +--ro ipv6-policy?   -> /oc-rpol:routing-policy/policy-definitions/policy-definition/config/name
```

Is the part being added (and the rest are part of the diff because there's an extra `|` at the start of the line).

## Feature Profile Testing

Some of the functionality / expected behavior can be found in [this feature profile PR](https://github.com/openconfig/featureprofiles/pull/5217).

### DIFF
--- a/release/models/aft/openconfig-aft-global-filter.yang
+++ b/release/models/aft/openconfig-aft-global-filter.yang
@@ -1,0 +1,72 @@
+submodule openconfig-aft-global-filter {
+  belongs-to "openconfig-aft" {
+    prefix "oc-aft";
+  }
+
+  import openconfig-extensions { prefix "oc-ext"; }
+  import openconfig-routing-policy { prefix "oc-rpol"; }
+
+  organization
+    "OpenConfig working group";
+
+  contact
+    "OpenConfig working group
+    www.openconfig.net";
+
+  description
+    "Submodule containing definitions of groupings for the global
+    filter configuration applicable to abstract forwarding tables.";
+
+  oc-ext:openconfig-version "3.3.0";
+
+  revision "2026-03-10" {
+    description
+      "Add global-filter container to support per-network-instance
+      AFT prefix filtering.";
+    reference "3.3.0";
+  }
+
+  grouping aft-global-filter-config {
+    description
+      "Configuration parameters for the AFT global filter.";
+
+    leaf policy-name {
+      type leafref {
+        path "/oc-rpol:routing-policy/oc-rpol:policy-definitions" +
+             "/oc-rpol:policy-definition/oc-rpol:config/oc-rpol:name";
+      }
+      description
+        "Reference to a routing policy that is used to filter the
+        prefixes that are installed into the AFT. Only prefixes that
+        are accepted by the referenced policy are installed.";
+    }
+  }
+
+  grouping aft-global-filter-structural {
+    description
+      "Structural grouping defining the schema for the global filter
+      applied to the abstract forwarding tables.";
+
+    container global-filter {
+      description
+        "Configuration and state for a global filter applied to the
+        AFT. The filter is used to restrict which prefixes are
+        installed into the AFT for the network instance.";
+
+      container config {
+        description
+          "Configuration parameters for the AFT global filter.";
+
+        uses aft-global-filter-config;
+      }
+
+      container state {
+        config false;
+        description
+          "Operational state parameters for the AFT global filter.";
+
+        uses aft-global-filter-config;
+      }
+    }
+  }
+}
--- a/release/models/aft/openconfig-aft-global-filter.yang
+++ b/release/models/aft/openconfig-aft-global-filter.yang
@@ -21,8 +21,8 @@ submodule openconfig-aft-global-filter {
 
   revision "2026-03-10" {
     description
-      "Add global-filter container to support per-network-instance
-      AFT prefix filtering.";
+      "Add global-filter container to support per-network-instance and per
+      AFT filtering, with routing policy split by address family.";
     reference "3.3.0";
   }
 
@@ -30,15 +30,26 @@ submodule openconfig-aft-global-filter {
     description
       "Configuration parameters for the AFT global filter.";
 
-    leaf policy-name {
+    leaf ipv4-policy {
       type leafref {
         path "/oc-rpol:routing-policy/oc-rpol:policy-definitions" +
              "/oc-rpol:policy-definition/oc-rpol:config/oc-rpol:name";
       }
       description
         "Reference to a routing policy that is used to filter the
-        prefixes that are installed into the AFT. Only prefixes that
-        are accepted by the referenced policy are installed.";
+        IPv4 prefixes that are installed into the AFT. Only prefixes
+        that are accepted by the referenced policy are installed.";
+    }
+
+    leaf ipv6-policy {
+      type leafref {
+        path "/oc-rpol:routing-policy/oc-rpol:policy-definitions" +
+             "/oc-rpol:policy-definition/oc-rpol:config/oc-rpol:name";
+      }
+      description
+        "Reference to a routing policy that is used to filter the
+        IPv6 prefixes that are installed into the AFT. Only prefixes
+        that are accepted by the referenced policy are installed.";
     }
   }
 
--- a/release/models/aft/openconfig-aft-global-filter.yang
+++ b/release/models/aft/openconfig-aft-global-filter.yang
@@ -15,14 +15,14 @@ submodule openconfig-aft-global-filter {
 
   description
     "Submodule containing definitions of groupings for the global
-    filter configuration applicable to abstract forwarding tables.";
+     filter configuration applicable to abstract forwarding tables.";
 
   oc-ext:openconfig-version "3.3.0";
 
   revision "2026-03-10" {
     description
       "Add global-filter container to support per-network-instance and per
-      AFT filtering, with routing policy split by address family.";
+       AFT filtering, with routing policy split by address family.";
     reference "3.3.0";
   }
 
@@ -36,11 +36,11 @@ submodule openconfig-aft-global-filter {
       }
       description
         "Reference to a routing policy that is used to filter the
-        IPv6 prefixes that are installed into the AFT. Only prefixes
-        that are accepted by the referenced policy are installed.
-        The policy may, for example, use an explicit prefix set to
-        match prefixes, or match prefixes that are more specific than
-        a given prefix.";
+         IPv6 prefixes that are present in the FIB. Only prefixes
+         that are accepted by the referenced policy are streamed.
+         The policy may, for example, use an explicit prefix set to
+         match prefixes, or match prefixes that are more specific than
+         a given prefix.";
     }
 
     leaf ipv4-policy {
@@ -50,11 +50,11 @@ submodule openconfig-aft-global-filter {
       }
       description
         "Reference to a routing policy that is used to filter the
-        IPv4 prefixes that are installed into the AFT. Only prefixes
-        that are accepted by the referenced policy are installed.
-        The policy may, for example, use an explicit prefix set to
-        match prefixes, or match prefixes that are more specific than
-        a given prefix.";
+         IPv4 prefixes that are in the FIB. Only prefixes
+         that are accepted by the referenced policy are streamed.
+         The policy may, for example, use an explicit prefix set to
+         match prefixes, or match prefixes that are more specific than
+         a given prefix.";
     }
   }
 
@@ -62,13 +62,13 @@ submodule openconfig-aft-global-filter {
   grouping aft-global-filter-structural {
     description
       "Structural grouping defining the schema for the global filter
-      applied to the abstract forwarding tables.";
+       applied to the abstract forwarding tables.";
 
     container global-filter {
       description
         "Configuration and state for a global filter applied to the
-        AFT. The filter is used to restrict which prefixes are
-        installed into the AFT for the network instance.";
+         AFT. The filter is used to restrict which prefixes are
+         installed into the AFT for the network instance.";
 
       container config {
         description
--- a/release/models/aft/openconfig-aft-global-filter.yang
+++ b/release/models/aft/openconfig-aft-global-filter.yang
@@ -29,6 +29,19 @@ submodule openconfig-aft-global-filter {
   grouping aft-global-filter-config {
     description
       "Configuration parameters for the AFT global filter.";
+    leaf ipv6-policy {
+      type leafref {
+        path "/oc-rpol:routing-policy/oc-rpol:policy-definitions" +
+             "/oc-rpol:policy-definition/oc-rpol:config/oc-rpol:name";
+      }
+      description
+        "Reference to a routing policy that is used to filter the
+        IPv6 prefixes that are installed into the AFT. Only prefixes
+        that are accepted by the referenced policy are installed.
+        The policy may, for example, use an explicit prefix set to
+        match prefixes, or match prefixes that are more specific than
+        a given prefix.";
+    }
 
     leaf ipv4-policy {
       type leafref {
@@ -43,21 +56,8 @@ submodule openconfig-aft-global-filter {
         match prefixes, or match prefixes that are more specific than
         a given prefix.";
     }
-
-    leaf ipv6-policy {
-      type leafref {
-        path "/oc-rpol:routing-policy/oc-rpol:policy-definitions" +
-             "/oc-rpol:policy-definition/oc-rpol:config/oc-rpol:name";
-      }
-      description
-        "Reference to a routing policy that is used to filter the
-        IPv6 prefixes that are installed into the AFT. Only prefixes
-        that are accepted by the referenced policy are installed.
-        The policy may, for example, use an explicit prefix set to
-        match prefixes, or match prefixes that are more specific than
-        a given prefix.";
-    }
   }
+
 
   grouping aft-global-filter-structural {
     description
--- a/release/models/aft/openconfig-aft-global-filter.yang
+++ b/release/models/aft/openconfig-aft-global-filter.yang
@@ -38,7 +38,10 @@ submodule openconfig-aft-global-filter {
       description
         "Reference to a routing policy that is used to filter the
         IPv4 prefixes that are installed into the AFT. Only prefixes
-        that are accepted by the referenced policy are installed.";
+        that are accepted by the referenced policy are installed.
+        The policy may, for example, use an explicit prefix set to
+        match prefixes, or match prefixes that are more specific than
+        a given prefix.";
     }
 
     leaf ipv6-policy {
@@ -49,7 +52,10 @@ submodule openconfig-aft-global-filter {
       description
         "Reference to a routing policy that is used to filter the
         IPv6 prefixes that are installed into the AFT. Only prefixes
-        that are accepted by the referenced policy are installed.";
+        that are accepted by the referenced policy are installed.
+        The policy may, for example, use an explicit prefix set to
+        match prefixes, or match prefixes that are more specific than
+        a given prefix.";
     }
   }
 
--- a/release/models/aft/openconfig-aft-global-filter.yang
+++ b/release/models/aft/openconfig-aft-global-filter.yang
@@ -6,7 +6,6 @@ module openconfig-aft-global-filter {
   import openconfig-network-instance { prefix "oc-ni"; }
   import openconfig-routing-policy { prefix "oc-rpol"; }
 
-
   organization
     "OpenConfig working group";
 
--- a/release/models/aft/openconfig-aft-global-filter.yang
+++ b/release/models/aft/openconfig-aft-global-filter.yang
@@ -36,9 +36,9 @@ module openconfig-aft-global-filter {
       }
       description
         "Reference to a routing policy that is used to filter the
-         IPv6 prefixes that are streamed from the FIB.  This does
-         not affect which prefixes are installed in the FIB.
-         The policy may, for example, use an explicit prefix set to
+         IPv6 prefixes that are streamed from the FIB. This does
+         not affect which prefixes are installed in the FIB. The
+         policy may, for example, use an explicit prefix set to
          match prefixes, or match prefixes that are more specific than
          a given prefix.";
     }
@@ -50,9 +50,9 @@ module openconfig-aft-global-filter {
       }
       description
         "Reference to a routing policy that is used to filter the
-         IPv4 prefixes that are in the FIB. Only prefixes
-         that are accepted by the referenced policy are streamed.
-         The policy may, for example, use an explicit prefix set to
+         IPv4 prefixes that are streamed from the FIB. This does
+         not affect which prefixes are installed in the FIB. The
+         policy may, for example, use an explicit prefix set to
          match prefixes, or match prefixes that are more specific than
          a given prefix.";
     }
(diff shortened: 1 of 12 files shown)


release/models/aft/openconfig-aft-global-filter.yang
```diff
@@ -0,0 +1,72 @@
+submodule openconfig-aft-global-filter {
```
earies: Is a submodule necessary here?  Reason I ask is more holistic and if this approach should be taken whatsoever going forward.

If a true extension of an anchor point that can also be optional, augments are the better approach imo.  Creates a clear modular divide whereas submodules are often unnecessary complexity just to break apart files but require them as mandatory

Otherwise, may as well contain these additions directly in the model defining the struct attachment point

earies (commented):
Some instance data examples might also assist here.  While a policy reference is meant to act as a filter that needs to evaluate to true, are all combinations of match/actions permitted here or will this focus on a subset in policy modeling (and if policy modeling expands over time)

earies:
Also in case its not obvious - if we think holistically, this could apply to many scenarios across subtrees where filtering is desirable on either policy matches or beyond - has there been consideration to other domains and if so, could this lead down a path of potential common filter block reuse

release/models/aft/openconfig-aft-global-filter.yang
```diff
@@ -0,0 +1,72 @@
...
+      "Add global-filter container to support per-network-instance
+      AFT prefix filtering.";
+    reference "3.3.0";
+  }
+
+  grouping aft-global-filter-config {
+    description
+      "Configuration parameters for the AFT global filter.";
+
+    leaf policy-name {
```
akalluru1: Its not clear whether this policy should be applied to only a specific container under `afts`, i.e., only for IPv4 routes (`ipv4-unicast`) or IPv6 routes (`ipv6-unicast`) or all the containers under `afts`.

Isn't it better to add a separate leaf for each container under `afts` to avoid having a broad filter that needs to cover various containers? My suggestion is to add:
`ipv4-policy-name` & `ipv6-policy-name` as that is the current main use case. when the need arises for other containers more leaves can be added under the `global-filter/config` & `global-filter/state`. This provide proper scope & clarity in my opinion.

release/models/aft/openconfig-aft-global-filter.yang
```diff
@@ -0,0 +1,89 @@
...
+
+    leaf ipv4-policy {
+      type leafref {
+        path "/oc-rpol:routing-policy/oc-rpol:policy-definitions" +
+             "/oc-rpol:policy-definition/oc-rpol:config/oc-rpol:name";
+      }
+      description
+        "Reference to a routing policy that is used to filter the
+        IPv4 prefixes that are installed into the AFT. Only prefixes
+        that are accepted by the referenced policy are installed.
```
akalluru1: is the word `installed` correct? Shouldn't it say `streamed` because this filter will affect streaming only, not the routes installed in hardware?

release/models/aft/openconfig-aft-global-filter.yang
```diff
@@ -0,0 +1,89 @@
...
+
+    leaf ipv6-policy {
+      type leafref {
+        path "/oc-rpol:routing-policy/oc-rpol:policy-definitions" +
+             "/oc-rpol:policy-definition/oc-rpol:config/oc-rpol:name";
+      }
+      description
+        "Reference to a routing policy that is used to filter the
+        IPv6 prefixes that are installed into the AFT. Only prefixes
+        that are accepted by the referenced policy are installed.
```
akalluru1: same comment as the one above.

release/models/aft/openconfig-aft-global-filter.yang
```diff
@@ -0,0 +1,89 @@
...
+  grouping aft-global-filter-structural {
+    description
+      "Structural grouping defining the schema for the global filter
+      applied to the abstract forwarding tables.";
+
+    container global-filter {
+      description
+        "Configuration and state for a global filter applied to the
+        AFT. The filter is used to restrict which prefixes are
+        installed into the AFT for the network instance.";
```
akalluru1: same comment as the one above

release/models/aft/openconfig-aft-global-filter.yang
ElodinLaarz: Yeah, makes sense. Should have made it an augment to start with. Updated.

release/models/aft/openconfig-aft-global-filter.yang
ElodinLaarz: Moved to be per AFI.

release/models/aft/openconfig-aft-global-filter.yang
ElodinLaarz: Updated.

release/models/aft/openconfig-aft-global-filter.yang
ElodinLaarz: Updated.

release/models/aft/openconfig-aft-global-filter.yang
ElodinLaarz: Updated.

ElodinLaarz:
> Some instance data examples might also assist here. While a policy reference is meant to act as a filter that needs to evaluate to true, are all combinations of match/actions permitted here or will this focus on a subset in policy modeling (and if policy modeling expands over time)

Added some info the description, and more details are in the [test plan](https://github.com/openconfig/featureprofiles/pull/5217).

ElodinLaarz:
> Also in case its not obvious - if we think holistically, this could apply to many scenarios across subtrees where filtering is desirable on either policy matches or beyond - has there been consideration to other domains and if so, could this lead down a path of potential common filter block reuse

In this case, we did do some internal discussion on whether or not it made sense to do the filtering at subscribe time, but in general it is a very hard problem-- in this case, we want to do the filtering at config time to give the box the ability to do some optimizations in matching against the policy-- otherwise, you need some sort of internal mechanism where the gnmi agents will need to have nontrivial interaction with other daemons on the device to learn how to do filtering. -- Also subscription time filtering would be less performant.

So, this change is purposely scoped to be very "narrow."

dplore (approved):
Reviewed in March 17, 2026 OC Operators meeting without objection.  Setting to last call for March 31, 2026

release/models/aft/openconfig-aft-global-filter.yang
```diff
@@ -0,0 +1,97 @@
...
+  grouping aft-global-filter-structural {
+    description
+      "Structural grouping defining the schema for the global filter
+       applied to the abstract forwarding tables.";
+
+    container global-filter {
+      description
+        "Configuration and state for a global filter applied to the
+         AFT. The filter is used to restrict which prefixes are
+         installed into the AFT for the network instance.";
```
akalluru1: `installed` keyword may give a wrong impression that it is controlling which prefixes are installed in the HW so  please consider something like `streamed` or anything you think that explains the intent better.

Also, Its better to explicitly call out the fact that all the gNMI clients subscribing to IPv4/IPv6 AFT will only get the data corresponding to the filtered/allowed prefixes.

release/models/aft/openconfig-aft-global-filter.yang
```diff
@@ -0,0 +1,103 @@
...
+        path "/oc-rpol:routing-policy/oc-rpol:policy-definitions" +
+             "/oc-rpol:policy-definition/oc-rpol:config/oc-rpol:name";
+      }
+      description
+        "Reference to a routing policy that is used to filter the
+         IPv6 prefixes that are present in the FIB. Only prefixes
+         that are accepted by the referenced policy are streamed.
+         The policy may, for example, use an explicit prefix set to
+         match prefixes, or match prefixes that are more specific than
+         a given prefix.";
```
dplore, line 43:
```suggestion
        "Reference to a routing policy that is used to filter the
         IPv6 prefixes that are streamed from the FIB.  This does
         not affect which prefixes are installed in the FIB.
         The policy may, for example, use an explicit prefix set to
         match prefixes, or match prefixes that are more specific than
         a given prefix.";
```

release/models/aft/openconfig-aft-global-filter.yang
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    leaf ipv4-policy {
+      type leafref {
+        path "/oc-rpol:routing-policy/oc-rpol:policy-definitions" +
+             "/oc-rpol:policy-definition/oc-rpol:config/oc-rpol:name";
+      }
+      description
+        "Reference to a routing policy that is used to filter the
+         IPv4 prefixes that are in the FIB. Only prefixes
```
dplore: Same comment as above to clarify this only affects the streaming of prefixes.

dplore (changes requested):

dplore (approved):

PR #1454, opened 2026-03-23
Add rtt to openconfig-wifi-mac.yang

### LINKED ISSUE #1450: Add enable-rtt leaf to WLAN SSID Configuration
This issue proposes an enhancement to the OpenConfig WLAN model by adding a new configuration leaf, enable-rtt, within the SSID configuration. This leaf will allow network operators to enable or disable IEEE 802.11mc Fine Timing Measurement (FTM), commonly known as Round Trip Time (RTT), at the SSID level.

Proposed OpenConfig Path:
The new leaf should be added under the existing /access-points/access-point/ssids/ssid/config/ container:

/access-points/access-point[hostname=hostname]/ssids/ssid[name=SSID]/config/enable-rtt

The ability to enable or disable RTT/FTM per SSID for Wi-Fi networks supporting location-based services. By enabling FTM, client devices can more accurately determine their distance to the Access Point, improving the precision of indoor positioning systems and location-aware applications.

istout-debug:
**Scope Update:**
- Initially, this PR introduced 802.11mc Fine Timing Measurement (RTT) as a standalone configuration leaf. 
- To better align with OpenConfig style guidelines and to future-proof the schema for additional RTT-related attributes (e.g., burst duration, timers, or thresholds), the model has been refactored. 
- RTT configuration and state are now encapsulated in a dedicated `rtt` container under the `ssid` list, featuring explicit `config` and `state` subcontainers. 
- The `openconfig-wifi-mac.yang` version has been bumped to `1.5.0` to reflect these additions.

### Related Issue
Closes #1450

### Platform Implementations
- Implementation A: Arista currently supports this via the CLI and through an OpenConfig vendor augment. The goal is to integrate this into the OpenConfig model for long-term support.
- Implementation B: Aruba currently supports this via the CLI. The goal is to integrate this into the OpenConfig model for long-term support.
### New Tree View

```diff
module: openconfig-wifi-mac
  +--rw ssids
     +--rw ssid* [name]
        +--rw name                 -> ../config/name
        +--rw config
        |  +--rw name?                      string
        |  +--rw enabled?                   boolean
        |  ...
        +--ro state
        |  ...
        +--ro bssids
        |  ...
        +--rw wmm
        |  ...
+        +--rw rtt
+           +--rw config
+           |  +--rw enabled   boolean
+           +--ro state
+              +--ro enabled   boolean

### DIFF
--- a/release/models/wifi/openconfig-wifi-mac.yang
+++ b/release/models/wifi/openconfig-wifi-mac.yang
@@ -26,7 +26,7 @@ module openconfig-wifi-mac {
   description
     "Model for managing MAC layer configuration of Radio interfaces.";
 
-  oc-ext:openconfig-version "1.4.3";
+  oc-ext:openconfig-version "1.5.0";
 
   revision "2026-03-23" {
     description "Add RTT-enable to SSID config.";
--- a/release/models/wifi/openconfig-wifi-mac.yang
+++ b/release/models/wifi/openconfig-wifi-mac.yang
@@ -612,7 +612,7 @@ module openconfig-wifi-mac {
     description
       "802.11mc RTT (Round Trip Time) configuration, per SSID.";
 
-    leaf rtt {
+    leaf enabled {
       type boolean;
       description
         "Configures whether IEEE 802.11mc Fine Timing Measurement (FTM),
--- a/release/models/wifi/openconfig-wifi-mac.yang
+++ b/release/models/wifi/openconfig-wifi-mac.yang
@@ -144,15 +144,15 @@ module openconfig-wifi-mac {
         "Whether this SSID IE is hidden within Beacons.";
     }
     leaf enable-rtt {
-          type boolean;
-          default false;
-          description
-            "Configures whether IEEE 802.11mc Fine Timing Measurement (FTM),
-             also known as Round Trip Time (RTT), is enabled for clients
-             associated with this SSID. When set to 'true', the Access Point
-             will respond to FTM requests from clients, allowing them to
-             perform accurate distance measurements. When 'false', FTM
-             responses are disabled for this SSID.";
+      type boolean;
+      default "false";
+      description
+        "Configures whether IEEE 802.11mc Fine Timing Measurement (FTM), 
+        also known as Round Trip Time (RTT), is enabled for clients 
+        associated with this SSID. When set to 'true', the Access Point 
+        will respond to FTM requests from clients, allowing them to 
+        perform accurate distance measurements. When 'false', FTM 
+        responses are disabled for this SSID.";
     }
 
     leaf default-vlan {
--- a/release/models/wifi/openconfig-wifi-mac.yang
+++ b/release/models/wifi/openconfig-wifi-mac.yang
@@ -32,7 +32,7 @@ module openconfig-wifi-mac {
     description "Add RTT-enable to SSID config.";
     reference "1.4.3";
   }
-  
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: release/models/wifi/openconfig-wifi-mac.yang [hunk 1]
```diff
@@ -26,7 +26,12 @@ module openconfig-wifi-mac {
   description
     "Model for managing MAC layer configuration of Radio interfaces.";
 
-  oc-ext:openconfig-version "1.4.2";
+  oc-ext:openconfig-version "1.4.3";
```
dplore:
```suggestion
  oc-ext:openconfig-version "1.5.0";
```

release/models/wifi/openconfig-wifi-mac.yang
```diff
@@ -26,7 +26,12 @@ module openconfig-wifi-mac {
...
+
+  revision "2026-03-23" {
+    description "Add RTT-enable to SSID config.";
+    reference "1.4.3";
```
dplore:
```suggestion
    reference "1.5.0";
```

jsnyder81:
@istout-debug I notice we don't align with other leaves in SSID.
other features, say broadcast-filter, multicast-filter do not use the word enabled as part of the leaf.

Also, if the plan is to add other features or configuration knobs, should we break this out into it's own container?  like say dot11r?

istout-debug:
> @istout-debug I notice we don't align with other leaves in SSID. other features, say broadcast-filter, multicast-filter do not use the word enabled as part of the leaf.
> 
> Also, if the plan is to add other features or configuration knobs, should we break this out into it's own container? like say dot11r?

I will update the leaf name to align with other features

I agree that a separate container makes sense for related features. My initial thought was to start with a simple boolean under the SSID, as this is an SSID-level configuration. The plan was to break out more advanced configurations into a dedicated container if vendors provide support for them. However, I'm open to creating a new container and placing the boolean within it. Would this structure it better for future additions?

istout-debug:
I have made the changes to this PR to reflect what was talked about, if I could please have this reviewed to ensure that this is correct and the workflows dont fail.

jsnyder81 (approved):
I like this approach better given we have ambitions for more configuration knobs.

ElodinLaarz:
Reviewed in OC Operators Meeting on March 31, 2026:

Seems to be consistent with other containers configuration e.g. `/ssids/ssid/dot11r/config/dot11r` etc.

If we plan to add more configuration leaves into the `rtt` container, seems to match precedent.

We can put this into last-call for 2 weeks from now (April 14th). If anyone has any comments we can address them then.

zpereyo (approved):
Reviewed and approved

release/models/wifi/openconfig-wifi-mac.yang
```diff
@@ -603,6 +608,22 @@ module openconfig-wifi-mac {
     }
   }
 
+  grouping rtt-config {
+    description
+      "802.11mc RTT (Round Trip Time) configuration, per SSID.";
+
+    leaf rtt {
```
dplore: THere is a great deal of style precedent to use <entity>/config/enabled to turn on or off a feature.  I suggest we use that here too:

```suggestion
    leaf enabled {
```

dplore (changes requested):

istout-debug (commented):
fixed

dplore (approved):

PR #1465, opened 2026-04-01
Available Memory Leaf

mrevang:
### Change Scope

* This change adds a new leaf for "available" memory: `/system/memory/state/available`

### Platform Implementations

 * https://man7.org/linux/man-pages/man1/free.1.html
 * https://learn.microsoft.com/en-us/powershell/scripting/learn/ps101/07-working-with-wmi?view=powershell-7.6

### Tree View

```
  +--rw system
     +--rw memory
     |  +--rw config
     |  +--ro state
     |     +--ro physical?    uint64
     |     +--ro reserved?    uint64
     |     +--ro used?        uint64
     |     +--ro free?        uint64
+    |     +--ro available?   uint64
```

### DIFF
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -1170,8 +1170,9 @@ module openconfig-system {
       type uint64;
       units bytes;
       description
-        "Memory that is available for allocation. This is largely the sum of
-        free memory + cache & buffer usage";
+        "Available memory includes free memory plus cached/buffered memory that
+        can be reclaimed instantly by the operating system for new
+        applications";
     }
   }
 
--- a/release/models/system/openconfig-system.yang
+++ b/release/models/system/openconfig-system.yang
@@ -1170,9 +1170,10 @@ module openconfig-system {
       type uint64;
       units bytes;
       description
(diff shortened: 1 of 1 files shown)


release/models/system/openconfig-system.yang
```diff
@@ -1159,6 +1165,16 @@ module openconfig-system {
...
     }
+
+    leaf available {
+      type uint64;
+      units bytes;
+      description
+        "The amount of memory that is available for use.
+        This includes memory that is currently unused, as well as memory
+        used for system caches and buffers that can be reclaimed by the
+        operating system.";
```
dplore: Reviewed in OC Operators meeting.  I suggest being even more direct about how this leaf is different from "free".    The linux definition is pretty good and has a specific statement about this.

mrevang: Updated

ElodinLaarz:
Discussed at OC Operators Meeting April 14, 2026:

Seems reasonable. Will move into last-call to see if anyone has an objections to the leaf addition.

ElodinLaarz (approved):

PR #1470, opened 2026-04-07
Extend LLDP model with global TTL, custom TLV, and per-interface enhancements

ndas7:
### Change Scope
*   This PR proposes enhancements to the existing LLDP model with the following changes:
 
    1. Adding a global ttl configuration/state leaf under /lldp.
    2. Extending interface LLDP config and state with port-description and interface-level custom-tlv-names.
    3. Introducing a shared custom-TLV grouping so both config and state reuse the same custom TLV fields.     
    4. Adding a top-level /lldp/custom-tlvs/tlv[name] list for global custom TLV definitions.     
    5. Adding interface state timestamp counters for LLDP packet events.

* This change is backwards compatible as it adds leaves to existing model.
 
### Platform Implementations
*   LLDPCLI, LLDPAD (LLDP daemons for Linux): https://lldpd.github.io/usage.html

```
lldpcli configure lldp custom-tlv oui 00,00,00 subtype 1 oui-info "your data"
```
*   SONiC: https://github.com/cshivashgit/SONiC/blob/lldp-custom-tlv-hld/doc/lldp_custom_tlv/lldp-custom-tlv-hld.md#config-db-schema 
*   Vendors use custom TLVs to support features like Cisco's [LLDP-MED](https://www.cisco.com/c/en/us/td/docs/switches/lan/catalyst9300/software/release/16-6/configuration_guide/int_and_hw/b_166_int_and_hw_9300_cg/b_166_int_and_hw_9300_cg_chapter_011.html) (Media Endpoint Discovery), which is itself a collection of organizationally specific TLVs. In Cisco [NX-OS](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/6-x/system_management/configuration/guide/b_Cisco_Nexus_9000_Series_NX-OS_System_Management_Configuration_Guide/sm_lldp.pdf), you can select which TLVs to include using the lldp tlv-select command. Aruba's [AOS-CX](https://www.youtube.com/watch?v=lPbtpFcTqgE) introduced a specific OUI (83A30) to help identify Aruba devices and automatically apply network profiles.

### Tree View
 
```diff
module: openconfig-lldp
  +--rw lldp
     +--rw config
     |  +--rw enabled?                      boolean
     |  +--rw hello-timer?                  uint64
+    |  +--rw ttl?                          uint16
     |  +--rw suppress-tlv-advertisement*   identityref
     |  +--rw system-name?                  string
     |  +--rw system-description?           string
     |  +--rw chassis-id?                   string
     |  +--rw chassis-id-type?              oc-lldp-types:chassis-id-type
     |  +--rw management-interface?         oc-if:interface-id
     +--ro state
     |  +--ro enabled?                      boolean
     |  +--ro hello-timer?                  uint64
+    |  +--ro ttl?                          uint16
     |  +--ro suppress-tlv-advertisement*   identityref
     |  +--ro system-name?                  string
     |  +--ro system-description?           string
     |  +--ro chassis-id?                   string
     |  +--ro chassis-id-type?              oc-lldp-types:chassis-id-type
     |  +--ro management-interface?         oc-if:interface-id
     |  +--ro counters
     |     +--ro frame-in?           oc-yang:counter64
     |     +--ro frame-out?          oc-yang:counter64
     |     +--ro frame-error-in?     oc-yang:counter64
     |     +--ro frame-discard?      oc-yang:counter64
     |     +--ro tlv-discard?        oc-yang:counter64
     |     +--ro tlv-unknown?        oc-yang:counter64
     |     +--ro last-clear?         oc-yang:date-and-time
     |     +--ro tlv-accepted?       oc-yang:counter64
     |     +--ro entries-aged-out?   oc-yang:counter64
     +--ro mgmt-addresses
     |  +--ro mgmt-address* [address]
     |     +--ro address    -> ../state/address
     |     +--ro state
     |        +--ro address?                    union
     |        +--ro interface-number?           uint32
     |        +--ro interface-number-subtype?   oc-lldp-types:mgmt-interface-number-subtype
     +--rw interfaces
     |  +--rw interface* [name]
     |     +--rw name         -> ../config/name
     |     +--rw config
     |     |  +--rw name?               oc-if:base-interface-ref
     |     |  +--rw enabled?            boolean
+    |     |  +--rw port-description?   string
+    |     |  +--rw custom-tlv-names*   -> ../../../../custom-tlvs/tlv/name
     |     +--ro state
     |     |  +--ro name?                        oc-if:base-interface-ref
     |     |  +--ro enabled?                     boolean
+    |     |  +--ro port-description?            string
+    |     |  +--ro custom-tlv-names*            -> ../../../../custom-tlvs/tlv/name
     |     |  +--ro counters
     |     |  |  +--ro frame-in?                    oc-yang:counter64
     |     |  |  +--ro frame-out?                   oc-yang:counter64
     |     |  |  +--ro frame-error-in?              oc-yang:counter64
     |     |  |  +--ro frame-discard?               oc-yang:counter64
     |     |  |  +--ro tlv-discard?                 oc-yang:counter64
     |     |  |  +--ro tlv-unknown?                 oc-yang:counter64
     |     |  |  +--ro last-clear?                  oc-yang:date-and-time
     |     |  |  +--ro frame-error-out?             oc-yang:counter64
+    |     |  +--ro last-packet-transmitted?     oc-types:timeticks64
+    |     |  +--ro last-good-packet-received?   oc-types:timeticks64
     |     +--ro neighbors
     |        +--ro neighbor* [id]
     |           +--ro id                -> ../state/id
     |           +--ro config
     |           +--ro state
     |           |  +--ro system-name?               string
     |           |  +--ro system-description?        string
     |           |  +--ro chassis-id?                string
     |           |  +--ro chassis-id-type?           oc-lldp-types:chassis-id-type
     |           |  +--ro management-interface?      oc-if:interface-id
     |           |  +--ro id?                        string
     |           |  +--ro age?                       uint64
     |           |  +--ro last-update?               int64
     |           |  +--ro ttl?                       uint16
     |           |  +--ro port-id?                   string
     |           |  +--ro port-id-type?              oc-lldp-types:port-id-type
     |           |  +--ro port-description?          string
     |           |  x--ro management-address?        string
     |           |  x--ro management-address-type?   string
     |           +--ro mgmt-addresses
     |           |  +--ro mgmt-address* [address]
     |           |     +--ro address    -> ../state/address
     |           |     +--ro state
     |           |        +--ro address?                    union
     |           |        +--ro interface-number?           uint32
     |           |        +--ro interface-number-subtype?   oc-lldp-types:mgmt-interface-number-subtype
     |           +--ro custom-tlvs
     |           |  +--ro tlv* [type oui oui-subtype]
     |           |     +--ro type           -> ../state/type
     |           |     +--ro oui            -> ../state/oui
     |           |     +--ro oui-subtype    -> ../state/oui-subtype
     |           |     +--ro config
     |           |     +--ro state
+    |           |        +--ro name?          string
     |           |        +--ro type?          int32
     |           |        +--ro oui?           string
     |           |        +--ro oui-subtype?   string
     |           |        +--ro value?         binary
     |           +--ro capabilities
     |              +--ro capability* [name]
     |                 +--ro name      -> ../state/name
     |                 +--ro config
     |                 +--ro state
     |                    +--ro name?      identityref
     |                    +--ro enabled?   boolean
+    +--rw custom-tlvs
+       +--rw tlv* [name]
+          +--rw name      -> ../config/name
+          +--rw config
+          |  +--rw name?          string
+          |  +--rw oui?           string
+          |  +--rw oui-subtype?   string
+          |  +--rw type?          int32
+          |  +--rw value?         binary
+          +--ro state
+             +--ro name?          string
+             +--ro oui?           string
+             +--ro oui-subtype?   string
+             +--ro type?          int32
+             +--ro value?         binary
```

### DIFF
--- a/release/models/lldp/openconfig-lldp.yang
+++ b/release/models/lldp/openconfig-lldp.yang
@@ -12,6 +12,7 @@ module openconfig-lldp {
   import openconfig-yang-types { prefix oc-yang; }
   import openconfig-inet-types { prefix oc-inet; }
   import openconfig-extensions { prefix oc-ext; }
+  import openconfig-types { prefix oc-types; }
 
 
   // meta
--- a/release/models/lldp/openconfig-lldp.yang
+++ b/release/models/lldp/openconfig-lldp.yang
@@ -25,7 +25,13 @@ module openconfig-lldp {
     "This module defines configuration and operational state data
     for the LLDP protocol.";
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "1.0.1";
+
+  revision "2026-04-06" {
+    description
+      "Add global TTL configuration/state leaf, interface port-description and custom-tlv-names leaves, custom TLV name leaf, global custom-tlvs list, and add interface counter timestamps for LLDP.";
+    reference "1.0.1";
+  }
 
   revision "2026-01-14" {
     description
@@ -143,6 +149,24 @@ module openconfig-lldp {
         description
           "The number of frame transmit errors on the
           interface.";
+      }
+
+      leaf last-packet-transmitted {
+        type oc-yang:date-and-time;
+        description
+          "Timestamp of the last LLDP packet transmitted on this interface.";
+      }
+
+      leaf last-good-packet-received {
+        type oc-yang:date-and-time;
+        description
+          "Timestamp of the last valid LLDP packet received on this interface.";
+      }
+
+      leaf loopback-packets-received {
+        type oc-yang:date-and-time;
+        description
+          "Timestamp of the last loopback LLDP packet received on this interface.";
       }
   }
 
@@ -441,11 +465,19 @@ module openconfig-lldp {
   grouping lldp-custom-tlv-config {
     description
       "Configuration data for custom LLDP TLVs";
+
+    uses lldp-custom-tlv-common;
   }
 
-  grouping lldp-custom-tlv-state {
+  grouping lldp-custom-tlv-common {
     description
-      "Operational state data for custom LLDP TLVs";
+      "Common data for custom LLDP TLVs";
+
+    leaf name {
+      type string;
+      description
+        "Name of the custom TLV.";
+    }
 
     leaf type {
       type int32;
@@ -479,6 +511,13 @@ module openconfig-lldp {
         "A variable-length octet-string containing the
         instance-specific information for this TLV.";
     }
+  }
+
+  grouping lldp-custom-tlv-state {
+    description
+      "Operational state data for custom LLDP TLVs";
+
+    uses lldp-custom-tlv-common;
   }
 
   grouping lldp-custom-tlv-top {
@@ -522,9 +561,7 @@ module openconfig-lldp {
 
         container config {
           description
-            "Configuration data ";
-
-          uses lldp-custom-tlv-config;
+            "Configuration data for custom LLDP TLVs";
         }
 
         container state {
@@ -534,7 +571,6 @@ module openconfig-lldp {
           description
             "Operational state data ";
 
-          uses lldp-custom-tlv-config;
           uses lldp-custom-tlv-state;
         }
       }
@@ -603,6 +639,23 @@ module openconfig-lldp {
       default "true";
       description
         "Enable or disable the LLDP protocol on the interface.";
+    }
+
+    leaf port-description {
+      type string;
+      description
+        "The binary string containing the actual port identifier for
+        the port which this LLDP PDU was transmitted. The source and
+        format of this field is defined by PtopoPortId from
+        RFC2922.";
+    }
+
+    leaf-list custom-tlv-names {
+      type leafref {
+        path "../../../../custom-tlvs/tlv/name";
+      }
+      description
+        "References to global custom TLV names associated with this interface.";
     }
   }
 
@@ -681,6 +734,13 @@ module openconfig-lldp {
         "System level hello timer for the LLDP protocol.";
     }
 
+    leaf ttl {
+      type uint64;
+      units "seconds";
+      description
+        "Time-to-live value for LLDP PDUs transmitted by this system.";
+    }
+
     leaf-list suppress-tlv-advertisement {
       type identityref {
         base oc-lldp-types:LLDP_TLV;
@@ -703,6 +763,47 @@ module openconfig-lldp {
         "Global LLDP counters";
 
       uses lldp-global-counters;
+    }
+  }
+
+  grouping lldp-global-custom-tlv-top {
+    description
+      "Top-level grouping for global custom LLDP TLVs";
+
+    container custom-tlvs {
+      description
+        "Enclosing container for list of global custom TLVs";
+
+      list tlv {
+        key "name";
+        description
+          "List of global custom LLDP TLVs";
+
+        leaf name {
+          type leafref {
+            path "../config/name";
+          }
+          description
+            "Reference to the list key";
+        }
+
+        container config {
+          description
+            "Configuration data for global custom TLVs";
+
+          uses lldp-custom-tlv-common;
+        }
+
+        container state {
+
+          config false;
+
+          description
+            "Operational state data for global custom TLVs";
+
+          uses lldp-custom-tlv-common;
+        }
+      }
     }
   }
 
@@ -736,6 +837,7 @@ module openconfig-lldp {
 
       uses lldp-management-addresses;
       uses lldp-interface-top;
+      uses lldp-global-custom-tlv-top;
     }
   }
 
--- a/release/models/lldp/openconfig-lldp.yang
+++ b/release/models/lldp/openconfig-lldp.yang
@@ -722,7 +722,7 @@ module openconfig-lldp {
 
         leaf name {
           type leafref {
-            path "../../../../../custom-tlvs/tlv/config/name";
+            path "../../../../../custom-tlvs/tlv/name";
           }
           description
             "Reference to the globally-configured custom TLV name. The override
--- a/release/models/lldp/openconfig-lldp.yang
+++ b/release/models/lldp/openconfig-lldp.yang
@@ -658,10 +658,10 @@ module openconfig-lldp {
         advertised port description, overriding any system-generated
         strings or the generic interface description found in
         /interfaces/interface/config/description.
-        
+
         If this leaf is not populated, the implementation
         should default to advertising the value of
-        /interfaces/interface/state/description.
+        /interfaces/interface/state/description.";
       reference
         "Sec 8.5.4 of IEEE Std 802.1AB-2016";
     }
@@ -767,7 +767,7 @@ module openconfig-lldp {
         Note: The TTL value is independent from the hello-timer interval.
         The hello-timer controls how frequently LLDP PDUs are transmitted,
         while the TTL value indicates the validity duration of information
-        contained in those PDUs at the receiving system.
+        contained in those PDUs at the receiving system.";
       reference
         "Sec 8.5.3 of IEEE Std 802.1AB-2016";
     }
--- a/release/models/lldp/openconfig-lldp.yang
+++ b/release/models/lldp/openconfig-lldp.yang
@@ -479,22 +479,6 @@ module openconfig-lldp {
     description
       "Common data for custom LLDP TLVs";
 
-    leaf name {
-      type string;
-      description
-        "Administrative name of the custom TLV for human identification
-        and reference purposes, allowing operators to define a template
-        once at the global level and then reference it by name elsewhere
-        rather than by their numeric type, OUI, and subtype values.";
-    }
-
-    leaf type {
-      type int32;
-      description
-        "The integer value identifying the type of information
-        contained in the value field.";
-    }
-
     leaf oui {
       type string;
       description
@@ -525,6 +509,13 @@ module openconfig-lldp {
   grouping lldp-custom-tlv-state {
     description
       "Operational state data for custom LLDP TLVs";
+
+    leaf type {
+      type int32;
+      description
+        "The integer value identifying the type of information
+        contained in the value field.";
+    }
 
     uses lldp-custom-tlv-common;
   }
@@ -694,6 +685,71 @@ module openconfig-lldp {
     uses lldp-interface-packet-timestamps;
   }
 
+  grouping lldp-interface-custom-tlv-override-config {
+    description
+      "Configuration data for per-interface custom TLV value overrides";
+
+    leaf value {
+      type binary;
+      description
+        "Override value for the custom TLV on this interface. When present,
+        this value overrides the global custom TLV value for transmission on
+        this interface. The override applies only if the TLV is present in
+        the interface's custom-tlv-names list.
+
+        A variable-length octet-string containing the instance-specific
+        information for this custom TLV override.";
+    }
+  }
+
+  grouping lldp-interface-custom-tlv-override-top {
+    description
+      "Top-level grouping for per-interface custom TLV value overrides";
+
+    container custom-tlv-value-overrides {
+      description
+        "Enclosing container for list of per-interface custom TLV value
+        overrides. Overrides only apply to TLVs that are (1) defined in the
+        global /lldp/custom-tlvs list and (2) referenced in this interface's
+        custom-tlv-names list.";
+
+      list tlv {
+        key "name";
+        description
+          "List of custom TLV value overrides for this interface. Each entry
+          corresponds to a globally-configured custom TLV and allows
+          per-interface customization of the transmitted value.";
+
+        leaf name {
+          type leafref {
+            path "../../../../../custom-tlvs/tlv/config/name";
+          }
+          description
+            "Reference to the globally-configured custom TLV name. The override
+            only applies if this name is also present in the interface's
+            custom-tlv-names list.";
+        }
+
+        container config {
+          description
+            "Configuration data for per-interface custom TLV value override";
+
+          uses lldp-interface-custom-tlv-override-config;
+        }
+
+        container state {
+
+          config false;
+
+          description
+            "Operational state data for per-interface custom TLV value override";
+
+          uses lldp-interface-custom-tlv-override-config;
+        }
+      }
+    }
+  }
+
   grouping lldp-interface-top {
     description
       "Top-level grouping ";
@@ -733,6 +789,7 @@ module openconfig-lldp {
           uses lldp-interface-state;
         }
 
+        uses lldp-interface-custom-tlv-override-top;
         uses lldp-neighbor-top;
       }
     }
@@ -804,7 +861,10 @@ module openconfig-lldp {
 
     container custom-tlvs {
       description
-        "Enclosing container for list of global custom TLVs";
+        "Enclosing container for list of global custom TLVs. 
+        Implementation Note: Custom TLV OUI/subtype combinations 
+        MUST NOT collide with standard LLDP TLVs. Configuration 
+        of conflicting OUI/subtype values is not permitted.";
 
       list tlv {
         key "name";
@@ -823,6 +883,15 @@ module openconfig-lldp {
           description
             "Configuration data for global custom TLVs";
 
+          leaf name {
+            type string;
+            description
+              "Administrative name of the custom TLV for human identification
+              and reference purposes, allowing operators to define a template
+              once at the global level and then reference it by name elsewhere
+              rather than by their numeric type, OUI, and subtype values.";
+          }
+
           uses lldp-custom-tlv-common;
         }
 
@@ -832,6 +901,15 @@ module openconfig-lldp {
 
           description
             "Operational state data for global custom TLVs";
+
+          leaf name {
+            type string;
+            description
+              "Administrative name of the custom TLV for human identification
+              and reference purposes, allowing operators to define a template
+              once at the global level and then reference it by name elsewhere
+              rather than by their numeric type, OUI, and subtype values.";
+          }
 
           uses lldp-custom-tlv-common;
         }
--- a/release/models/lldp/openconfig-lldp.yang
+++ b/release/models/lldp/openconfig-lldp.yang
@@ -689,6 +689,16 @@ module openconfig-lldp {
     description
       "Configuration data for per-interface custom TLV value overrides";
 
+    leaf name {
+      type leafref {
+        path "/lldp/custom-tlvs/tlv/config/name";
+      }
+      description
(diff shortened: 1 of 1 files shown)


release/models/lldp/openconfig-lldp.yang
```diff
@@ -25,7 +25,13 @@ module openconfig-lldp {
     "This module defines configuration and operational state data
     for the LLDP protocol.";
 
-  oc-ext:openconfig-version "1.0.0";
+  oc-ext:openconfig-version "1.0.1";
```
earies:
```suggestion
  oc-ext:openconfig-version "1.1.0";
```

release/models/lldp/openconfig-lldp.yang
```diff
@@ -144,6 +150,24 @@ module openconfig-lldp {
           "The number of frame transmit errors on the
           interface.";
       }
+
+      leaf last-packet-transmitted {
+        type oc-yang:date-and-time;
```
earies:
```suggestion
        type oc-types:timeticks64;
```

I know there is already a leaf w/ `oc-yang:date-and-time` here but imo this was incorrect per common and more programmatic friendly representation of time since unix epoch

release/models/lldp/openconfig-lldp.yang
```diff
@@ -144,6 +150,24 @@ module openconfig-lldp {
...
+        description
+          "Timestamp of the last LLDP packet transmitted on this interface.";
+      }
+
+      leaf last-good-packet-received {
+        type oc-yang:date-and-time;
```
earies: Same as above although I would question more non-counter leafs under this `counters` container.  Seems appropriate to move these up to a parent level

release/models/lldp/openconfig-lldp.yang
```diff
@@ -604,6 +640,22 @@ module openconfig-lldp {
       description
         "Enable or disable the LLDP protocol on the interface.";
     }
+
+    leaf port-description {
```
earies, line 647: I think some clarification would need to surround this leaf as most implementations would derive this TLV from other areas (e.g. interface description) by default and your intention here appears to be a TLV specific override

release/models/lldp/openconfig-lldp.yang
```diff
@@ -681,6 +733,13 @@ module openconfig-lldp {
         "System level hello timer for the LLDP protocol.";
     }
 
+    leaf ttl {
```
earies, line 754: How will this field conflict or relate to `hello-timer` for where in implementations this is derived/calculated?

release/models/lldp/openconfig-lldp.yang
```diff
@@ -441,11 +465,19 @@ module openconfig-lldp {
...
+      "Common data for custom LLDP TLVs";
+
+    leaf name {
+      type string;
+      description
+        "Name of the custom TLV.";
+    }
 
     leaf type {
       type int32;
```
earies, line 486: I know you are just reusing what was previously defined but this is an incorrect type for what is a 7-bit unsigned field per spec definition.

Suggest proposing/refactoring to a scoped `uint8` to go in accordance to 8021AB-2016

release/models/lldp/openconfig-lldp.yang
earies, line 493: Would suggest considering typedef w/ constraints for these fields

earies, line 504: Would suggest considering typedef w/ constraints for these fields

earies, line 510: Since you are proposing the ability for r/w in here, this probably should be revisited.  If we maintain `type binary` this complicates the reading/writing for users

release/models/lldp/openconfig-lldp.yang
```diff
@@ -604,6 +640,22 @@ module openconfig-lldp {
...
+        Description TLV (Type 4) for LLDP PDUs transmitted on this
+        interface.";
+    }
+
+    leaf-list custom-tlv-names {
+      type leafref {
+        path "../../../../custom-tlvs/tlv/name";
+      }
+      description
+        "References to global custom TLV names associated with this interface.";
```
dplore: Suggest being very specific here and state something like,  if present in the list, the tlv is advertised and if not, the tlv will not be advertised

release/models/lldp/openconfig-lldp.yang
```diff
@@ -604,6 +640,22 @@ module openconfig-lldp {
...
         "Enable or disable the LLDP protocol on the interface.";
     }
+
+    leaf port-description {
+      type string;
+      description
+        "The description of the port to be advertised in the Port
+        Description TLV (Type 4) for LLDP PDUs transmitted on this
+        interface.";
+    }
```
dplore: I would expect that implementations today would use this leaf as the LLDP port-description.  
[/interfaces/interface/state/description](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-interfaces.html#interfaces-interface-state-description)

Discussing this, the desire is to override the interface description with an LLDP specific string.  Please update the description to indicate  what happens if this leaf is not populated and if it is.  Thanks!

release/models/lldp/openconfig-lldp.yang
```diff
@@ -441,11 +465,19 @@ module openconfig-lldp {
...
-  grouping lldp-custom-tlv-state {
+  grouping lldp-custom-tlv-common {
     description
-      "Operational state data for custom LLDP TLVs";
+      "Common data for custom LLDP TLVs";
+
+    leaf name {
+      type string;
+      description
+        "Name of the custom TLV.";
```
dplore: Please elaborate that this is for ease for human identification of the LLDP type.

release/models/lldp/openconfig-lldp.yang
ndas7: Agreed, I've updated the description - The dedicated port-description leaf acts as a specific override for the LLDP TLV, allowing operators to advertise a standardized identifier to neighbors while maintaining a different, private alias in the standard interface model. It decouples protocol advertisement from local configuration, providing better control that defaults to the interface description only when this specific override is not populated.

release/models/lldp/openconfig-lldp.yang
ndas7: The dedicated port-description leaf acts as a specific override for the LLDP TLV, allowing operators to advertise a standardized identifier to neighbors while maintaining a different, private alias in the standard interface model. It decouples protocol advertisement from local configuration, providing better control that defaults to the interface description only when this specific override is not populated.

release/models/lldp/openconfig-lldp.yang
ndas7: The ttl field will act as an explicit override that decouples the transmitted ttl value from the traditional derived calculation of (hello-timer × hold-multiplier). While implementations typically derive ttl to ensure neighbors don't timeout between updates, this leaf allows operators to force a specific value for non-standard requirements. When populated, the manual ttl takes precedence; when omitted, the system falls back to the dynamic derived value.

release/models/lldp/openconfig-lldp.yang
ndas7: Discussed in the weekly sync before, but since changing an existing type would be a breaking change for downstream systems, this can be proposed in a separate PR.

release/models/lldp/openconfig-lldp.yang
ndas7: Similar as above - since changing an existing type would be a breaking change for downstream systems, this can be proposed in a separate PR.

ndas7: Similar as above - since changing an existing type would be a breaking change for downstream systems, this can be proposed in a separate PR.

ndas7: I agree with this - changing the type to string would be simpler to use. However since changing an existing type would be a breaking change for downstream systems, discussed in the meeting to be proposed in a separate PR.

ElodinLaarz:
Discussed at the OC Operators Meeting April 21, 2026:

The only open comments are about making a breaking addition (to the types, for instance), which seems reasonable to split into a separate PR (if we need to introduce it at all.)

If there are no other comments, we can see about moving this into last call to get a final review from folks. (May 5th, 2026)

release/models/lldp/openconfig-lldp.yang
```diff
@@ -144,6 +151,29 @@ module openconfig-lldp {
           "The number of frame transmit errors on the
           interface.";
       }
+
+      leaf loopback-packets-received {
```
yogeshg-arista: Could you clarify the motivation for this?
LLDP uses the multicast DMAC 01:80:C2:00:00:0E, which per IEEE 802.1D is not forwarded by bridges. A "loopback" would require a physical loopback (interface cabled to itself) or a device reflecting the frame. Even then, the received LLDPDU would simply be stored as a neighbor entry.
What is the operational use case for tracking loopback packets in LLDP specifically?
Is this based on a standard or a common implementation practice?

release/models/lldp/openconfig-lldp.yang
```diff
@@ -441,11 +471,22 @@ module openconfig-lldp {
...
+      "Common data for custom LLDP TLVs";
+
+    leaf name {
+      type string;
+      description
+        "Administrative name of the custom TLV for human identification
+        and reference purposes, allowing operators to define a template
+        once at the global level and then reference it by name elsewhere
+        rather than by their numeric type, OUI, and subtype values.";
+    }
```
yogeshg-arista, line 483: How are implementations expected to populate this field?
A received custom TLV contains only an OUI, subtype, and value. If the OUI/subtype combination matches a TLV that the local device is configured to transmit, the implementation could echo back the configured name. But if the neighbor sends a TLV with an OUI/subtype that the local device has no knowledge of, there is no information in the received LLDPDU from which to derive a human-readable name.
Should this leaf be optional with clear guidance on when it can be empty?

release/models/lldp/openconfig-lldp.yang
```diff
@@ -604,6 +649,35 @@ module openconfig-lldp {
...
+        path "../../../../custom-tlvs/tlv/name";
+      }
+      description
+        "References to global custom TLV names to be advertised on this
+        interface. If a custom TLV name is present in this list, the
+        corresponding custom TLV will be advertised in LLDP PDUs transmitted
+        on this interface. If a custom TLV name is not present in this list,
+        it will not be advertised on this interface.";
+    }
   }
```
yogeshg-arista, line 675: The PR places custom TLV definitions at the global /lldp/custom-tlvs/tlv level, with per-interface name references under /lldp/interfaces/interface/custom-tlvs/tlv/config/name. This gives per-interface presence (which TLVs to send on which interface) but not per-interface values.
In practice, many org-defined TLVs carry per-interface values:
Port VLAN ID — each interface may be in a different VLAN
Link Aggregation — aggregation status and port ID differ per physical member
Power via MDI / PoE — power class, priority, etc are per-port

If the intent is that these custom TLVs are strictly for values that are identical across all interfaces, then is the per-interface custom-tlv-names reference necessary — they can be transmitted globally like Chassis ID. But if per-interface selection is needed (as the custom-tlv-names suggests), then per-interface values will likely be needed too.
Would it make sense to place the custom TLV value configuration at the interface level instead?

release/models/lldp/openconfig-lldp.yang
```diff
@@ -706,6 +798,47 @@ module openconfig-lldp {
...
+
+          description
+            "Operational state data for global custom TLVs";
+
+          uses lldp-custom-tlv-common;
+        }
+      }
+    }
+  }
+
```
yogeshg-arista, line 838: What is the expected behavior when a user configures a custom TLV to be transmitted whose OUI and subtype match a TLV that the implementation already transmits as part of standard protocol behavior?
Examples:
A user configures a custom TLV with OUI 00-80-C2, subtype 1 (Port VLAN ID) — the implementation may already be transmitting this TLV based on the interface's VLAN configuration.
A user configures OUI 00-12-0F, subtype 2 (Power via MDI) — the implementation may already be transmitting PoE information from hardware state.
A user configures OUI 00-12-0F, subtype 3 (Link Aggregation) — the implementation may already advertise LAG membership.
In each case: does the custom TLV replace the standard one? Are both transmitted (resulting in duplicate OUI/subtype in the LLDPDU)? Is it an error?
Should the PR specify conflict resolution behavior when a custom TLV's OUI/subtype matches a standard TLV?

release/models/lldp/openconfig-lldp.yang
ndas7: Good point, I've included an optional separate per-interface value override mechanism at /lldp/interfaces/interface/custom-tlv-value-overrides/tlv[name]/config/value. When this override is present, the interface uses it; otherwise uses it uses the global value. This ensures:
* Global templates remain at /lldp/custom-tlvs/tlv[name]/config/value
* Per-interface presence controlled by custom-tlv-names
* Per-interface value overrides (if needed) go in a separate override structure, not duplicating the template container

release/models/lldp/openconfig-lldp.yang
ndas7: Even though the standard doesn't mandate transmitting or tracking a separate loopback packet counter, IEEE 802.1AB does define loopback detection — if a device receives an LLDPDU on an interface containing its own chassis-id and port-id, it indicates a loop. This is implementation-specific but operationally useful.
Operational use cases:

- Self-identification detection: When a device receives an LLDPDU containing its own chassis-id/port-id on an interface, indicating a physical loopback (e.g., port cabled to itself for testing)
- Debug and diagnostics — Network operators use loopback testing to verify interface health, bidirectional connectivity, or detect misconfiguration
- Monitoring — Tracking loopback occurrences helps identify unintended loops (e.g., accidental interface loops during maintenance)

release/models/lldp/openconfig-lldp.yang
ndas7: Yes, the name leaf should be optional or absent in the neighbor custom TLV. I've refactored the neighbor custom TLVs to use lldp-custom-tlv-state which includes type but not name. Only the global configured TLVs have the name leaf (and I've removed the type from there)

release/models/lldp/openconfig-lldp.yang
ndas7: Good point, I've updated the PR to include a comment on the conflict resolution: If a user configures a custom TLV with OUI/subtype matching a known standard TLV, the implementation should reject the config with a validation error.

release/models/lldp/openconfig-lldp.yang
dplore: @ndas7 if you can provide some references to implementation for `loopback-packets-received` that will help justify its inclusion in your model update.

release/models/lldp/openconfig-lldp.yang
yogeshg-arista: Thanks for refactoring the neighbor custom TLVs. However, I think the Tx (global) and Rx (neighbor) custom TLV models should be symmetrical. The refactoring removed `type` from the global custom TLV config, but the neighbor custom TLV state still has `type` in its list key. This creates an asymmetry: on the Rx side we know the TLV type, but on the Tx side we don't specify it. An implementation might assume type 127 (organizationally-specific) since OUI and OUI-subtype are configured, but this isn't stated anywhere in the model. Adding `type` to the global custom TLV config would make the two sides consistent and remove the implicit assumption.

On the `name` leaf for neighbor custom TLVs — I think it's fine to keep it, as long as the leaf description clearly states when implementations are expected to populate it (e.g., when the received OUI/subtype matches a locally-configured custom TLV definition) and when it can be empty (e.g., when the TLV is unknown to the local device). That guidance would address my original concern about how the field gets populated.

release/models/lldp/openconfig-lldp.yang
yogeshg-arista: Thanks for adding the per-interface value override. However, the model now has three layers for custom TLVs: global templates, per-interface name filtering, and per-interface value overrides. It's worth considering whether this complexity is justified.

Placing custom TLVs directly under `/lldp/interfaces/interface/custom-tlvs` would be simpler, consistent with how most TLV values are inherently per-interface and if someone wants the same value everywhere, they just configure it on every interface. It also eliminates the question of what happens when a per-interface value conflicts with a global value.

release/models/lldp/openconfig-lldp.yang
yogeshg-arista: Could you clarify what "known standard TLV" means here? There are several possible interpretations:

1. Only TLVs defined in IEEE 802.1AB itself (types 0–8)?
2. TLVs defined in commonly referenced standards — Port VLAN ID, Link Aggregation, Serial Number, etc.?
3. All TLVs the vendor's implementation transmits, including vendor-proprietary ones (e.g., Arista's SWAG or ZTP Boot VLAN TLVs)?
4. Only TLVs currently enabled on the switch — meaning a user could configure a custom TLV that matches a vendor-supported TLV as long as that TLV is currently disabled?

The behavior is quite different depending on which scope is intended.

release/models/lldp/openconfig-lldp.yang
ndas7: Thanks for the feedback. I couldn't find existing implementations of loopback-packets-received in other vendor models or open-source projects to reference. This counter is currently used only in our company's specific production use case, and given the lack of broader adoption or standardization around it, I understand the concern about including it in the model.

I'm happy to remove it from this update to keep the model focused on widely-supported features. We can add it back later once we have more evidence of implementation or adoption, or alternatively propose it as an augment module if we want to keep it separate from the core model.

release/models/lldp/openconfig-lldp.yang
ndas7: I agree about keeping the container symmetrical, that was the original intent - I have reverted it back to the common container with name and type leafs with the descriptions updated.

release/models/lldp/openconfig-lldp.yang
ndas7: We discussed this in the OpenConfig operators review today and I've removed the overrides container for simplicity. From the discussion, the global custom TLVs with reference at the per-interface level is preferred for scalability and efficiency since having per-interface custom tlv container is not efficient if there are thousands of interfaces with the same custom tlvs.

release/models/lldp/openconfig-lldp.yang
ndas7: By known standard TLV, this includes TLVs with existing defined implementations such as
1. Only TLVs defined in IEEE 802.1AB itself (types 0–8)
2. TLVs defined in commonly referenced standards — Port VLAN ID, Link Aggregation, Serial Number, etc.

dplore (approved):

PR #1480, opened 2026-04-24
Clarify enabled/admin-status descriptions for coherent transceiver low-power-mode

earies:
* (M) release/models/interfaces/openconfig-interfaces.yang
    - Update `enabled` leaf description to indicate that disabling
      an interface SHOULD place a pluggable coherent transceiver
      (e.g. ZR/ZR+) into low-power-mode, including channelized
      cases where all channels sharing a module must be disabled
    - Mirror clarification on `admin-status` leaf and DOWN enum
    - Scope behavior to coherent optics; gray (direct-detect) and
      non-pluggable interfaces are unaffected
    - Increment version to 3.9.0

### Change Scope

This PR amends descriptions only.  No structural or type changes are made
to the model.

The `enabled` configuration leaf and `admin-status` operational state leaf
(including the `DOWN` enum) are clarified to indicate that, on interfaces
backed by a pluggable coherent transceiver (e.g. ZR/ZR+), administratively
disabling the interface SHOULD additionally place the transceiver into
low-power-mode.  When a coherent module is shared by multiple channelized
interfaces, low-power-mode is only entered once all channels sharing the
module have been disabled.

This behavior is explicitly scoped to coherent optics.  Gray
(direct-detect) optics and non-pluggable interfaces are not in scope, even
where the underlying module would otherwise support a low-power mode.

While these are description-only changes, behavior is being clarified for
a pre-existing leaf and may require implementations to adjust.  The model
is therefore bumped as a minor version `3.8.1` -> `3.9.0`.  If folks feel
otherwise, this could be treated as a patch update - e.g. `3.8.2`

### Platform Implementations

N/A

### Tree View

No tree changes - only description updates

### DIFF
--- a/release/models/interfaces/openconfig-interfaces.yang
+++ b/release/models/interfaces/openconfig-interfaces.yang
@@ -51,7 +51,15 @@ module openconfig-interfaces {
     Section 4.c of the IETF Trust's Legal Provisions Relating
     to IETF Documents (http://trustee.ietf.org/license-info).";
 
-  oc-ext:openconfig-version "3.8.1";
+  oc-ext:openconfig-version "3.9.0";
+
+  revision "2026-04-23" {
+      description
+        "Clarify enabled and admin-status descriptions for
+        coherent transceiver low-power-mode behavior.";
+      reference
+        "3.9.0";
+  }
 
   revision "2026-01-06" {
       description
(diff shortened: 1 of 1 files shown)


ahsaanyousaf (approved):

earies:
@dplore @ElodinLaarz for review/comment

Issue #1405, opened 2025-11-25
Add `hierarchical-fec` bool leaf to Static NHG


romeyod:
A hierarchical forwarding equivalence class (FEC) is a system for organizing and classifying IP packets into groups that can be forwarded identically through a network, typically using a tiered or hierarchical structure to handle different levels of traffic priority or type. This allows for more complex and granular control over network traffic, especially for quality of service (QoS) and differentiated services, by creating multiple layers of FECs, where an outer layer might classify a broad category of traffic and an inner layer can further subdivide it. 

We want to introduce a new boolean leaf to control hierarchical forwarding equivalence class (FEC) behavior on next-hop groups:
  - `/network-instances/network-instance/static/next-hop-groups/next-hop-group/config/hierarchical-fec`
  - `/network-instances/network-instance/static/next-hop-groups/next-hop-group/state/hierarchical-fec`

When `True`: The Next-Hop Group (NHG) preserves its hierarchical structure, explicitly pointing to the resolving NHG for each next-hop.

When `False`: The NHG contents are flattened, or a single random next-hop is selected.

This provides operators with fine-grained control over the hierarchical FEC structure, facilitating advanced Traffic Engineering and QoS strategies.

Reference:
https://www.arista.com/en/support/toi/eos-4-20-1f/13934-hierarchical-forwarding-equivalence-class

dplore:
Hi @romeyod , the link requires a userid and password.  Can you share something that is publicly available?  Then it will be easier to make more comments on this issue.

nupkanoi:
Hi, IMO the proposed paths doesn't align with the next-hop group yang paths-
Proposal-
```
/network-instances/network-instance/static/next-hop-groups/next-hop-group/config/hierarchical-fec
/network-instances/network-instance/static/next-hop-groups/next-hop-group/state/hierarchical-fec
```

I suggest below paths(aligned with existing OC for next-group config)-
```
/network-instances/network-instance/static/next-hops/next-hop/encap-headers/encap-header/config/hierarchical-fec, STRING, BOOLEAN
/network-instances/network-instance/static/next-hops/next-hop/encap-headers/encap-header/state/hierarchical-fec, STRING, BOOLEAN
```

nupkanoi:
Hi, synced with Arista offline and LGTM the below paths- 

Proposal-
```
/network-instances/network-instance/static/next-hop-groups/next-hop-group/config/hierarchical-fec
[...]

PR #1497, opened 2026-05-20
Add `hierarchical-fec` bool leaf to static nexthop-group

romeyod:
**Change Scope**
- Add `hierarchical-fec` bool leaf to Static nexthop-group to control hierarchical forwarding equivalence class (FEC) behavior on next-hop groups:
`/network-instances/network-instance/static/next-hop-groups/next-hop-group/config/hierarchical-fec`
`/network-instances/network-instance/static/next-hop-groups/next-hop-group/state/hierarchical-fec`
When True: The nexthop-group  preserves its hierarchical structure, explicitly pointing to the resolving NHG for each next-hop.
When False: The nexthop-group contents are flattened, or a single random next-hop is selected.

**Context:** 
[Add hierarchical-fec bool leaf to Static NHG](https://github.com/openconfig/public/issues/1405)

**Tree View**
```
  +--rw network-instances
           +--rw static
           |  +--rw next-hop-groups
           |  |     +--rw config
  +        |  |     |  +--rw hierarchical-fec?   boolean
           |  |     +--ro state
  +        |  |     |  +--ro hierarchical-fec?   boolean
           |  |     +--rw next-hops
           |  |           +--rw config
           |  |           +--ro state
```

### DIFF
--- a/release/models/network-instance/openconfig-network-instance-static.yang
+++ b/release/models/network-instance/openconfig-network-instance-static.yang
@@ -22,7 +22,13 @@ module openconfig-network-instance-static {
   description
     "Static configurations associated with a network instance";
 
-  oc-ext:openconfig-version "0.2.1";
+  oc-ext:openconfig-version "0.3.0";
+
+  revision "2026-05-20" {
+    description
+      "Add hierarchical-fec leaf to static next-hop-group.";
+    reference "0.3.0";
+  }
 
   revision "2025-08-05" {
     description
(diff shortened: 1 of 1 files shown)


romeyod:
@nupkanoi for visibility

robshakir:
@dplore @navaneethyv @ElodinLaarz -- can we please make sure that we review this carefully with other vendors since this sounds like it has the risk of being very platform specific.

romeyod:
@rgwilton @earies for other vendor inputs

rgwilton:
We have discussed this internally and we believe that this behaviour should be managed by the system without requiring explicit user configuration.  I.e., this feels too implementation (and perhaps hardware) specific.  E.g., certainly the implication that if you don't set this flag then that would mean that you wouldn't automatically be able to make these hierarchical would be unhelpful.

So, in conclusion we are not supportive of this change.

PR #1499, opened 2026-05-21
openconfig keychain send-and-receive description update.

### LINKED ISSUE #1476: keychain model send-and-receive expected behaviour
Hi,

In the key-chain model we have this leaf send-and-receive with following description
```
When this is set to true (the default value), the specified send lifetime is also used in the receive direction. When set to false, the device should use the specified receive-lifetime for the receive direction (asymmetric mode). If send-and-receive is false, and the device does not support asymmetric configuration, the config should be rejected as unsupported.
```

We have a question regarding the expected behavior in the following use case, assuming the device supports asymmetric send-and-receive lifetimes:

**When send-and-receive is set to true and the user configures a new receive-lifetime that does not match the send-lifetime, should the device:**
         a) Accept the configuration, but continue using the send-lifetime for the receive direction for all the subscribers of this keychain secret until send-and-receive is set to false? Once set to false, the device would then start using the configured receive-lifetime.
         b) Reject the configuration, allowing an asymmetric receive-lifetime only when send-and-receive is set to false?

ishwarbnaik:
**Change Scope**

- The current description of this leaf does not clearly state that the configuration to receive-lifetime should be rejected when send-and-receive is set to "True". This description should be updated to clearly state that when this leaf is set to "True", configuration of a distinct receive-lifetime is not valid and the server should reject it as an invalid configuration. ( https://github.com/openconfig/public/issues/1476 ).

**Related Issues**
Fixes https://github.com/openconfig/public/issues/1476

**Platform Implementations**
NA

Tree View
No changes to tree view

### DIFF
--- a/release/models/keychain/openconfig-keychain.yang
+++ b/release/models/keychain/openconfig-keychain.yang
@@ -120,7 +120,7 @@ module openconfig-keychain {
          as an invalid configuration. When set to false, the device uses the
          configured receive-lifetime for the receive direction (asymmetric mode).
          If send-and-receive is false and the device does not support asymmetric
-         configuration, the server should reject the config as unsupported."
+         configuration, the server should reject the config as unsupported.";
     }
   }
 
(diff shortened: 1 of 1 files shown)


navaneethyv (approved):

navaneethyv:
OC operator review : June 23, approving.

PR #1507, opened 2026-06-15
Update documentation for AFT Limited Prefix Streaming and aft-summarry behavior

### LINKED ISSUE #1503: Update documentation for AFT Limited Prefix Streaming and aft-summaries behavior
AFT global-filter was introduced [openconfig/public#1441](https://github.com/openconfig/public/pull/1441) and the initial testing READMEs were added in [openconfig/featureprofiles#5217](https://github.com/openconfig/featureprofiles/pull/5217).

My understanding is that aft-summaries should reflect the total AFT entries in the HW FIB, not the filtered summary. Specifically, the Route Summary (per protocol route count in the FIB) should remain untouched, meaning it will still contain the counts related to all prefixes.
The YANG paths in question are:
/network-instances/network-instance[name]/afts/aft-summaries/ipv4-unicast/protocols/protocol[origin-protocol]/state/counters/aft-entries
/network-instances/network-instance[name]/afts/aft-summaries/ipv6-unicast/protocols/protocol[origin-protocol]/state/counters/aft-entries

To ensure this expected behavior is clear, we need to document this expectation directly in the public YANG model descriptions.

romeyod:
**Change Scope**

- The current description that aft-summaries should reflect the total AFT entries in the HW FIB, not the filtered summary. Specifically, the Route Summary (per protocol route count in the FIB) should remain untouched, meaning it will still contain the counts related to all prefixes. 

**Related Issues**
Fixes https://github.com/openconfig/public/issues/1503

**Platform Implementations**
NA

Tree View
No changes to tree view

### DIFF
--- a/release/models/aft/openconfig-aft-summary.yang
+++ b/release/models/aft/openconfig-aft-summary.yang
@@ -20,7 +20,15 @@ module openconfig-aft-summary {
     "This module provides summary of aft entry counts per protocol type for each network
     instance.";
 
-  oc-ext:openconfig-version "0.2.1";
+  oc-ext:openconfig-version "0.2.2";
+
+  revision "2026-06-15" {
+    description
+      "Clarify that aft-summaries counters reflect total unfiltered
+       entries in the hardware FIB, independent of any global-filter
+       policy.";
+    reference "0.2.2";
+  }
 
   revision "2026-03-11" {
     description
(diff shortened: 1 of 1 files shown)


navaneethyv (approved):

navaneethyv:
Jul 7th OC operator review: thanks for the PR moving it to last call.

PR #1492, opened 2026-05-12
Adding Ipv6 header encapsulation support to static GUE tunnel

nupkanoi:
## Change Scope
Adding Ipv6 header encapsulation support to static GUE tunnel

Support for UDP-V4 already exist. 

### Tree View

```
          |              +--rw udp-v4
          |              |  +--rw config
          |              |  |  +--rw src-ip?         oc-inet:ipv4-address
          |              |  |  +--rw dst-ip?         oc-inet:ipv4-address
          |              |  |  +--rw dscp?           oc-inet:dscp
          |              |  |  +--rw src-udp-port?   oc-inet:port-number
          |              |  |  +--rw dst-udp-port?   oc-inet:port-number
          |              |  |  +--rw ip-ttl?         uint8
          |              |  +--ro state
          |              |     +--ro src-ip?         oc-inet:ipv4-address
          |              |     +--ro dst-ip?         oc-inet:ipv4-address
          |              |     +--ro dscp?           oc-inet:dscp
          |              |     +--ro src-udp-port?   oc-inet:port-number
          |              |     +--ro dst-udp-port?   oc-inet:port-number
          |              |     +--ro ip-ttl?         uint8
+         |              +--rw udp-v6
+         |              |  +--rw config
+         |              |  |  +--rw src-ip?         oc-inet:ipv6-address
+         |              |  |  +--rw dst-ip?         oc-inet:ipv6-address
+         |              |  |  +--rw dscp?           oc-inet:dscp
+         |              |  |  +--rw src-udp-port?   oc-inet:port-number
+         |              |  |  +--rw dst-udp-port?   oc-inet:port-number
+         |              |  |  +--rw ip-ttl?         uint8
+         |              |  +--ro state
+         |              |     +--ro src-ip?         oc-inet:ipv6-address
+         |              |     +--ro dst-ip?         oc-inet:ipv6-address
+         |              |     +--ro dscp?           oc-inet:dscp
+         |              |     +--ro src-udp-port?   oc-inet:port-number
+         |              |     +--ro dst-udp-port?   oc-inet:port-number
+         |              |     +--ro ip-ttl?         uint8
          |              +--rw gre
          |              |  +--rw config
          |              |  |  +--rw src-ip?   oc-inet:ip-address
          |              |  |  +--rw dst-ip?   oc-inet:ip-address
          |              |  |  +--rw ttl?      uint8
          |              |  +--ro state
          |              |     +--ro src-ip?   oc-inet:ip-address

```

PR #1511, opened 2026-07-07
Update the description of ip-ttl and dscp

tasaleh:
Update the description of the following leaves:

`ip-ttl`: Since this leave models both TTL (IPv4) and Hop Limit (IPv6) values, then, we need to indicate this explicitly in the leaf's description.

`dscp`: Update the description of the behavior when the leaf is not configured. When the leaf is not configured, the value of the inner packet will be copied over as the outer packet's value during encapsulation.

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -843,8 +843,10 @@ submodule openconfig-aft-common {
     leaf dscp {
       type oc-inet:dscp;
       description
-        "DSCP value to use for the UDP header of the encapsulated
-         packet.";
+        "This leaf reflects the configured/default DSCP value that is used
+        in the outer header during packet encapsulation. When this leaf is
+        not set, the DSCP value of the inner packet is copied over as the
+        outer packet's DSCP value during encapsulation.";
     }
 
     leaf src-udp-port {
@@ -875,10 +877,10 @@ submodule openconfig-aft-common {
     leaf ip-ttl {
       type uint8;
       description
-        "This leaf reflects the configured/default IP TTL value that is used
+        "This leaf reflects the configured/default IP TTL / Hop Limit value that is used
          in the outer header during packet encapsulation. When this leaf is
-         not set, the TTL value of the inner packet is copied over as the
-         outer packet's IP TTL value during encapsulation.";
+         not set, the IP TTL / Hop Limit value of the inner packet is copied over as the
+         outer packet's IP TTL / Hop Limit value during encapsulation.";
     }
   }
 


navaneethyv:
OC operator review July 14th.

Thanks for your contribution.

Can you add references to this in the PR description? Do we have an RFC that specifics this or 2 vendor implementations that support this behavior?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -875,10 +877,10 @@ submodule openconfig-aft-common {
     leaf ip-ttl {
       type uint8;
       description
-        "This leaf reflects the configured/default IP TTL value that is used
+        "This leaf reflects the configured/default IP TTL / Hop Limit value that is used
          in the outer header during packet encapsulation. When this leaf is
-         not set, the TTL value of the inner packet is copied over as the
-         outer packet's IP TTL value during encapsulation.";
+         not set, the IP TTL / Hop Limit value of the inner packet is copied over as the
```
navaneethyv, line 882: HI @tasaleh ,

if your intent is towards v6 version of IP ttl , can we add it as a different field so that it is consistent with v6 terminology ?

release/models/aft/openconfig-aft-common.yang
```diff
@@ -843,8 +843,10 @@ submodule openconfig-aft-common {
     leaf dscp {
       type oc-inet:dscp;
       description
-        "DSCP value to use for the UDP header of the encapsulated
-         packet.";
+        "This leaf reflects the configured/default DSCP value that is used
```
navaneethyv, line 846: This makes some assumptions about the behavior of  DSCP value  in this leaf. Can you please add 2 supporting implementations for this ?

navaneethyv:
OC review 28th. 

Added a few comments, + requesting for supporting implementations from 2 vendors.

Can you please sign this to ensure the check passes [ this is a presubmit check]
https://github.com/openconfig/public/pull/1511/checks?check_run_id=87139774474

release/models/aft/openconfig-aft-common.yang
tasaleh: Hello @navaneethyv,

We have raised this point in this [PR 1492](https://github.com/openconfig/public/pull/1492). The response was to keep using `ip-ttl` with V6. Please see the comments from @earies and @ ElodinLaarz
This is the reason why we propose this description change to capture the new usage.

sachendras: I would prefer this to be consistent to the rfc8200 for IPv6. If it is "Hop Limit" for IPv6, I feel the OC should match.

sachendras: However, just saw that https://github.com/openconfig/public/pull/1492 is merged now after multiple discussions with the following reasoning and looks fine to me.


Summary: The deviation is purely a YANG modeling compromise. In pull/1492, the slightly inaccurate naming (ip-ttl instead of hop-limit for IPv6) was accepted to maximize code reuse by importing an existing module's structural grouping. Doing something different for IPv6 will add additional complexities without much benefit

navaneethyv: Thanks for the clarification. nothing the exact comment precedence here.

https://github.com/openconfig/public/pull/1492#issuecomment-4489971914

release/models/aft/openconfig-aft-common.yang
```diff
@@ -843,8 +843,10 @@ submodule openconfig-aft-common {
...
+        in the outer header during packet encapsulation. When this leaf is
+        not set, the DSCP value of the inner packet is copied over as the
```
akalluru1, line 848: This behavior may be platform dependent where new/existing platforms may not be able to support this statement. should there be a mention of `behavior depends on ASIC/platform functionality`?

release/models/aft/openconfig-aft-common.yang
akalluru1, line 882: same comment as above.

akvasu: This is same thing defined in the AFT already, the only correction here is to mention that ttl is also used for the IPv6 hop-limit. Agree that ASIC's might work different, but the config is defining default (i.e. no config) as uniform and a configured value as pipe. ASIC's need to support either of the options and if only pipe mode is supported, then the config has to be specified.

release/models/aft/openconfig-aft-common.yang
akvasu: The AFT's original intent was - "configured dscp MUST be used in udp encap" but as you state, not all asic's would support it. Hence the phrasing is changed to make it optional - config can be used if the asic supports it or it can be left blank to let asic default behavior work (assumed to be uniform mode)

release/models/aft/openconfig-aft-common.yang
akvasu: The original statement made assumptions - like specifying this value means we use it in the encap - not sure if there are supporting implementations exactly this way. Ideally QoS should be used to control this field. The clarification was just to make this field behavior somewhat optional - provide value for pipe mode or leave it unset for asic default

navaneethyv:
Thanks for @akalluru1  from Cisco for the review.

PR #1515, opened 2026-07-14
Deprecate cpu-utilization and add cpu-utilization-pct for multi-core support

charanjith-anet:
## Summary
- Deprecate the existing `cpu-utilization` leaf (`oc-types:percentage`, uint8 range 0-100)
- Add new `cpu-utilization-pct` leaf with `uint32` type to support multi-core systems where per-process CPU utilization can exceed 100%
- Bump module version from 0.4.0 to 0.5.0

## Background
The `cpu-utilization` leaf under `/system/processes/process/state` is backed by `/proc/PID/stat` in Linux-based implementations. On multi-core systems, a process using multiple cores can report CPU utilization exceeding 100% (up to 100% per core). The previous type `oc-types:percentage` restricts values to 0-100, which is insufficient.

A new leaf is added instead of modifying the existing type to preserve backward compatibility — downstream consumers may have strict type handling or run mixed model versions.

## Test plan
- [ ] Verify YANG model compiles with `pyang`
- [ ] Validate backward compatibility with existing consumers of `cpu-utilization`

### DIFF
--- a/release/models/system/openconfig-procmon.yang
+++ b/release/models/system/openconfig-procmon.yang
@@ -25,7 +25,15 @@ module openconfig-procmon {
     "This module provides data definitions for process health
     monitoring of one or more processes running on the system.";
 
-  oc-ext:openconfig-version "0.4.0";
+  oc-ext:openconfig-version "0.5.0";
+
+  revision "2026-07-13" {
+    description
+      "Change cpu-utilization type from oc-types:percentage to
+      uint16 to support multi-core systems where CPU utilization
+      can exceed 100 percent.";
+    reference "0.5.0";
+  }
 
   revision "2019-03-15" {
     description
@@ -153,9 +161,11 @@ module openconfig-procmon {
     }
 
     leaf cpu-utilization {
-      type oc-types:percentage;
+      type uint16;
       description
-        "The percentage of CPU that is being used by the process.";
+        "The percentage of CPU that is being used by the process.
+        On multi-core systems, this value may exceed 100 percent,
+        up to 100 percent per core.";
     }
 
     leaf memory-usage {
--- a/release/models/system/openconfig-procmon.yang
+++ b/release/models/system/openconfig-procmon.yang
@@ -27,10 +27,10 @@ module openconfig-procmon {
 
   oc-ext:openconfig-version "0.5.0";
 
-  revision "2026-07-13" {
+  revision "2026-07-14" {
     description
-      "Change cpu-utilization type from oc-types:percentage to
-      uint16 to support multi-core systems where CPU utilization
+      "Deprecate cpu-utilization and add cpu-utilization-pct as
+      uint32 to support multi-core systems where CPU utilization
       can exceed 100 percent.";
     reference "0.5.0";
   }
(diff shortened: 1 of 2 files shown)


navaneethyv:
OC operator review  Jul 14: 
Changing an existing representation of a data type might break assumptions. 

Can you please look  at having cpu-utilization-uint32 as an alternative ?

earies:
> OC operator review Jul 14: Changing an existing representation of a data type might break assumptions.
> 
> Can you please look at having cpu-utilization-uint32 as an alternative ?

Agreed - widening a type is probably safe in most scenarios but any strict handling or hybrid env running different versions of the modeling complicates matters that we honestly cannot probably fully know, especially as time goes on and adoption increases.

As far as the proposal, I agree this is a gap in the current definition that needs to be fixed.

release/models/system/openconfig-procmon.yang
```diff
@@ -153,9 +161,22 @@ module openconfig-procmon {
...
       type oc-types:percentage;
       description
-        "The percentage of CPU that is being used by the process.";
+        "The percentage of CPU that is being used by the process.
+        This leaf is deprecated because its type constrains values
+        to 0-100, which is insufficient for multi-core systems.
+        Use cpu-utilization-pct instead.";
+    }
+
+    leaf cpu-utilization-pct {
```
navaneethyv, line 173: pct is non descriptive.

navaneethyv: @earies do you have any suggestions?

earies: We are likely to run into more cases of needing to fix a prior definition while wanting to retain compatibility for the old name which leads to likely unnatural naming for the new nodes.

`pct` isn't entirely bad since its an abbreviation that is fairly known and self descriptive and using abbreviations or acronyms is permitted if that is the case generally (or has proper spec definitions, etc..).

Otherwise, off the top of my head a few suggestions for comment (I don't feel strongly about any one of them)

* cpu-utilization-percent (spelling it out)
* cpu-utilization-total (mean to represent the aggregate across cores)
* cpu-utilization-aggregate (similar to above)

charanjith-anet: I didn't want to spell out completely because the units is "percentage" and I ended up choosing abbreviation since there's precedence to it. I considered -total as well but it sounded a bit redundant. I'm fine with -total or -aggregate as well.

navaneethyv: reference : https://github.com/openconfig/public/blob/94f589629e1e88d0a8dbd2b3024f829196801f2e/release/models/bgp/openconfig-bgp-common-multiprotocol.yang#L722

navaneethyv:
OC review : July 28th  moving ahead.

navaneethyv:
OC operator's review August 4th.

Taking it further to merge.

navaneethyv (approved):
Approving.

release/models/system/openconfig-procmon.yang
```diff
@@ -153,9 +161,22 @@ module openconfig-procmon {
...
+      type uint32;
```
earies: One additional thought that I think would be helpful here for various renderings is make this a typedef, take the type, units, description verbatim - call the typedef something like `oc-types:multi-core-percentage` then we get a descriptive type and reuse as this will probably come up in other future uses

navaneethyv:
OC review August 11th.

@charanjith-anet  can you please take a look at the comment and add a typedef  ? WDYT ?

release/models/system/openconfig-procmon.yang
charanjith-anet: Sounds good. I'll incorporate it.

charanjith-anet:
@navaneethyv I incorporated the comments. Is it good to merge now?

PR #1531, opened 2026-08-03
Update openconfig-system-grpc.yang

danielbarney:
## Parameter Mapping

| Device Configuration | OpenConfig YANG Path | Type / Range | Configured Value | Default Value |
| :--- | :--- | :--- | :--- | :--- |
| `tls mutual` | `/system/grpc-servers/grpc-server/config/tls-mutual` | `boolean` | `true` | `false` |
| `keepalive time 5` | `/system/grpc-servers/grpc-server/keepalive/config/time` | `uint32` (range `1..max`, seconds) | `5` | Unset / system default |
| `keepalive min-keepalive-interval 5` | `/system/grpc-servers/grpc-server/keepalive/config/min-keepalive-interval` | `uint32` (range `1..max`, seconds) | `5` | Unset / system default |
| `max-concurrent-streams 128` | `/system/grpc-servers/grpc-server/config/max-concurrent-streams` | `uint32` (range `1..max`) | `128` | Unset / system default |

---

## Configuration Example

```json
{
  "openconfig-system:system": {
    "openconfig-system-grpc:grpc-servers": {
      "grpc-server": [
        {
          "name": "gnxi-mgmt",
          "config": {
            "name": "gnxi-mgmt",
            "port": 9339,
            "transport-security": true,
            "tls-mutual": true,
            "max-concurrent-streams": 128,
          },
          "keepalive": {
            "config": {
              "time": 5,
              "min-keepalive-interval": 5
            }
          }
        }
      ]
    }
  }
}
```

```
 module: openconfig-system-grpc
 
   augment /oc-sys:system:
     +--rw grpc-servers
        +--rw grpc-server* [name]
           +--rw name          -> ../config/name
           +--rw config
           |  +--rw services*                  identityref
           |  +--rw enable?                    boolean
           |  +--rw port?                      oc-inet:port-number
           |  +--rw transport-security?        boolean
+          |  +--rw tls-mutual?                boolean
           |  +--rw certificate-id?            string
           |  +--rw metadata-authentication?   boolean
+          |  +--rw max-concurrent-streams?    uint32
           |  +--rw listen-addresses*          union
           |  +--rw network-instance?          oc-ni:network-instance-ref
           +--ro state
           |  +--ro name?                      string
           |  +--ro services*                  identityref
           |  +--ro enable?                    boolean
           |  +--ro port?                      oc-inet:port-number
           |  +--ro transport-security?        boolean
+          |  +--ro tls-mutual?                boolean
           |  +--ro certificate-id?            string
           |  +--ro metadata-authentication?   boolean
+          |  +--ro max-concurrent-streams?    uint32
           |  +--ro listen-addresses*          union
           |  +--ro network-instance?          oc-ni:network-instance-ref
+          +--rw keepalive
+          |  +--rw config
+          |  |  +--rw time?                   uint32
+          |  |  +--rw min-keepalive-interval? uint32
+          |  |  +--rw timeout?                uint32
+          |  +--ro state
+          |     +--ro time?                   uint32
+          |     +--ro min-keepalive-interval? uint32
+          |     +--ro timeout?                uint32
           +--ro connections
              +--ro connection* [address port]
                 +--ro address     -> ../state/address
                 +--ro port        -> ../state/port
                 +--ro state
                 |  +--ro address?    oc-inet:ip-address
                 |  +--ro port?       oc-inet:port-number
                 |  +--ro counters
                 |     +--ro bytes-sent?        oc-yang:counter64
                 |     +--ro packets-sent?      oc-yang:counter64
                 |     +--ro data-send-error?   oc-yang:counter64
                 +--ro counters
                    +--ro bytes-sent?        oc-yang:counter64
                    +--ro packets-sent?      oc-yang:counter64
                    +--ro data-send-error?   oc-yang:counter64

```

### Nokia Implementation

Nokia implementation relays on TLS profile being configured with `authenticate-client`

```
{
  "system": {
    "tls": {
      "profile": [
        {
          "name": "tls",
          "key": " ... ",
          "certificate": "-----BEGIN CERTIFICATE-----\n ... \n-----END CERTIFICATE-----",
          "authenticate-client": true      <<<<<<<<<<< mTLS enabled
        }
      ]
    },
    "grpc-server": [
      {
        "name": "mgmt",
        "admin-state": "enable",
        "rate-limit": 65000,
        "session-limit": 1024,
        "metadata-authentication": true,
        "yang-models": "openconfig",
        "tls-profile": "tls",
        "network-instance": "mgmtVrf",
        "port": 9339,
        "services": [
          "gnmi",
          "gnoi",
          "gnsi"
        ],
        "gnmi": {
          "commit-save": true
        },
        "unix-socket": {
          "admin-state": "enable"
        }
      }
    ]
  }
}
```

### Cisco Implementation

Cisco implementations relays not just on having the profile saved to harddisk (the configuration of the profile does not shows in the configuration) rather the gRPC server needs to be aligned to the content of the profile and adding mtls dual under each server configuration. 

````
grpc 
server gnxi-mgmt
  vrf mgmtVrf
  dscp cs4
  port 9339
  tls mutual <<<<<<<  + Profile in harddisk
  services GNMI
  services GNOI
  services GNSI
  services ATTESTZ
  services ENROLLZ
  ssl-profile-id tls
  metadata-authentication disable
 !

````

### Arista implementation

```
management security
   ssl profile tls
      certificate eea339da.10.crt key eea339da.10.key
      trust certificate a69360bc.11.crt
      chain certificate 377050fb.9.crt
!
management api gnmi
   transport grpc default
      ssl profile tls
      port 9339
      vrf mgmtVrf
      qos dscp 32
      accounting requests
      aaa config-commands disabled
      authentication username priority x509-spiffe x509-common-name
```

### CODE REVIEW

**Comment by danielbarney** [comment]
What is the suggested naming for enforcing mTLS under the grpc-server

/system/grpc-servers/grpc-server/authenticate-client
/system/grpc-servers/grpc-server/tls-mutual

File: release/models/system/openconfig-system-grpc.yang [hunk 1]
```diff
@@ -167,6 +175,18 @@ module openconfig-system-grpc {
         are not supported, such as lab testing.";
     }
 
+    leaf tls-mutual {
```
macb2625, line 178: i would add one more container "tls" under config so that we can extend it to add ssl/tls-profile(certz reference) and other TLS specific configs which can help me control version, ciphers, signature algorithms etc. Please check the feasibility of this approach

danielbarney: @navaneethyv What do you think?

earies: As it stands, this leaf is not enough for mutual auth.  The schema needs to be mindful of both any gNSI approaches as well as schema-driven profiles/references (the latter being the predominant deployment across secured gRPC transports today)

Agree that TLS is in need of its own hierarchy to consolidate all TLS related parameters

e.g. JUNOS/EVO

```
# set system services http servers server S1 tls ?            
Possible completions:
  <[Enter]>            Execute this command
+ apply-groups         Groups from which to inherit configuration data
+ apply-groups-except  Don't inherit configuration data from these groups
  local-certificate    Name of local X.509 certificate to use
> mutual-authentication  Enable TLS mutual authentication
  |                    Pipe through a command
[edit]
# set system services http servers server S1 tls mutual-authentication ?
Possible completions:
  <[Enter]>            Execute this command
+ apply-groups         Groups from which to inherit configuration data
+ apply-groups-except  Don't inherit configuration data from these groups
  authentication-type  Specify requirements for client certificate
  certificate-authority  Certificate authority profile
  |                    Pipe through a command
[edit]
# set system services http servers server S1 tls mutual-authentication authentication-type ?
Possible completions:
  dont-request-cert    Client certificate is not requested
  request-cert-but-dont-verify  Request certificate from client, do not verify
  request-cert-and-verify  Request certificate from client, verify if provided
  request-and-require-cert-but-dont-verify  Client certificate is mandatory, do not verify
  request-and-require-cert-and-verify  Client certificate is mandatory, certificate is verified
```

release/models/system/openconfig-system-grpc.yang
```diff
@@ -208,6 +237,70 @@ module openconfig-system-grpc {
...
+        range "1..max";  
+      }  
+      units "seconds";  
+      description  
+        "Interval in seconds at which the server will send keepalive PING  
+        frames to the client to verify connection health when no activity  
+        is detected.";  
+    }  
+
+    leaf min-keepalive-interval {
```
earies:
```suggestion
    leaf min-interval {  
```
Already under `keepalive` container

release/models/system/openconfig-system-grpc.yang
```diff
@@ -167,6 +175,18 @@ module openconfig-system-grpc {
         are not supported, such as lab testing.";
     }
 
+    leaf tls-mutual {  
+      type boolean;  
+      default false;  
+      must "not(../transport-security) or ../transport-security = 'true'" {
```
earies, line 181: This might be the time to question `transport-security` defaulting to `true`.  Great idea to enforce secure channels but there are dependencies that need to be fulfilled prior to that assumption.

This should be able to be conveyed with just:
```suggestion
      must "../transport-security = 'true'" {  
```
However default=true is imo problematic in reality

release/models/system/openconfig-system-grpc.yang
```diff
@@ -208,6 +237,70 @@ module openconfig-system-grpc {
     }
   }
 
+  grouping grpc-server-keepalive-config {  
+    description  
+      "Configuration parameters relating to gRPC server keepalive.";  
+
+    leaf time {
```
earies:
```suggestion
    leaf interval {  
```

release/models/system/openconfig-system-grpc.yang
LimeHat, line 178: In two of the referenced implementations (Nokia and Arista), mTLS is configured exclusively at the TLS-profile level.

Although a case can be made for placing it at the per-server level, I don’t see any rationale for choosing this approach in the description.

LimeHat (changes requested):

release/models/system/openconfig-system-grpc.yang
```diff
@@ -208,6 +237,70 @@ module openconfig-system-grpc {
     }
   }
 
+  grouping grpc-server-keepalive-config {  
+    description  
+      "Configuration parameters relating to gRPC server keepalive.";  
+
+    leaf time {  
+      type uint32 {  
+        range "1..max";
```
LimeHat, line 246: This probably can start at 0 (where `0` means `disabled`)?

release/models/system/openconfig-system-grpc.yang
danielbarney: What about 
/system/grpc-servers/grpc-server/tls/tls-type = Enum: mtual / Local
/system/grpc-servers/grpc-server/tls/authentication-type = ENUM request-cert-but-dont-verify ... ETC
/system/grpc-servers/grpc-server/tls/profile-name
/system/grpc-servers/grpc-server/tls/profile-location = Path to where the cert are saved in harddisk

This just enforces the

release/models/system/openconfig-system-grpc.yang
danielbarney: SG

release/models/system/openconfig-system-grpc.yang
danielbarney: Nokia and Arista have in running config the profile configured, but this details are in the cert itself. I dont think we want to model in OC the cert data, what we want to model is how that cert would be used ... specifically if it would be used in mutual tls. 

Cisco in this /system/tls does not have it in running config as it is expected in their implementation the files are located in a well know directory

PR #1537, opened 2026-08-26
Fix typo in packets-per-meaurement-period leaf

hung1806vn:
Fix typo in packets-per-meaurement-period leaf


romeyod:
renaming the leaf is a backward incompatible change in this case. Not sure if worth deprecation and adding a new leaf

@dplore @navaneethyv

earies:
> renaming the leaf is a backward incompatible change in this case. Not sure if worth deprecation and adding a new leaf
> 
> @dplore @navaneethyv

agreed - one would also think it's caught by the first implementor vs. blindly implemented w/ a typo as well.  No harm in introducing a new leaf, marking the existing as `status deprecated` just incase.

release/models/oam/openconfig-oam-cfm.yang
earies, line 32: Add a new revision + increment version/reference per comments on compatibility

release/models/oam/openconfig-oam-cfm.yang
```diff
@@ -1249,7 +1249,7 @@ module openconfig-oam-cfm {
         "MEF-SOAM-PM-MIB.mefSoamLmCfgNumIntervalsStored";
     }
 
-    leaf packets-per-meaurement-period {
+    leaf packets-per-measurement-period {
       type uint16;
```
earies, line 1253: Whether defining a new leaf or modifying existing, I'd probably suggest filling in any missing `units` statements that should otherwise be there.

dplore:
> > renaming the leaf is a backward incompatible change in this case. Not sure if worth deprecation and adding a new leaf
> > @dplore @navaneethyv
> 
> agreed - one would also think it's caught by the first implementor vs. blindly implemented w/ a typo as well. No harm in introducing a new leaf, marking the existing as `status deprecated` just incase.

Thanks for the PR @hung1806vn .  I support this change but , please refactor this PR to make it a non-breaking change.

PR #1541, opened 2026-09-03
Create new YANG path for triggering backup-next-hop-group.

aayueshbarui:
### Change Scope

* Adding a new path to trigger backup-next-hop-group. Currently backup for a next-hop-group is triggered automatically by the switch. With this new AFT path, the backup could be triggered by the switch client as well through gRIBI. 
* [Please indicate whether this change is backwards compatible.] - Yes

### Platform Implementations

This is a new feature to be implemented by vendors. We already have communicated this feature with a few vendors and have got confirmation from Arista, Juniper and Nokia.

### Tree View
```diff
module: openconfig-network-instance
   +--rw network-instances
      +--rw network-instance* [name]
         +--rw afts
+        |  +--ro backup-activations
+        |  |  +--ro backup-activation* [next-hop-group]
+        |  |     +--ro next-hop-group    -> ../state/next-hop-group
+        |  |     +--ro state
+        |  |        +--ro next-hop-group?   -> ../../../../next-hop-groups/next-hop-group/state/id
```

### DIFF
--- a/release/models/aft/openconfig-aft-common.yang
+++ b/release/models/aft/openconfig-aft-common.yang
@@ -23,7 +23,13 @@ submodule openconfig-aft-common {
     "Submodule containing definitions of groupings that are re-used
     across multiple contexts within the AFT model.";
 
-  oc-ext:openconfig-version "3.4.0";
+  oc-ext:openconfig-version "3.5.0";
+
+  revision "2026-09-01" {
+    description
+      "Add backup activation state for a next-hop-groups.";
+    reference "3.5.0";
+  }
 
   revision "2026-03-10" {
     description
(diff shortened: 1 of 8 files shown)


release/models/aft/openconfig-aft-common.yang
```diff
@@ -1086,6 +1092,48 @@ submodule openconfig-aft-common {
...
+        }
+
+        container state {
+          config false;
+
+          description
+            "Operational state parameters relating to backup activation.";
+
+          leaf next-hop-group {
+            type uint64;
```
earies: Would suggest expanding your tree view in the description for surrounding context as it is hard to tell for reviewers but would assume this should rather be a leafref back to an existing nh group?

aayueshbarui: Done. Please see description.

earies: The type is still not addressed - is there a reason this is not a leafref and if so, expand the tree context for associated references

earies:
Comment https://github.com/openconfig/public/pull/1541/changes#r3950922016 is not resolved (unable to unresolve)

release/models/aft/openconfig-aft-common.yang
aayueshbarui: The reason of not doing reference to NHG is because its possible that controller tries deleting NHG even when the backup is activated, in that case NHG deletion should not be blocked.

earies: That sounds like implementation detail that should not affect the modeling.  The scenario you describe is entirely possible in various other modeling but the relationship should remain intact to where the source of truth of NHG lay vs. creating a completely separate dangling ID

aayueshbarui: Done. Thanks

release/models/aft/openconfig-aft-common.yang
```diff
@@ -1086,6 +1092,50 @@ submodule openconfig-aft-common {
...
+          "An operational entry indicating that the backup-next-hop-group has been
+           activated for the specified next-hop-group.";
+
+        leaf next-hop-group {
+          type leafref {
+            path "../state/next-hop-group";
+          }
+
+          description
+            "Reference to the target next-hop-group whose backup path is activated.";
```
nandanarista: nit: should we say "next-hop-group" instead of "path" here?

nandanarista (approved):

dplore:
@aayueshbarui please fix the versioning for submodules.   We'll investigate the failure with the pyangbind check which appears to be something wrong on the backend here.

dplore (approved):

PR #1542, opened 2026-09-09
gnsi-pathz: add aggregate per-rule match histogram for pathz policy

mur7uza:
## Summary

Add an aggregate per-rule match histogram to `openconfig-gnsi-pathz` that reflects the actual pathz evaluation model. Existing pathz counters are tagged as deprecated to retain backward-compatibility.

## Problem

The current pathz accounting model (v0.3.0) inherits the authz counter template — counters keyed by OpenConfig path with read/write × accept/reject splits. This works for authz where RPCs are finite and defined by proto, but breaks for pathz:

| | authz (per-RPC) | pathz (per-path) |
|---|---|---|
| Key cardinality | Finite (proto-defined) | **Unbounded** (any accessed path) |
| Rule mapping | 1:1 (rule → RPC) | Many:many (wildcards, overlap) |
| Key semantics | RPC name IS the subject | Path is one attribute of a rule |

**Result:** Cannot answer basic operator questions — *Which rules are firing? Are any rules unused? Did policy rotation change behavior?*

## Solution

The pathz evaluation model is `f(user, path, mode) → rule`. Counters should be keyed by the rule that determined the outcome, not the data model (paths).

### Data tree (proposed)

```
gnmi-pathz-policies/policies/policy[instance='ACTIVE']
  └── rules/rule[id]            ← AuthorizationRule.id
      └── state/
          ├── matches           ← aggregate counter, summed across all users
          └── last-match-time   ← stale rule detection
```

### Key design decisions

| Decision | Rationale |
|----------|-----------|
| **Single `matches` counter** | Rule carries `ACTION_PERMIT` / `ACTION_DENY`. A match on a rule IS the outcome. Separate accept/reject counters are redundant. |
| **`DEFAULT_DENY_RULE`** | Well-known rule-id string constant for the implicit default deny when no explicit rule matched. Ensures every evaluation is accounted for. |
| **No read/write split** | Rule already encodes `MODE_READ` / `MODE_WRITE`. Counter-level split duplicates policy information. |
| **No per-user dimension** | Counter footprint is fixed at O(rules) — bounded by policy size, not by traffic volume or active user population. |
| **Active instance only** | Sandbox is transient during `Rotate()`. Sandbox validation uses the `Probe()` RPC. Enforced via `when "state/instance = 'ACTIVE'"`. |

### Operator workflows enabled

| Workflow | v0.3.0 (per-path) | Proposed (aggregate per-rule) |
|----------|-------------------|-------------------------------|
| Rule utilization signal | Impossible (no rule dimension) | `rules/rule[id]/matches` — a starting point for auditing, not standalone proof (see Caveats) |
| Rule utilization histogram | Impossible | Native — `rules/rule[id]/matches` |
| Post-rotation drift detection | Compare path counts | Compare rule histograms before/after, as a post-rotation usage signal — `Probe()` remains the authoritative pre-rotation validation mechanism (see Caveats) |
| Stale rule detection | Not available | `last-match-time` per rule (same tree-instantiation caveat) |
| Per-user attribution | Not available | Not available (unchanged trade-off) |
| Per-request correlation | Not available | Not available (unchanged trade-off — see Caveats) |

### Caveats

- **Histogram is accurate for the YANG tree instantiation at the time of policy evaluation.** Since tree instantiation can change with the addition of new nodes, a rule with `matches = 0` need not be a definitive dead-rule indicator — it may simply target a path with no currently-provisioned instance.
- **`Probe()` is the primary, authoritative mechanism to validate pathz policy accuracy** (e.g. ahead of a rotation, against the sandboxed instance). The policy histogram is a post-rotation policy usage signal, not a substitute for `Probe()`-based validation.
- **Pathz enforcement is recursive** — a `Get` at a container path applies enforcement to each node under the subtree, so a single client request may result in multiple pathz evaluations across different rules, with no 1:1 correlation back to the request.

Overall, the histogram is meant to provide an aggregate policy match distribution / usage signal — not a definitive dead-rule proof or a per-request validation tool.

### Scalability

- **Fixed footprint:** O(rules) — 47 rules → 47 entries, regardless of traffic or fleet size
- **No lazy instantiation needed:** all rule entries exist from policy load, since the key space is the policy itself
- **Simple collection:** gNMI Subscribe on `rules`, no per-user fan-out

## Changes

- **Deprecated:** `counters`, `gnmi-pathz-policy-success-failure-counters`, `gnmi-pathz-policy-xpath-success-failure-counters` groupings and the per-gRPC-server counter augment that uses them — marked `status deprecated`, kept in the schema for backwards compatibility
- **Added:** `pathz-policy-rule-counters` grouping (`matches`, `last-match-time`)
- **Added:** `pathz-policy-rules` grouping (`rules/rule[id]`, aggregate counters, no user dimension)
- **Version:** 0.3.0 → 0.4.0

### Tree View

```diff
 module: openconfig-gnsi-pathz
 
   augment /oc-sys:system:
     +--ro gnmi-pathz-policies
        +--ro policies
           +--ro policy* [instance]
              +--ro instance    -> ../state/instance
              +--ro state
              |  +--ro instance?     enumeration
              |  +--ro version?      version
              |  +--ro created-on?   created-on
+             +--ro rules
+                +--ro rule* [id]
+                   +--ro id       -> ../state/id
+                   +--ro state
+                      +--ro id?                string
+                      +--ro matches?           oc-yang:counter64
+                      +--ro last-match-time?   oc-types:timeticks64
   augment /oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server/oc-sys-grpc:state:
     +--ro gnmi-pathz-policy-version?      version
     +--ro gnmi-pathz-policy-created-on?   created-on
   augment /oc-sys:system/oc-sys-grpc:grpc-servers/oc-sys-grpc:grpc-server:
     +--ro gnmi-pathz-policy-counters
        +--ro paths
           +--ro path* [name]
              +--ro name     -> ../state/name
              +--ro state
                 +--ro name?     string
                 +--ro reads
                 |  x--ro access-rejects?       oc-yang:counter64
                 |  x--ro last-access-reject?   oc-types:timeticks64
                 |  x--ro access-accepts?       oc-yang:counter64
                 |  x--ro last-access-accept?   oc-types:timeticks64
                 +--ro writes
                    x--ro access-rejects?       oc-yang:counter64
                    x--ro last-access-reject?   oc-types:timeticks64
                    x--ro access-accepts?       oc-yang:counter64
                    x--ro last-access-accept?   oc-types:timeticks64
```

## Out of scope

This change intentionally does not add per-user attribution or per-request forensics. It answers "which rules are firing" but not "who triggered them" or "which exact path was denied." Those can be pursued independently in follow-up proposals if there is sufficient interest, without blocking this fix.


brianneville:
I think this is better than counters per-xpath, but im still not 100% sure about this, or pathz counters in general.

> Result: Cannot answer basic operator questions — Which rules are firing? Are any rules unused? Did policy rotation change behavior?

Even under this proposal its difficult to determine this without knowing for sure what the state of the real YANG tree is at the time of request. 
eg. if you have some rules for a user like `rule A - PERMIT READ /a`, `rule B - DENY READ /a/b` and then make a request for `/a` , you may assume that rule B is unused if it doesnt fire, but it may be the case that `/a/b` does not actually exist yet.
If all rules are on top level paths though (e.g. `/components`, `/system`) , then this would not be an issue, but nothing in the spec limits rules to top-level paths.

>Did policy rotation change behavior?

I think the question of "Did policy rotation change behavior?" should be delegated to the Probe RPC, and should be determined concretely during the rotation while the new policy is still sandboxed, rather than being something that we uncover by having clients run requests and examining the counters.

>The pathz evaluation model is f(user, path, mode) → rule. Counters should be keyed by the rule that determined the outcome, not the data model (paths).

Each yang path does have this `f(user, path, mode) → rule` mapping, but there may be many paths relevant for a given request since the enforcement should recurse into the relevant tree (as per https://github.com/openconfig/gnsi/blob/4b7fb6120826a243752ad2d98cef9bc513266f73/pathz/authorization-README.md?plain=1#L50-L51 "For portions of the tree for which the user has no access no data will be returned". This means that many rules will be "best matches" under a given gNMI request unless it is to a leaf directly. 
So we don't have a clear association here where 1 request = 1 rule match.

Network operators would see that a single request would lead to many counters incrementing, and those counters would increment potentially multiple times, based on the state of the YANG tree. On the client side if the rules are complex it could be difficult to associate each path from the request with the subset of rules that were relevant for it.

Is the expectation here that the network operator should be able to look at the counters and determine if they are incrementing as-expected given the policy rules? If so it feels like they'd almost need their own pathz enforcement engine to validate the result for each path, unless the rules are all simple, like being on top-level paths.

mur7uza:
> I think this is better than counters per-xpath, but im still not 100% sure about this, or pathz counters in general.
> 
> > Result: Cannot answer basic operator questions — Which rules are firing? Are any rules unused? Did policy rotation change behavior?
> 
> Even under this proposal its difficult to determine this without knowing for sure what the state of the real YANG tree is at the time of request. eg. if you have some rules for a user like `rule A - PERMIT READ /a`, `rule B - DENY READ /a/b` and then make a request for `/a` , you may assume that rule B is unused if it doesnt fire, but it may be the case that `/a/b` does not actually exist yet. If all rules are on top level paths though (e.g. `/components`, `/system`) , then this would not be an issue, but nothing in the spec limits rules to top-level paths.
>
Histogram is accurate for the YANG tree instantiation at the time of policy evaluation. Since tree instantiation can change with addition of new nodes, policy rule with matches=0 need not be definitive dead-rule indicator.

> > Did policy rotation change behavior?
> 
> I think the question of "Did policy rotation change behavior?" should be delegated to the Probe RPC, and should be determined concretely during the rotation while the new policy is still sandboxed, rather than being something that we uncover by having clients run requests and examining the counters.
>
Agree, Probe() is the primary authoritative mechanism to validate pathz policy accuracy. Where as policy histogram would be post rotation policy usage signal. I can update PR description to reflect this.

> > The pathz evaluation model is f(user, path, mode) → rule. Counters should be keyed by the rule that determined the outcome, not the data model (paths).
> 
> Each yang path does have this `f(user, path, mode) → rule` mapping, but there may be many paths relevant for a given request since the enforcement should recurse into the relevant tree (as per https://github.com/openconfig/gnsi/blob/4b7fb6120826a243752ad2d98cef9bc513266f73/pathz/authorization-README.md?plain=1#L50-L51 "For portions of the tree for which the user has no access no data will be returned". This means that many rules will be "best matches" under a given gNMI request unless it is to a leaf directly. So we don't have a clear association here where 1 request = 1 rule match.
> 
Right, this originates from pathz enforcement characteristic which is recursive in nature e.g. 'get' at the top level means pathz enforcement being applied for each KV under the subtree. Basically single request may results in multiple pathz evaluations.

> Network operators would see that a single request would lead to many counters incrementing, and those counters would increment potentially multiple times, based on the state of the YANG tree. On the client side if the rules are complex it could be difficult to associate each path from the request with the subset of rules that were relevant for it.
> 
> Is the expectation here that the network operator should be able to look at the counters and determine if they are incrementing as-expected given the policy rules? If so it feels like they'd almost need their own pathz enforcement engine to validate the result for each path, unless the rules are all simple, like being on top-level paths.

Overall, histogram is meant to provide an aggregate policy match distribution / usage signal — not a definitive dead-rule proof or a per-request validation tool.

PR #1551, opened 2026-09-16
Add a system-wide network-instance leaf to NTP config and state

vishak-mut:
* (M) release/models/system/openconfig-system.yang
  - Add a system-wide (global) NTP network instance leaf at
    `/system/ntp/config/network-instance`.
  - Clarify that the existing per-server
    `/system/ntp/servers/server/config/network-instance` overrides the
    system-wide value.

### Change Scope

Today the NTP model only allows a network instance (VRF) to be specified
per-server, via `/system/ntp/servers/server/config/network-instance`
(added in `0.17.0`). Several network operating systems instead model NTP
VRF membership as a **global** property of the NTP client: the NTP
process runs in (and is bound to) a single VRF, and every server
association is established over that VRF. There is currently no way to
express this in OpenConfig other than repeating the same network instance
on every configured server, which also cannot represent NTP traffic that
is not associated with a configured server (for example, servers learned
via DHCP, or NTP server/listening mode).

This change adds `/system/ntp/config/network-instance` with the same type
(`oc-ni:network-instance-ref`) as the existing per-server leaf, with the
following semantics:

* The system-wide value is the default network instance used for NTP
  operation, including reaching any server in the `ntp/servers/server`
  list, and for NTP packets that are not associated with a specific
  configured server.
* Implementations that run a single NTP client bound to one network
  instance use only this leaf, and do not support the per-server leaf.
* Where per-server selection is supported, the per-server leaf takes
  precedence, and the system-wide leaf applies to any server that does
  not specify one.
* If neither is configured, the default network instance is used.

This keeps existing per-server behaviour fully intact, while allowing
operators to express "all NTP runs in VRF X" with a single leaf, which is
how the platforms below expose the feature.

The change is backwards-compatible and purely additive (minor version bump
`3.3.0` -> `3.4.0`).

### Platform Implementations

* **SONiC** [NTP design document](https://github.com/sonic-net/SONiC/blob/master/doc/ntp/ntp-design.md)
  - NTP VRF is a global property of the NTP client, not a per-server one:
    - CLI: `config ntp vrf default` / `config ntp vrf mgmt`
    - CONFIG_DB: `NTP|global|vrf`
    - YANG: `sonic-ntp/NTP/global/vrf`, described as "NTP can be enabled
      only in one VRF at a time."
  - Servers are configured independently with
    `config ntp server add <host> [--type|--key|--version]`, and all
    server associations are established over the globally configured VRF.
    The global source interface must also reside in that VRF.
  - Verification: `show ntp global` (reports `VRF: <vrf>`), `show ntp`,
    `show ntp server`

* **HPE Aruba Networking (AOS-CX)** [documentation](https://arubanetworking.hpe.com/techdocs/AOS-CX/10.13/HTML/fundamentals_8400/Content/Chp_IniCfg/NTP_cmds/ntp-ser.htm)
  - Global NTP VRF:
    - `ntp vrf <VRF-NAME>`: sets the VRF the NTP client uses to reach all
      configured servers (`no ntp vrf` reverts to the default VRF).
  - Servers are configured with
    `ntp server <FQDN|IPV4-ADDR|IPV6-ADDR> [key-id <KEY-ID>] [version <VERSION>]
    [prefer] [burst | iburst] [minpoll <POLL>] [maxpoll <POLL>]`,
    and use the globally configured VRF.
  - Verification: `show ntp status`, `show ntp associations`,
    `show ntp servers`

### Tree View

```diff
 module: openconfig-system
   +--rw system
      +--rw ntp
         +--rw config
         |  +--rw enabled?            boolean
         |  +--rw enable-ntp-auth?    boolean
+        |  +--rw network-instance?   oc-ni:network-instance-ref
         +--ro state
         |  +--ro enabled?            boolean
         |  +--ro enable-ntp-auth?    boolean
+        |  +--ro network-instance?   oc-ni:network-instance-ref
         |  +--ro auth-mismatch?      oc-yang:counter64
         +--rw ntp-keys
         |  +--rw ntp-key* [key-id]
         |     +--rw key-id    -> ../config/key-id
         |     +--rw config
         |     |  +--rw key-id?      uint16
         |     |  +--rw key-type?    identityref
         |     |  +--rw key-value?   string
         |     +--ro state
         |        ...
         +--rw servers
            +--rw server* [address]
               +--rw address    -> ../config/address
               +--rw config
               |  +--rw address?            oc-inet:host
               |  +--rw port?               oc-inet:port-number
               |  +--rw version?            uint8
               |  +--rw association-type?   enumeration
               |  +--rw iburst?             boolean
               |  +--rw prefer?             boolean
               |  +--rw network-instance?   oc-ni:network-instance-ref
               |  +--rw source-address?     oc-inet:ip-address
               |  +--rw key-id?             -> ../../../../ntp-keys/ntp-key/key-id
               +--ro state
                  ...
```


earies:
General comments:

This problem statement exists in quite a few areas today and what you are proposing can be achieved by rather some implementation detail/guardrails to enforce all instances share the same `network-instance-ref` - no model changes necessary, single pattern.

If this pattern is introduced, it means there are > 1 patterns to accomodate implementations.  We probably have various cases of this already but isn't the intended approach.  The inverse of domains referencing associated network-instances (vs. nested under a network-instance) also comes with it's limitations in the case of `servers` in that the key identifier of an `address` must remain unique no matter the NI attachment.

If we were to take an approach as suggested here, I'd propose a uniform approach for this problem statement with clear definition of inheritance (any domain in which children can inherit or precede parent attributes) - without that, we'll end up seeing different design patterns and solutions throughout the model set.

